PR #29, opened 2016-02-15
Parse Mutation block from GraphQL

manishrjain:
GraphQL doesn't have a self-sufficient way to provide mutations. So, I created something which should work well, using RDF as the input format. This is how a valid mutation would look like:
```
    mutation {
      set {
        <name> <is> <something> .
        <hometown> <is> <san francisco> .
      }
      delete {
        <name> <is> <something-else> .
      }
    }
    query {
      me(_xid_: tomhanks) {
        name
        hometown
      }
    }
```

Mutation supports 2 operations: `set`, and `delete`. The data inside the `set` and `delete` blocks is in RDF NQuad format; the same format that our bulk loader parses as well.

Mutations can be independent, but a `mutation` block can also be paired with a `query` block. The order in which these would run would be mutation first, then query.

[Review on Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/29)

### DIFF
--- a/gql/parser.go
+++ b/gql/parser.go
@@ -36,6 +36,11 @@ type GraphQuery struct {
 	Children []*GraphQuery
 }
 
+type Mutation struct {
+	Set string
+	Del string
+}
+
 func run(l *lex.Lexer) {
 	for state := lexText; state != nil; {
 		state = state(l)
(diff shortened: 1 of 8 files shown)


ashwin95r:
Reviewed 8 of 8 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dgraph-io/dgraph/29)*

ashwin95r:
:lgtm: 

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dgraph-io/dgraph/29#-:-KAYKjVzJx6htgid28B3:980077423)*

PR #146, opened 2016-08-03
Try to serve Go client and http clients on the same port

pawanrawal:
Doesn't work, but wanted to get a second view on how I am doing things here. Using the same listener, either the serve method for HTTP takes up all calls or that for Grpc. So only one of them works.

Also wanted to check houndci.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/146)

### DIFF
--- a/cmd/dgraph/main.go
+++ b/cmd/dgraph/main.go
@@ -310,18 +310,13 @@ func (s *server) Query(ctx context.Context,
 
 // This function register a Dgraph grpc server on the address, which is used
 // exchanging protocol buffer messages.
-func runGrpcServer(address string) {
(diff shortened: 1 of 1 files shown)


manishrjain:
Hmm.. makes sense. HoundCI would only trigger if it finds a lint issue in the diff. Okay, let's stick to the original 2 port version.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/146#-:-KOF4AvoCDHdc1jDdyHG:-2096162252)*

PR #169, opened 2016-08-21
Feature/filter

jchiu0:
First attempt at filters / name matching.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/169)

### DIFF
--- a/gql/parser.go
+++ b/gql/parser.go
@@ -168,32 +168,8 @@ func Parse(input string) (gq *GraphQuery, mu *Mutation, rerr error) {
 		if err := gq.expandFragments(fmap); err != nil {
 			return nil, nil, err
 		}
-		// Validate query.
-		if err := validate(gq); err != nil {
-			return nil, nil, err
-		}
 	}
 	return gq, mu, nil
-}
-
-func validate(gq *GraphQuery) error {
-	attrMap := make(map[string]bool)
-	for _, child := range gq.Children {
-		if child.Attr != "" {
-			attrMap[child.Attr] = true
-		}
-	}
-	for _, f := range gq.Filters {
-		if !attrMap[f.Key] {
-			return fmt.Errorf("Filter on missing attr: %s", f.Key)
-		}
-	}
-	for _, child := range gq.Children {
-		if err := validate(child); err != nil {
-			return err
-		}
-	}
-	return nil
 }
 
 // getFragment parses a fragment definition (not reference).
--- a/gql/parser.go
+++ b/gql/parser.go
@@ -35,6 +35,7 @@ type GraphQuery struct {
 	Offset   int
 	After    uint64
 	Children []*GraphQuery
+	Filters  []Pair
 
 	// Internal fields below.
 	// If gq.fragment is nonempty, then it is a fragment reference / spread.
@@ -47,8 +48,8 @@ type Mutation struct {
 	Del string
 }
 
-// pair denotes the key value pair that is part of the GraphQL query root in parenthesis.
-type pair struct {
+// Pair denotes the key value pair that is part of the GraphQL query root in parenthesis.
+type Pair struct {
 	Key string
 	Val string
 }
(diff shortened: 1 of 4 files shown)


manishrjain:
Love your pace of work! I think the design could use work. Searching and parsing each and every value is slow and inefficient, look at Bleve instead. Let's discuss this tomorrow maybe. There're certain issues that need to be tackled, for, ​e.g., how do we deal with migration of string indexes to other machines -- or do we regenerate them from scratch every time, etc.

---

Reviewed 4 of 4 files at r1.

---

*[query/query.go, line 666 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/169#-KPgPkFsOwcoXVJH28SA:-KPgPkFsOwcoXVJH28SB:bvbt3p0) ([raw file](https://github.com/dgraph-io/dgraph/blob/8728565b40542834ccf8aae3ed5a491279176222/query/query.go#L666)):*
> ```Go
> }
> 
> func (sg *SubGraph) applyFilters(ctx context.Context, sorted []uint64) ([]uint64, error) {
> ```

This seems to iterate over all the results and check their values. Instead, what you want to do is to use somethine like Bleve: http://www.blevesearch.com/ and generate an index up front for a given predicate. Then, on search, ideally this generates a list of sorted uids, which you can then use.

---

*[query/query.go, line 731 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/169#-KPgQ9eqIJKdzyDhzg98:-KPgQ9erwrM-31ijnrrv:b3hdj8a) ([raw file](https://github.com/dgraph-io/dgraph/blob/8728565b40542834ccf8aae3ed5a491279176222/query/query.go#L731)):*
> ```Go
> 		sg.keepUIDMap = make(map[uint64]struct{})
> 		for _, uid := range newSorted {
> 			sg.keepUIDMap[uid] = struct{}{}
> ```

Don't understand the reasoning behind this. Don't you want to intersect this sorted list with another sorted list? If so, why use a map, they're inefficient.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/169#-:-KPgQR56_Io-uhFsC8Gx:bf577on)*

jchiu0:

---

*[query/query.go, line 731 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/169#-KPgQ9eqIJKdzyDhzg98:-KPg_u-ZgWAnIDKuTVmN:b-12spu5) ([raw file](https://github.com/dgraph-io/dgraph/blob/8728565b40542834ccf8aae3ed5a491279176222/query/query.go#L731)):*
> Don't understand the reasoning behind this. Don't you want to intersect this sorted list with another sorted list? If so, why use a map, they're inefficient.
</details>
Yes, you are right about this. This looks more like a hack to get the feature out.. I am quite keen to do a rewrite of the "query" package for several reasons..

See doc on some thoughts on other things and this as well:
https://docs.google.com/document/d/19dp4IFD72NMJw3m1Yf6IChmZCJMrpNuqsAg0PkaAHsk/edit#

Will reply to more comments tomorrow.. Thanks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/169)*

manishrjain:

---

*[query/query.go, line 731 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/169#-KPgQ9eqIJKdzyDhzg98:-KPh3S-4WMyc9WmFz8bP:bf4dokg) ([raw file](https://github.com/dgraph-io/dgraph/blob/8728565b40542834ccf8aae3ed5a491279176222/query/query.go#L731)):*
> Yes, you are right about this. This looks more like a hack to get the feature out.. I am quite keen to do a rewrite of the "query" package for several reasons..
> 
> See doc on some thoughts on other things and this as well:
> https://docs.google.com/document/d/19dp4IFD72NMJw3m1Yf6IChmZCJMrpNuqsAg0PkaAHsk/edit#
> 
> Will reply to more comments tomorrow.. Thanks.
</details>
Sure. Just a general advice fwiw: Don't be so keen to rewrite packages before truly understanding what's going on there. I've done many refactorings at Google for code written back in 2003. The general rule of thumb that we followed for refactorings was that we had at least 2-3 engineers who agreed what would be output look like and why that's better than the current status. I'm a big proponent of refactorings myself -- as long as they're justified. It's easy to fall into the trap of, I don't understand this, and hence this is bad. So, avoid that.

Having a look at the doc. The great thing that you've done here is that you've written down your thoughts. This is amazing, and a great step to engage in communication to ensure that you understand the code base and then can accurately figure out its flaws. Once the flaws are obvious and agreed upon, that'd let you figure out what's the best option here to proceed -- fix / refactor / rewrite, etc.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/169)*

PR #170, opened 2016-08-21
Temp solution to get blockCacheSize

jchiu0:
Piping block cache size from RocksDB.

(It is probably forbidden to modify vendor's code.. I suppose I need to create a PR against gorocksdb, but before that, I might need to do a PR against facebook rocksdb? It is going to take a while)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/170)


manishrjain:
Hey Jiawei, 2 things:
- Add the results of your testing with this change? Can we accurately calculate the size of dgraph process with this change?
- No point putting things in vendor, they'll get overridden with a sync. So, instead send a PR to gorocksdb and check if they're willing to accept it. In most likeliness, they should be quick to integrate with a little​ bit of persistence.

---

Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/170#-:-KPgNeC4r_fvyvZWLcu-:bip8yuu)*

manishrjain:
One more thing, just read that you'd need to change Facebook's rocksdb code. Why's that? Don't they already have a way to determine how much memory is being consumed by RocksDB process? I'm sure there should be. So, would make sense to clarify it with them as well -- they're pretty responsive.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/170#-:-KPgW4B_54QHLKMK2Qtp:b-8kasee)*

jchiu0:
(I haven't added you as a reviewer since it's weekend.. but yes, I will do that.) Yes, rocksdb has C++ code to provide block cache size. But staring at c.h and c.cc, I don't think this is exposed as a C function. Hence, gorocksdb cannot access it, I believe.

I will be back with more numbers for size of dgraph process.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/170#-:-KPgZl9u6M85Xlycg3sW:b-syzcz7)*

PR #177, opened 2016-08-24
Predicate data rpc

pawanrawal:
This PR provides an RPC method using which an instance can request data corresponding to a predicate from another instance.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/177)


ashwin95r:
Review status: 0 of 7 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[worker/worker.go, line 157 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPwTV3Hv9NC_J4Dq0qt:-KPwTV3Hv9NC_J4Dq0qu:bpjy2pq) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/worker.go#L157)):*
> ```Go
> 
> 	it := dataStore.GetIterator()
> 	it.Seek(qp)
> ```

In case this machine doesn't have the required predicate, could throw an error here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/177)*

ashwin95r:
Review status: 0 of 7 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[worker/worker.go, line 156 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPwVpYKCPoxMh3WoXWv:-KPwVpYLrgUKMaqEEL8T:b1p47h) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/worker.go#L156)):*
> ```Go
> 	defer dataStore.ReleaseSnapshot()
> 
> 	it := dataStore.GetIterator()
> ```

Should the iterator be obtained from the db or the snapshot object? GetIterator returns the db one. Just wanted to be sure here.

---

*[worker/worker.go, line 162 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPwUqv4amfoemYdT88H:-KPwUqv5T5Vpyev6Ku-T:b-2lggfu) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/worker.go#L162)):*
> ```Go
> 	for ; it.Valid(); it.Next() {
> 		k, v := it.Key(), it.Value()
> 		defer k.Free()
> ```

Instead of defer for every (k,v) which would be called after the function is out of scope, this could be done at the end of loop. Will free them earlier.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/177)*

manishrjain:
Reviewed 7 of 7 files at r1.
Review status: all files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[store/store.go, line 130 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPz_jsXiR0FhAFdGe5k:-KPz_jsXiR0FhAFdGe5l:brvpcfp) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/store/store.go#L130)):*
> ```Go
> 
> // WriteBatch performs a batch write of key value pairs to RocksDB.
> func (s *Store) WriteBatch(kv []KV) error {
> ```

Don't cache everything in memory before applying it to RocksDB. This would be very heavy on memory usage. Instead, use streaming and channels, and create batches of a certain size, say 32MB or something.

---

*[worker/payload.proto, line 14 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPza-NfCib9YWPgp-3b:-KPza-NgXg0BrHwwAnDT:b-h1bb6j) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/payload.proto#L14)):*
> ```protobuf
> 	rpc Mutate (Payload) returns (Payload) {}
> 	rpc ServeTask (Payload) returns (Payload) {}
>        rpc PredicateData (Payload) returns (stream Payload) {}
> ```

This is an odd one out compared to the others. The others use tab, this uses spaces. Tabs vs spaces, fight!

---

*[worker/predicate.go, line 29 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPza8aEaaCSllHb1-7n:-KPza8aEaaCSllHb1-7o:bpvjhji) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/predicate.go#L29)):*
> ```Go
> 
> // Data represents key-value data stored in RocksDB.
> type Data struct {
> ```

Why not have a KV struct in x.go? That way you can share it here.

---

*[worker/predicate.go, line 36 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPzaJbGiL9T1F36pHDv:-KPzaJbGiL9T1F36pHDw:b-kf9zxs) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/predicate.go#L36)):*
> ```Go
> func (d *Data) encode() ([]byte, error) {
> 	var b bytes.Buffer
> 	enc := gob.NewEncoder(&b)
> ```

gob encode would be slow. Don't do it. Instead use flatbuffers. Directly write the key retrieved from rocksdb to a flatbuffer, then you can read it without decoding at the receiving end.

---

*[worker/predicate.go, line 80 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPzarzCta6pHa2VJSKu:-KPzarzCta6pHa2VJSKv:b-22danb) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/predicate.go#L80)):*
> ```Go
> 
> 		d := new(Data)
> 		if err := d.decode(b.Data); err != nil {
> ```

Avoid unnecessary encoding/decoding. Read the size of the buffer, read the buffer, set it to key, then size again, read the buffer, set it to value, and then pass it to rocksdb batch writer.

---

*[worker/predicate.go, line 83 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPzair4wrQE5hKp8ItY:-KPzair4wrQE5hKp8ItZ:bv7knew) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/predicate.go#L83)):*
> ```Go
> 			return err
> 		}
> 		kvs = append(kvs, store.KV{K: d.Key, V: d.Value})
> ```

Don't store them in a list. Bring the write batch call here directly, or pass it a channel or something.

---

*[worker/worker.go, line 153 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPzb7zm0FIl9XAbWZBd:-KPzb7zm0FIl9XAbWZBe:b-uybp2l) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/worker.go#L153)):*
> ```Go
> 	// RocksDB library we are using.
> 	// http://rocksdb.org/blog/2609/use-checkpoints-for-efficient-snapshots/
> 	dataStore.SetSnapshot()
> ```

During a snapshot can we still do writes to RocksDB? Can you add a wiki page about RocksDB and how we're using it? Add a task for yourself to do so.

---

*[worker/worker.go, line 157 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPwTV3Hv9NC_J4Dq0qt:-KPzbKX8RTDMmG_MOdy2:b-vvhipj) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/worker.go#L157)):*
> In case this machine doesn't have the required predicate, could throw an error here.
</details>
Always ensure that you're handling all the errors. Never ignore an error return.

---

*[worker/worker.go, line 169 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPzbYfVbqy3IG7d90Ag:-KPzbYfWuDciTbRr6EiU:bj8poft) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/worker.go#L169)):*
> ```Go
> 		// As keys are sorted, when we get a pred that doesn't match the query pred
> 		// we can return.
> 		if !bytes.Equal(qp, pred) {
> ```

You shouldn't need to check this. You can get a range iterator directly from RocksDB. Range could be [pred, pred~]  (with a tilde at the end, look at ascii table).

---

*[worker/worker.go, line 173 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPzbiQslJv9TED7M0aH:-KPzbiQslJv9TED7M0aI:b-pe4k4y) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/worker.go#L173)):*
> ```Go
> 		}
> 
> 		d := &Data{Key: k.Data(),
> ```

Move Key to next time.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/177)*

pawanrawal:
Review status: 3 of 15 files reviewed at latest revision, 32 unresolved discussions, some commit checks failed.

---

*[store/store.go, line 130 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPz_jsXiR0FhAFdGe5k:-KQ0-JCS9_LRTYv1fCll:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/store/store.go#L130)):*
> Don't cache everything in memory before applying it to RocksDB. This would be very heavy on memory usage. Instead, use streaming and channels, and create batches of a certain size, say 32MB or something.
</details>
Done.

---

*[worker/payload.proto, line 14 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPza-NfCib9YWPgp-3b:-KQ-QyRBERTzyh-7TKjz:b1u6eyd) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/payload.proto#L14)):*
> This is an odd one out compared to the others. The others use tab, this uses spaces. Tabs vs spaces, fight!
</details>
Oops, tabs it is.

---

*[worker/predicate.go, line 29 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPza8aEaaCSllHb1-7n:-KQ0-PxefdnBIxJQ3Ud_:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/predicate.go#L29)):*
> Why not have a KV struct in x.go? That way you can share it here.
</details>
Done.

---

*[worker/predicate.go, line 36 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPzaJbGiL9T1F36pHDv:-KQ-R0hTb1iU7waDQxDQ:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/predicate.go#L36)):*
> gob encode would be slow. Don't do it. Instead use flatbuffers. Directly write the key retrieved from rocksdb to a flatbuffer, then you can read it without decoding at the receiving end.
</details>
Done.

---

*[worker/predicate.go, line 80 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPzarzCta6pHa2VJSKu:-KQ0-RIlslh9YaQMcskp:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/predicate.go#L80)):*
> Avoid unnecessary encoding/decoding. Read the size of the buffer, read the buffer, set it to key, then size again, read the buffer, set it to value, and then pass it to rocksdb batch writer.
</details>
Done.

---

*[worker/predicate.go, line 83 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPzair4wrQE5hKp8ItY:-KQ0-S_qfILL7ZSFFavR:bbz6e4y) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/predicate.go#L83)):*
> Don't store them in a list. Bring the write batch call here directly, or pass it a channel or something.
</details>
Thanks, didn't realize this wastage here. Memory is precious one has to be cautious while using it.

---

*[worker/worker.go, line 153 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPzb7zm0FIl9XAbWZBd:-KQ-JGpWb39zQm9VyoHa:bjbdzze) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/worker.go#L153)):*
> During a snapshot can we still do writes to RocksDB? Can you add a wiki page about RocksDB and how we're using it? Add a task for yourself to do so.
</details>
Didn't find anything about this. I will search and get back to you. Here is the task. https://app.asana.com/0/158897884215291/172133660950368

---

*[worker/worker.go, line 156 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPwVpYKCPoxMh3WoXWv:-KQ-4BJm2dM4vkY1eTxx:b-8qre2g) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/worker.go#L156)):*
> Should the iterator be obtained from the db or the snapshot object? GetIterator returns the db one. Just wanted to be sure here.
</details>
So GetIterator gets the iterator from `s.ropt` to which Snapshot info is stored when we call `SetSnapshot`

---

*[worker/worker.go, line 157 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPwTV3Hv9NC_J4Dq0qt:-KQ-3evteSzjFWwPnLtT:b-1lfu6r) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/worker.go#L157)):*
> Always ensure that you're handling all the errors. Never ignore an error return.
</details>
Seek method doesn't return an error. Should I explicitly check if and return an error if we didn't find a key with the prefix?

---

*[worker/worker.go, line 162 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPwUqv4amfoemYdT88H:-KQ007fPkJt0CWvN8FPX:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/worker.go#L162)):*
> Instead of defer for every (k,v) which would be called after the function is out of scope, this could be done at the end of loop. Will free them earlier.
</details>
Done.

---

*[worker/worker.go, line 169 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPzbYfVbqy3IG7d90Ag:-KQ-4Rc7jURnAUbg_6vd:b-tr4q0l) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/worker.go#L169)):*
> You shouldn't need to check this. You can get a range iterator directly from RocksDB. Range could be [pred, pred~]  (with a tilde at the end, look at ascii table).
</details>
Ok, changed the check to `it.ValidForPrefix()` which is gorocksdb for the recommended way at the end of the post here. https://github.com/facebook/rocksdb/wiki/Prefix-Seek-API-Changes. 

---

*[worker/worker.go, line 173 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPzbiQslJv9TED7M0aH:-KQ-4nXl--S6l_Om4HQP:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/worker.go#L173)):*
> Move Key to next time.
</details>
Done.

---

*[worker/worker.go, line 176 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPwR1r9h3Mb-gt2UJtl:-KQ0096vAGC-ZeLGZht8:bepyuov) ([raw file](https://github.com/dgraph-io/dgraph/blob/b5ec52a06125e97de8dcdae8c06041e98b24fc68/worker/worker.go#L176)):*
> Will it be better if we define a type in proto file which has key and value bytes? It will avoid this encoding and decoding step and gogoproto marshalling/unmarshalling​ is quite faster than gob I guess.
</details>
Using flatbuffers now. Hope they are fastest :-)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/177)*

manishrjain:
Reviewed 9 of 12 files at r2, 1 of 3 files at r3, 2 of 2 files at r4.
Review status: all files reviewed at latest revision, 35 unresolved discussions, some commit checks failed.

---

*[store/store.go, line 128 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ0Vq9r4-VAHH7vLhy9:-KQ0Vq9sD90GCNAgH3ri:b-9z4pkg) ([raw file](https://github.com/dgraph-io/dgraph/blob/8e829a420c51d12c45cef104f8418722d466e05e/store/store.go#L128)):*
> ```Go
> 
> // WriteBatch performs a batch write of key value pairs to RocksDB.
> func (s *Store) WriteBatch(kv chan x.KV, che chan error) {
> ```

Also pass in a context, so you can add some tracing here. The other thing is, if this takes too long, you want to be able to cancel the operation.

---

*[store/store.go, line 132 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ0WdkbgQ_zclefNEdT:-KQ0Wdkc64YpYXkyleqB:bp5apiy) ([raw file](https://github.com/dgraph-io/dgraph/blob/8e829a420c51d12c45cef104f8418722d466e05e/store/store.go#L132)):*
> ```Go
> 	for i := range kv {
> 		wb.Put(i.Key, i.Val)
> 		if len(wb.Data()) > 32*MB {
> ```

wb.Data() is going to be expensive. It returns a serialized version of the data, IIUC. Instead, do the byte count over key and val, which would be fast.

---

*[store/store.go, line 141 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ0X-QbGpVtuc0ajB0D:-KQ0X-QcB1xwRKqUHEFU:bxkm3h9) ([raw file](https://github.com/dgraph-io/dgraph/blob/8e829a420c51d12c45cef104f8418722d466e05e/store/store.go#L141)):*
> ```Go
> 	// After channel is closed the above loop would exit, we write the data in
> 	// write batch here.
> 	if len(wb.Data()) > 0 {
> ```

use wb.Count() here, instead of Data() which is slow.

---

*[worker/predicate.go, line 29 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ0XL4Px9ykyi39Ovl4:-KQ0XL4Px9ykyi39Ovl5:b-88n2f) ([raw file](https://github.com/dgraph-io/dgraph/blob/8e829a420c51d12c45cef104f8418722d466e05e/worker/predicate.go#L29)):*
> ```Go
> 
> // Predicate gets data for a predicate p from another instance and writes it to RocksDB.
> func Predicate(p string, idx int) error {
> ```

PopulateShard(pred string, serverId int) error

---

*[worker/predicate.go, line 51 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ0X6_6nnEA7Br2zI4H:-KQ0X6_6nnEA7Br2zI4I:b-4bncu4) ([raw file](https://github.com/dgraph-io/dgraph/blob/8e829a420c51d12c45cef104f8418722d466e05e/worker/predicate.go#L51)):*
> ```Go
> 	}
> 
> 	kvs := make(chan x.KV, 10000)
> ```

1000 should be sufficient.

---

*[worker/predicate.go, line 56 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ0XecQPqDIFHzJAnbw:-KQ0XecQPqDIFHzJAnbx:bg1g1a8) ([raw file](https://github.com/dgraph-io/dgraph/blob/8e829a420c51d12c45cef104f8418722d466e05e/worker/predicate.go#L56)):*
> ```Go
> 
> 	for {
> 		b, err := stream.Recv()
> ```

what does b stand for here?

---

*[worker/predicate.go, line 68 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ0XotXSVjbioX5zug6:-KQ0XotYZJMUCyqPKwGo:b6s7wlk) ([raw file](https://github.com/dgraph-io/dgraph/blob/8e829a420c51d12c45cef104f8418722d466e05e/worker/predicate.go#L68)):*
> ```Go
> 		kv := new(task.KV)
> 		kv.Init(b.Data, uo)
> 		kvs <- x.KV{Key: kv.KeyBytes(), Val: kv.ValBytes()}
> ```

Why do you need to convert a flatbuffer KV to an x.KV? You can just directly pass the flatbuffers KV to the store. Also, store isn't the right place for that APi. It should lie closer to where it's being called from.

Ensure that the KV pointers remain valid when passed onto a channel. I'm just wondering if b would not become eligible for GC once the loop continues.

---

*[worker/worker.go, line 153 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ0ZZHiEifSMIsxXFXn:-KQ0ZZHjNnjGS2MTLj5v:b3j6bij) ([raw file](https://github.com/dgraph-io/dgraph/blob/8e829a420c51d12c45cef104f8418722d466e05e/worker/worker.go#L153)):*
> ```Go
> 	// RocksDB library we are using.
> 	// http://rocksdb.org/blog/2609/use-checkpoints-for-efficient-snapshots/
> 	dataStore.SetSnapshot()
> ```

What does snapshot do? Add a comment here to explain why you use them.

---

*[x/x.go, line 150 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ0ZfCzEchvhJl9ccbZ:-KQ0ZfCzEchvhJl9ccb_:bfi1t75) ([raw file](https://github.com/dgraph-io/dgraph/blob/8e829a420c51d12c45cef104f8418722d466e05e/x/x.go#L150)):*
> ```Go
> 
> // KV is used to represent a key value pair in rocksdb.
> type KV struct {
> ```

Doesn't look like you need this anymore.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/177)*

pawanrawal:
Review status: 11 of 15 files reviewed at latest revision, 42 unresolved discussions, some commit checks failed.

---

*[store/store.go, line 124 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPwAwhCZkQQImxhO-j4-r4-124:-KQ4oi4QUsKqdAZ0yb4Q:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/8e829a420c51d12c45cef104f8418722d466e05e/store/store.go#L124)):*
> exported const MB should have comment (or a comment on this block) or be unexported
</details>
Done.

---

*[store/store.go, line 128 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ0Vq9r4-VAHH7vLhy9:-KQ4nmi6mEZrqQ0d-sgD:bpxbll8) ([raw file](https://github.com/dgraph-io/dgraph/blob/8e829a420c51d12c45cef104f8418722d466e05e/store/store.go#L128)):*
> Also pass in a context, so you can add some tracing here. The other thing is, if this takes too long, you want to be able to cancel the operation.
</details>
Done with support for cancellation. Moved this to `worker/predicate.go`.

---

*[store/store.go, line 132 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ0WdkbgQ_zclefNEdT:-KQ3lNJBqwDTNcenAPJ9:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/8e829a420c51d12c45cef104f8418722d466e05e/store/store.go#L132)):*
> wb.Data() is going to be expensive. It returns a serialized version of the data, IIUC. Instead, do the byte count over key and val, which would be fast.
</details>
Done.

---

*[store/store.go, line 141 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ0X-QbGpVtuc0ajB0D:-KQ3lOTJpr4K-Vxb1o18:b-woogp8) ([raw file](https://github.com/dgraph-io/dgraph/blob/8e829a420c51d12c45cef104f8418722d466e05e/store/store.go#L141)):*
> use wb.Count() here, instead of Data() which is slow.
</details>
Using batchSize.

---

*[store/store.go, line 127 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPwAwhCZkQQImxhO-j4-r5-127:-KQ4pntk6-LzGWTWIZrw:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/bb4d395f3a5eb60038f31e5c1bd9a690890f4396/store/store.go#L127)):*
> exported method Store.WriteBatch should have comment or be unexported
</details>
Done.

---

*[worker/predicate.go, line 29 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ0XL4Px9ykyi39Ovl4:-KQ3YWOh2q0Es41ouR8s:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/8e829a420c51d12c45cef104f8418722d466e05e/worker/predicate.go#L29)):*
> PopulateShard(pred string, serverId int) error
</details>
Done.

---

*[worker/predicate.go, line 51 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ0X6_6nnEA7Br2zI4H:-KQ3YZR24IbhNjFdG_U3:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/8e829a420c51d12c45cef104f8418722d466e05e/worker/predicate.go#L51)):*
> 1000 should be sufficient.
</details>
Done.

---

*[worker/predicate.go, line 56 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ0XecQPqDIFHzJAnbw:-KQ3fcZe174VRR2mOZPf:b-2h6v14) ([raw file](https://github.com/dgraph-io/dgraph/blob/8e829a420c51d12c45cef104f8418722d466e05e/worker/predicate.go#L56)):*
> what does b stand for here?
</details>
It's just the payload which internally have KV flatbuffers. Changed it to reflect the name.

---

*[worker/predicate.go, line 68 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ0XotXSVjbioX5zug6:-KQ4oT_HIjSP7AwOBbsa:b-7wky9y) ([raw file](https://github.com/dgraph-io/dgraph/blob/8e829a420c51d12c45cef104f8418722d466e05e/worker/predicate.go#L68)):*
> Why do you need to convert a flatbuffer KV to an x.KV? You can just directly pass the flatbuffers KV to the store. Also, store isn't the right place for that APi. It should lie closer to where it's being called from.
> 
> Ensure that the KV pointers remain valid when passed onto a channel. I'm just wondering if b would not become eligible for GC once the loop continues.
</details>
Well bad that I didn't realize that I could just pass the flatbuffers. Doing that now. The pointers do remain valid and data is transferred to the goroutine. Even if payload is collected by the GC, kv shouldnt be right?

---

*[worker/worker.go, line 153 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ0ZZHiEifSMIsxXFXn:-KQ3fSNAu8Png-3F52NK:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/8e829a420c51d12c45cef104f8418722d466e05e/worker/worker.go#L153)):*
> What does snapshot do? Add a comment here to explain why you use them.
</details>
Done.

---

*[x/x.go, line 150 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ0ZfCzEchvhJl9ccbZ:-KQ3fZl_8JuHQei-rqsZ:bss4a2b) ([raw file](https://github.com/dgraph-io/dgraph/blob/8e829a420c51d12c45cef104f8418722d466e05e/x/x.go#L150)):*
> Doesn't look like you need this anymore.
</details>
Removed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/177)*

pawanrawal:
Realized another thing that for the snapshot and the iteration thereafter I shouldn't be using `s.db.ropt`.  I should create another read options and set snapshot for that only. Otherwise if we just mutate `s.db.ropt` it would fix affect the Get operations after taking the snapshot. Making this change.

---

Review status: 11 of 15 files reviewed at latest revision, 42 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-:-KQ599ycRjxp1sd8nsl4:b-tazhve)*

manishrjain:
Reviewed 1 of 4 files at r5, 1 of 2 files at r6, 2 of 2 files at r8.
Review status: all files reviewed at latest revision, 41 unresolved discussions, some commit checks failed.

---

*[store/store.go, line 102 \[r8\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ5mw2CoUyHiaiEnUZc:-KQ5mw2DMT1sOBXOvXhF:bqjlpoe) ([raw file](https://github.com/dgraph-io/dgraph/blob/bcfe0e5ce887ea569e626cf11f1dec56f746cf7c/store/store.go#L102)):*
> ```Go
> 	// while doing bulk scans.
> 	ro.SetFillCache(false)
> 	s.ropt.SetSnapshot(snap)
> ```

There's a race condition here. Also, if multiple callers want a snapshot, they'll end up overwriting each other. Also, why set snapshot in s.ropt?

---

*[store/store.go, line 104 \[r8\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ5mGssmR_SZZV-zNSI:-KQ5mGssmR_SZZV-zNSJ:bbgvc2t) ([raw file](https://github.com/dgraph-io/dgraph/blob/bcfe0e5ce887ea569e626cf11f1dec56f746cf7c/store/store.go#L104)):*
> ```Go
> 	s.ropt.SetSnapshot(snap)
> 	// So that we can release it later.
> 	s.snap = snap
> ```

Setting s.snap here can be overwritten by another call to GetSnapIterator() -- and you'd then end up with a snap which never gets released. Also, checking if there's already a snap present is also no good here. You'll have to return a snapshot and let the caller create a new read option.

---

*[store/store.go, line 123 \[r8\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ5nOKLNx4s2JlK6RM1:-KQ5nOKMjelk3z0BGoHY:bevovb4) ([raw file](https://github.com/dgraph-io/dgraph/blob/bcfe0e5ce887ea569e626cf11f1dec56f746cf7c/store/store.go#L123)):*
> ```Go
> 
> // ReleaseSnapshot releases a snapshot.
> func (s *Store) ReleaseSnapshot() {
> ```

Currently, your snapshots are specific to callers, not to store. So, you can't store them here or release them from here.

---

*[worker/predicate.go, line 68 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ0XotXSVjbioX5zug6:-KQ5pwr-iHt9v8KJFdn2:b-i0y873) ([raw file](https://github.com/dgraph-io/dgraph/blob/8e829a420c51d12c45cef104f8418722d466e05e/worker/predicate.go#L68)):*
> Well bad that I didn't realize that I could just pass the flatbuffers. Doing that now. The pointers do remain valid and data is transferred to the goroutine. Even if payload is collected by the GC, kv shouldnt be right?
</details>
kv is a pointer to that payload, so the payload must stay valid.

---

*[worker/predicate.go, line 42 \[r8\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ5sntXGf3fC44CX6Mc:-KQ5sntXGf3fC44CX6Md:b-gsm423) ([raw file](https://github.com/dgraph-io/dgraph/blob/bcfe0e5ce887ea569e626cf11f1dec56f746cf7c/worker/predicate.go#L42)):*
> ```Go
> 		// We write in batches of size 32MB.
> 		if batchSize >= 32*MB {
> 			if err := dataStore.WriteBatch(wb); err != nil {
> ```

Pass context in here and do a trace whenever you do a writebatch. That way we can figure out how long does it take for the writes.

---

*[worker/predicate.go, line 55 \[r8\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ5oua43n7LX0fhkbSz:-KQ5oua43n7LX0fhkbT-:b-bzyo4n) ([raw file](https://github.com/dgraph-io/dgraph/blob/bcfe0e5ce887ea569e626cf11f1dec56f746cf7c/worker/predicate.go#L55)):*
> ```Go
> 	// After channel is closed the above loop would exit, we write the data in
> 	// write batch here.
> 	if batchSize > 0 {
> ```

if batchSize > 0 { che <- dataStore.WriteBatch(wb) }
else che <- nil

---

*[worker/predicate.go, line 120 \[r8\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ5qerlV2a4oi4bM-tx:-KQ5qerm4mAiR8Go3LUO:b1kys7r) ([raw file](https://github.com/dgraph-io/dgraph/blob/bcfe0e5ce887ea569e626cf11f1dec56f746cf7c/worker/predicate.go#L120)):*
> ```Go
> 	close(kvs)
> 
> 	if err := <-che; err != nil {
> ```

x.Trace for error as well.

---

*[worker/worker.go, line 154 \[r8\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ5qr-UaAmmA_lgSK4G:-KQ5qr-UaAmmA_lgSK4H:b-7g5jbn) ([raw file](https://github.com/dgraph-io/dgraph/blob/bcfe0e5ce887ea569e626cf11f1dec56f746cf7c/worker/worker.go#L154)):*
> ```Go
> 	// http://rocksdb.org/blog/2609/use-checkpoints-for-efficient-snapshots/
> 	// Snapshots allow us to create a point-in-time view of the database.
> 	it := dataStore.GetSnapIterator()
> ```

Get snapshot, then create iterator here. Then you can release the snapshot here. But, honestly, I'm not sure if we need to worry about snapshot -- we'll have to replay all the updates from an arbitrary point anyway.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/177)*

pawanrawal:
Review status: 14 of 15 files reviewed at latest revision, 43 unresolved discussions, some commit checks failed.

---

*[store/store.go, line 102 \[r8\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ5mw2CoUyHiaiEnUZc:-KQ79kWP_Ye-6J6P2dy6:b-er44p3) ([raw file](https://github.com/dgraph-io/dgraph/blob/bcfe0e5ce887ea569e626cf11f1dec56f746cf7c/store/store.go#L102)):*
> There's a race condition here. Also, if multiple callers want a snapshot, they'll end up overwriting each other. Also, why set snapshot in s.ropt?
</details>
Right, We can probably use just iterators. 
RocksDB doc says `Short lived scans are best done via an iterator while long-running scans are better done via a snapshot.`. Our scan is a very short lived scan anyway.

---

*[store/store.go, line 104 \[r8\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ5mGssmR_SZZV-zNSI:-KQ79hBZRHpeRrOYvT_L:baw6cmz) ([raw file](https://github.com/dgraph-io/dgraph/blob/bcfe0e5ce887ea569e626cf11f1dec56f746cf7c/store/store.go#L104)):*
> Setting s.snap here can be overwritten by another call to GetSnapIterator() -- and you'd then end up with a snap which never gets released. Also, checking if there's already a snap present is also no good here. You'll have to return a snapshot and let the caller create a new read option.
</details>
Very true, we can probably use just iterators here?

---

*[store/store.go, line 123 \[r8\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ5nOKLNx4s2JlK6RM1:-KQ79aXF2p9DrGAuuCFv:b-p8fuac) ([raw file](https://github.com/dgraph-io/dgraph/blob/bcfe0e5ce887ea569e626cf11f1dec56f746cf7c/store/store.go#L123)):*
> Currently, your snapshots are specific to callers, not to store. So, you can't store them here or release them from here.
</details>
I guess we can do without snapshots and just use iterators.

---

*[worker/predicate.go, line 68 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ0XotXSVjbioX5zug6:-KQ783H_8pzB40I2Fs5W:b-b3oihq) ([raw file](https://github.com/dgraph-io/dgraph/blob/8e829a420c51d12c45cef104f8418722d466e05e/worker/predicate.go#L68)):*
> kv is a pointer to that payload, so the payload must stay valid.
</details>
Ok, but it does remain valid AFAIK.

---

*[worker/predicate.go, line 42 \[r8\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ5sntXGf3fC44CX6Mc:-KQ77i6g9hewxIwgtKMY:btn617p) ([raw file](https://github.com/dgraph-io/dgraph/blob/bcfe0e5ce887ea569e626cf11f1dec56f746cf7c/worker/predicate.go#L42)):*
> Pass context in here and do a trace whenever you do a writebatch. That way we can figure out how long does it take for the writes.
</details>
Done. Should we check for timeout here too? I guess since we already do that in the parent, we don't require it here.

---

*[worker/predicate.go, line 55 \[r8\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ5oua43n7LX0fhkbSz:-KQ7801CJ1nSi4PWnPaT:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/bcfe0e5ce887ea569e626cf11f1dec56f746cf7c/worker/predicate.go#L55)):*
> if batchSize > 0 { che <- dataStore.WriteBatch(wb) }
> else che <- nil
</details>
Done.

---

*[worker/predicate.go, line 120 \[r8\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ5qerlV2a4oi4bM-tx:-KQ795Lb7exEX8AHIP3G:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/bcfe0e5ce887ea569e626cf11f1dec56f746cf7c/worker/predicate.go#L120)):*
> x.Trace for error as well.
</details>
Done.

---

*[worker/worker.go, line 154 \[r8\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ5qr-UaAmmA_lgSK4G:-KQ76HKiqpUXhmY9ByhN:b7sc9rb) ([raw file](https://github.com/dgraph-io/dgraph/blob/bcfe0e5ce887ea569e626cf11f1dec56f746cf7c/worker/worker.go#L154)):*
> Get snapshot, then create iterator here. Then you can release the snapshot here. But, honestly, I'm not sure if we need to worry about snapshot -- we'll have to replay all the updates from an arbitrary point anyway.
</details>
hmm, yeah. 
Maybe we can just use an iterator here? Also, I guess I should pass in a timestamp to writeBatch to write on PL who ts>updateTs.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/177)*

pawanrawal:
Review status: 12 of 15 files reviewed at latest revision, 43 unresolved discussions.

---

*[store/store.go, line 123 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPwAwhCZkQQImxhO-j4-r5-123:-KQBpwk1-_RA901190Bd:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/bb4d395f3a5eb60038f31e5c1bd9a690890f4396/store/store.go#L123)):*
> exported method Store.NewWriteBatch should have comment or be unexported
</details>
Done.

---

*[store/store.go, line 102 \[r8\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ5mw2CoUyHiaiEnUZc:-KQBpkpiKLIioFZUPuh_:be02p1m) ([raw file](https://github.com/dgraph-io/dgraph/blob/bcfe0e5ce887ea569e626cf11f1dec56f746cf7c/store/store.go#L102)):*
> Right, We can probably use just iterators. 
> RocksDB doc says `Short lived scans are best done via an iterator while long-running scans are better done via a snapshot.`. Our scan is a very short lived scan anyway.
</details>
Using iterators.

---

*[store/store.go, line 104 \[r8\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ5mGssmR_SZZV-zNSI:-KQBpoCh0_KE7sInyLmu:be02p1m) ([raw file](https://github.com/dgraph-io/dgraph/blob/bcfe0e5ce887ea569e626cf11f1dec56f746cf7c/store/store.go#L104)):*
> Very true, we can probably use just iterators here?
</details>
Using iterators.

---

*[store/store.go, line 123 \[r8\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ5nOKLNx4s2JlK6RM1:-KQBprE0OUYbk7UpZ1GJ:bovqcrz) ([raw file](https://github.com/dgraph-io/dgraph/blob/bcfe0e5ce887ea569e626cf11f1dec56f746cf7c/store/store.go#L123)):*
> I guess we can do without snapshots and just use iterators.
</details>
Using iterators so this isn't needed anymore.

---

*[worker/predicate.go, line 66 \[r6\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KPwAwhCZkQQImxhO-jA-r6-66:-KQBq3DMV4_JZYCI9izS:b6a3pq) ([raw file](https://github.com/dgraph-io/dgraph/blob/530712d003de7bf3bb9842c4a977375a33cba8f4/worker/predicate.go#L66)):*
> func parameter serverId should be serverID
</details>
Not doing this because Id looks better.

---

*[worker/worker.go, line 154 \[r8\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ5qr-UaAmmA_lgSK4G:-KQBpZ-77cEWtZpFy2iF:b-wsli24) ([raw file](https://github.com/dgraph-io/dgraph/blob/bcfe0e5ce887ea569e626cf11f1dec56f746cf7c/worker/worker.go#L154)):*
> hmm, yeah. 
> Maybe we can just use an iterator here? Also, I guess I should pass in a timestamp to writeBatch to write on PL who ts>updateTs.
> 
</details>
Using iterators now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/177)*

manishrjain:
:lgtm_strong: Couple more comments, so address them before submitting. Good iterations here! Also, run this new API and ensure that the predicate shard gets copied completely. Add a task for that.

---

Reviewed 1 of 1 files at r9, 2 of 2 files at r10.
Review status: all files reviewed at latest revision, 35 unresolved discussions.

---

*[store/store.go, line 88 \[r10\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQDYtqgOG6rlGacKB_P:-KQDYtqgOG6rlGacKB_Q:b2lsids) ([raw file](https://github.com/dgraph-io/dgraph/blob/1207d7d38f7bd6959bd593b4d8bae5fe3cfdd0b4/store/store.go#L88)):*
> ```Go
> 
> // GetIterator initializes a new iterator and returns it.
> func (s *Store) GetIterator() *rocksdb.Iterator {
> ```

NewIterator() instead of GetIterator()

---

*[worker/predicate.go, line 42 \[r8\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQ5sntXGf3fC44CX6Mc:-KQDZJJMUJJ7wVkD-T-d:brkl6ei) ([raw file](https://github.com/dgraph-io/dgraph/blob/bcfe0e5ce887ea569e626cf11f1dec56f746cf7c/worker/predicate.go#L42)):*
> Done. Should we check for timeout here too? I guess since we already do that in the parent, we don't require it here.
</details>
OK. Yeah, I think you close the channel in case of timeout, so we should be able to return from here as well.

---

*[worker/predicate.go, line 42 \[r10\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQDZVv1jmWIydaZDSXl:-KQDZVv1jmWIydaZDSXm:b-ujn835) ([raw file](https://github.com/dgraph-io/dgraph/blob/1207d7d38f7bd6959bd593b4d8bae5fe3cfdd0b4/worker/predicate.go#L42)):*
> ```Go
> 		// We write in batches of size 32MB.
> 		if batchSize >= 32*MB {
> 			x.Trace(ctx, "Starting batch write.")
> ```

Instead of having two lines of tracing here, just have one called: "Doing batch write %d", where you have a counter for the number of batch writes done so far. This way, we can easily see how much data was written and how long did it take.

---

*[worker/predicate.go, line 59 \[r10\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQDZnQIHlQMhcN2E3SS:-KQDZnQIHlQMhcN2E3ST:bul9gw2) ([raw file](https://github.com/dgraph-io/dgraph/blob/1207d7d38f7bd6959bd593b4d8bae5fe3cfdd0b4/worker/predicate.go#L59)):*
> ```Go
> 	// write batch here.
> 	if batchSize > 0 {
> 		x.Trace(ctx, "Starting batch write.")
> ```

Same as above here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-:-KQD_liPy5CsuxW8WKXx:b-5xj9b)*

pawanrawal:
Yeah, have been running this API after doing each iteration and ensuring predicate is copied completely. Looking to add a test which can mock some of the worker functions and simulate this, though it might take some time. Do you think that would be good?

---

Review status: 12 of 16 files reviewed at latest revision, 35 unresolved discussions.

---

*[store/store.go, line 88 \[r10\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQDYtqgOG6rlGacKB_P:-KQFX1YLXJQU4UvCPp0K:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/1207d7d38f7bd6959bd593b4d8bae5fe3cfdd0b4/store/store.go#L88)):*
> NewIterator() instead of GetIterator()
</details>
Done.

---

*[worker/predicate.go, line 42 \[r10\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQDZVv1jmWIydaZDSXl:-KQFY4OrMeMcEoVG5TTe:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/1207d7d38f7bd6959bd593b4d8bae5fe3cfdd0b4/worker/predicate.go#L42)):*
> Instead of having two lines of tracing here, just have one called: "Doing batch write %d", where you have a counter for the number of batch writes done so far. This way, we can easily see how much data was written and how long did it take.
</details>
Done.

---

*[worker/predicate.go, line 59 \[r10\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-KQDZnQIHlQMhcN2E3SS:-KQFYA3vIUUyXMt68SMO:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/1207d7d38f7bd6959bd593b4d8bae5fe3cfdd0b4/worker/predicate.go#L59)):*
> Same as above here.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-:-KQFYEUmEJXiRec56Hbi:b-ezz1ug)*

manishrjain:
Depends on how much work it is. We could also extend our integration tests on Travis to test this. Might be easier to set up. Feel free to submit! :lgtm_strong: 

---

Reviewed 4 of 4 files at r11.
Review status: all files reviewed at latest revision, 32 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/177#-:-KQFkkGvRHkYl2Psx6L1:bdwtarn)*

PR #172, opened 2016-08-22
add basic types, sample schema and dummy schema validation skeleton

akhiltak:
ToDo (before merge to master):
 

- [x] complete validation and execution step
- [x] add tests for new methods
- [ ] squash additional commits (if any)
- [x] fix formatting issues (text editor bugs)
- [x] use git rebase while merging from master branch  
- [ ] get LGTM

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/172)

PR #185, opened 2016-08-31
Feature/supporting types graphql

akhiltak:
New PR for https://github.com/dgraph-io/dgraph/pull/172

Need to implement/discuss:
- [x] Use init() functionality added by Jay
- [x] Address the ParseType func return type (it's interface{}, which should be something stricter)
- [x] About Schema file, I am thinking it will have to be lexed using existing function and a write a parser for it. That would catch any declaration errors and form a json internally from which Schema Object Types could be constructed. Or, could there be a better way. (Not thinking directly reading a file because will have to support [] and nulltypes later and file reading code would be messy).
- [x] https://github.com/golang/go/issues/2632 (for type inference/conversion directly from a []byte)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/185)


akhiltak:
Review status: 0 of 8 files reviewed at latest revision, 4 unresolved discussions.

---

*[types/scalar_types.go, line 29 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVM_t7H_N90QRUICby-r1-29:-KQVTbSs9_kNX5fcz8c9:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/353566361685b8927e038a5dc3d52c1bd68ac06f/types/scalar_types.go#L29)):*
> exported var Int should have comment or be unexported
</details>
Done.

---

*[types/scalar_types.go, line 42 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVM_t7H_N90QRUICby-r1-42:-KQVTdnzp1kpGKalfzkP:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/353566361685b8927e038a5dc3d52c1bd68ac06f/types/scalar_types.go#L42)):*
> if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)
</details>
Done.

---

*[types/scalar_types.go, line 31 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVM_t7H_N90QRUICby-r2-31:-KQVTbEagmSipv0R8T2x:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/f62617dd944e26ca01c7ac2a45f097db66d2626c/types/scalar_types.go#L31)):*
> exported var Float should have comment or be unexported
</details>
Done.

---

*[types/scalar_types.go, line 44 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVTlw5TuKhKH4m9tzD:-KQVTlw5TuKhKH4m9tzE:b-ajkbjj) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/scalar_types.go#L44)):*
> ```Go
> // NOTE:
> // - Although, currently, input is of string type, later it will change to []byte
> func coerceInt(input string) (interface{}, error) {
> ```

NOTE: here, return type is still interface{}. After much deliberation, have thought of keeping it this way (although don't like the loose interface type, must be something stricter). This way ensures that we don't have to do type checking/switching while coercing query results leaf values.
Another way would be to have stricter return types but then type checking will be involved which will bloat the code.
@manish: Would love to hear your take on this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/185)*

manishrjain:
Reviewed 7 of 8 files at r1, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 21 unresolved discussions.

---

*[cmd/dgraph/main.go, line 100 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVd2k9IytD4pCO3JcY:-KQVd2k9IytD4pCO3JcZ:bsvi4au) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/cmd/dgraph/main.go#L100)):*
> ```Go
> 	// xidToUID is used to store ids which are not uids. It is sent to the instance
> 	// which has the xid <-> uid mapping to get uids.
> 	xidToUID := make(map[string]uint64)
> ```

You don't need to make this change. We ignore the ID related lint warnings.

---

*[cmd/dgraph/main.go, line 342 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVdRskuEkZ74EoXmn3:-KQVdRskuEkZ74EoXmn4:bphoxhm) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/cmd/dgraph/main.go#L342)):*
> ```Go
> // TODO(akhil): move out to types package and initialize there AFTER ensuring flags are initialized
> func init() {
> 	types.LoadSchema()
> ```

Shouldn't this be called from main? Something like load schema seems like a thing for main func.

---

*[types/definitions.go, line 2 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVm5ikpROBz8bvCV-d:-KQVm5ikpROBz8bvCV-e:bb6ht2g) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/definitions.go#L2)):*
> ```Go
> /*
>  * Copyright 2015 DGraph Labs, Inc.
> ```

2016

---

*[types/definitions.go, line 31 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVnJFcTwV0719XrkrH:-KQVnJFcTwV0719XrkrI:b-e80bae) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/definitions.go#L31)):*
> ```Go
> // Almost all scalar types can also act as input types.
> // Scalars (along with Enums) form leaf nodes of request or input values to arguements.
> type GraphQLScalar struct {
> ```

You can drop the GraphQL prefix from here and everywhere else. I doubt it's needed.

---

*[types/definitions.go, line 33 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVsPXNRIiatlay3Qqq:-KQVsPXNRIiatlay3Qqr:bxxld25) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/definitions.go#L33)):*
> ```Go
> type GraphQLScalar struct {
> 	Name        string // name of scalar type
> 	Description string // short description, could be used for documentation in GraphiQL
> ```

How is it used in GraphiQL?

---

*[types/definitions.go, line 38 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVmi3brjGfefa17Ywj:-KQVmi3brjGfefa17Ywk:b-ydw1x7) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/definitions.go#L38)):*
> ```Go
> 
> // ParseTypeFunc is a function type that parses and does coercion for GraphQL Scalar types.
> type ParseTypeFunc func(input string) (interface{}, error)
> ```

This should be a byte array, which is what we store in FB for values.

---

*[types/definitions.go, line 50 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVsNxbzj5teyxSMboC:-KQVsNxbzj5teyxSMboD:b4tpl5c) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/definitions.go#L50)):*
> ```Go
> type GraphQLObject struct {
> 	Name   string
> 	Desc   string
> ```

This was named Description above.

---

*[types/definitions.go, line 55 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVnYdn3QIntd_E1p-e:-KQVnYdn3QIntd_E1p-f:bljx0f9) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/definitions.go#L55)):*
> ```Go
> 
> // FieldMap maps field names to their corresponding types.
> type FieldMap map[string]GraphQLType
> ```

Think you're using the GraphQL spec terminology here. But, it would be better to bring it back to Dgraph's terminology. If the field name = attribute, then call it so.

---

*[types/definitions.go, line 64 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVo0FoqBK501UHN6qX:-KQVo0FpkUdPTAModPfP:b-3foutf) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/definitions.go#L64)):*
> ```Go
> // GraphQLList type defines a wrapper for other grapql object types.
> // Mostly used while defining object fields
> type GraphQLList struct {
> ```

Is this to represent a list of things? Do we have a need for this in Dgraph?

---

*[types/scalar_types.go, line 2 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVoHiGnnO8i-0wg-NQ:-KQVoHiGnnO8i-0wg-NR:blncpc2) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/scalar_types.go#L2)):*
> ```Go
> /*
>  * Copyright 2015 DGraph Labs, Inc.
> ```

2016 here and everywhere else.

---

*[types/scalar_types.go, line 44 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVTlw5TuKhKH4m9tzD:-KQVp3GfswP84Li1BMcS:bb5zcfp) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/scalar_types.go#L44)):*
<details><summary><i>Previously, akhiltak (Akhil Tak) wrote…</i></summary>
> NOTE: here, return type is still interface{}. After much deliberation, have thought of keeping it this way (although don't like the loose interface type, must be something stricter). This way ensures that we don't have to do type checking/switching while coercing query results leaf values.
> Another way would be to have stricter return types but then type checking will be involved which will bloat the code.
> @manish: Would love to hear your take on this.
</details>
Yeah, return type makes sense. The input should be []byte though.

---

*[types/scalar_types.go, line 78 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVpIXDYpSOvrnPXdWD:-KQVpIXDYpSOvrnPXdWE:bdtia65) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/scalar_types.go#L78)):*
> ```Go
> 
> // loadScalarTypes defines and initializes all scalar types in system and checks for errors
> func loadScalarTypes() {
> ```

This doesn't need to be in a function. This could be consts or vars global to the module.

---

*[types/schema.go, line 21 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVroO9VxEUvsrfgTpA:-KQVroO9VxEUvsrfgTpB:b-5wry3w) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/schema.go#L21)):*
> ```Go
> // GraphQLSchema declares the schema structure the GraphQL queries.
> type GraphQLSchema struct {
> 	Query    GraphQLObject
> ```

Why are queries and mutations different objects?

---

*[types/schema.go, line 30 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVt4daAs-80qjVwCEO:-KQVt4dbF0EZyRp8VhOL:byckz9x) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/schema.go#L30)):*
> ```Go
> // LoadSchema loads the schema
> func LoadSchema() {
> ```

Unnecessary space here.

---

*[types/schema.go, line 31 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVt6IT6NgDpzdNZkob:-KQVt6IT6NgDpzdNZkoc:b-m6mhqs) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/schema.go#L31)):*
> ```Go
> func LoadSchema() {
> 
> 	// load scalar types
> ```

This comment doesn't add any value.

---

*[types/schema.go, line 41 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVtLE5Lg7IAkjjRtef:-KQVtLE5Lg7IAkjjRteg:bukmszs) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/schema.go#L41)):*
> ```Go
> 		Desc: "object to represent a person type",
> 		Fields: FieldMap{
> 			"name":          String,
> ```

This should be done via a JSON schema. Because then you'll have to deal with the values here, i.e. String, Int, Float etc., which are currently Go vars.

In fact, the tests should provide this schema when loading and retrieving the data.

---

*[types/schema.go, line 50 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVs1NyQBRPtB5n2aJx:-KQVs1NyQBRPtB5n2aJy:b-tdc1zu) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/schema.go#L50)):*
> ```Go
> 
> 	// defined this field separately to let personType instantiate properly
> 	personType.Fields["friend"] = GraphQLList{HasType: personType}
> ```

`friend` would correspond to more entities, so sounds like GraphQLList here = list of entities?

---

*[types/schema.go, line 59 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVtjA4lY2339Lc2L5X:-KQVtjA4lY2339Lc2L5Y:b7w2j8a) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/schema.go#L59)):*
> ```Go
> 		},
> 	}
> 	Schema = GraphQLSchema{Query: queryType}
> ```

The `types` allocated to attributes shouldn't be different based on whether they fall in a query, or in a mutation. So, I don't see the value of this struct.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/185)*

akhiltak:
Review status: all files reviewed at latest revision, 21 unresolved discussions, some commit checks failed.

---

*[cmd/dgraph/main.go, line 100 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVd2k9IytD4pCO3JcY:-KQd8HzqHqBHxh5OICnP:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/cmd/dgraph/main.go#L100)):*
> You don't need to make this change. We ignore the ID related lint warnings.
</details>
Done.

---

*[cmd/dgraph/main.go, line 342 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVdRskuEkZ74EoXmn3:-KQ_I39ipbvXSTnUkoKC:bwnivz5) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/cmd/dgraph/main.go#L342)):*
> Shouldn't this be called from main? Something like load schema seems like a thing for main func.
</details>
Yep. Earlier was doing error handling for schema generation here (for future case, when user will provide a schema and could potentially make errors) and wanted to stop execution of main if schema had errors. Since I'm not doing that anymore (based on earlier review), will move it to main.
Should fit in nicely after Jay's init() changes.

---

*[types/definitions.go, line 2 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVm5ikpROBz8bvCV-d:-KQ_Iw_A05EW8LIvKYvz:byvvme6) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/definitions.go#L2)):*
> 2016
</details>
oops! I thought we want to keep it to 2015 (guessing that's when company was registered), so purposefully left it to 2015 (most files have this year but I guess the reason is they were created in 2015!). Changed. Sorry, didn't knew how copyright comment work, googled it up now.

---

*[types/definitions.go, line 31 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVnJFcTwV0719XrkrH:-KQ_KwqrUqIUA0RJgeOA:bq682x0) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/definitions.go#L31)):*
> You can drop the GraphQL prefix from here and everywhere else. I doubt it's needed.
</details>
Done. Wanted to denote it's for graphql spec but I guess it's not necessary. Comments would suffice.

---

*[types/definitions.go, line 33 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVsPXNRIiatlay3Qqq:-KQ_LInjXgwTgQvpCveG:b1lskza) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/definitions.go#L33)):*
> How is it used in GraphiQL?
</details>
Found it mentioned somewhere that GraphiQL pulls the description for objects into it's UI. In autocomplete, it displays objects' name and description.

---

*[types/definitions.go, line 38 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVmi3brjGfefa17Ywj:-KQd-hPF-5GcjEhiKqmf:b-n784w1) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/definitions.go#L38)):*
> This should be a byte array, which is what we store in FB for values.
</details>
Yes, but currently we are not doing type inference while storing values in FB. So, values are converted from string types and stored in FB, won't the values be incorrect if we directly infer []byte slice as an int or a float?
Easy to verify, will check it out and update accordingly.

---

*[types/definitions.go, line 50 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVsNxbzj5teyxSMboC:-KQd0ZvS3FEto97nzc3J:bsj646i) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/definitions.go#L50)):*
> This was named Description above.
</details>
Corrected.

---

*[types/definitions.go, line 55 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVnYdn3QIntd_E1p-e:-KQd0dCvpKBIjmHXt_a7:bz69u8b) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/definitions.go#L55)):*
> Think you're using the GraphQL spec terminology here. But, it would be better to bring it back to Dgraph's terminology. If the field name = attribute, then call it so.
</details>
Good advice. It will keep naming consistent. 
How about replacing 'Fields' with 'Attributes' sound(or 'Attrs' to keep it short), and 'AttrMap' for 'FieldMap'

---

*[types/definitions.go, line 64 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVo0FoqBK501UHN6qX:-KQd1fLVh4J8NtcTuUdk:b-lkty9e) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/definitions.go#L64)):*
> Is this to represent a list of things? Do we have a need for this in Dgraph?
</details>
Yes, it's to denote a list of objects. It will be used when posting list could potentially have more than one elements.
For e.g., 
{
   name
   age
   friends {
        name
   }
}

here, friends will be a list of objects in result struct. So, schema uses List to denote this. (As opposed to 'name' where attr-id combo will just return single entry in posting list)
Couldn't directly use a slice here in schema definition because we want FieldMap to have values of type GraphQLType.

---

*[types/scalar_types.go, line 2 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVoHiGnnO8i-0wg-NQ:-KQd2rzc1aqUeD19n7H2:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/scalar_types.go#L2)):*
> 2016 here and everywhere else.
</details>
Done.

---

*[types/scalar_types.go, line 44 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVTlw5TuKhKH4m9tzD:-KQd2wgVpVMJzKlj50VU:b-p5gp7p) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/scalar_types.go#L44)):*
> Yeah, return type makes sense. The input should be []byte though.
</details>
As stated above, will check the consistency of getting an int out of FB when it's stored as a string. If consistency checks out, will replace string with []byte. 

---

*[types/scalar_types.go, line 78 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVpIXDYpSOvrnPXdWD:-KQd3DS6cC7p0jMY2Ksx:b-959h3r) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/scalar_types.go#L78)):*
> This doesn't need to be in a function. This could be consts or vars global to the module.
</details>
Yes. Had put them in a function for error handling (in case when users are allowed to define schema and his/her custom types).
Assuming, we won't be doing them for now, will move it in global scope.

---

*[types/schema.go, line 21 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVroO9VxEUvsrfgTpA:-KQd4Ndb0kX0T9qoKrAt:b-en85tf) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/schema.go#L21)):*
> Why are queries and mutations different objects?
</details>
Two reasons:
1. To account for separate schema validation required in case of mutations since they change values in DB and accept input params to do so (according to graphql spec[https://facebook.github.io/relay/docs/graphql-mutations.html])

2. In our case, we don't follow graphql syntax for mutations. To account for different way of specifying mutations, will have a new kind of struct to denote it. Currently, settled for GraphQLObject as a placeholder.
Could remove it for now, what are your thoughts?

---

*[types/schema.go, line 41 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVtLE5Lg7IAkjjRtef:-KQdAuGn2S6HyTb1gyES:b-nha6gz) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/schema.go#L41)):*
> This should be done via a JSON schema. Because then you'll have to deal with the values here, i.e. String, Int, Float etc., which are currently Go vars.
> 
> In fact, the tests should provide this schema when loading and retrieving the data.
</details>
Yep, that's the plan. I am thinking the Json will just have a map like this:
(a part of schema input file)
```javascript
{  
   "actor":{  
      "name":"String",
      "age":"Int",
      "friend":[  
         "List",
         "actor"
      ]
   }
}
```
where the array for friend will be interpreted as a "list" of "actor" types.
This is my next task in the pipeline.

Earlier, I was thinking lexer/parser for this, but I guess that's be like killing an ant with a bazooka for now. :P

---

*[types/schema.go, line 50 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVs1NyQBRPtB5n2aJx:-KQd7XhRJxHV2cgVnhl1:b2nhswy) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/schema.go#L50)):*
> `friend` would correspond to more entities, so sounds like GraphQLList here = list of entities?
</details>
Yes

---

*[types/schema.go, line 59 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVtjA4lY2339Lc2L5X:-KQd7dFUMyIovn_N4gcJ:b3v7t46) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/schema.go#L59)):*
> The `types` allocated to attributes shouldn't be different based on whether they fall in a query, or in a mutation. So, I don't see the value of this struct.
</details>
As mentioned above, it's for the validation of entire query/mutation based on the specified schema. For e.g., for a mutation  in GraphQL, we validate if input arguments are present and they are of the correct types while we just have to do the later part for a query.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/185)*

manishrjain:
Reviewed 4 of 9 files at r4, 4 of 5 files at r5.
Review status: all files reviewed at latest revision, 28 unresolved discussions, some commit checks failed.

---

*[query/query.go, line 122 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQyYdMG0jEGXWZyBlyg:-KQyYdMG0jEGXWZyBlyh:boag2nx) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/query/query.go#L122)):*
> ```Go
> type SubGraph struct {
> 	Attr     string
> 	AttrType types.Type
> ```

Not a string?

---

*[query/query.go, line 151 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQyYhHd7znX0NsHFqbO:-KQyYhHd7znX0NsHFqbP:bunjwzh) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/query/query.go#L151)):*
> ```Go
> }
> 
> // postTraverse traverses the subgraph recursively and returns final result for the query
> ```

full stop at the end.

---

*[query/query.go, line 268 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQya9qXFsREdv80zlC4:-KQya9qXFsREdv80zlC5:b3b3tep) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/query/query.go#L268)):*
> ```Go
> 			m[sg.Attr] = string(val)
> 		} else {
> 			stype := sg.AttrType.(types.Scalar)
> ```

It's better to check if the type is scalar, before doing the casting. Also raise error if this is the object type.

---

*[query/query.go, line 473 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQyaTsrZLNA39N6OCik:-KQyaTssSByANJl4x0ia:bxkvsbt) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/query/query.go#L473)):*
> ```Go
> 			isDebug:  sg.isDebug,
> 			Attr:     gchild.Attr,
> 			AttrType: types.GetTypeFromSchema(gchild.Attr),
> ```

Couldn't this throw an error? Also, drop the `Get` prefix. Call it `types.SchemaType(gchild.Attr)`. Or, gql.SchemaType, if you move the code to that package.

---

*[query/query_test.go, line 559 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQyb3XVyI23vyvexxr1:-KQyb3XWiObsZxWEa6xZ:b-mwh2qw) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/query/query_test.go#L559)):*
> ```Go
> 	s := `
> 		{
> 			"_uid_": "id",
> ```

This is internal. User supplied predicate shouldn't be called `_uid_`.

---

*[query/query_test.go, line 563 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQybJyjgKVMEMTPNDbM:-KQybJykN_bUkFl-deNs:bxpvk4j) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/query/query_test.go#L563)):*
> ```Go
> 			"name": "string",
> 			"age": "int",
> 			"friend": "object",
> ```

Where do you define actor and friend? Object is just a category right -- sort of like scalar. I think the schema should define these object categories here.

---

*[query/query_test.go, line 574 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQyb04-yUKqCZLhh-LU:-KQyb04-yUKqCZLhh-LV:beb192b) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/query/query_test.go#L574)):*
> ```Go
> 
> 	// set test schema file path
> 	flag.Set("schemaF", file.Name())
> ```

Avoid ending var name with uppercase.

---

*[query/query_test.go, line 640 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQybjjyY5gqNagl7X4D:-KQybjjzdxd_jC_I0wD3:bqmd4yh) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/query/query_test.go#L640)):*
> ```Go
> 			t.Errorf("Expected type coercion to float64 for: %v\n", actorMap["survival_rate"])
> 		}
> 		friendMap := actorMap["friend"].([]interface{})
> ```

Maybe we should check the internal fields in `friend` list. And ensure that they all contain the edges that they're supposed to.

---

*[types/definitions.go, line 33 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVsPXNRIiatlay3Qqq:-KQyc0gdwXBYXBQgBPPR:b32w0x) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/definitions.go#L33)):*
<details><summary><i>Previously, akhiltak (Akhil Tak) wrote…</i></summary>
> Found it mentioned somewhere that GraphiQL pulls the description for objects into it's UI. In autocomplete, it displays objects' name and description.
</details>
OK

---

*[types/definitions.go, line 55 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVnYdn3QIntd_E1p-e:-KQycVEDxSvtvbKAfyWR:b-t9bt1w) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/definitions.go#L55)):*
<details><summary><i>Previously, akhiltak (Akhil Tak) wrote…</i></summary>
> Good advice. It will keep naming consistent. 
> How about replacing 'Fields' with 'Attributes' sound(or 'Attrs' to keep it short), and 'AttrMap' for 'FieldMap'
> 
</details>
AttrMap sounds good. But, I doubt you need to define this type at all.

---

*[types/definitions.go, line 64 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVo0FoqBK501UHN6qX:-KQycfUVNtgXztT6GCL5:b4vp8d5) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/definitions.go#L64)):*
<details><summary><i>Previously, akhiltak (Akhil Tak) wrote…</i></summary>
> Yes, it's to denote a list of objects. It will be used when posting list could potentially have more than one elements.
> For e.g., 
> {
>    name
>    age
>    friends {
>         name
>    }
> }
> 
> here, friends will be a list of objects in result struct. So, schema uses List to denote this. (As opposed to 'name' where attr-id combo will just return single entry in posting list)
> Couldn't directly use a slice here in schema definition because we want FieldMap to have values of type GraphQLType.
</details>
Get rid of this based on our discussion.

---

*[types/definitions.go, line 17 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQyZB-Dt0GbiM8JOWMr:-KQyZB-Dt0GbiM8JOWMs:b-re1sc7) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/definitions.go#L17)):*
> ```Go
>  */
> 
> package types
> ```

Shouldn't this go to gql package?

---

*[types/definitions.go, line 25 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQyZDiPI8lRm2UZu8fW:-KQyZDiPI8lRm2UZu8fX:bfv78e9) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/definitions.go#L25)):*
> ```Go
> // Type interface is the wrapper interface for all types
> type Type interface {
> 	// TODO(akhil): discuss and add a method to implement this interface (didn't want to add dummy method)
> ```

IsScalar() bool, IsObject() bool?

---

*[types/definitions.go, line 51 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQyZ-klvqAYkJfQTpn8:-KQyZ-klvqAYkJfQTpn9:b-1lnh8d) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/definitions.go#L51)):*
> ```Go
> 	Name        string
> 	Description string
> 	Fields      FieldMap
> ```

Attributes map[string]Type

---

*[types/schema.go, line 31 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQydTM38wUaACvQ5GMp:-KQydTM38wUaACvQ5GMq:b-rtoxgu) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/schema.go#L31)):*
> ```Go
> var (
> 	schema  = make(map[string]Type)
> 	schemaF = flag.String("schemaF", "../../types/schema.json", "Path to the file that specifies schema in json format")
> ```

Avoid upper case in var names, and in flag names.

---

*[types/schema.go, line 39 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQydfJhBVDBkPjf6vKm:-KQydfJhBVDBkPjf6vKn:bunjwzh) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/schema.go#L39)):*
> ```Go
> }
> 
> // LoadSchema loads the schema and checks for errors
> ```

full stop at the end.

---

*[types/schema.go, line 41 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQyddW6EwtZb_wC3K2Z:-KQyddW6EwtZb_wC3K2_:bo29poj) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/schema.go#L41)):*
> ```Go
> // LoadSchema loads the schema and checks for errors
> func LoadSchema() {
> ```

Unnecessary vertical space.

---

*[types/schema.go, line 46 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQydhm2SX0GIf7gRJ_i:-KQydhm2SX0GIf7gRJ_j:b-uzrfnh) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/schema.go#L46)):*
> ```Go
> 		log.Fatalf("Schema load error:%v", err)
> 	}
> 	sch := make(map[string]string)
> ```

schema := 

---

*[types/schema.go, line 73 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQyg0n2ZKCKZuPVmEQm:-KQyg0n3eRY-0f0aunFe:b-6gh3tu) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/schema.go#L73)):*
> ```Go
> 			schema[k] = idType
> 		case "object":
> 			schema[k] = objectType
> ```

We shouldn't have an object type like this. The user should define the fields for the predicates.

---

*[types/schema.json, line 11 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQygIxhz9KhsIQI0UL2:-KQygIxit2FcUUJ3mNck:bcr097f) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/schema.json#L11)):*
> ```JSON
> 	"actor": "object",
> 	"friend": "object"
> }
> ```

This is for testing right? Then push this within schema_test.go using a string.

---

*[types/schema_test.go, line 45 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQygVKM2B_pZu-50yT0:-KQygVKM2B_pZu-50yT1:bmd1ibx) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/schema_test.go#L45)):*
> ```Go
> // TestLoadSchema tests schema reading and parsing from input schema file
> func TestLoadSchema(t *testing.T) {
> ```

Vertical space.

---

*[types/schema_test.go, line 57 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQygamS5Lk6V-do75bM:-KQygamS5Lk6V-do75bN:b57j36h) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/schema_test.go#L57)):*
> ```Go
> 
> 	// load schema from json file
> 	LoadSchema()
> ```

You could take a filepath argument here. Then you wouldn't have to set the flag.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/185)*

akhiltak:
Review status: all files reviewed at latest revision, 27 unresolved discussions, some commit checks failed.

---

*[query/query.go, line 122 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQyYdMG0jEGXWZyBlyg:-KQzH4WVzSm-oqs6LyhN:be4570e) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/query/query.go#L122)):*
> Not a string?
</details>
No, to directly do type coercion using the Type. 
One negative aspect of this I could think could be allocation size of `SubGraph` will increase since it will store a type object. Is that a big concern?

---

*[query/query.go, line 268 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQya9qXFsREdv80zlC4:-KQzHcDbURWsr7Zu4lYi:boknta0) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/query/query.go#L268)):*
> It's better to check if the type is scalar, before doing the casting. Also raise error if this is the object type.
</details>
thanks, that would be the go way (type assertion) and more stringent error handling!

---

*[query/query.go, line 473 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQyaTsrZLNA39N6OCik:-KQzIFIRl3BJ3f_61qAp:b7ph1ap) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/query/query.go#L473)):*
> Couldn't this throw an error? Also, drop the `Get` prefix. Call it `types.SchemaType(gchild.Attr)`. Or, gql.SchemaType, if you move the code to that package.
</details>
No, as we discussed, we are not stopping the execution if client asks for entities whose types are not defined in the schema. We will just ignore them and log the error. Please take a look at `SchemaType(...)`.

---

*[query/query_test.go, line 559 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQyb3XVyI23vyvexxr1:-KQzJRAHVMkOLn3e_vjd:b-c57ysc) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/query/query_test.go#L559)):*
> This is internal. User supplied predicate shouldn't be called `_uid_`.
</details>
Okay, had included them for testing the ID type conversion, if required.
How about using xid instead, or will just remove it since it's similar to string for now.

---

*[query/query_test.go, line 563 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQybJyjgKVMEMTPNDbM:-KQzK4WwPI0jOlfQBw9O:bq0bkri) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/query/query_test.go#L563)):*
> Where do you define actor and friend? Object is just a category right -- sort of like scalar. I think the schema should define these object categories here.
</details>
As we discussed, keeping just the object category for now and not the more composite actors/friends type objects. Will do that in next iteration because the current system does not relies on or checks for the presence of specific object fields (which in turn become predicates), we keep global predicate types.
Please let me know if you think we should have those specific definitions before we start using them in the next step (to have a blueprint present).

---

*[query/query_test.go, line 574 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQyb04-yUKqCZLhh-LU:-KQzLbY6F3YeJ-yLk9Z6:b25laaf) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/query/query_test.go#L574)):*
> Avoid ending var name with uppercase.
</details>
noted, thanks!

---

*[query/query_test.go, line 640 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQybjjyY5gqNagl7X4D:-KQzNv_KAn_G6kl8Lgha:b-nrb7ax) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/query/query_test.go#L640)):*
> Maybe we should check the internal fields in `friend` list. And ensure that they all contain the edges that they're supposed to.
</details>
As mentioned earlier, this would come when we have composite types like person/actor/friend defined. Then, we will check if the predicates following friend in the query belong to edges defined in `friendType` in schema.

---

*[types/definitions.go, line 55 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVnYdn3QIntd_E1p-e:-KQzS_IcpUOiJEyEEt_P:b-b5qb3w) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/definitions.go#L55)):*
> AttrMap sounds good. But, I doubt you need to define this type at all.
</details>
Just wanted to keep it a little modular. :P

---

*[types/definitions.go, line 17 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQyZB-Dt0GbiM8JOWMr:-KQzPTPk3BCoaE00A5rr:b-sbnxq3) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/definitions.go#L17)):*
> Shouldn't this go to gql package?
</details>
Had kept this package separate based on our initial discussion that in starting, we would keep the type/schema system separate and then move it if we feel like it.
I also think that since we will be developing type/schema system more from Dgrpah's perspective and not strictly adhering to the GraphQL spec (but more as a guidelines), it would be okay to keep it in a separate module.
Please let me know if you think it would still serve us better to classify type system under gql package, will move it.

---

*[types/definitions.go, line 25 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQyZDiPI8lRm2UZu8fW:-KQzkfLY_8V4tv3etX_H:b-9cx57o) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/definitions.go#L25)):*
> IsScalar() bool, IsObject() bool?
</details>
Wow, that a great idea. It will work as interface implementation method as well as for assertions (in tests).
But if I use `IsScalar() bool, IsObject() bool`, will have to implement both methods (and prob. more) in all structs that implement this interface which will be too much.
Rather, it gives me an idea to use `OfType(string) bool`. How about it?

---

*[types/schema.go, line 31 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQydTM38wUaACvQ5GMp:-KQzTjuedsqJDdkiu7s5:b-3xpemo) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/schema.go#L31)):*
> )
>  
Hmm, can't seem to think of a better descriptive name, can't just name it `schema` since we already denote actual schema with that name internally. 

Will keep this in mind about naming in future but for now, `sfile` seems like closest to what I want the flag to mean here (but won't that break camel-case convention?)
Also, we already have flags like `postingDir` and `workerPort` which follow camel-case and not the lower-case convention.

Sorry about this question, but asking this because I have come to believe in last few years that names should be descriptive to quickly and clearly indicate what values they carry. 
But I figure golang wants code to be as clean and quickly readable as possible, which is why shorter names are encouraged?
Have been caught between these differing views and had a hard time naming stuff in this PR. :P
So, would be great to know your further take on this, thanks!

---

*[types/schema.go, line 46 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQydhm2SX0GIf7gRJ_i:-KQzWaift4jg1XnCFSK-:b-rc3exh) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/schema.go#L46)):*
> schema := 
</details>
Oops, again caught better Go's shorter names vs longer descriptive names! `schema` is already a global variable, so would just name it `s`.

---

*[types/schema.go, line 73 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQyg0n2ZKCKZuPVmEQm:-KQz_nj4TrNOrfrDNPzT:bmwvqr8) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/schema.go#L73)):*
> We shouldn't have an object type like this. The user should define the fields for the predicates.
</details>
Again, that would be when the composite objects are defined in schema. Let's have a quick call about this, where I could explain what I have in mind and we could go over it to iron out inconsistencies. Please let me know if that would be possible?

---

*[types/schema.json, line 11 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQygIxhz9KhsIQI0UL2:-KQzc0QAYxLfcdkFESV0:bwtwuqn) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/schema.json#L11)):*
> This is for testing right? Then push this within schema_test.go using a string.
</details>
This is the default file for schema in case no input file is being provided using the flag. 
In the test, already pushing schema with a string.

---

*[types/schema_test.go, line 57 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQygamS5Lk6V-do75bM:-KQzcNvxgQWO_TsdZJ2Y:b-etplko) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/schema_test.go#L57)):*
> You could take a filepath argument here. Then you wouldn't have to set the flag.
</details>
Yep, I was indeed taking filepath as argument in my first implementation. But that wouldn't work with Jay's init() functionality (func AddInit() there only accepts non-parameter functions)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/185)*

manishrjain:
Getting closer.

---

Reviewed 6 of 6 files at r6.
Review status: 4 of 9 files reviewed at latest revision, 25 unresolved discussions, some commit checks broke.

---

*[query/query.go, line 268 \[r7\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KR2nWwHHTFxoKs1SB0f:-KR2nWwHHTFxoKs1SB0g:bhexspj) ([raw file](https://github.com/dgraph-io/dgraph/blob/9b2ef3c6c7e16fc8271138d55dfb51ec50fafc09/query/query.go#L268)):*
> ```Go
> 			m[sg.Attr] = string(val)
> 		} else {
> 			// the values should always be of scalar types here, do type assertion
> ```

Form proper sentences.

---

*[query/query_test.go, line 563 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQybJyjgKVMEMTPNDbM:-KR1n7vQA2GRiBFLBwFm:b-84at88) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/query/query_test.go#L563)):*
<details><summary><i>Previously, akhiltak (Akhil Tak) wrote…</i></summary>
> As we discussed, keeping just the object category for now and not the more composite actors/friends type objects. Will do that in next iteration because the current system does not relies on or checks for the presence of specific object fields (which in turn become predicates), we keep global predicate types.
> Please let me know if you think we should have those specific definitions before we start using them in the next step (to have a blueprint present).
</details>
`object` to me, isn't a type. It's a category. If this is for the next PR, then make it part of the next PR, and delete from here. Otherwise, convert it to a predicate-scalar map or something.

---

*[query/query_test.go, line 640 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQybjjyY5gqNagl7X4D:-KR1nR-g3Wz77kR-oAAi:bdrqig5) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/query/query_test.go#L640)):*
<details><summary><i>Previously, akhiltak (Akhil Tak) wrote…</i></summary>
> As mentioned earlier, this would come when we have composite types like person/actor/friend defined. Then, we will check if the predicates following friend in the query belong to edges defined in `friendType` in schema.
</details>
Remove this for now then.

---

*[types/definitions.go, line 31 \[r3\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQVnJFcTwV0719XrkrH:-KR1oBxdFM0tNwO2MPWZ:b2mmke0) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c853075c50bbaf3f74753588f6a49ed7ac2e411/types/definitions.go#L31)):*
<details><summary><i>Previously, akhiltak (Akhil Tak) wrote…</i></summary>
> Done. Wanted to denote it's for graphql spec but I guess it's not necessary. Comments would suffice.
</details>
Also, if you move this to gql, then `gql.Scalar`, `gql.Schema`, etc.

---

*[types/definitions.go, line 17 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQyZB-Dt0GbiM8JOWMr:-KR1nVoPAvaCCOHrYgCx:b-p03uqg) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/definitions.go#L17)):*
<details><summary><i>Previously, akhiltak (Akhil Tak) wrote…</i></summary>
> Had kept this package separate based on our initial discussion that in starting, we would keep the type/schema system separate and then move it if we feel like it.
> I also think that since we will be developing type/schema system more from Dgrpah's perspective and not strictly adhering to the GraphQL spec (but more as a guidelines), it would be okay to keep it in a separate module.
> Please let me know if you think it would still serve us better to classify type system under gql package, will move it.
</details>
types is already a package, though internal to posting package. I think this can just lie within gql itself. I doubt we can use this in other languages, at least based on current understanding.

---

*[types/definitions.go, line 25 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQyZDiPI8lRm2UZu8fW:-KR1nycfOxe0gPOHybXD:b-mrm1ux) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/definitions.go#L25)):*
<details><summary><i>Previously, akhiltak (Akhil Tak) wrote…</i></summary>
> Wow, that a great idea. It will work as interface implementation method as well as for assertions (in tests).
> But if I use `IsScalar() bool, IsObject() bool`, will have to implement both methods (and prob. more) in all structs that implement this interface which will be too much.
> Rather, it gives me an idea to use `OfType(string) bool`. How about it?
</details>
OfType string requies one to pass a string, which can be anything. If you want to keep it to just one function, then just use IsScalar() for now.

---

*[types/definitions.go, line 42 \[r6\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KR1oLBkC-bOlxiDFgSw:-KR1oLBkC-bOlxiDFgSx:b-yeu00r) ([raw file](https://github.com/dgraph-io/dgraph/blob/5a21cc17885295117958a2ccb6ca352ca92d0779/types/definitions.go#L42)):*
> ```Go
> // String function to implement string interface
> func (s Scalar) String() string {
> 	return fmt.Sprint(s.Name)
> ```

What is this function for? 

---

*[types/definitions.go, line 47 \[r6\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KR1oVBbg2f872yF8G7K:-KR1oVBbg2f872yF8G7L:byv2m7q) ([raw file](https://github.com/dgraph-io/dgraph/blob/5a21cc17885295117958a2ccb6ca352ca92d0779/types/definitions.go#L47)):*
> ```Go
> // OfType function to assert scalar type
> func (s Scalar) OfType(name string) bool {
> 	return name == "scalar"
> ```

yeah, that isn't very intuitive to the caller.

---

*[types/definitions.go, line 61 \[r6\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KR1oemtRNZ84VWY9TgF:-KR1oemtRNZ84VWY9TgG:b-98qzwv) ([raw file](https://github.com/dgraph-io/dgraph/blob/5a21cc17885295117958a2ccb6ca352ca92d0779/types/definitions.go#L61)):*
> ```Go
> // String function to implement string interface
> func (o Object) String() string {
> 	return fmt.Sprint(o.Name)
> ```

o.Name is already a string.

---

*[types/schema.go, line 31 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQydTM38wUaACvQ5GMp:-KR1orv0lFaRP_vPnk96:b6epy4x) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/schema.go#L31)):*
<details><summary><i>Previously, akhiltak (Akhil Tak) wrote…</i></summary>
> > )
> >  
> Hmm, can't seem to think of a better descriptive name, can't just name it `schema` since we already denote actual schema with that name internally. 
> 
> Will keep this in mind about naming in future but for now, `sfile` seems like closest to what I want the flag to mean here (but won't that break camel-case convention?)
> Also, we already have flags like `postingDir` and `workerPort` which follow camel-case and not the lower-case convention.
> 
> Sorry about this question, but asking this because I have come to believe in last few years that names should be descriptive to quickly and clearly indicate what values they carry. 
> But I figure golang wants code to be as clean and quickly readable as possible, which is why shorter names are encouraged?
> Have been caught between these differing views and had a hard time naming stuff in this PR. :P
> So, would be great to know your further take on this, thanks!
</details>
Why not schemaFile internally, and just schema for the flag name. That's descriptive, but without the ending capital case.

instanceIdx is the only flag name which has camelcase. And that needs to be fixed. In fact, it should just be idx.

---

*[types/schema.go, line 73 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQyg0n2ZKCKZuPVmEQm:-KR1pNX71EEX7xjna1he:b1snrtk) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/schema.go#L73)):*
<details><summary><i>Previously, akhiltak (Akhil Tak) wrote…</i></summary>
> Again, that would be when the composite objects are defined in schema. Let's have a quick call about this, where I could explain what I have in mind and we could go over it to iron out inconsistencies. Please let me know if that would be possible?
</details>
Sure, we can have a chat. But, the first PR can just be about the scalar types, as was the initial plan anyway. So, let's just make it about them.

---

*[types/schema.json, line 11 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQygIxhz9KhsIQI0UL2:-KR1pYXFmyIcYwDR8xFZ:b-d71d4q) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/schema.json#L11)):*
<details><summary><i>Previously, akhiltak (Akhil Tak) wrote…</i></summary>
> This is the default file for schema in case no input file is being provided using the flag. 
> In the test, already pushing schema with a string.
</details>
We shouldn't set default values, except possibly for internal predicates, but those are internal and so they don't need to be checked via schema. So, maybe we don't need this file. This can go as an example in wiki.

---

*[types/schema_test.go, line 57 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQygamS5Lk6V-do75bM:-KR1prYgcemAPTJV_G2i:b-3p3fsv) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/schema_test.go#L57)):*
<details><summary><i>Previously, akhiltak (Akhil Tak) wrote…</i></summary>
> Yep, I was indeed taking filepath as argument in my first implementation. But that wouldn't work with Jay's init() functionality (func AddInit() there only accepts non-parameter functions)
</details>
Why didn't we just call `LoadSchema` in main? That's how we do everything else.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-:-KR1q-THUk4PGTpg2Z3B:b-b3xrxp)*

akhiltak:
Review status: 4 of 9 files reviewed at latest revision, 24 unresolved discussions, some commit checks broke.

---

*[types/definitions.go, line 42 \[r6\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KR1oLBkC-bOlxiDFgSw:-KR3aH1T6CKD815-OdwC:ba8uwbi) ([raw file](https://github.com/dgraph-io/dgraph/blob/5a21cc17885295117958a2ccb6ca352ca92d0779/types/definitions.go#L42)):*
> What is this function for? 
</details>
to implement string interface{}, so that while printing types, just their names are printed and not the entire struct.

---

*[types/schema.json, line 11 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQygIxhz9KhsIQI0UL2:-KR3bCckH34vFoXsFzdv:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/schema.json#L11)):*
> We shouldn't set default values, except possibly for internal predicates, but those are internal and so they don't need to be checked via schema. So, maybe we don't need this file. This can go as an example in wiki.
</details>
Done.

---

*[types/schema_test.go, line 57 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KQygamS5Lk6V-do75bM:-KR3YMw0KGNv9AMbkTV2:bf73n4h) ([raw file](https://github.com/dgraph-io/dgraph/blob/5d9ccc091719d08a1ae034366aeea46823548a48/types/schema_test.go#L57)):*
> Why didn't we just call `LoadSchema` in main? That's how we do everything else.
</details>
Was doing that earlier but moved it out to make it a part of package init() func. Reverted my code to initial state.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/185)*

manishrjain:
Reviewed 11 of 12 files at r8, 1 of 1 files at r9.
Review status: all files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[cmd/dgraph/main.go, line 65 \[r9\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KR7YW1mUx8xBMXAnjOo:-KR7YW1mUx8xBMXAnjOp:b-wtpoa1) ([raw file](https://github.com/dgraph-io/dgraph/blob/ad7cb7481ba797dde69bc0572ddf5b8aaa46b504/cmd/dgraph/main.go#L65)):*
> ```Go
> 	nomutations = flag.Bool("nomutations", false, "Don't allow mutations on this server.")
> 	tracing     = flag.Float64("trace", 0.5, "The ratio of queries to trace.")
> 	schemaFile  = flag.String("schema", "", "Path to the file that specifies schema in json format")
> ```

hmm.. wouldn't the loader need this? If so, this should lie in the package where it's used.

---

*[gql/schema.go, line 65 \[r9\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KR7ZYjXphp8z4wUuyih:-KR7ZYjYFR6mVJMpJjWL:b9wvip0) ([raw file](https://github.com/dgraph-io/dgraph/blob/ad7cb7481ba797dde69bc0572ddf5b8aaa46b504/gql/schema.go#L65)):*
> ```Go
> 	v, present := schema[p]
> 	if !present {
> 		x.Trace(ctx, "Schema does not have type definition for:%v", p)
> ```

This is a very simplistic function. This doesn't even need to be a function. Also, the trace adds little value here, because most predicates won't have an type def.

---

*[query/query.go, line 490 \[r9\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KR7_8SAPhVm2_cXf8aL:-KR7_8SBblVpp3Oy48HN:bc0nrd7) ([raw file](https://github.com/dgraph-io/dgraph/blob/ad7cb7481ba797dde69bc0572ddf5b8aaa46b504/query/query.go#L490)):*
> ```Go
> 			isDebug:  sg.isDebug,
> 			Attr:     gchild.Attr,
> 			AttrType: gql.SchemaType(ctx, gchild.Attr),
> ```

The context is being passed just for SchemaType, which is a map lookup. I don't think this adds much value. Remove this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/185)*

akhiltak:
Review status: all files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[cmd/dgraph/main.go, line 65 \[r9\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KR7YW1mUx8xBMXAnjOo:-KR7c0ZKnUp9S7hslua_:beuwb4k) ([raw file](https://github.com/dgraph-io/dgraph/blob/ad7cb7481ba797dde69bc0572ddf5b8aaa46b504/cmd/dgraph/main.go#L65)):*
> hmm.. wouldn't the loader need this? If so, this should lie in the package where it's used.
</details>
This is just being used here in the main func.
I took you advice. So, now I am calling LoadSchema from main instead of init().

---

*[gql/schema.go, line 65 \[r9\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KR7ZYjXphp8z4wUuyih:-KR7cOuJG4ZpmhgJHR3p:b-x0ly7w) ([raw file](https://github.com/dgraph-io/dgraph/blob/ad7cb7481ba797dde69bc0572ddf5b8aaa46b504/gql/schema.go#L65)):*
> This is a very simplistic function. This doesn't even need to be a function. Also, the trace adds little value here, because most predicates won't have an type def.
</details>
Made this a function to keep `schema` as a package variable and not pollute global scope with more variables.
Earlier, had logged this warning in logs but you said we should not be doing it but use Trace instead.

I want this warning to be logged somewhere, even though I know most predicates won't have type def. As a client, it would be good to know if you missed specifying a `predicate->type` mapping in your schema file for and edge that points to a value (and not an entity).
What do you suggest? Please let me know. :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/185)*

manishrjain:
Reviewed 2 of 12 files at r8, 3 of 3 files at r11.
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[gql/schema.go, line 60 \[r11\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KR8HldhPIpf0Paw3Zrh:-KR8HldhPIpf0Paw3Zri:bbzp9rf) ([raw file](https://github.com/dgraph-io/dgraph/blob/9d3188bc9fddad14c24daaf94a9d3e86f4550f7b/gql/schema.go#L60)):*
> ```Go
> // SchemaType fetches types for a predicate from schema map
> func SchemaType(p string) Type {
> 	return schema[p]
> ```

what happens if p isn't inside the schema map?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/185)*

akhiltak:
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[gql/schema.go, line 60 \[r11\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KR8HldhPIpf0Paw3Zrh:-KR8MlJjMyOvIEywilA-:bfi0wy9) ([raw file](https://github.com/dgraph-io/dgraph/blob/9d3188bc9fddad14c24daaf94a9d3e86f4550f7b/gql/schema.go#L60)):*
> what happens if p isn't inside the schema map?
</details>
the return value will be set to nil, and will be ignored later when we validate for types (e.g., for non scalar type predicates) 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/185)*

akhiltak:
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[gql/schema.go, line 60 \[r11\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-KR8HldhPIpf0Paw3Zrh:-KR8RPK_iFxeQksP7_fR:b-bq559z) ([raw file](https://github.com/dgraph-io/dgraph/blob/9d3188bc9fddad14c24daaf94a9d3e86f4550f7b/gql/schema.go#L60)):*
<details><summary><i>Previously, akhiltak (Akhil Tak) wrote…</i></summary>
> the return value will be set to nil, and will be ignored later when we validate for types (e.g., for non scalar type predicates) 
</details>
I check for `nil` values when I validate types in query package

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/185)*

manishrjain:
:lgtm: Thanks for all the iterations! Feel free to submit.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/185#-:-KR8aai8VgwEPrdxRC8G:b8oey31)*

akhiltak:
Got LGTM. Closing as merged to master.

query/query.go
```diff
@@ -260,7 +260,22 @@ func postTraverse(sg *SubGraph) (map[uint64]interface{}, error) {
 		if sg.GetUid || sg.isDebug {
 			m["_uid_"] = fmt.Sprintf("%#x", q.Uids(i))
 		}
-		m[sg.Attr] = string(val)
+		if sg.AttrType == nil {
+			// No type defined for attr in type system/schema, hence return string value
+			m[sg.Attr] = string(val)
+		} else {
+			// type assertion for scalar type values
+			if !sg.AttrType.IsScalar() {
```
rstorr, line 268: @codelingobot capture Check Type Before Casting

rstorr:
@CodeLingoBot capture Check Type Before Casting

rstorr:
@codelingobot capture Avoid Ending Flag Arg Name In UpperCase

rstorr:
@CodeLingoBot capture Avoid Ending Flag Arg Name In UpperCase

rstorr:
@Codelingobot capture Don't Sprint values that are already strings

rstorr:
@CodeLingoBot capture Don't Sprint values that are already strings

rstorr:
@codelingobot capture comments end in a full stop

rstorr:
@CodeLingoBot capture comments end in a full stop

PR #191, opened 2016-09-03
Feature/index3

jchiu0:
Added index directory and x/serialize.go. Everything else should be found in the other branch feature/index2, so do not review, e.g., algo directory. Thanks.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/191)


manishrjain:
Okay, reviewed a lot of code, not all. The general feeling I have is, the code is very cleanly written, but you can easily get rid of a bunch of files which contain little logic and a bunch of concepts like multiple indexers; while making the whole thing a lot more readable. I see a tendency to go towards indirection -- you should try to keep that tendency in check. It's better to write something concrete now, and then generalize later; than to generalize first.

My advice to you is to *Design for the future, then implement it with today's concrete use cases.*

---

Reviewed 6 of 34 files at r1.
Review status: 6 of 18 files reviewed at latest revision, 29 unresolved discussions, some commit checks failed.

---

*[cluster/predicates_test.go, line 34 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmk-NoO1UPg2bejJP3:-KQmk-NpWx3AdVuQZDk6:b-p5zfyb) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/cluster/predicates_test.go#L34)):*
> ```Go
> func TestGetPredicate(t *testing.T) {
> 	attr := "friends"
> 	uidB, err := x.EncodeUint64(12345)
> ```

why not uid? It's encoded, but it's still a uid. And you have no conflicts here with any other uid.

---

*[cluster/predicates_test.go, line 39 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmjnY58-FGEUj2Mh6h:-KQmjnY58-FGEUj2Mh6i:bp2wbqo) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/cluster/predicates_test.go#L39)):*
> ```Go
> 		return
> 	}
> 	buf := bytes.NewBufferString(attr + "|")
> ```

attr + "|" => this would end up creating a whole new string, copying their contents into it. Which negates the whole point of using bytes.Buffer. Instead do:

bytes.NewBufferString(attr)
bytes.WriteRune('|')
bytes.Write(uidB)

---

*[index/config.go, line 33 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmkas2NhfKPIkXjETU:-KQmkas3bz5aEQxt9zgD:b-vzbowp) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/config.go#L33)):*
> ```Go
> var (
> 	// Every directory containing index data will have this file.
> 	defaultConfigFile = flag.String("index_config", "config.json",
> ```

Why this defaut prefix, if this is a flag? Also, maybe just call the flag string, `indexconfig`, because we haven't used underscores in all the other flags.

---

*[index/config.go, line 40 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmkqOlryUrfXVquEBv:-KQmkqOlryUrfXVquEBw:b-2pjm2g) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/config.go#L40)):*
> ```Go
> type Configs struct {
> 	Cfg     []*Config `json:"Config"`
> 	Indexer string    // Which indexer to use, e.g., memtable.
> ```

I think this should be less about the exact indexer to use, and more about what tokenizer to use. We probably would only have one indexer, but we'd need many tokenizers to handle sub strings, exact strings, dates, etc.

---

*[index/config.go, line 72 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmlD1H2KRnKfAYr7Rl:-KQmlD1H2KRnKfAYr7Rm:bbi1dr6) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/config.go#L72)):*
> ```Go
> 
> func (c *Configs) validate() error {
> 	// TODO(jchiu): Add more checks here in the future.
> ```

Be a bit more specific.

// TODO: This needs more checks to ensure that x, y and z are valid.

---

*[index/config.go, line 87 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmlhT4GnLFP7RdsZ8A:-KQmlhT4GnLFP7RdsZ8B:bp1zsub) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/config.go#L87)):*
> ```Go
> // storing an internal copy of the given config, so that we get rid of any
> // extraneous fields and do some formatting / normalization.
> func (c *Configs) write(dir string) error {
> ```

I don't see the point of this function. If we're able to parse the user provided config and validate it, then why do we need to store it again?

---

*[index/index.go, line 133 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQhxqw6o9UvUX10RZ0C-r1-133:-KQmhs3MlpD2GOVs0g1K:b-wdwknx) ([raw file](https://github.com/dgraph-io/dgraph/blob/2c99af217527260083708738b8df2cf895cba032/index/index.go#L133)):*
> comment on exported method Indices.Backfill should be of the form "Backfill ..."
</details>
You don't need this divider.

---

*[index/index.go, line 47 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmkLbZSSU1ORtuZgUO:-KQmkLbZSSU1ORtuZgUP:b-4fnz1e) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/index.go#L47)):*
> ```Go
> type Indices struct {
> 	dir        string
> 	cfg        *Configs
> ```

So you have access to an entire Configs struct, and yet you need to have the pred below? Couldn't the Configs struct give you the specific Config, given a predicate?

---

*[index/lookup.go, line 23 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmhgY51zCs91xxpA3y:-KQmhgY51zCs91xxpA3z:b-uw368b) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/lookup.go#L23)):*
> ```Go
> )
> 
> // LookupSpec defines a index query. Given a value, what are the matching keys.
> ```

If you implement all the suggestions I have down here, you'd realize that you don't really need this file. It's just a simple function that can lie along with the other adds and deletes etc.

---

*[index/lookup.go, line 36 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmfmzHAMlO-6-fqg9B:-KQmfmzHAMlO-6-fqg9C:b-q5a6cq) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/lookup.go#L36)):*
> ```Go
> 
> // Lookup does a lookup using underlying Indexer object.
> func (s *Indices) Lookup(li *LookupSpec) *LookupResult {
> ```

I think your code would be more readable if you just use the 2 vars directly in the query and result. If it goes beyond 3 vars, that's when it makes sense to define a struct to hold them.

---

*[index/lookup.go, line 37 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmg0iHtfjgac7Xtp9r:-KQmg0iIzuHYqLDODbP7:b-xpst9) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/lookup.go#L37)):*
> ```Go
> // Lookup does a lookup using underlying Indexer object.
> func (s *Indices) Lookup(li *LookupSpec) *LookupResult {
> 	if s == nil {
> ```

This could have been:

if s == nil {
  return []uint64, x.Errorf("Indices is nil")
}

Just more easily readable IMHO.

---

*[index/lookup.go, line 42 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmgTksRZ82l3RHPW7X:-KQmgTksRZ82l3RHPW7Y:biyzlh8) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/lookup.go#L42)):*
> ```Go
> 		}
> 	}
> 	r, err := s.indexer.Query(li.Attr, li.Value)
> ```

why not s.Query?

---

*[index/lookup.go, line 48 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmgZsz9HKXShADulXJ:-KQmgZt-c3tqrq8hGUjr:bjgxrfm) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/lookup.go#L48)):*
> ```Go
> 	// Extract UIDs from array of strings r.
> 	uid := make([]uint64, len(r))
> 	for i, k := range r {
> ```

There are 2 slow things here:

1. Using channel to communicate uint64 results, when you can just return a []uint64 directly from the function. A channel would have been useful, if multiple goroutines were involved in this.
2. Decoding it. This should ideally be stored in uint64 format, if possible. Though, I reckon you're doing this because the underlying data in RocksDB is stored in bytes?

---

*[index/worker.go, line 40 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmfJGxnJvS1PioKGMB:-KQmfJGyQz9tJ6Sqg5xu:b-kvxurq) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/worker.go#L40)):*
> ```Go
> 
> // WorkerLookup looks up the indices and push the result to the given channel.
> func WorkerLookup(li *LookupSpec, results chan *LookupResult) {
> ```

I don't think there's enough logic here to justify a new file called worker. We have workers in Dgraph, because they do the heavy job of network communication and provide network APIs etc. This is just wrapping one function with another function.

---

*[index/worker.go, line 46 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmf31_TV6aIl_Fw0MX:-KQmf31_TV6aIl_Fw0MY:by0bld8) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/worker.go#L46)):*
> ```Go
> 
> // FrontfillAdd inserts with overwrite (replace) key, value into our indices.
> func FrontfillAdd(ctx context.Context, attr string, uid uint64, val string) error {
> ```

I think you can avoid the prefix `Frontfill`. It can just be Add and Del.

---

*[index/indexer/indexer.go, line 32 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQminlHuzh9eM_bebkI:-KQminlHuzh9eM_bebkJ:b-jsy0k) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/indexer/indexer.go#L32)):*
> ```Go
> // Register adds an Indexer constructor to registry. If you need to call this
> // function, we recommend you do it in your package init.
> func Register(name string, f func() Indexer) {
> ```

Why do we allow multiple indexers?

---

*[index/indexer/indexer.go, line 52 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmiD0Ngb7CbrmZCM9z:-KQmiD0ONxsf-vLlspyX:bv96a59) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/indexer/indexer.go#L52)):*
> ```Go
> 	Size() int
> 	Reset()
> 	Insert(pred, key, val string) error
> ```

It would help if you used a bit more specific terminologies here. For e..g, I think your key is a UID. Then call it so. And make it uint64. Your val is an indexing term, then call it so.

---

*[index/indexer/indexer.go, line 53 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmiU3Q5vcSmLfV2OPq:-KQmiU3Q5vcSmLfV2OPr:brhrjx9) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/indexer/indexer.go#L53)):*
> ```Go
> 	Reset()
> 	Insert(pred, key, val string) error
> 	Remove(pred, key string) error
> ```

Okay, now I feel your key is the indexing term... Again would help the readability of your code if you were to use specific terms.

---

*[index/indexer/indexer.go, line 65 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmcaOgrPQudrlYYQrJ:-KQmcaOhuEuEwoFu6RBH:beyd7lj) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/indexer/indexer.go#L65)):*
> ```Go
> 
> 	// Updates.
> 	Insert(pred, key, val string) error
> ```

We can simplify the indexer interface by only having batches of updates. If someone needs to do only one insert or update, they can create a batch holding just one.

---

*[index/indexer/indexer.go, line 68 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmcneVfHQej3fmKb-I:-KQmcneVfHQej3fmKb-J:b86u3p2) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/indexer/indexer.go#L68)):*
> ```Go
> 	Remove(pred, key string) error
> 	NewBatch() (Batch, error)
> 	Batch(b Batch) error
> ```

Apply(batch) error

---

*[posting/list.go, line 36 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmRzmvD7TQcRUBAzJL:-KQmRzmvD7TQcRUBAzJM:b-vuh9te) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/posting/list.go#L36)):*
> ```Go
> 	"github.com/dgraph-io/dgraph/store"
> 	"github.com/dgraph-io/dgraph/x"
> 	"github.com/dgryski/go-farm"
> ```

I think the grouping was: standard Go libs, then external, then internal.

---

*[posting/list.go, line 130 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmaU6GSCphvkFXswC7:-KQmaU6GSCphvkFXswC8:b-v3ao5f) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/posting/list.go#L130)):*
> ```Go
> 	attr = attr[:len(attr)-1]
> 	var uid uint64
> 	err = binary.Read(buf, binary.LittleEndian, &uid)
> ```

This logic seems flawed. binary.Read would read from the beginning of buf, which is the attribute.

---

*[x/serialize.go, line 25 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmakO01k_78KUU7fld:-KQmakO01k_78KUU7fle:b-5hfskt) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/x/serialize.go#L25)):*
> ```Go
> 
> // EncodeUint64 encodes value using little endian.
> func EncodeUint64(v uint64) ([]byte, error) {
> ```

I didn't know that big endian encoding causes lexicographical sorting of integers. If that's right, then we can just use big endian everywhere. We don't need to use little endian.

---

*[x/serialize.go, line 44 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmQic9l4cYIgRsEpN5:-KQmQic9l4cYIgRsEpN6:b5iw6ir) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/x/serialize.go#L44)):*
> ```Go
> 
> // EncodeUint64Ordered encodes value using big endian. String ordering will imply
> // that the uint64s sorted.
> ```

that the uint64s 'are' sorted.

---

*[x/serialize.go, line 53 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmQ2rsb-Q5sYPw80yb:-KQmQ2rsb-Q5sYPw80yc:bcxq0mm) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/x/serialize.go#L53)):*
> ```Go
> }
> 
> // DecodeUint64Ordered decodes value using little endian.
> ```

big endian.

---

*[x/serialize_test.go, line 5 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmQSyzRYx7TOuR9fLk:-KQmQSyzRYx7TOuR9fLl:b1qlsbs) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/x/serialize_test.go#L5)):*
> ```Go
> import (
> 	//"bytes"
> 	//	"fmt"
> ```

remove these two lines?

---

*[x/serialize_test.go, line 23 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmQa5ihcHGshi8oCHi:-KQmQa5ihcHGshi8oCHj:b-w5a25f) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/x/serialize_test.go#L23)):*
> ```Go
> 		}
> 		if w != v {
> 			t.Errorf("Encode . Decode not identity %d %v %d", v, b, w)
> ```

Grammar?

---

*[x/serialize_test.go, line 43 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmQcY6_73mx5I153GZ:-KQmQcY6_73mx5I153G_:b-1kzc9c) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/x/serialize_test.go#L43)):*
> ```Go
> 		}
> 		if w != v {
> 			t.Error("Encode . Decode not identity %d %v %d", v, b, w)
> ```

And here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-:-KQmlwW38NDg-cyxQqc_:b-7ztag6)*

jchiu0:
I think the Indexer interface is a good way for us to swap out different underlying implementation later on. We discussed earlier that there would be some Indexer later on, but we could not decide which one yet.

Although it is extra code, it feels like a very clean way to delineate our codebase from this incoming Indexer implementation. I think of it as a scaffold, like if I want to build a beautiful mechanism, we can first decide where the bones are, then fill in the details later. Why not design for the future, and sketch it? I can try to further reduce the footprint of Indexer, but I think it is good to have it now.

It may be discomforting to see a lot of these interfaces lying around. I will refrain from doing it. And it might help to have more diagrams, more documentation as well.

Also, indexer.go is actually very short. I don't think it adds that much more. Maybe the registry more makes it look really bloated. Without it, it is really just a few lines:

```
type Batch interface {
	Size() int
	Reset()
	Insert(pred, key, val string) error
	Remove(pred, key string) error
}
type Indexer interface {
	Open(dir string) error
	Close() error
	Create(dir string) error
	NewBatch() (Batch, error)
	Apply(b Batch) error
	Query(pred, val string) ([]string, error)
}
```

---

Review status: 6 of 18 files reviewed at latest revision, 29 unresolved discussions, some commit checks failed.

---

*[cluster/predicates_test.go, line 34 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmk-NoO1UPg2bejJP3:-KQp3KOavQ-fp7Mbj1x4:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/cluster/predicates_test.go#L34)):*
> why not uid? It's encoded, but it's still a uid. And you have no conflicts here with any other uid.
</details>
Done.

---

*[cluster/predicates_test.go, line 39 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmjnY58-FGEUj2Mh6h:-KQp3LNzTR2GdroduJ3G:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/cluster/predicates_test.go#L39)):*
> attr + "|" => this would end up creating a whole new string, copying their contents into it. Which negates the whole point of using bytes.Buffer. Instead do:
> 
> bytes.NewBufferString(attr)
> bytes.WriteRune('|')
> bytes.Write(uidB)
</details>
Done.

---

*[index/config.go, line 33 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmkas2NhfKPIkXjETU:-KQoaJnT-HOqRx_onDw0:bmuy4r3) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/config.go#L33)):*
> Why this defaut prefix, if this is a flag? Also, maybe just call the flag string, `indexconfig`, because we haven't used underscores in all the other flags.
</details>
I rather this be a constant because it is for internal use only, but you suggested having a flag in an earlier review (branch feature/index). Can I change it back to a constant?

---

*[index/config.go, line 40 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmkqOlryUrfXVquEBv:-KQoacO6Nofzi54p8-xc:b-tblboy) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/config.go#L40)):*
> I think this should be less about the exact indexer to use, and more about what tokenizer to use. We probably would only have one indexer, but we'd need many tokenizers to handle sub strings, exact strings, dates, etc.
</details>
The indexer can be Bleve or we build our own or this very simple "memtable". I wanted to be able to switch it out easily for another implementation.

---

*[index/config.go, line 72 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmlD1H2KRnKfAYr7Rl:-KQoavIVvip4drgyU6r0:bv6sw9v) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/config.go#L72)):*
> Be a bit more specific.
> 
> // TODO: This needs more checks to ensure that x, y and z are valid.
</details>
Removed the TODO. It used to be the case that you need to specify the type of field here and number of child indices, which we would want to validate. But these are gone now.

---

*[index/config.go, line 87 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmlhT4GnLFP7RdsZ8A:-KQob3kn9FkQ-M1TPl3e:b-xuo3b5) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/config.go#L87)):*
> I don't see the point of this function. If we're able to parse the user provided config and validate it, then why do we need to store it again?
</details>
Let me explain the way indices are created and updated, in this implementation.

Indices <-> directory. The directory contains a config file which describes the indexed preds, and also perhaps some additional state information. We don't have that yet, but we might in the future.

index.NewIndices(dir) is the function that opens up the config file in this directory and initializes everything according to the config. It however does not do any backfilling. The user has to explicitly call index.Backfill to start this potentially time-consuming process.

However, for a Bleve index (just take this as an example), before you can open the object, you need to make an additional "create" call. Hence, when this code is first written, we need the user to explicitly call "index.CreateIndices" as a one-off. I don't really like this extra step. It would be nicer to have something like "NewOrCreateIndices". Back to the story, so when you call CreateIndices, we currently copy your config to the given directory, hence the need for this function.

Just to further explain my thinking: I thought it might be nice to make the config in the index directory totally internal to Dgraph. The user will create CreateIndices on a config file somewhere else, and we will "merge" it into the internal config that we have. This is one way to allow users to modify the indices config. In retrospect, it is not that great either. It might be nicer if the server just exposes some endpoints that can modify the indices.

Let's chat about this. One possibility is that the index directory will have a separate state file which describes the columns that have been indexed. If file is missing, then nothing is indexed. The user can modify the config file as he likes. Remove the function CreateIndices. Have only NewIndices(dir). This will check the internal state file and the config file. If there is no state file or if the state file says "totally uninitialized", then we will ask Bleve index (or any other index) to initialize itself for the first time.

---

*[index/index.go, line 103 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQhxqw6o9UvUX10RZ0C-r1-103:-KQp3ZFD3xTQv6zTHGA2:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/2c99af217527260083708738b8df2cf895cba032/index/index.go#L103)):*
> comment on exported method Indices.FrontfillAdd should be of the form "FrontfillAdd ..."
</details>
Done.

---

*[index/index.go, line 133 \[r1\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQhxqw6o9UvUX10RZ0C-r1-133:-KQp3XDzHczXxOVlERHs:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/2c99af217527260083708738b8df2cf895cba032/index/index.go#L133)):*
> You don't need this divider.
</details>
Done.

---

*[index/index.go, line 47 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmkLbZSSU1ORtuZgUO:-KQofgG3dc6xmNuMQnaN:b-hgm5uj) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/index.go#L47)):*
> So you have access to an entire Configs struct, and yet you need to have the pred below? Couldn't the Configs struct give you the specific Config, given a predicate?
</details>
Configs and Config should be named JSONConfigs and JSONConfig. They are there for JSON to do its work. I am not sure if we can put a map inside JSONConfigs. It might be possible. The JSON might look like:

"type.name.object.en": {
  blah blah blah
}

Let me take a look at JSON parser.

---

*[index/lookup.go, line 23 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmhgY51zCs91xxpA3y:-KQp3aQuNo9ONA098ksC:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/lookup.go#L23)):*
> If you implement all the suggestions I have down here, you'd realize that you don't really need this file. It's just a simple function that can lie along with the other adds and deletes etc.
</details>
Done.

---

*[index/lookup.go, line 36 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmfmzHAMlO-6-fqg9B:-KQogihegCJrO4Vj7GrA:b-txwx5j) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/lookup.go#L36)):*
> I think your code would be more readable if you just use the 2 vars directly in the query and result. If it goes beyond 3 vars, that's when it makes sense to define a struct to hold them.
</details>
Sorry, the code used to be more complicated with child indices, pred indices, some channels here and there. These are some remnants of it. I will clear them up.

---

*[index/lookup.go, line 37 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmg0iHtfjgac7Xtp9r:-KQp3cH9IjfXkzVau2Zl:b-sjzzoe) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/lookup.go#L37)):*
> This could have been:
> 
> if s == nil {
>   return []uint64, x.Errorf("Indices is nil")
> }
> 
> Just more easily readable IMHO.
</details>
Done. (Just forgot to clean up completely)

---

*[index/lookup.go, line 42 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmgTksRZ82l3RHPW7X:-KQp1ygYcAvaWMdAudzQ:bpay4po) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/lookup.go#L42)):*
> why not s.Query?
</details>
Forgot to clean up. There used to be channels here.

---

*[index/lookup.go, line 48 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmgZsz9HKXShADulXJ:-KQp24hrOTL4TaSOS3AD:b64b29d) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/lookup.go#L48)):*
> There are 2 slow things here:
> 
> 1. Using channel to communicate uint64 results, when you can just return a []uint64 directly from the function. A channel would have been useful, if multiple goroutines were involved in this.
> 2. Decoding it. This should ideally be stored in uint64 format, if possible. Though, I reckon you're doing this because the underlying data in RocksDB is stored in bytes?
</details>
(1) Yes, you are right. Not sure why it was there. I think there used to be one channel per child index or per predicate. This is removed.
(2) The underlying data in the Indexer (which can be Bleve or anything else) is assumed to be []byte.

---

*[index/worker.go, line 40 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmfJGxnJvS1PioKGMB:-KQp2ZC5lc4OP1NhChu-:bfc6txi) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/worker.go#L40)):*
> I don't think there's enough logic here to justify a new file called worker. We have workers in Dgraph, because they do the heavy job of network communication and provide network APIs etc. This is just wrapping one function with another function.
</details>
This is more like a stub for future work. Let me just remove it for now.

---

*[index/worker.go, line 46 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmf31_TV6aIl_Fw0MX:-KQp3EKwzPHBDDuX4-2k:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/worker.go#L46)):*
> I think you can avoid the prefix `Frontfill`. It can just be Add and Del.
</details>
Done.

---

*[index/indexer/indexer.go, line 32 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQminlHuzh9eM_bebkI:-KQp3jrM6wMyZ1XkE4hk:btyx98c) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/indexer/indexer.go#L32)):*
> Why do we allow multiple indexers?
</details>
I thought we agree that while we're still not sure if we want to build our own Indexer or use Bleve or some other solution, let's just put something in place. That way I chose to do this is to have some interface in place, as some kind of scaffold.

---

*[index/indexer/indexer.go, line 52 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmiD0Ngb7CbrmZCM9z:-KQp3yTHVuithHFxYT0p:bqgnnw9) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/indexer/indexer.go#L52)):*
> It would help if you used a bit more specific terminologies here. For e..g, I think your key is a UID. Then call it so. And make it uint64. Your val is an indexing term, then call it so.
</details>
I was thinking of a generic Indexer here and the point of having this interface is that when we have some new Indexer, we can think from its perspective and easily hook it up to the rest of the codebase without thinking about something as specific as UIDs.

---

*[index/indexer/indexer.go, line 53 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmiU3Q5vcSmLfV2OPq:-KQp4EpWtVfQWbL72LQP:b92eayg) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/indexer/indexer.go#L53)):*
> Okay, now I feel your key is the indexing term... Again would help the readability of your code if you were to use specific terms.
</details>
I could perhaps write some more comments about what keys and values mean for an Indexer.

---

*[index/indexer/indexer.go, line 65 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmcaOgrPQudrlYYQrJ:-KQp4cvdaJaNBPBHcMaD:btr2nsb) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/indexer/indexer.go#L65)):*
> We can simplify the indexer interface by only having batches of updates. If someone needs to do only one insert or update, they can create a batch holding just one.
</details>
Done. Updated tests and some other code.

---

*[index/indexer/indexer.go, line 68 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmcneVfHQej3fmKb-I:-KQp816C8nbMlbiElOmT:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/index/indexer/indexer.go#L68)):*
> Apply(batch) error
</details>
Done.

---

*[posting/list.go, line 36 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmRzmvD7TQcRUBAzJL:-KQp89GRlJ7C35QbxFXq:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/posting/list.go#L36)):*
> I think the grouping was: standard Go libs, then external, then internal.
</details>
Done.

---

*[posting/list.go, line 130 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmaU6GSCphvkFXswC7:-KQp8SSvHTwXrx9oFgKd:btu4c45) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/posting/list.go#L130)):*
> This logic seems flawed. binary.Read would read from the beginning of buf, which is the attribute.
</details>
The tests passed, in particular TestDecodeKey in list_test.go.

I think buf (bytes.NewBuffer) does have a read pointer.

---

*[x/serialize.go, line 25 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmakO01k_78KUU7fld:-KQp8kwoYrE1fs47kr7R:b-yz4qk0) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/x/serialize.go#L25)):*
> I didn't know that big endian encoding causes lexicographical sorting of integers. If that's right, then we can just use big endian everywhere. We don't need to use little endian.
</details>
Let me just confirm: You would like me to change to big endian in this PR?

---

*[x/serialize.go, line 44 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmQic9l4cYIgRsEpN5:-KQp91S3RtuLF_F_X79J:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/x/serialize.go#L44)):*
> that the uint64s 'are' sorted.
</details>
Done.

---

*[x/serialize.go, line 53 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmQ2rsb-Q5sYPw80yb:-KQp94iH5mrnYm9Bj3Nn:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/x/serialize.go#L53)):*
> big endian.
</details>
Done.

---

*[x/serialize_test.go, line 5 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmQSyzRYx7TOuR9fLk:-KQp98dqfCOnKXDHjhrc:b-qsx7u6) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/x/serialize_test.go#L5)):*
> remove these two lines?
</details>
Done. Oops.

---

*[x/serialize_test.go, line 23 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmQa5ihcHGshi8oCHi:-KQp9BUXPDbV02KRqS3i:b-j3z5uc) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/x/serialize_test.go#L23)):*
> Grammar?
</details>
"." is supposed to mean function composition..

---

*[x/serialize_test.go, line 43 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-KQmQcY6_73mx5I153GZ:-KQp9K-_3nW5bdPOasZ_:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/97b66a4220a8a1f2ba17b1a24268d2285ea921e9/x/serialize_test.go#L43)):*
> And here.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/191#-:-KQp9ME9cdPKyfoKxY_1:b-nmdad6)*

PR #196, opened 2016-09-06
Feature/indexpl

jchiu0:
Use posting lists to create indices.

We are working mostly on the backend, modifying packages "worker", "posting" and the flatbuffer "Query".

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/196)


manishrjain:
Review status: 0 of 24 files reviewed at latest revision, 19 unresolved discussions, some commit checks broke.

---

*[posting/list.go, line 432 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR1jrdytLKxgF84uvCh:-KR1jrdytLKxgF84uvCi:b-2m15oc) ([raw file](https://github.com/dgraph-io/dgraph/blob/042da9f74f763d0c17c477e01623ec6d4130ec71/posting/list.go#L432)):*
> ```Go
> // NOTE: We have a tricky situation that can easily lead to deadlocks. Caller of
> // mergeMutation is addMutation which locks the current list l while we are
> // inside mergeMutation. If mergeMutation calls addMutation via a channel or any
> ```

mergeMutation should only work on itself. Not on other PLs. If you're doing that, you're doing something wrong.

---

*[posting/list.go, line 456 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR1k1RNLlPMUEz7du2C:-KR1k1RNLlPMUEz7du2D:bvqesbb) ([raw file](https://github.com/dgraph-io/dgraph/blob/042da9f74f763d0c17c477e01623ec6d4130ec71/posting/list.go#L456)):*
> ```Go
> 	mlink.posting = mp
> 
> 	indexJobs := make([]*indexJob, 0, 2)
> ```

This is very internal to PL. You should not be modifying the code here.

---

*[posting/list.go, line 715 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR1l-DkhsQ8TDoECZEU:-KR1l-DlRIniUMWVakCC:b-nhfi57) ([raw file](https://github.com/dgraph-io/dgraph/blob/042da9f74f763d0c17c477e01623ec6d4130ec71/posting/list.go#L715)):*
> ```Go
> 
> 	mpost := x.NewPosting(mbuf)
> 	indexJobs := l.mergeMutation(mpost)
> ```

You're calling merge on every AddMutation? That's wrong.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/196)*

jchiu0:
Review status: 0 of 28 files reviewed at latest revision, 35 unresolved discussions, some commit checks failed.

---

*[posting/list.go, line 432 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR1jrdytLKxgF84uvCh:-KR2C4syfdN2zQ5Iey24:b-n3rvkl) ([raw file](https://github.com/dgraph-io/dgraph/blob/042da9f74f763d0c17c477e01623ec6d4130ec71/posting/list.go#L432)):*
> mergeMutation should only work on itself. Not on other PLs. If you're doing that, you're doing something wrong.
</details>
Done. (Was making the problem more complicated than it was.)

---

*[posting/list.go, line 715 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR1l-DkhsQ8TDoECZEU:-KR2CCzkEBgM2yB7JqTx:bugjgiy) ([raw file](https://github.com/dgraph-io/dgraph/blob/042da9f74f763d0c17c477e01623ec6d4130ec71/posting/list.go#L715)):*
> You're calling merge on every AddMutation? That's wrong.
</details>
AddMutation has always been calling l.mergeMutation.

https://github.com/dgraph-io/dgraph/blob/master/posting/list.go#L610

I believe the merge here means merging to mutation layers. Maybe we should rename this function.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/196)*

manishrjain:
Really happy to see how simple this has become! A few more iterations and we should be there.

---

Reviewed 19 of 24 files at r1, 1 of 2 files at r2, 7 of 8 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 57 unresolved discussions, some commit checks broke.

---

*[cmd/dgraph/main.go, line 369 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3GcwZarce6mhdrng0:-KR3Gcw_KgTumsCun1NI:ba6kdi2) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/cmd/dgraph/main.go#L369)):*
> ```Go
> 
> 	posting.InitIndex(ps)
> 	defer posting.CloseIndex()
> ```

Just an FYI: Expect this to never be called.

---

*[posting/index.go, line 36 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3SVLZOZlss2H7ew9M:-KR3SVLZOZlss2H7ew9N:bgm3wde) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L36)):*
> ```Go
> 	// Posting list keys are prefixed with this byte if it is a mutation meant for
> 	// the index.
> 	indexByte = ':'
> ```

think this is a `rune`.

---

*[posting/index.go, line 51 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3SyN2cxbtA_EzQSdi:-KR3SyN3CeqWSjI5QO-p:bgv5vcx) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L51)):*
> ```Go
> 	x.AddInit(func() {
> 		x.Assert(indexConfigFile != nil && len(*indexConfigFile) > 0)
> 		f, err := ioutil.ReadFile(*indexConfigFile)
> ```

if indexConfigFile is being used here for init, define it in this file as well. *UPDATE*: Just saw that you're doing this below. Move them before any functions.

---

*[posting/index.go, line 80 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3VeTvuStd-mJjHYJu:-KR3VeTwrCWe06pMGDTt:b-fnr1hp) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L80)):*
> ```Go
> }
> 
> var (
> ```

Ah.. just saw this. Move this to the top.

---

*[posting/index.go, line 95 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3W-N90obO6H02Wc2i:-KR3W-N90obO6H02Wc2j:b-9cvfm3) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L95)):*
> ```Go
> 	}
> 	indexStore = ds
> 	indexJobC = make(chan indexJob, 100)
> ```

Make it a 1000, so it doesn't get blocked.

---

*[posting/index.go, line 102 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3W67cwmFJxMbNn6Hq:-KR3W67cwmFJxMbNn6Hr:b-83sjo4) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L102)):*
> ```Go
> func CloseIndex() {
> 	close(indexJobC)
> 	<-indexDone
> ```

Nice.

---

*[posting/index.go, line 105 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3_Ka8S2Gjmm7Zi01R:-KR3_Ka8S2Gjmm7Zi01S:b-tfho5g) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L105)):*
> ```Go
> }
> 
> func IndexKey(attr string, value []byte) []byte {
> ```

(attr string, term []byte)

---

*[posting/index.go, line 110 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3ZKwlDLKysePUJXPl:-KR3ZKwlDLKysePUJXPm:bngcbd) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L110)):*
> ```Go
> 	_, err := buf.WriteString(attr)
> 	x.Check(err)
> 	_, err = buf.WriteString("|")
> ```

buf.WriteRune('|')?

---

*[posting/index.go, line 119 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3_eDb9PHce-5xVhJV:-KR3_eDb9PHce-5xVhJW:boi714g) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L119)):*
> ```Go
> // processIndexJob consumes and processes jobs from indexJobC. It can create new
> // posting lists and will add new mutations.
> func processIndexJob() {
> ```

This would make a great candidate for trace's eventlog. You can create one inside this, and then as you iterate over the jobs, you can add some statements there.

---

*[posting/index.go, line 120 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3_ria3J-C7ErGsfqH:-KR3_ria3J-C7ErGsfqI:b-b7kf3q) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L120)):*
> ```Go
> // posting lists and will add new mutations.
> func processIndexJob() {
> 	for job := range indexJobC {
> ```

`for job := range indexJobs` is more intuitive. Also, avoid ending var names with a capital case.

---

*[posting/index.go, line 130 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3a-Hw0BS9vtQynTGP:-KR3a-Hw0BS9vtQynTGQ:b-37ghr0) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L130)):*
> ```Go
> 		x.Assertf(plist != nil, "plist is nil [%s] %d %s", key, edge.ValueId, edge.Attribute)
> 		if job.del {
> 			plist.AddMutation(context.Background(), edge, Del)
> ```

Maybe you can just create a context.Background() once before loop starts, and reuse it.

---

*[posting/index.go, line 140 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3aFeBgllKvuHiN8vL:-KR3aFeBgllKvuHiN8vM:b-qds4uy) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L140)):*
> ```Go
> // AddMutationWithIndex is AddMutation with support for indexing.
> func (l *List) AddMutationWithIndex(ctx context.Context, t x.DirectedEdge, op byte) error {
> 	doUpdateIndex := t.Value != nil &&
> ```

This should never be called with indexer keys. So, instead of doUpdateIndex, this should be a check. All the indexer keys should be called with `AddMutation()`.

---

*[posting/index.go, line 142 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3ae5Ks-f8GQaW51UE:-KR3ae5Ks-f8GQaW51UF:bl8vdav) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L142)):*
> ```Go
> 	doUpdateIndex := t.Value != nil &&
> 		(len(l.key) > 0 && l.key[0] != indexByte) && indexedAttr[l.keyAttr]
> 	var lastPost *types.Posting
> ```

This doesn't need to be a pointer.

---

*[posting/index.go, line 146 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3aXqMLX5zcYJVkPUX:-KR3aXqMLX5zcYJVkPUY:b-kwoesg) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L146)):*
> ```Go
> 		// Check last posting for original value BEFORE any mutation actually happens.
> 		if l.Length() >= 1 {
> 			lastPost = new(types.Posting)
> ```

You don't need to allocate the memory on a heap.

---

*[posting/index.go, line 147 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3agaVVSbDGOBG7Klw:-KR3agaVVSbDGOBG7Klx:btsuddr) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L147)):*
> ```Go
> 		if l.Length() >= 1 {
> 			lastPost = new(types.Posting)
> 			x.Assert(l.Get(lastPost, l.Length()-1))
> ```

... `l.Get(&lastPost, l.Length() -1)`

---

*[posting/index.go, line 161 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3b4g2wk_PeuISOStJ:-KR3b4g3QoIP1FzfFwnW:b-9hae7t) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L161)):*
> ```Go
> 
> 	if lastPost != nil && lastPost.ValueBytes() != nil {
> 		l.indexDel(lastPost.ValueBytes())
> ```

You can call l.AddMutation(..) here. AddMutation is pretty fast, so we can just do it here.

---

*[posting/index.go, line 164 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3bCGmGzuNr83aK55b:-KR3bCGmGzuNr83aK55c:bapp3ch) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L164)):*
> ```Go
> 	}
> 	if op == Set {
> 		l.indexAdd(t.Value)
> ```

Just write to channel here? You don't need the extra one-liner functions.

---

*[posting/index.go, line 171 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3bQq8YMxcn9e9EhhE:-KR3bQq8YMxcn9e9EhhF:b41umhu) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L171)):*
> ```Go
> // indexAdd adds to jobs an indexJob that is an insert.
> func (l *List) indexAdd(newValue []byte) {
> 	if indexJobC != nil {
> ```

if indexJobC was nil, you can avoid all the logic to generate the data. So, check this during AddMutationWithIndex.

---

*[posting/list.go, line 715 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR1l-DkhsQ8TDoECZEU:-KR3cTdqnk0yNRIOkDOJ:b-fw24hs) ([raw file](https://github.com/dgraph-io/dgraph/blob/042da9f74f763d0c17c477e01623ec6d4130ec71/posting/list.go#L715)):*
> AddMutation has always been calling l.mergeMutation.
> 
> https://github.com/dgraph-io/dgraph/blob/master/posting/list.go#L610
> 
> I believe the merge here means merging to mutation layers. Maybe we should rename this function.
</details>
Arrr.. I see. My bad! yeah, we should call the push to rocksdb, `commit` or something.

---

*[posting/list.go, line 70 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3botVTEPGzXZ68Wim:-KR3botWoHkyrkQiJfB-:b-o2hf5i) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/list.go#L70)):*
> ```Go
> 	wg          sync.WaitGroup
> 	deleteMe    int32
> 	keyAttr     string // Might start with ":" for index-related mutations.
> ```

Why do you need these 2 fields?

---

*[posting/list.go, line 127 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3eWO2vygwpTHTyvKT:-KR3eWO2vygwpTHTyvKU:bb76mu4) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/list.go#L127)):*
> ```Go
> 
> // DecodeKey deserializes the uid and attr.
> func DecodeKey(b []byte) (uint64, string) {
> ```

Why?

---

*[posting/list.go, line 260 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3dZJg833p_qx5UZr4:-KR3dZJhSHz-94m8pP0k:b8tfl65) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/list.go#L260)):*
> ```Go
> 	err := clog.StreamEntries(posting.CommitTs()+1, l.hash,
> 		func(hdr commit.Header, buffer []byte) {
> 			m := x.NewPosting(buffer)
> ```

The difference between writing m := new(types.Posting) here, v/s creating that memory in another function, is where the memory gets allocated. In the first case, if m isn't being used outside this func, it would be put on stack, which is very cheap. In the latter case, the memory would be put on heap, because it needs to be passed outside the creator function. Given the number of entries can be large, we can end up with a lot of memory allocated on heap really fast, which the GC would then have to sweep through.

I had a similar case in rdf.Parse, which significantly impacted the memory and cpu profiling.

In this case, we're going to get rid of this code anyway. So, nothing much to do here, but something to keep in mind in the future. Memory allocation is tricky business, so do it carefully.

---

*[posting/list.go, line 286 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3c4aZk8o4qDx2Cpru:-KR3c4aZk8o4qDx2Cprv:b-36bv1w) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/list.go#L286)):*
> ```Go
> 			nbuf.d = make([]byte, len(empty))
> 			// C API does one copy to move from std::string to C char pointer.
> 			// We do another copy here.
> ```

Not really. Read the code again, I'll wait :-).

---

*[posting/list.go, line 489 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3ceY1StpM0wiQYHks:-KR3ceY2pAurWWZrqx6P:bppf47i) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/list.go#L489)):*
> ```Go
> 				if ok := plist.Postings(&cp, pi); ok {
> 					if samePosting(&cp, mp) {
> 						// POTENTIAL BUG: What if you add a mutation which is valid first,
> ```

I'm not sure if a comment is the best way to address bugs. Either bring it up for a discussion on communication channels or write a test case to prove it or deny it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-:-KR3edqn4d5Rg9kCJI7I:b-lybvxx)*

jchiu0:
Thanks for pointing out how useful "Get" is. Somehow I thought it only returns the underlying PL from database without checking the code.

---

Review status: 17 of 28 files reviewed at latest revision, 59 unresolved discussions, some commit checks broke.

---

*[cmd/dgraph/main.go, line 369 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3GcwZarce6mhdrng0:-KR44R41ZEj0GqAmzEYW:b-91j3es) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/cmd/dgraph/main.go#L369)):*
> Just an FYI: Expect this to never be called.
</details>
No more channel.

---

*[posting/index.go, line 36 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3SVLZOZlss2H7ew9M:-KR44yEgGTIe6nvLc-yO:b-we8vcv) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L36)):*
> think this is a `rune`.
</details>
Done. Renamed.

---

*[posting/index.go, line 51 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3SyN2cxbtA_EzQSdi:-KR45DT3Wa0eaZIp7OkH:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L51)):*
> if indexConfigFile is being used here for init, define it in this file as well. *UPDATE*: Just saw that you're doing this below. Move them before any functions.
</details>
Done.

---

*[posting/index.go, line 80 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3VeTvuStd-mJjHYJu:-KR45EMvHihZZS4-CQyT:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L80)):*
> Ah.. just saw this. Move this to the top.
</details>
Done.

---

*[posting/index.go, line 95 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3W-N90obO6H02Wc2i:-KR45I-539FNNkafOsYi:baoaj9s) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L95)):*
> Make it a 1000, so it doesn't get blocked.
</details>
Done. No more channel.

---

*[posting/index.go, line 105 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3_Ka8S2Gjmm7Zi01R:-KR45WRNybMKLkAQ-lTn:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L105)):*
> (attr string, term []byte)
</details>
Done.

---

*[posting/index.go, line 110 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3ZKwlDLKysePUJXPl:-KR45XNYyjZVEoxw3CUv:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L110)):*
> buf.WriteRune('|')?
</details>
Done.

---

*[posting/index.go, line 119 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3_eDb9PHce-5xVhJV:-KR6oN3xh5JAWPWdJRp7:b-c2b7jo) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L119)):*
> This would make a great candidate for trace's eventlog. You can create one inside this, and then as you iterate over the jobs, you can add some statements there.
</details>
Added event log but not too sure what's the best way to use it. See processIndexJob..

---

*[posting/index.go, line 120 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3_ria3J-C7ErGsfqH:-KR45jQvCfsQLw2GZlkK:baoaj9s) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L120)):*
> `for job := range indexJobs` is more intuitive. Also, avoid ending var names with a capital case.
</details>
Done. No more channel.

---

*[posting/index.go, line 130 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3a-Hw0BS9vtQynTGP:-KR46MHW8IVm-D1-zrcT:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L130)):*
> Maybe you can just create a context.Background() once before loop starts, and reuse it.
</details>
Done.

---

*[posting/index.go, line 140 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3aFeBgllKvuHiN8vL:-KR46ngJ2athTxT76E2O:b-vw93y0) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L140)):*
> This should never be called with indexer keys. So, instead of doUpdateIndex, this should be a check. All the indexer keys should be called with `AddMutation()`.
</details>
Done. Good point. You are so awake.

---

*[posting/index.go, line 142 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3ae5Ks-f8GQaW51UE:-KR46uztwX9YbIg9tprm:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L142)):*
> This doesn't need to be a pointer.
</details>
Done.

---

*[posting/index.go, line 146 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3aXqMLX5zcYJVkPUX:-KR47CE8w4DGqLB6BZke:bnmwwqv) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L146)):*
> You don't need to allocate the memory on a heap.
</details>
Done. (thought golang escape analysis might be smart enough)

---

*[posting/index.go, line 147 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3agaVVSbDGOBG7Klw:-KR47DhmmXUgUYDDPG2O:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L147)):*
> ... `l.Get(&lastPost, l.Length() -1)`
</details>
Done.

---

*[posting/index.go, line 161 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3b4g2wk_PeuISOStJ:-KR6ovCiQv5C_4T2ISFP:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L161)):*
> You can call l.AddMutation(..) here. AddMutation is pretty fast, so we can just do it here.
</details>
Done.

---

*[posting/index.go, line 164 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3bCGmGzuNr83aK55b:-KR47UVKjYXgLdUB2hIg:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L164)):*
> Just write to channel here? You don't need the extra one-liner functions.
</details>
Done.

---

*[posting/index.go, line 171 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3bQq8YMxcn9e9EhhE:-KR47Y87PoSEu7N67jEu:btv8xnv) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L171)):*
> if indexJobC was nil, you can avoid all the logic to generate the data. So, check this during AddMutationWithIndex.
</details>
Done. (remnants of old code)

---

*[posting/list.go, line 456 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR1k1RNLlPMUEz7du2C:-KR49DNM3NS5Va2MLDHC:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/042da9f74f763d0c17c477e01623ec6d4130ec71/posting/list.go#L456)):*
> This is very internal to PL. You should not be modifying the code here.
</details>
Done.

---

*[posting/list.go, line 715 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR1l-DkhsQ8TDoECZEU:-KR48sV5RER1wE6Wi5UK:b-qgzwkd) ([raw file](https://github.com/dgraph-io/dgraph/blob/042da9f74f763d0c17c477e01623ec6d4130ec71/posting/list.go#L715)):*
> Arrr.. I see. My bad! yeah, we should call the push to rocksdb, `commit` or something.
</details>
Ack. No prob!

---

*[posting/list.go, line 70 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3botVTEPGzXZ68Wim:-KR6oy6paO4TrFTrlbKk:b-6pw3sr) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/list.go#L70)):*
> Why do you need these 2 fields?
</details>
Thought they might be useful to others. But realize I don't need them also.

---

*[posting/list.go, line 127 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3eWO2vygwpTHTyvKT:-KR6vF6bBzc3vnLWcl_D:b5ujawf) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/list.go#L127)):*
> Why?
</details>
Decided that there is no need to use it. We can get it from DirectedEdge. Have to modify worker_test to propagate this info in DirectedEdge. Function deleted.

---

*[posting/list.go, line 260 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3dZJg833p_qx5UZr4:-KR47qQV2sLTq-7By-TR:b-ydm282) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/list.go#L260)):*
> The difference between writing m := new(types.Posting) here, v/s creating that memory in another function, is where the memory gets allocated. In the first case, if m isn't being used outside this func, it would be put on stack, which is very cheap. In the latter case, the memory would be put on heap, because it needs to be passed outside the creator function. Given the number of entries can be large, we can end up with a lot of memory allocated on heap really fast, which the GC would then have to sweep through.
> 
> I had a similar case in rdf.Parse, which significantly impacted the memory and cpu profiling.
> 
> In this case, we're going to get rid of this code anyway. So, nothing much to do here, but something to keep in mind in the future. Memory allocation is tricky business, so do it carefully.
</details>
I see.. it's subtle. But a lot of x=new(..) and then x.Init(..) is kind of ugly. If you like to be explicit about using the stack, why not use a local var?

(Miss C++ and RAII https://en.wikipedia.org/wiki/Resource_Acquisition_Is_Initialization)

---

*[posting/list.go, line 286 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3c4aZk8o4qDx2Cpru:-KR48DK4-S-8BLjqk0-4:b-s1uujt) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/list.go#L286)):*
> Not really. Read the code again, I'll wait :-).
</details>
haha.. The word "copy" is misleading.. Ack.

---

*[posting/list.go, line 489 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3ceY1StpM0wiQYHks:-KR48Pew4egiUid2X3pl:bzd0kzm) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/list.go#L489)):*
> I'm not sure if a comment is the best way to address bugs. Either bring it up for a discussion on communication channels or write a test case to prove it or deny it.
</details>
Yes, I have done that. Comment deleted.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-:-KR44DH477ou-qRnOvgn:b-l3huz2)*

manishrjain:
Reviewed 11 of 11 files at r5.
Review status: all files reviewed at latest revision, 38 unresolved discussions, some commit checks broke.

---

*[posting/index.go, line 110 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3ZKwlDLKysePUJXPl:-KR7TLznmrBEyURDnfVo:bvizsuv) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L110)):*
> Done.
</details>
Emm.. don't see the change. Maybe you missed it or forgot to push?

---

*[posting/index.go, line 110 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR7Tz9dqVF-UrMyJSZf:-KR7Tz9dqVF-UrMyJSZg:bmr8uyc) ([raw file](https://github.com/dgraph-io/dgraph/blob/550fee09c8647d6f35dcd8a01cf5acb6b047dc8f/posting/index.go#L110)):*
> ```Go
> 
> // processIndexJob adds mutations to maintain our index.
> func processIndexJob(attr string, uid uint64, term []byte, del bool) {
> ```

processIndexTerm

---

*[posting/index.go, line 132 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR7TYM7rAuHaqh5QJAS:-KR7TYM7rAuHaqh5QJAT:b-13tenn) ([raw file](https://github.com/dgraph-io/dgraph/blob/550fee09c8647d6f35dcd8a01cf5acb6b047dc8f/posting/index.go#L132)):*
> ```Go
> 	}
> 	numOp := numSet + numDel
> 	if (numOp % 100000) == 0 {
> ```

Hmm... I'd have just printed it on every mutation. Along with the attr, term and the uid. This can be a great way to figure out what's going on right now in the system. Also, then you don't need the counters.

I think the concern that you have which might be valid is this might slow things a bit. But, trace library is generally meant to tackle that issue -- so we'll thin it out if it turns out to be slow in profiling.

---

*[posting/list.go, line 260 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3dZJg833p_qx5UZr4:-KR7VWGhjIBcdtM-Oudj:bhghb9i) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/list.go#L260)):*
> I see.. it's subtle. But a lot of x=new(..) and then x.Init(..) is kind of ugly. If you like to be explicit about using the stack, why not use a local var?
> 
> (Miss C++ and RAII https://en.wikipedia.org/wiki/Resource_Acquisition_Is_Initialization)
</details>
Yeah, local vars allocated on the stack are the best. But at times, it's more convenient to have pointers.. The main point is to understand how Go memory allocation works.

Also, you'll have a conflict here. So, when you do, remember to delete the x.NewPosting(buffer) function as well. Because it would have the problem that I mentioned above.

Also: I know I told you that you could do this when you were complaining about Flatbuffers, but in this case, we're choosing cheaper memory allocation over 3 lines of repetitive code.

---

*[posting/list.go, line 489 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3ceY1StpM0wiQYHks:-KR7WU0i5brI_295MCDZ:bvpfyjm) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/list.go#L489)):*
> Yes, I have done that. Comment deleted.
</details>
Don't see the delete. Maybe reviewable​ issue.

---

*[posting/list.go, line 252 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR7WJ_E8bwPlzBnFTes:-KR7WJ_E8bwPlzBnFTet:b-1n9lws) ([raw file](https://github.com/dgraph-io/dgraph/blob/550fee09c8647d6f35dcd8a01cf5acb6b047dc8f/posting/list.go#L252)):*
> ```Go
> // Key returns the key of posting list. You should try to use DirectedEdge if you
> // can for attribute, entity and other info.
> func (l *List) Key() []byte {
> ```

Actually, I feel you might not need this. Let me read the code below, and see if you can avoid this.

**Update**: Yeah, couldn't find any usage of this. Also, I doubt you really need to double check that the key in gotomic hash = key in list. So, remove it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/196)*

jchiu0:
Review status: 27 of 28 files reviewed at latest revision, 41 unresolved discussions.

---

*[posting/index.go, line 110 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3ZKwlDLKysePUJXPl:-KR7YrsWd5AVPrVqSJxv:b-9203ne) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L110)):*
> Emm.. don't see the change. Maybe you missed it or forgot to push?
</details>
It's there? Maybe you forgot to refresh?

---

*[posting/index.go, line 110 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR7Tz9dqVF-UrMyJSZf:-KR7Z46VE2bH6URZh0he:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/550fee09c8647d6f35dcd8a01cf5acb6b047dc8f/posting/index.go#L110)):*
> processIndexTerm
</details>
Done.

---

*[posting/index.go, line 132 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR7TYM7rAuHaqh5QJAS:-KR7ZjzFW_8nLqHeEc5R:bko04kc) ([raw file](https://github.com/dgraph-io/dgraph/blob/550fee09c8647d6f35dcd8a01cf5acb6b047dc8f/posting/index.go#L132)):*
> Hmm... I'd have just printed it on every mutation. Along with the attr, term and the uid. This can be a great way to figure out what's going on right now in the system. Also, then you don't need the counters.
> 
> I think the concern that you have which might be valid is this might slow things a bit. But, trace library is generally meant to tackle that issue -- so we'll thin it out if it turns out to be slow in profiling.
</details>
The string <-> []byte copying always feel bad to me. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/196)*

jchiu0:
Review status: 26 of 28 files reviewed at latest revision, 41 unresolved discussions.

---

*[posting/index.go, line 110 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3ZKwlDLKysePUJXPl:-KR7amK8Jw4gl_fP91RW:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/index.go#L110)):*
> It's there? Maybe you forgot to refresh?
</details>
Done.

---

*[posting/list.go, line 260 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3dZJg833p_qx5UZr4:-KR7aR0iKb6mxpTv3Qdw:bmnrtav) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/list.go#L260)):*
> Yeah, local vars allocated on the stack are the best. But at times, it's more convenient to have pointers.. The main point is to understand how Go memory allocation works.
> 
> Also, you'll have a conflict here. So, when you do, remember to delete the x.NewPosting(buffer) function as well. Because it would have the problem that I mentioned above.
> 
> Also: I know I told you that you could do this when you were complaining about Flatbuffers, but in this case, we're choosing cheaper memory allocation over 3 lines of repetitive code.
</details>
Sorry, I will try to stop complaining.

Switched to using x.ParsePosting.

---

*[posting/list.go, line 489 \[r4\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR3ceY1StpM0wiQYHks:-KR7bOa2KONXyWAFJgl-:b-ln87d9) ([raw file](https://github.com/dgraph-io/dgraph/blob/f6d140783d3ddf97697886d09e96ce938aa95674/posting/list.go#L489)):*
> Don't see the delete. Maybe reviewable​ issue.
</details>
My bad. I lost a lot of intermediate changes this morning when my machine crashed just now.

---

*[posting/list.go, line 252 \[r5\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-KR7WJ_E8bwPlzBnFTes:-KR7aqIB-Jt-HyiKwfMU:b-am2rv6) ([raw file](https://github.com/dgraph-io/dgraph/blob/550fee09c8647d6f35dcd8a01cf5acb6b047dc8f/posting/list.go#L252)):*
> Actually, I feel you might not need this. Let me read the code below, and see if you can avoid this.
> 
> **Update**: Yeah, couldn't find any usage of this. Also, I doubt you really need to double check that the key in gotomic hash = key in list. So, remove it.
</details>
Done. Deleted.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/196)*

manishrjain:
:lgtm: Alright, Alright, Alright! Submission time! Glad to see how simple this whole code and set up has become. Great job, and thanks for doing all these iterations!

---

Reviewed 2 of 2 files at r6.
Review status: all files reviewed at latest revision, 33 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-:-KR87QUTUTgBdl7SO-KA:b-t1yf0x)*

jchiu0:
Thanks for going through this with me. I learned a lot!

---

Review status: all files reviewed at latest revision, 33 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/196#-:-KR89FiJrNnFAOjkNRRo:b1pmb9r)*

PR #200, opened 2016-09-07
Rdbwrap

jchiu0:
Get rid of gorocksdb. I think we can also get rid of Cockroach?

Would require user to install RocksDB on their own, the usual way. Binaries will work fine. Brew install will work fine.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/200)


manishrjain:
:lgtm_strong: I noticed that the license is there. Should be good enough. Feel free to submit.

---

Reviewed 38 of 50 files at r2, 1 of 1 files at r3, 11 of 11 files at r4.

---

*[rdb/cache.go, line 1 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/200#-KRDolGlLrsFHk6XCEHp:-KRDolGmnRYPe2bA7etB:bjx4nmx) ([raw file](https://github.com/dgraph-io/dgraph/blob/859bda431468fa6393bb03ea58186f342b529452/rdb/cache.go#L1)):*
> ```Go
> package rdb
> ```

Looks like you're using a lot of gorocksdb code. In that case, please preserve the LICENSE.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/200#-:-KRSumOwMwNuI-AgO315:bohj5jr)*

PR #201, opened 2016-09-07
Avoid importing a new package and a second logic to do the same thing.

manishrjain:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/201)

### DIFF
--- a/uid/assigner.go
+++ b/uid/assigner.go
@@ -17,16 +17,15 @@
 package uid
 
 import (
-	"crypto/rand"
+	"context"
 	"errors"
 	"fmt"
 	"log"
 	"math"
+	"math/rand"
 	"strings"
 	"sync"
 	"time"
-
-	"context"
 
 	"github.com/dgryski/go-farm"
 
@@ -99,6 +98,7 @@ func (lm *lockManager) clean() {
 
 // package level init
 func init() {
+	rand.Seed(time.Now().UnixNano())
 	lmgr = new(lockManager)
 	lmgr.locks = make(map[string]*entry)
 	// TODO(manishrjain): This map should be cleaned up.
--- a/uid/assigner.go
+++ b/uid/assigner.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"log"
 	"math"
-	mrand "math/rand"
 	"strings"
 	"sync"
 	"time"
@@ -38,7 +37,6 @@ import (
 )
 
 var lmgr *lockManager
-var letterRunes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ "
 var uidStore *store.Store
 var eidPool = sync.Pool{
 	New: func() interface{} {
@@ -118,20 +116,26 @@ func Init(ps *store.Store) {
 // until the obtained integer is unique.
 func allocateUniqueUid(xid string, instanceIdx uint64,
 	numInstances uint64) (uid uint64, rerr error) {
-	mrand.Seed(time.Now().UnixNano())
 	mod := math.MaxUint64 / numInstances
 	minIdx := instanceIdx * mod
+
 	txid := []byte(xid)
 	val := xid
 	if strings.HasPrefix(xid, "_new_:") {
 		if _, err := rand.Read(txid); err != nil {
 			return 0, err
 		}
-
 		val = "_new_"
 	}
 
-	for ; ; txid = append(txid, letterRunes[mrand.Intn(len(letterRunes))]) {
+	suffix := make([]byte, 1)
+	for i := 0; ; i++ {
+		if i > 0 {
+			_, err := rand.Read(suffix)
+			x.Check(err)
+			txid = append(txid, suffix[0])
+		}
+
 		uid1 := farm.Fingerprint64(txid) // Generate from hash.
 		uid = (uid1 % mod) + minIdx
 		if uid == math.MaxUint64 {


ashwin95r:
Umm, based on some reading, I found this: 
`Your benchmark and discussion of math/rand and crypto/rand seems to completely miss the point that these serve completely different needs and should not arbitrarily be interchanged. I.e. it's an apples vs oranges comparison.

If you're doing cryptography (e.g. you need a nonce, an IV, an AES key, whatever) then you damn well better be using crypto/rand (or some other cryptographically secure (pseudo)random number source) and you shouldn't care that it's slower than math/rand; if you use math/rand for crypto stuff you're implementation will not be secure!

On the other hand, if you don't need cryptographic quality random numbers then you'd have no business using crypto/rand (except for limited quantities, e.g. it's an great idea to use crypto/rand once to seed a PRNG rather than using time). One reason is that some system's cryptographic random sources have a limited pool of entropy and consuming that can either slow down the source or turn it into something more like a PRNG. (Although modern good OSes doen't really suffer much from this, it's still a concern).`

Because​ of that I wanted to avoid using a lot of crypto rand. Otherwise, no issues.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/201#-:-KR4_CgH0yvcp_bjETip:bk9kopz)*

manishrjain:
That's a fair point. But if you were convinced that crypto is an overkill, you should swap it out completely. Having both the libraries imported and doing some from crypto and some from rand is just hacky to say the least.

We could switch over to math rand, that should be alright. I'll make the changes.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/201#-:-KR6PXBSWBQmlB87-7s8:b7wdkax)*

ashwin95r:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[uid/assigner.go, line 125 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/201#-KRCoxAPX0sNXxSxb6nR:-KRCoxAPX0sNXxSxb6nS:bt1oj6v) ([raw file](https://github.com/dgraph-io/dgraph/blob/8960a5c01392545408c0ca0866debab0935a0b8e/uid/assigner.go#L125)):*
> ```Go
> 	val := xid
> 	if strings.HasPrefix(xid, "_new_:") {
> 		if _, err := rand.Read(txid); err != nil {
> ```

Looks like err is always nil. 
```
Read generates len(p) random bytes from the default Source
and writes them into p. It always returns len(p) and a nil error. 
Read, unlike the Rand.Read method, is safe for concurrent use.
```
So, shall avoid this check maybe?

https://golang.org/pkg/math/rand/#Read

---

*[uid/assigner.go, line 135 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/201#-KRCpeObITtOUdsTVO6E:-KRCpeObITtOUdsTVO6F:b-je9b1u) ([raw file](https://github.com/dgraph-io/dgraph/blob/8960a5c01392545408c0ca0866debab0935a0b8e/uid/assigner.go#L135)):*
> ```Go
> 		if i > 0 {
> 			_, err := rand.Read(suffix)
> 			x.Check(err)
> ```

Could avaoid this check as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/201)*

manishrjain:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[uid/assigner.go, line 125 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/201#-KRCoxAPX0sNXxSxb6nR:-KRCpynGH3em6Z1ffchF:br3t6q3) ([raw file](https://github.com/dgraph-io/dgraph/blob/8960a5c01392545408c0ca0866debab0935a0b8e/uid/assigner.go#L125)):*
> Looks like err is always nil. 
> ```
> Read generates len(p) random bytes from the default Source
> and writes them into p. It always returns len(p) and a nil error. 
> Read, unlike the Rand.Read method, is safe for concurrent use.
> ```
> So, shall avoid this check maybe?
> 
> https://golang.org/pkg/math/rand/#Read
</details>
We still need to check it.

---

*[uid/assigner.go, line 135 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/201#-KRCpeObITtOUdsTVO6E:-KRCq06jldSay2jw-ZVR:bc3wqxo) ([raw file](https://github.com/dgraph-io/dgraph/blob/8960a5c01392545408c0ca0866debab0935a0b8e/uid/assigner.go#L135)):*
> Could avaoid this check as well.
</details>
Still need to check for errors.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/201)*

ashwin95r:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---

*[uid/assigner.go, line 125 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/201#-KRCoxAPX0sNXxSxb6nR:-KRCq5nAyFMz6y6O09E5:b2nhll9) ([raw file](https://github.com/dgraph-io/dgraph/blob/8960a5c01392545408c0ca0866debab0935a0b8e/uid/assigner.go#L125)):*
> We still need to check it.
</details>
Ok.

---

*[uid/assigner.go, line 135 \[r2\]](https://reviewable.io:443/reviews/dgraph-io/dgraph/201#-KRCpeObITtOUdsTVO6E:-KRCq7Z9SIfYI16QZtE_:b2nhll9) ([raw file](https://github.com/dgraph-io/dgraph/blob/8960a5c01392545408c0ca0866debab0935a0b8e/uid/assigner.go#L135)):*
> Still need to check for errors.
</details>
Ok.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/201#-:-KRCq9sXi2m2837pDUNq:bnfp4nl)*

PR #207, opened 2016-09-12
Fix/worker race

jchiu0:
1) Previously, there are race conditions around the array "pools" in worker.State. I added a "harmless looking" lock around it.
2) Not good to build too much logic around globals, in this case "ws". We let "Connect" become a class method for easier testing.
3) Not good to use ws as a global and also as a "self" variable. We don't want any variable shadowing.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/207)

### DIFF
--- a/cmd/dgraph/main.go
+++ b/cmd/dgraph/main.go
@@ -547,8 +547,10 @@ func main() {
 	}
 
 	posting.Init()
+	var ws *worker.State
 	if *instanceIdx != 0 {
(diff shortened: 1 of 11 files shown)


manishrjain:
:lgtm:

---

Reviewed 11 of 11 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/207#-:-KRTSycavZJmEM58viUi:bnfp4nl)*

PR #213, opened 2016-09-13
  add type validation in mutations

akhiltak:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/213)


manishrjain:
:lgtm_strong: Looks nice, clean and simple. One small comment. Address that, and feel free to submit.

---

Reviewed 2 of 2 files at r1.

---

*[cmd/dgraph/main.go, line 263 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/213#-KRd3ET7oFILb6gOpQSQ:-KRd3ET7oFILb6gOpQSR:b7gxhir) ([raw file](https://github.com/dgraph-io/dgraph/blob/716e6bda6ad4104f8a21205bb09c522200a3cb46/cmd/dgraph/main.go#L263)):*
> ```Go
> 	}
> 	m := set
> 	for _, nquad := range del {
> ```

Do we need to validate the deletes?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/213#-:-KRd3x7Re31-dlmo4m_5:bmmeozp)*

akhiltak:

---

*[cmd/dgraph/main.go, line 263 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/213#-KRd3ET7oFILb6gOpQSQ:-KRhoGxjMtHjiWdeUHno:b-44prrc) ([raw file](https://github.com/dgraph-io/dgraph/blob/716e6bda6ad4104f8a21205bb09c522200a3cb46/cmd/dgraph/main.go#L263)):*
> Do we need to validate the deletes?
</details>
Not really, since for scalars we can give just about any value in place of `object` in rdf format here.
like 
```
mutation{
    delete {
    	  <person> <age> "asdf" .
    }
}
```
I think we kept this placeholder object for some valid reasons (one of which might be to keep parsing into RDF consistent)

But, I believe we shouldn't allow very loose queries on our database. Validating the delete will disallow queries like following which doesn't make sense. Currently, these are valid.
```
mutation{
    delete {
    	  <brandon> <friend> "asdf" .
    }
}
// this results in no change
OR
mutation{
    delete {
    	  <brandon> <age> "asdf" .
    }
}
// this results in deletion of age attribute for brandon with value "40".
```
But if we have some other way to keep these kind of queries in check, please let me know. Then, we wouldn't need to validate delete.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/213)*

manishrjain:

---

*[cmd/dgraph/main.go, line 263 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/213#-KRd3ET7oFILb6gOpQSQ:-KRifg5irRW-Zd-130fW:b-ggwzb5) ([raw file](https://github.com/dgraph-io/dgraph/blob/716e6bda6ad4104f8a21205bb09c522200a3cb46/cmd/dgraph/main.go#L263)):*
<details><summary><i>Previously, akhiltak (Akhil Tak) wrote…</i></summary>
> Not really, since for scalars we can give just about any value in place of `object` in rdf format here.
> like 
> ```
> mutation{
>     delete {
>     	  <person> <age> "asdf" .
>     }
> }
> ```
> I think we kept this placeholder object for some valid reasons (one of which might be to keep parsing into RDF consistent)
> 
> But, I believe we shouldn't allow very loose queries on our database. Validating the delete will disallow queries like following which doesn't make sense. Currently, these are valid.
> ```
> mutation{
>     delete {
>     	  <brandon> <friend> "asdf" .
>     }
> }
> // this results in no change
> OR
> mutation{
>     delete {
>     	  <brandon> <age> "asdf" .
>     }
> }
> // this results in deletion of age attribute for brandon with value "40".
> ```
> But if we have some other way to keep these kind of queries in check, please let me know. Then, we wouldn't need to validate delete.
> 
</details>
// this results in deletion of age attribute for brandon with value "40".
That doesn't sound right. If that's happening, can you file a bug report?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/213)*

akhiltak:

---

*[cmd/dgraph/main.go, line 263 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/213#-KRd3ET7oFILb6gOpQSQ:-KRjVlhZvH9Cn69hghqj:b7skeph) ([raw file](https://github.com/dgraph-io/dgraph/blob/716e6bda6ad4104f8a21205bb09c522200a3cb46/cmd/dgraph/main.go#L263)):*
> // this results in deletion of age attribute for brandon with value "40".
> That doesn't sound right. If that's happening, can you file a bug report?
</details>
I also thought it was weird but there is only a single edge in that case, so thought it was intended.
Filling a bug for it. 
So, what do you think about adding validation for del operation in mutation?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/213)*

akhiltak:

---

*[cmd/dgraph/main.go, line 263 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/213#-KRd3ET7oFILb6gOpQSQ:-KRlBTILEEuLbvL00cbb:bjqxzb9) ([raw file](https://github.com/dgraph-io/dgraph/blob/716e6bda6ad4104f8a21205bb09c522200a3cb46/cmd/dgraph/main.go#L263)):*
<details><summary><i>Previously, akhiltak (Akhil Tak) wrote…</i></summary>
> I also thought it was weird but there is only a single edge in that case, so thought it was intended.
> Filling a bug for it. 
> So, what do you think about adding validation for del operation in mutation?
</details>
@mrjn - My bad, it seems like bug is no longer present. Tried reproducing the steps using few older commits but no luck. Maybe I missed the correct output due to some old garbage data and delete was working fine all along. :-/
Question remains though, do we want to label clients mutations like this valid?
```
mutation{
    delete {
    	 <brandon> <age> "dummy" .
    }
}
```
although it won't change the data internally.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/213)*

manishrjain:
Is this submitted?

---

---

*[cmd/dgraph/main.go, line 263 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/213#-KRd3ET7oFILb6gOpQSQ:-KS1WI5cBoNoI0tTo83p:b-3re9bo) ([raw file](https://github.com/dgraph-io/dgraph/blob/716e6bda6ad4104f8a21205bb09c522200a3cb46/cmd/dgraph/main.go#L263)):*
<details><summary><i>Previously, akhiltak (Akhil Tak) wrote…</i></summary>
> @mrjn - My bad, it seems like bug is no longer present. Tried reproducing the steps using few older commits but no luck. Maybe I missed the correct output due to some old garbage data and delete was working fine all along. :-/
> Question remains though, do we want to label clients mutations like this valid?
> ```
> mutation{
>     delete {
>     	 <brandon> <age> "dummy" .
>     }
> }
> ```
> although it won't change the data internally.
</details>
Yeah, good point. We should validate this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/213#-:-KS1WOEFIkT5KFZNkhHI:bpx1dn9)*

akhiltak:
merged to master.

PR #218, opened 2016-09-15
Improve/protobuf

pawanrawal:
This makes the uid optional and also fixes a bug in which for leaf nodes uid was also sent as part of properties.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/218)


pawanrawal:
Merged

PR #222, opened 2016-09-19
Feature/hashsharded

jchiu0:
On my machine, using 32 shards for lhmap takes about 5 mins. Using one shard takes about 7 mins.
(Before this change, it takes about 8 mins)

We can add unit tests but I want to make sure the overall design is accepted first.

The main changes are:
1) We stop using gotomic hash.
2) For lhmap, we delete as we iterate, and push *List to a channel. Goroutines which consume this channel do not lock lhmap at all.
3) For dirtymap, we use a channel to control updates to it. Then every 5s, we do a gentle merge. There can be at most 18 gentle merges happening at the same time.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/222)


manishrjain:
Reviewed 1 of 4 files at r1, 1 of 3 files at r2, 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 17 unresolved discussions.

---

*[posting/hash.go, line 41 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS0w8cPArROQyHpudcA:-KS0w8cRR1OVRXhYiblY:b6rpw08) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/hash.go#L41)):*
> ```Go
> 
> // PutIfMissing puts item into list if it is missing. Otherwise, do nothing.
> func (s *listMapShard) PutIfMissing(key uint64, val *List) bool {
> ```

If this returned the last value, that'd avoid another call to Get.

---

*[posting/hash.go, line 71 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS0wy0J2W8h6Op6wYWa:-KS0wy0K_I2tS9m-7gee:bspbnzf) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/hash.go#L71)):*
> ```Go
> // MultiGet returns number of items written to buf. Number of items written is
> // guaranteed to be <= size.
> func (s *listMapShard) MultiGet(buf []uint64, size int) int {
> ```

You probably don't need this func.

---

*[posting/hash.go, line 123 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS0xAHtTmzCMKITsScI:-KS0xAHuDQRe0dGDJ9YX:b-ngtumq) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/hash.go#L123)):*
> ```Go
> // MultiGet returns number of items written to buf. Number of items written is
> // guaranteed to be <= size.
> func (s *listMap) MultiGet(buf []uint64, size int) int {
> ```

And this one. Also, this isn't MultiGet, because you're not passing in the keys to retrieve. It's more like a RetrieveNKeys(..)

---

*[posting/lists.go, line 44 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS0xSAJqWue8ao__epK:-KS0xSAK5-jRBEpugJYu:b-m7cjf) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L44)):*
> ```Go
> 		"If RAM usage exceeds this, we stop the world, and flush our buffers.")
> 
> 	keysBufferSize = flag.Int("keysbuffer", 10000, "Number of keys to be buffered for deletion.")
> ```

Why not do it based on percentage of the size? If we have few mutations, we don't want to merge everything every 5 seconds. That'd unnecessarily push more work on all resources (cpu, ram, rocksdb).

---

*[posting/lists.go, line 48 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS0xlJLUGSxHD8Y4ZGO:-KS0xlJLUGSxHD8Y4ZGP:b-wnv1pr) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L48)):*
> ```Go
> 	dirtyMap       map[uint64]struct{} // Made global for log().
> 	dirtyChannel   chan uint64         // Puts to dirtyMap has to go through this channel.
> 	resetDirtyChan chan struct{}       // To clear dirtyMap, push to this channel.
> ```

hmm... do we need this? Will read below and check again.

---

*[posting/lists.go, line 114 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS13foZ-jUPv9sAM8Yw:-KS13foZ-jUPv9sAM8Yx:b-2az718) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L114)):*
> ```Go
> 
> // mergeAndUpdateKeys calls mergeAndUpdate for each key in array "keys".
> func mergeAndUpdateKeys(keys []uint64, gentleMergeChan chan struct{}) {
> ```

if len(keys) == 0 return

---

*[posting/lists.go, line 129 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS13ifwNCaxcezFtouq:-KS13ifwNCaxcezFtour:b3faxdf) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L129)):*
> ```Go
> 	}
> 	ctr.log()
> 	<-gentleMergeChan
> ```

This could be a defer at the top, if you're using the above return.

---

*[posting/lists.go, line 137 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS0ybWJxY1D3uVLpyJL:-KS0ybWJxY1D3uVLpyJM:btn0q4b) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L137)):*
> ```Go
> 	// Max number of gentle merges in parallel.
> 	gentleMergeChan := make(chan struct{}, 18)
> 	timer := time.Tick(5 * time.Second)
> ```

If you're assigning a var to this, might as well use time.NewTicker(..)

---

*[posting/lists.go, line 144 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS0yj_8z0JGWmGgQU0T:-KS0yj_8z0JGWmGgQU0U:bo76kum) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L144)):*
> ```Go
> 			case gentleMergeChan <- struct{}{}:
> 				n := int(float64(len(dirtyMap)) * 0.07) // Clear 7% of dirty lists.
> 				keysBuffer := make([]uint64, 0, 10000)
> ```

You can use `n` insted of hardcoding 10K capacity. Also, you could increase the 7% to 10%, if you think that's better.

---

*[posting/lists.go, line 161 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS0zNP2YF4wqXIWnkH0:-KS0zNP2YF4wqXIWnkH1:bvowjq1) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L161)):*
> ```Go
> 
> 		case <-resetDirtyChan:
> 			dirtyMap = make(map[uint64]struct{}, 10000)
> ```

This would recreate a new dirtyMap, whole capacity might be much lower than what the existing dirtyMap reached. It might seem unclean, but this is actually better, I'd think:

for k := range dirtyMap {
  delete(dirtyMap, k)
}

This way, the capacity might stay the same, and there'd be no need for creating a new map and reallocating all that space.

---

*[posting/lists.go, line 228 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS139rrRpniqSmUTOUQ:-KS139rrRpniqSmUTOUR:b8vhmyi) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L228)):*
> ```Go
> 		totMemory := getMemUsage()
> 		if totMemory > *maxmemory {
> 			aggressivelyEvict()
> ```

So, you removed gentlyMerge from here. Would that continue to be called even when aggressive merging is run?

---

*[posting/lists.go, line 242 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS11Ei65sbl-h3JD_Tn:-KS11Ei65sbl-h3JD_To:b-knt67y) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L242)):*
> ```Go
> 	lhmap = newShardedListMap(*lhmapNumShards)
> 	dirtyChannel = make(chan uint64, 10000)
> 	dirtyMap = make(map[uint64]struct{}, 10000)
> ```

You can set it to something smaller, like a 1000. In a typical dgraph server run, things shouldn't be so crazy. Also, this map can then grow to accommodate the size required.

---

*[posting/lists.go, line 243 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS11RV4rr8Cg20kpHLy:-KS11RV5YgG4hjM6t070:b-80kjcl) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L243)):*
> ```Go
> 	dirtyChannel = make(chan uint64, 10000)
> 	dirtyMap = make(map[uint64]struct{}, 10000)
> 	resetDirtyChan = make(chan struct{})
> ```

This is unbuffered channel, it would be blocking at both read and write. I think having cap=1 there would help with that.

---

*[posting/lists.go, line 297 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS11mAogfJlhqwBbFf7:-KS11mAogfJlhqwBbFf8:b-2x4ky) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L297)):*
> ```Go
> 	const grainSize = 100
> 	for {
> 		last := atomic.AddUint64(idx, grainSize) // Take a batch of elements.
> ```

What is this logic? Why so many lines of code? This seems like a lot of lines were added from the original range over a channel, which seems simpler in comparison.

I think processOne has it's benefits, in it's simpler. You can call it from an iterator instead of a channel, if you're trying to avoid having a channel.

---

*[posting/lists.go, line 322 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS12UzagI30GnHg__7M:-KS12UzbCuuSV5wh-S3-:b-jucinz) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L322)):*
> ```Go
> 	keysBuffer := make([]uint64, *keysBufferSize) // Better to allocate locally.
> 	for lhmap.Size() > 0 {
> 		size := lhmap.MultiGet(keysBuffer, *keysBufferSize)
> ```

I think you need a `lhmap.GetAll`. You can use an iterator instead of a channel if you want.

---

*[posting/lists.go, line 327 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS12KOGLGuj-o9Cy4SD:-KS12KOGLGuj-o9Cy4SE:b3ovb9j) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L327)):*
> ```Go
> 		for i := 0; i < numRoutines; i++ {
> 			wg.Add(1)
> 			go process(keysBuffer[:size], &idx, c, &wg)
> ```

What's the rationale behind sending in these batches? It makes for more complex code, and they're not even going over the wire.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/222)*

jchiu0:
Review status: 1 of 3 files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[posting/hash.go, line 41 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS0w8cPArROQyHpudcA:-KS1k_Ux0XOUyIW04NwB:b-qu6n94) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/hash.go#L41)):*
> If this returned the last value, that'd avoid another call to Get.
</details>
Good point. Done.

---

*[posting/hash.go, line 71 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS0wy0J2W8h6Op6wYWa:-KS1krPvLdJiXqpMF3mp:b5q6fat) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/hash.go#L71)):*
> You probably don't need this func.
</details>
This function has been removed. Instead, we now have streamUntilCap. I keep listMapShard.streamUntilCap because of the defer call. If you would like me to remove listMapShard.streamUntilCap, do you prefer me to remove the defer call or add an anonymous function ?

---

*[posting/hash.go, line 123 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS0xAHtTmzCMKITsScI:-KS1ncoKTn8VQeZs9DfR:bmwv8zw) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/hash.go#L123)):*
> And this one. Also, this isn't MultiGet, because you're not passing in the keys to retrieve. It's more like a RetrieveNKeys(..)
</details>
All replaced with streamUntilCap.

---

*[posting/lists.go, line 44 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS0xSAJqWue8ao__epK:-KS4nVfyQdA6siauzlGR:bcl1m4m) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L44)):*
> Why not do it based on percentage of the size? If we have few mutations, we don't want to merge everything every 5 seconds. That'd unnecessarily push more work on all resources (cpu, ram, rocksdb).
</details>
I thought the objective of MergeLists is to commit everything 

---

*[posting/lists.go, line 48 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS0xlJLUGSxHD8Y4ZGO:-KS4nyM5bW863r_h2US2:b-ne6zja) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L48)):*
> hmm... do we need this? Will read below and check again.
</details>
The verdict is?

---

*[posting/lists.go, line 114 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS13foZ-jUPv9sAM8Yw:-KS1olEODwxp6NtAGccA:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L114)):*
> if len(keys) == 0 return
</details>
Done.

---

*[posting/lists.go, line 129 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS13ifwNCaxcezFtouq:-KS1omrR4Dk1m3gjeixt:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L129)):*
> This could be a defer at the top, if you're using the above return.
</details>
Done.

---

*[posting/lists.go, line 137 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS0ybWJxY1D3uVLpyJL:-KS1pG7dsKyphQLZScsB:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L137)):*
> If you're assigning a var to this, might as well use time.NewTicker(..)
</details>
Done.

---

*[posting/lists.go, line 144 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS0yj_8z0JGWmGgQU0T:-KS1pO7ke5zBqo4Q53Bl:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L144)):*
> You can use `n` insted of hardcoding 10K capacity. Also, you could increase the 7% to 10%, if you think that's better.
</details>
Done.

---

*[posting/lists.go, line 161 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS0zNP2YF4wqXIWnkH0:-KS1paCzkglt0DRm6IN4:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L161)):*
> This would recreate a new dirtyMap, whole capacity might be much lower than what the existing dirtyMap reached. It might seem unclean, but this is actually better, I'd think:
> 
> for k := range dirtyMap {
>   delete(dirtyMap, k)
> }
> 
> This way, the capacity might stay the same, and there'd be no need for creating a new map and reallocating all that space.
</details>
Done.

---

*[posting/lists.go, line 228 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS139rrRpniqSmUTOUQ:-KS1qHCQxy-vXsKrzYh3:b-ogoop3) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L228)):*
> So, you removed gentlyMerge from here. Would that continue to be called even when aggressive merging is run?
</details>
Yes it continues to happen at its own pace. I did a printf under "gentleMergeChan <- struct{}{}" and it does print while an aggressive evict is going on.

---

*[posting/lists.go, line 242 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS11Ei65sbl-h3JD_Tn:-KS1r5B-AJN95oRu0O0y:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L242)):*
> You can set it to something smaller, like a 1000. In a typical dgraph server run, things shouldn't be so crazy. Also, this map can then grow to accommodate the size required.
</details>
Done.

---

*[posting/lists.go, line 243 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS11RV4rr8Cg20kpHLy:-KS1r922a2B-ey7Yj1jr:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L243)):*
> This is unbuffered channel, it would be blocking at both read and write. I think having cap=1 there would help with that.
</details>
Done.

---

*[posting/lists.go, line 297 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS11mAogfJlhqwBbFf7:-KS4jUaGqz9B0eKa_Ykr:b4rzp4p) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L297)):*
> What is this logic? Why so many lines of code? This seems like a lot of lines were added from the original range over a channel, which seems simpler in comparison.
> 
> I think processOne has it's benefits, in it's simpler. You can call it from an iterator instead of a channel, if you're trying to avoid having a channel.
</details>
I was trying to reduce the number of atomic adds by 100 times. But there is really little gain in reality so this has been removed.

---

*[posting/lists.go, line 322 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS12UzagI30GnHg__7M:-KS4ln6zRKT4FNDGfWU-:b-ogyn0j) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L322)):*
> I think you need a `lhmap.GetAll`. You can use an iterator instead of a channel if you want.
</details>
You might be pleased that I am opted for a channel here. I timed it several times and find that using a channel doesn't slow things down. Probably most of the time is spent with RocksDB and the tiny overhead here doesn't matter. A channel does make the code neater.

---

*[posting/lists.go, line 327 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS12KOGLGuj-o9Cy4SD:-KS4kf1xhvVaTzgjF0Or:b-6nvgnd) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L327)):*
> What's the rationale behind sending in these batches? It makes for more complex code, and they're not even going over the wire.
</details>
I added some comments. Search for "main idea". There are two things I want to say. I write a lot, so bear with me.

1) When iterating, we want to get all the keys out asap, with minimal interruption. I have tried pushing and pulling at the same time, but that seems to slow things down from 5 mins to ~7 mins.  I think it is because iterating locks the map a lot and doesn't play well when you are trying to call lhmap.Delete in processOne. Hence, the code is structured such that we focus on one thing at a time.

2) It would be nice if we can push all the keys into one buffer, but I don't want to crash the machine by taking up all its memory. Hence,  I split the keys into smaller batches.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/222)*

manishrjain:
Reviewed 1 of 2 files at r6, 1 of 1 files at r7.

---

*[posting/hash.go, line 52 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS6pwRCjZehj98iAp_q:-KS6pwRCjZehj98iAp_r:b-dgfd66) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c8f8f21181c2f83e2255a95ed56bc13f971bf77/posting/hash.go#L52)):*
> ```Go
> 	s.RLock()
> 	defer s.RUnlock()
> 	val, found := s.m[key]
> ```

return s.m[key]

---

*[posting/hash.go, line 107 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS6qotyFCkL-jp-qRxw:-KS6qotyFCkL-jp-qRxx:b-fyimwf) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c8f8f21181c2f83e2255a95ed56bc13f971bf77/posting/hash.go#L107)):*
> ```Go
> 	for key := range s.m {
> 		out <- key
> 		if len(out) == cap(out) {
> ```

This is called from within the stop the world block, so you don't really need to stop if cap is reached. You can just block there. Alternatively, you can take in an iterator function, which you can call with the key repeatedly. Would be faster to iterate than a channel, theoretically.

---

*[posting/lists.go, line 44 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS0xSAJqWue8ao__epK:-KS6rE_CpYSFbJsCN3P3:bp49l1l) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L44)):*
> I thought the objective of MergeLists is to commit everything 
</details>
MergeLists is called from aggressive merging which happens from a stop the world block. The 7% is the gentle merging which happens in the live system. Two different things.

Though reading your implementation below, seems like you're using this for aggressive merging. You don't really need a flag for aggressive merging, there's no real choice here because you're anyway deleting everything.

Instead, having one for gentle merging makes sense -- because that affects how much disk activity happens in a live system.

---

*[posting/lists.go, line 48 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS0xlJLUGSxHD8Y4ZGO:-KS6rQnWieVikZS81ngF:b-85i63v) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L48)):*
> The verdict is?
</details>
LGTM.

---

*[posting/lists.go, line 228 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS139rrRpniqSmUTOUQ:-KS6sLIl1Zgf_9CDswyv:blgb2n7) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L228)):*
> Yes it continues to happen at its own pace. I did a printf under "gentleMergeChan <- struct{}{}" and it does print while an aggressive evict is going on.
</details>
It shouldn't really be called when aggressive evict is going on, because the latter is going to take care of all the keys. gentleMerge shouldn't run for that period. It might end up redoing some of the keys that are in the buffer of aggressiveEvict; which means one of the two is porbably going to bring the PL back to memory from Rocksdb after having merged it.

The full interaction is hard to imagine, but in any case, it's better to only stop gentle merge when aggressive one is happening. That's also the current logic.

---

*[posting/lists.go, line 322 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS12UzagI30GnHg__7M:-KS6v2_vuwA67K8uqUe3:bb6ywmx) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L322)):*
> You might be pleased that I am opted for a channel here. I timed it several times and find that using a channel doesn't slow things down. Probably most of the time is spent with RocksDB and the tiny overhead here doesn't matter. A channel does make the code neater.
</details>
LGTM

---

*[posting/lists.go, line 44 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS6y4AhIiSE8v_6glXJ:-KS6y4AiDAd5v_tkCIKA:b2fdc5i) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c8f8f21181c2f83e2255a95ed56bc13f971bf77/posting/lists.go#L44)):*
> ```Go
> 		"If RAM usage exceeds this, we stop the world, and flush our buffers.")
> 
> 	keysBufferFrac = flag.Float64("keysbuffer", 0.07, "Number of keys to be buffered for deletion as fraction of lhmap.")
> ```

Ratio of dirty posting lists to merge every 5 seconds.

---

*[posting/lists.go, line 147 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS6rrKHzef8FQl7pc61:-KS6rrKHzef8FQl7pc62:b-x8t9kn) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c8f8f21181c2f83e2255a95ed56bc13f971bf77/posting/lists.go#L147)):*
> ```Go
> 			select {
> 			case gentleMergeChan <- struct{}{}:
> 				n := int(float64(len(dirtyMap)) * 0.10) // Clear 10% of dirty lists.
> ```

You added a flag for the ratio above, use it here.

---

*[posting/lists.go, line 271 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS6tGniiaTlQ2HsCqP4:-KS6tGnju95wwPrwR03t:b-4biv5e) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c8f8f21181c2f83e2255a95ed56bc13f971bf77/posting/lists.go#L271)):*
> ```Go
> 		l.init(key, pstore)
> 		return l
> 	}
> ```

FYI. When you merge: Remember to do a ref decrement if lp != l. Also, before returning the pointer to caller, increment the ref.

---

*[posting/lists.go, line 309 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS6vnLb2QmOxO5QJ5Nb:-KS6vnLb2QmOxO5QJ5Nc:bmua4c2) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c8f8f21181c2f83e2255a95ed56bc13f971bf77/posting/lists.go#L309)):*
> ```Go
> 
> 	// Read n items each time by iterating over list map.
> 	n := int(*keysBufferFrac * float64(lhmap.Size()))
> ```

FYI.

I think you mixed the two concepts. keysBufferFrac was for gentle merging, so we don't do too much work too quickly in a live system.

In this case, you've already stopped everything, so you can choose to bite big chunks. Or, just set n to some high enough value.

---

*[posting/lists.go, line 319 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS6vW8fi4lQ6qVJaLdd:-KS6vW8fi4lQ6qVJaLde:b-t9yyrr) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c8f8f21181c2f83e2255a95ed56bc13f971bf77/posting/lists.go#L319)):*
> ```Go
> 	// trying to lock parts of the map. On one machine, the loading times goes from
> 	// 8 mins to <5 mins. We have also tried pushing to keysChan while consuming it
> 	// with processOne calls.  That seems to increase running time back to 7 mins.
> ```

FYI. Interesting.

---

*[posting/lists.go, line 324 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS6vfIgSLoV7dZFkObG:-KS6vfIgSLoV7dZFkObH:b-6actm0) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c8f8f21181c2f83e2255a95ed56bc13f971bf77/posting/lists.go#L324)):*
> ```Go
> 	defer close(keysChan)
> 	for lhmap.Size() > 0 {
> 		lhmap.streamUntilCap(keysChan) // Don't close keysChan yet. We can reuse it.
> ```

Essentially, doing the whole thing in batches? Seems like a very interesting​ approach.

Mention the word "batch" in a comment somewhere here, or "waves" or something to explain what's going on. You're creating gorotines to deal with a batch, then returning from them, then creating a new batch and redoing the steps.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/222)*

manishrjain:
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[posting/lists.go, line 319 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS6vW8fi4lQ6qVJaLdd:-KS9HDL-FRmhD-IUBaYF:buhvrlk) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c8f8f21181c2f83e2255a95ed56bc13f971bf77/posting/lists.go#L319)):*
> FYI. Interesting.
</details>
I was thinking about this a bit more. And I think this approach would suffer from stragglers. A couple of stragglers in each "wave", would block that wave. No more PLs can be batched up until that entire wave is completed fully. All the goroutines would return, and we'd just be waiting for these few straggler goroutines to finish even though we could be doing a lot more work concurrently.

I've seen this happen quite a lot in Mapreduce, and they had elaborate mechanisms to deal with this problem. Disk can act in weird ways, so it's possible we'll see the same problem for bigger PLs here as well.

I'd recommend modifying the implementation to deal with this issue.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/222)*

jchiu0:
Review status: 0 of 3 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[posting/hash.go, line 71 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS0wy0J2W8h6Op6wYWa:-KSC8MzKk-FugJy44h2I:b-lhsbzk) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/hash.go#L71)):*
> This function has been removed. Instead, we now have streamUntilCap. I keep listMapShard.streamUntilCap because of the defer call. If you would like me to remove listMapShard.streamUntilCap, do you prefer me to remove the defer call or add an anonymous function ?
</details>
Actually this function is no longer needed. I am tempted to just delete it. Also helps enforce that we only delete stuff from lhmap via aggressive evict.

---

*[posting/hash.go, line 52 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS6pwRCjZehj98iAp_q:-KSABBsV24UnlxwpeDK4:b-9pci8b) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c8f8f21181c2f83e2255a95ed56bc13f971bf77/posting/hash.go#L52)):*
> return s.m[key]
</details>
I have tried this before. It surprisingly doesn't compile. The error is "not enough arguments to return".....

---

*[posting/hash.go, line 107 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS6qotyFCkL-jp-qRxw:-KSC4HAuAFRBBRGBzLMb:b-vfwqhq) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c8f8f21181c2f83e2255a95ed56bc13f971bf77/posting/hash.go#L107)):*
> This is called from within the stop the world block, so you don't really need to stop if cap is reached. You can just block there. Alternatively, you can take in an iterator function, which you can call with the key repeatedly. Would be faster to iterate than a channel, theoretically.
</details>
Replaced with a EachWithDelete. This resembles the gotomic Each function, except that it deletes before calling the user function.

---

*[posting/lists.go, line 44 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS0xSAJqWue8ao__epK:-KSC4Na7BIM7fGlW6ApV:b49e0hk) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L44)):*
> MergeLists is called from aggressive merging which happens from a stop the world block. The 7% is the gentle merging which happens in the live system. Two different things.
> 
> Though reading your implementation below, seems like you're using this for aggressive merging. You don't really need a flag for aggressive merging, there's no real choice here because you're anyway deleting everything.
> 
> Instead, having one for gentle merging makes sense -- because that affects how much disk activity happens in a live system.
</details>
Thanks for explaining in such detail. Yes, I am aware of the differences between the two, but I got mixed up between what I want to do with this flag. You have a good point here. Can I call the flag "gentlemerge"?

---

*[posting/lists.go, line 48 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS0xlJLUGSxHD8Y4ZGO:-KSC5cNNzAnvfp75JXC9:blhor0t) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L48)):*
> LGTM.
</details>
I made the channel a list of dirtyMapOp and use it to do reset or gentle merges. Is that ok?

---

*[posting/lists.go, line 228 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS139rrRpniqSmUTOUQ:-KSC5s0buWyShpQKhZjs:b-q64kyp) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7b076fb1ea78f5c4abb88f4297f0fce6fc0bf0e/posting/lists.go#L228)):*
> It shouldn't really be called when aggressive evict is going on, because the latter is going to take care of all the keys. gentleMerge shouldn't run for that period. It might end up redoing some of the keys that are in the buffer of aggressiveEvict; which means one of the two is porbably going to bring the PL back to memory from Rocksdb after having merged it.
> 
> The full interaction is hard to imagine, but in any case, it's better to only stop gentle merge when aggressive one is happening. That's also the current logic.
</details>
Thanks for explaining in detail. I have modified the code to make sure we don't start gentle merge when aggressive evict starts, and have verified that.

---

*[posting/lists.go, line 44 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS6y4AhIiSE8v_6glXJ:-KSC9EJqhVf-4XyPNocV:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c8f8f21181c2f83e2255a95ed56bc13f971bf77/posting/lists.go#L44)):*
> Ratio of dirty posting lists to merge every 5 seconds.
</details>
Done.

---

*[posting/lists.go, line 147 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS6rrKHzef8FQl7pc61:-KSC5lGeHpbv4L8-1CJ7:b-p9kisl) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c8f8f21181c2f83e2255a95ed56bc13f971bf77/posting/lists.go#L147)):*
> You added a flag for the ratio above, use it here.
</details>
Done. Flag is currently called gentleMergeFrac..

---

*[posting/lists.go, line 271 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS6tGniiaTlQ2HsCqP4:-KSC6M5fwdCPGotZk65s:bcjq8pa) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c8f8f21181c2f83e2255a95ed56bc13f971bf77/posting/lists.go#L271)):*
> FYI. When you merge: Remember to do a ref decrement if lp != l. Also, before returning the pointer to caller, increment the ref.
</details>
Noted. I will be very careful with the merge. In fact, after everything here is ok, I can merge master into this branch first and let you take a quick look.

---

*[posting/lists.go, line 309 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS6vnLb2QmOxO5QJ5Nb:-KSC6htwTgBLU1MXr0v_:b7v5shp) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c8f8f21181c2f83e2255a95ed56bc13f971bf77/posting/lists.go#L309)):*
> FYI.
> 
> I think you mixed the two concepts. keysBufferFrac was for gentle merging, so we don't do too much work too quickly in a live system.
> 
> In this case, you've already stopped everything, so you can choose to bite big chunks. Or, just set n to some high enough value.
</details>
Yeah, keysBufferFrac is very confusing. I rename it as gentleMergeFrac.

---

*[posting/lists.go, line 319 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS6vW8fi4lQ6qVJaLdd:-KSC6bDNpk-49xMh5F3v:b41vz5o) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c8f8f21181c2f83e2255a95ed56bc13f971bf77/posting/lists.go#L319)):*
> I was thinking about this a bit more. And I think this approach would suffer from stragglers. A couple of stragglers in each "wave", would block that wave. No more PLs can be batched up until that entire wave is completed fully. All the goroutines would return, and we'd just be waiting for these few straggler goroutines to finish even though we could be doing a lot more work concurrently.
> 
> I've seen this happen quite a lot in Mapreduce, and they had elaborate mechanisms to deal with this problem. Disk can act in weird ways, so it's possible we'll see the same problem for bigger PLs here as well.
> 
> I'd recommend modifying the implementation to deal with this issue.
</details>
The concern here is very valid. I have modified the code. It looks much simpler now and it runs even faster!

---

*[posting/lists.go, line 324 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KS6vfIgSLoV7dZFkObG:-KSC7QWizyStwLxZZLQB:bpekg3a) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c8f8f21181c2f83e2255a95ed56bc13f971bf77/posting/lists.go#L324)):*
> Essentially, doing the whole thing in batches? Seems like a very interesting​ approach.
> 
> Mention the word "batch" in a comment somewhere here, or "waves" or something to explain what's going on. You're creating gorotines to deal with a batch, then returning from them, then creating a new batch and redoing the steps.
</details>
Code is much simpler now. Thanks for all the nice suggestions!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/222)*

manishrjain:
Small comments here and there. Otherwise, it looks amazing. Absolutely love the refactoring, and the simplicity of code. Do you want to address the comments and do a merge with master, so I can have a final look before you submit?

---

Reviewed 3 of 3 files at r8.

---

*[posting/hash.go, line 85 at r8](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KSEh_IJxpvBYofQ3ZjM:-KSEh_IJxpvBYofQ3ZjN:b1w3216) ([raw file](https://github.com/dgraph-io/dgraph/blob/87e63e9ce8787ca76b27053dcc4399f6a5b8c385/posting/hash.go#L85)):*
> ```Go
> 	for k, v := range s.m {
> 		delete(s.m, k)
> 		f(k, v)
> ```

LGTM. Nice, simple and clean, love it!

---

*[posting/lists.go, line 158 at r8](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KSEiwDJ2HjkERIXmhno:-KSEiwDJ2HjkERIXmhnp:b-n0i71c) ([raw file](https://github.com/dgraph-io/dgraph/blob/87e63e9ce8787ca76b27053dcc4399f6a5b8c385/posting/lists.go#L158)):*
> ```Go
> 				case gentleMergeChan <- struct{}{}:
> 					n := int(float64(len(dirtyMap)) * *gentleMergeFrac)
> 					keysBuffer := make([]uint64, 0, n)
> ```

You can potentially move this outside of for loop, and only resize the slice if it's cap is less than n. Also, another memory opt might be to resize it to max of (2*current cap, n) -- to allow for n to vary a bit, without having to reallocate memory.

---

*[posting/lists.go, line 247 at r8](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KSEiRL45gC4szvUPT8k:-KSEiRL5G-Nrc_e7ZGmk:b-hdxwrw) ([raw file](https://github.com/dgraph-io/dgraph/blob/87e63e9ce8787ca76b27053dcc4399f6a5b8c385/posting/lists.go#L247)):*
> ```Go
> 		} else {
> 			// Gentle merge should not happen during an aggressive evict.
> 			dirtyMapOpChan <- dirtyMapOpGentleMerge
> ```

This might block until a current merge returns. Which is alright, because the ticker would then drop a tick. But, just mention it here, so it's clear.

---

*[posting/lists.go, line 306 at r8](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KSEjh1NB4dABF2saEYH:-KSEjh1NB4dABF2saEYI:b-bv341k) ([raw file](https://github.com/dgraph-io/dgraph/blob/87e63e9ce8787ca76b27053dcc4399f6a5b8c385/posting/lists.go#L306)):*
> ```Go
> func MergeLists(numRoutines int) {
> 	// We're merging all the lists, so just create a new dirtymap.
> 	dirtyMapOpChan <- dirtyMapOpReset
> ```

Mention here that this operation would block in case we've a pending gentle merging, because OpChan has a capacity of 1.

---

*[posting/lists.go, line 335 at r8](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KSEkRIYncw9L1OK53p8:-KSEkRIYncw9L1OK53p9:b-4x8mva) ([raw file](https://github.com/dgraph-io/dgraph/blob/87e63e9ce8787ca76b27053dcc4399f6a5b8c385/posting/lists.go#L335)):*
> ```Go
> 	})
> 	close(workChan)
> 	wg.Wait()
> ```

LGTM. Love it! Simple and elegant.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-:-KSEkpP2YQJuLog8EHvh:b-89nags)*

jchiu0:
Review status: 2 of 3 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[posting/lists.go, line 158 at r8](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KSEiwDJ2HjkERIXmhno:-KSEnllywx-cSkqtjItf:b-r069oy) ([raw file](https://github.com/dgraph-io/dgraph/blob/87e63e9ce8787ca76b27053dcc4399f6a5b8c385/posting/lists.go#L158)):*
> You can potentially move this outside of for loop, and only resize the slice if it's cap is less than n. Also, another memory opt might be to resize it to max of (2*current cap, n) -- to allow for n to vary a bit, without having to reallocate memory.
</details>
Done. Great memory opt.

---

*[posting/lists.go, line 247 at r8](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KSEiRL45gC4szvUPT8k:-KSEpRrEYDa3JvY6OV6n:b-bizqh8) ([raw file](https://github.com/dgraph-io/dgraph/blob/87e63e9ce8787ca76b27053dcc4399f6a5b8c385/posting/lists.go#L247)):*
> This might block until a current merge returns. Which is alright, because the ticker would then drop a tick. But, just mention it here, so it's clear.
</details>
Added comment.

---

*[posting/lists.go, line 306 at r8](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KSEjh1NB4dABF2saEYH:-KSEpfznX8S5cRGkNq7H:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/87e63e9ce8787ca76b27053dcc4399f6a5b8c385/posting/lists.go#L306)):*
> Mention here that this operation would block in case we've a pending gentle merging, because OpChan has a capacity of 1.
</details>
Done.

---

*[posting/lists.go, line 335 at r8](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KSEkRIYncw9L1OK53p8:-KSEpkchPRGY27PQMhB9:b-ucfknh) ([raw file](https://github.com/dgraph-io/dgraph/blob/87e63e9ce8787ca76b27053dcc4399f6a5b8c385/posting/lists.go#L335)):*
> LGTM. Love it! Simple and elegant.
</details>
Thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/222)*

manishrjain:
Reviewed 11 of 11 files at r9.

---

*[posting/lists.go, line 162 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KSFW3XaHWP3yOkhFAKV:-KSFW3XaHWP3yOkhFAKW:b-ijurdw) ([raw file](https://github.com/dgraph-io/dgraph/blob/dc9b18d22594ebaab7bb41bb891de7a364ec5bfe/posting/lists.go#L162)):*
> ```Go
> 						// Resize keysBuffer to max(2*current cap, n).
> 						newCap := n
> 						if cap(keysBuffer) > newCap {
> ```

logical flaw? You just checked that cap is less than n, and newCap = n, hence cap should remain less than n.

---

*[posting/lists.go, line 309 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KSFWjVPMrckXtD9VLIR:-KSFWjVPMrckXtD9VLIS:b-5n93hw) ([raw file](https://github.com/dgraph-io/dgraph/blob/dc9b18d22594ebaab7bb41bb891de7a364ec5bfe/posting/lists.go#L309)):*
> ```Go
> 	l := getNew() // This retrieves a new *List and increments its ref count.
> 	lp = lhmap.PutIfMissing(fp, l)
> 	// Actually we could do lp.incr here, and remove the two incr calls below, but
> ```

Yeah, let's do the incr here. I think might be easier to read.

So,
lp = lhmap.Put..
lp.Inr()

if lp == l {
  lp.init(..)  // Just added a new one, so init.
} else {
l.decr() // unable to insert l, so let it go back to pool.
}

return lp, lp.decr

---

*[posting/lists.go, line 360 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KSFXxgGCCFuac_mDPiG:-KSFXxgGCCFuac_mDPiH:b-agkutk) ([raw file](https://github.com/dgraph-io/dgraph/blob/dc9b18d22594ebaab7bb41bb891de7a364ec5bfe/posting/lists.go#L360)):*
> ```Go
> 			for l := range workChan {
> 				l.SetForDeletion() // No more AddMutation.
> 				mergeAndUpdate(l, c)
> ```

Add l.decr() after this line. You're done with it, so it's ready for reuse.

---

*[posting/lists.go, line 369 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KSFXVzgNd1cx1l0J2O2:-KSFXVzgNd1cx1l0J2O3:b-7jabth) ([raw file](https://github.com/dgraph-io/dgraph/blob/dc9b18d22594ebaab7bb41bb891de7a364ec5bfe/posting/lists.go#L369)):*
> ```Go
> 			return
> 		}
> 		l.decr() // List is now deleted from lhmap. Decrement reference counter.
> ```

I think you shouldn't do this. You removed it from map, but then gave it to workChan. So, essentially, it's one decr() and then one incr(). Which is essentially zero.

if you do the decr upfront, then the l becomes eligible for reuse, because nothing else is holding a reference to it. The user isn't defined yet.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/222)*

jchiu0:
Review status: 11 of 12 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[posting/lists.go, line 162 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KSFW3XaHWP3yOkhFAKV:-KSFaAK7urT-e-hus__N:bxaml80) ([raw file](https://github.com/dgraph-io/dgraph/blob/dc9b18d22594ebaab7bb41bb891de7a364ec5bfe/posting/lists.go#L162)):*
> logical flaw? You just checked that cap is less than n, and newCap = n, hence cap should remain less than n.
</details>
I forgot the factor 2. It should have been 2*cap(keysBuffer) > newCap. Fixed.

---

*[posting/lists.go, line 309 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KSFWjVPMrckXtD9VLIR:-KSFdDSwA2wPMBEVIwgg:bbst1r3) ([raw file](https://github.com/dgraph-io/dgraph/blob/dc9b18d22594ebaab7bb41bb891de7a364ec5bfe/posting/lists.go#L309)):*
> Yeah, let's do the incr here. I think might be easier to read.
> 
> So,
> lp = lhmap.Put..
> lp.Inr()
> 
> if lp == l {
>   lp.init(..)  // Just added a new one, so init.
> } else {
> l.decr() // unable to insert l, so let it go back to pool.
> }
> 
> return lp, lp.decr
> 
</details>
Nice refactoring!

---

*[posting/lists.go, line 360 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KSFXxgGCCFuac_mDPiG:-KSFcf7LNvyIK_xjghzy:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/dc9b18d22594ebaab7bb41bb891de7a364ec5bfe/posting/lists.go#L360)):*
> Add l.decr() after this line. You're done with it, so it's ready for reuse.
</details>
Done.

---

*[posting/lists.go, line 369 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KSFXVzgNd1cx1l0J2O2:-KSFcfdJ3hH4prf1aTsv:bcti8yq) ([raw file](https://github.com/dgraph-io/dgraph/blob/dc9b18d22594ebaab7bb41bb891de7a364ec5bfe/posting/lists.go#L369)):*
> I think you shouldn't do this. You removed it from map, but then gave it to workChan. So, essentially, it's one decr() and then one incr(). Which is essentially zero.
> 
> if you do the decr upfront, then the l becomes eligible for reuse, because nothing else is holding a reference to it. The user isn't defined yet.
</details>
Great catch! (Thanks for explaining in so much detail.) I wasn't careful enough here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/222)*

manishrjain:
Review status: 11 of 12 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[posting/lists.go, line 356 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KSFldHt3Cyjw423jqYK:-KSFldHt3Cyjw423jqYL:b-m38kc7) ([raw file](https://github.com/dgraph-io/dgraph/blob/36e27f3eeed9aad181119a3da222a6268c267238/posting/lists.go#L356)):*
> ```Go
> 			for l := range workChan {
> 				l.SetForDeletion() // No more AddMutation.
> 				l.decr()
> ```

This should be after you run the merge and update, not before. The reason it didn't throw an error is because we're not allowing any new lists to be created; otherwise, you'd have seen the reuse by someone else.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/222)*

jchiu0:
Review status: 11 of 12 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[posting/lists.go, line 356 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-KSFldHt3Cyjw423jqYK:-KSFnmL1cBKbh57XPzQd:b-vruya9) ([raw file](https://github.com/dgraph-io/dgraph/blob/36e27f3eeed9aad181119a3da222a6268c267238/posting/lists.go#L356)):*
> This should be after you run the merge and update, not before. The reason it didn't throw an error is because we're not allowing any new lists to be created; otherwise, you'd have seen the reuse by someone else.
</details>
Sorry, it should be after. Fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/222)*

manishrjain:
:lgtm: Nice work! Feel free to submit. Glad to see us move away from gotomic.

---

Reviewed 1 of 1 files at r12, 1 of 1 files at r13.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-:-KSGBgLGJREWazULV2dL:b-r39gm2)*

jchiu0:
Thanks for the review!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/222#-:-KSGCle6M0KG1lntSQoI:b-9qdoz3)*

PR #227, opened 2016-09-21
Feature/filterdirective

jchiu0:
Changes are restricted to gql package in this PR.

Support @filter in gql.

Sample filters:
@filter( f("aaa", "bbb") || g("ccc") && d() )

We do respect operator precedence. Now, there are only two operators: && and ||. And && takes precedence.

We currently do not support functions inside functions, e.g., @filter(f(g("aa"))).

The output of the parser is a FilterTree object inside a GraphQuery object. This is a parse tree and it should be quite easy to use it from the query package. We defer modifications in query package to another PR.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/227)


gql/parser.go
```diff
@@ -562,20 +580,182 @@ func parseArguments(l *lex.Lexer) (result []pair, rerr error) {
...
+// String converts FilterTree to a string. Good for testing, debugging.
+func (t *FilterTree) String() string {
+	buf := bytes.NewBuffer(make([]byte, 0, 20))
+	t.stringHelper(buf)
+	return buf.String()
+}
+
+// stringHelper does simple DFS to convert FilterTree to string.
+func (t *FilterTree) stringHelper(buf *bytes.Buffer) {
+	if t == nil {
```
manishrjain: This makes little sense. A function on t would crash, if t is nil, so this would never be called.

gql/parser.go
```diff
@@ -562,20 +580,182 @@ func parseArguments(l *lex.Lexer) (result []pair, rerr error) {
 	return result, nil
 }
 
+// String converts FilterTree to a string. Good for testing, debugging.
+func (t *FilterTree) String() string {
```
manishrjain: Is this for debugging purposes? Because you can avoid string conversion in production run.

If this is for debug purposes, then call it debugString(), and make it internal.

gql/parser_test.go
```diff
@@ -827,3 +835,267 @@ func TestParseVariablesError7(t *testing.T) {
...
 }
+
+func TestParseFilter_simplest(t *testing.T) {
+	query := `
+	query {
+		me(_uid_:0x0a) {
+			friends @filter() {
+				name @filter(namefilter())
+			}
+			gender @filter(eq()),age @filter(neq("a", "b"))
```
manishrjain: We shouldn't have @filter inside @filter, unless there's a particular case that can't be solved by boolean functions.

gql/parser_test.go
```diff
@@ -827,3 +835,267 @@ func TestParseVariablesError7(t *testing.T) {
...
+		return
+	}
+}
+
+// Test operator precedence. More elaborate brackets.
+func TestParseFilter_brac(t *testing.T) {
+	query := `
+	query {
+		me(_uid_:0x0a) {
+			friends @filter(  a() || b() && (c() || (d() || e())) && f()){
```
manishrjain: Would be good to also have args in some of these. Will test the logic a bit more thoroughly.

gql/state.go
```diff
@@ -131,12 +140,128 @@ func lexInside(l *lex.Lexer) lex.StateFn {
...
+func lexFilterInside(l *lex.Lexer) lex.StateFn {
+	for {
+		r := l.Next()
+		switch {
+		case r == leftRound:
+			l.Emit(itemLeftRound)
+			l.FilterDepth++
+		case r == rightRound:
+			if l.FilterDepth == 0 {
+				return l.Errorf("Unexpected )")
```
manishrjain: Just call it "right round bracket", makes the error a bit more clear.

gql/state.go
```diff
@@ -131,12 +140,128 @@ func lexInside(l *lex.Lexer) lex.StateFn {
...
+
+// lexDirective is called right after we see a @.
+func lexDirective(l *lex.Lexer) lex.StateFn {
+	r := l.Next()
+	if !isNameBegin(r) {
+		return l.Errorf("Unrecognized character in lexDirective: %#U", r)
+	}
+
+	l.Backup()
+	buf := bytes.NewBuffer(make([]byte, 0, 15))
```
manishrjain: Add a comment about why this buf wouldn't exceed 15 bytes (or can it resize the underlying slice automatically?)

x/init.go
```diff
@@ -64,3 +67,13 @@ To say hi to the community       , visit https://dgraph.slack.com.
 		os.Exit(0)
 	}
 }
+
+// Printf does a log.Printf. We often do printf for debugging but has to keep
+// adding import "fmt" or "log" and removing them after we are done.
+// Let's add Printf to "x" and include "x" almost everywhere. Caution: Do remember
+// to call x.Init. For tests, you need a TestMain that calls x.Init.
+func Printf(format string, args ...interface{}) {
```
manishrjain: Or, you could just use your debugflag and make all the traces do a log.Printf(..), which is what I did recently. That way, you'd only add traces always, even while debugging -- which is what we should be using. Don't have any strong feelings about this, so just a thought. Feel free to ignore.

gql/parser.go
jchiu0: Done. Yes, it is internal and should be debugString.

gql/parser.go
jchiu0: A function on nil wouldn't crash in Go.. (recall the proverb: Make the zero value useful.)
It was there when I print out DebugStrings for all filters (even if they are nil).
It is no longer needed. Can I do an assert here to be sure?

gql/parser_test.go
jchiu0: I suppose you are referring to name @filter inside friends @filter.

I see that as a useful construct. You apply a filter on friends to select friends you want to view. And then a friend might have 1000 names, so maybe you want to add another filter on the name predicate. I thought that was included in the design in https://discuss.dgraph.io/t/filters-in-graphql/831/26 ?

gql/parser_test.go
jchiu0: Done.

gql/state.go
jchiu0: Done.

gql/state.go
jchiu0: Yes, it can resize automatically. I tested it out. Added a comment.

x/init.go
jchiu0: Thanks for the suggestion. I agree we should move towards using one single log function. However, x.Trace needs a ctx object and a lot of our functions don't have that yet.... And sometimes, for long running processes in the background, we use something else. So, I am not sure what is the best long term solution here.. For now, let me just use this one? We can do that in a separate PR..

manishrjain:
:lgtm: Looks good. Well thought!

---

Reviewed 1 of 5 files at r1, 1 of 1 files at r2, 3 of 3 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[gql/parser.go, line 592 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/227#-KSA0lLV3lh03hSQgiFq-r2-592:-KS_6iYVZlafPbhhkIT1:b-73e5tm) ([raw file](https://github.com/dgraph-io/dgraph/blob/8529b779efd6aec8c9fe577347a053edaca8b72b/gql/parser.go#L592)):*
> A function on nil wouldn't crash in Go.. (recall the proverb: Make the zero value useful.)
> 
> It was there when I print out DebugStrings for all filters (even if they are nil).
> 
> It is no longer needed. Can I do an assert here to be sure?
</details>
LGTM. Wao, filed under TIL. Assert sounds good. Though, this doesn't explain the panics that we see sometimes if a struct is nil and not properly initialized.

---

*[gql/parser_test.go, line 846 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/227#-KSA0lLWn8nSvkPaboBS-r2-846:-KS_8z5tC8RsEt6H3nZ5:b-jywflq) ([raw file](https://github.com/dgraph-io/dgraph/blob/8529b779efd6aec8c9fe577347a053edaca8b72b/gql/parser_test.go#L846)):*
> I suppose you are referring to name @filter inside friends @filter.
> 
> 
> 
> I see that as a useful construct. You apply a filter on friends to select friends you want to view. And then a friend might have 1000 names, so maybe you want to add another filter on the name predicate. I thought that was included in the design in https://discuss.dgraph.io/t/filters-in-graphql/831/26 ?
</details>
OK. Sorry my bad. I read the age filter as if it was inside the gender filter. friends and name is obviously ok.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/227#-:-KS_9SWzKwD6w1aUt3F3:bq4lj5c)*

PR #238, opened 2016-09-26
Miscel/xtracestacktrace

jchiu0:
Introducing `x.TraceError` which takes in an `error` object, and if in debug mode, will produce a shortened stack trace.

We also replaced `x.Trace(ctx, fmt.Errorf("Some error: %v", err))` with `x.TraceError(ctx, x.Wrapf(err, "Some error"))`

(Note that in non-debug mode, this will print the same error string: `x.Wrapf(fmt.Errorf("Original error"), "Wrapped").Error()` is equal to `Wrapped: Original error`.

We also replaced `x.Trace(ctx, fmt.Errorf("..."))` with `x.TraceError(ctx, x.Errorf("..."))`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/238)


jchiu0:

---

*[x/error.go, line 111 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/238#-KSdZ9nyIbR4iiwwsQfJ:-KSeB2PJK7Fg_-qz9_8t:b-5n5ah1) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c061d346a3429f34dd733342e6fda21c7b6d6bf/x/error.go#L111)):*
> Is 40 the max characters this can take? We may exceed this if there are errors from external libraries, but I guess that should be ok.
</details>
This is just the initial buffer. When it exceeds, the buffer will be "grown" like a standard array.

---

*[x/error.go, line 148 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/238#-KSdYlPb8knXLDvIDE9s:-KSeAfGyVWVBZTSucTG4:b5ir71u) ([raw file](https://github.com/dgraph-io/dgraph/blob/9c061d346a3429f34dd733342e6fda21c7b6d6bf/x/error.go#L148)):*
> Just wondering if we would have this situtation. Won't lineref and !lineref alternate?
</details>
With x.Wrapf, sometimes you add additional error messages to one stack trace. If you print out, you will first see two or more lines followed by the stack trace. We need to handle this, and I thought why not handle the slightly more general case as well..

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/238)*

jchiu0:
@pawan Thanks for the review!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/238#-:-KSeBJ9f2dmp0SYNTiTa:bukepck)*

PR #245, opened 2016-09-27
Feature/exactmatch

jchiu0:
`gql.FilterTree` is available. In this PR, we use it to apply the filter to our results. Main functions modified are in query.go, particularly `processGraph`, `postTraverse` and `preTraverse`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/245)


manishrjain:
Reviewed 3 of 7 files at r1, 6 of 6 files at r2.

---

*[algo/sorted.go, line 51 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSpz8_rKJ9HxzCfOnUS:-KSpz8_rKJ9HxzCfOnUT:bbjmd8h) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/algo/sorted.go#L51)):*
> ```Go
> 
> // String returns the UIDList as a string. Good for printing the list.
> func (ul *UIDList) String() string {
> ```

Maybe call it DebugString?, or just Debug()?

---

*[algo/sorted.go, line 80 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSq6fCSUQQKDXlgIinc:-KSq6fCSUQQKDXlgIind:b-14k863) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/algo/sorted.go#L80)):*
> ```Go
> 
> // Size returns number of lists.
> func (s PlainUintLists) Size() int { return len(s) }
> ```

I'd say avoid using PlainUintLists -- they encourage using slices instead of encoded FB lists directly. I have some suggestions that might apply to your implementation which can help avoid this.

---

*[gql/state.go, line 64 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSq0P6YTTDNO09OFPEV:-KSq0P6YTTDNO09OFPEW:bnlppbg) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/gql/state.go#L64)):*
> ```Go
> 	itemAlias                                   // Alias for a field
> 	itemDirectiveName                           // Name starting with @
> 	ItemFilterAnd                               // And inside a filter.
> ```

These are very specific to lexer state. I don't think these should be exposed.

---

*[query/query.go, line 148 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSq-npnYUv4D5Tujbf2:-KSq-npnYUv4D5Tujbf3:bed3gp8) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/query/query.go#L148)):*
> ```Go
> 	// matrix for this node does not take into account the filtering. Hence, during
> 	// postTraverse or preTraverse, we need to intersect with sg.sorted.
> 	sorted []uint64
> ```

Why do we need to store it? I guess what you're doing is, looking at the sorted to check if the result should be shown or not. But, the absence of the UID in the SG children can just as well be used to realize that, I think.

---

*[query/query.go, line 244 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSq1farRF-49lZINKOz:-KSq1farRF-49lZINKP-:b32tj58) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/query/query.go#L244)):*
> ```Go
> 		for j := 0; j < ul.UidsLength(); j++ {
> 			uid := ul.Uids(j)
> 			for ; sortedIdx < len(sg.sorted) && sg.sorted[sortedIdx] < uid; sortedIdx++ {
> ```

I think an absence could be used:

if ival, present := cResult[uid]; !present {
  // ignore. instead of append.
}

---

*[query/query.go, line 657 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSq-RwMI1WSA7ISmDBz:-KSq-RwMI1WSA7ISmDC-:btv4bn0) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/query/query.go#L657)):*
> ```Go
> 
> // createTaskQuery generates the query buffer.
> func createTaskQuery(sg *SubGraph, uids []uint64, terms []string) []byte {
> ```

Why is terms an argument? That seems specific to term matching -- which is just one of the operations available in filters. How would this extend later, when we have say date ranges or other things?

Have a think about it. Might be that this is the best way to build right now, which would be fine, as long as we have an eye towards how the design would handle all the various searches that could be done (that we know about right now).

---

*[query/query.go, line 803 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSq2oGzwrgNos2UdhDU:-KSq2oGzwrgNos2UdhDV:b1n1gwo) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/query/query.go#L803)):*
> ```Go
> // filterHelper traverses filter tree and produce a filtered list of UIDs.
> // Input "sorted" is the very original list of destination UIDs of a SubGraph.
> func filterHelper(ctx context.Context, sorted []uint64, filter *gql.FilterTree) ([]uint64, error) {
> ```

runFilter is probably a better name for this function.

---

*[query/query.go, line 811 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSq67Ad9h8zexsmA25O:-KSq67Ad9h8zexsmA25P:b-sbopsb) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/query/query.go#L811)):*
> ```Go
> 		attr := filter.FuncArgs[0]
> 		sg := &SubGraph{Attr: attr}
> 		sg.Query = createTaskQuery(sg, nil, []string{filter.FuncArgs[1]})
> ```

If you sent `sorted` to ProcessGraph via the query, then the intersection can happen directly on the server. This would result in potentially lot fewer results being sent back.

---

*[query/query.go, line 812 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSq3GQKWAA45lq7hZus:-KSq3GQLP_zfocp9OYjP:b71k5ro) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/query/query.go#L812)):*
> ```Go
> 		sg := &SubGraph{Attr: attr}
> 		sg.Query = createTaskQuery(sg, nil, []string{filter.FuncArgs[1]})
> 		sgChan := make(chan error)
> ```

Have at least a buffer size of 1, otherwise, the input and output would block for each other.

---

*[query/query.go, line 827 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSq5qa7WFliIv5r2idi:-KSq5qa8UCcepH5ngk16:btdczxs) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/query/query.go#L827)):*
> ```Go
> 		}
> 		// Create a GenericLists. Note there is no copying here.
> 		lists := algo.GenericLists{[]algo.Uint64List{algo.PlainUintList(sorted), ul}}
> ```

It looks like a nice implementation. But, if we can avoid converting the FB uid results to slices, that could save a lot of memory. In fact, if you send `sorted` in the query to the server itself, then the result is already an intersection and hence you can just pick it up straight away.

---

*[query/query.go, line 832 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSpzkm0VNjXSWtctyMe:-KSpzkm0VNjXSWtctyMf:bah93cc) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/query/query.go#L832)):*
> ```Go
> 
> 	// For now, we only handle AND and OR.
> 	if filter.Op != gql.ItemFilterAnd && filter.Op != gql.ItemFilterOr {
> ```

This feels strange -- query package is depending upon a lexer state within gql. I think using strings is better.

---

*[query/query.go, line 838 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSq-3yQCyOMcCEhgusf:-KSq-3yQCyOMcCEhgusg:b-lkfv7r) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/query/query.go#L838)):*
> ```Go
> 	// Get UIDs for child filters.
> 	type resultPair struct {
> 		uid []uint64
> ```

Don't you get results encoded in Flatbuffer list?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/245)*

jchiu0:
Review status: 3 of 26 files reviewed at latest revision, 49 unresolved discussions, some commit checks failed.

---

*[algo/sorted.go, line 51 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSpz8_rKJ9HxzCfOnUS:-KStsDpv7NEavuzA63Gb:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/algo/sorted.go#L51)):*
> Maybe call it DebugString?, or just Debug()?
</details>
Done.

---

*[algo/sorted.go, line 80 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSq6fCSUQQKDXlgIinc:-KTCFWKXmOWVtYTvmZQO:bv0w5qe) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/algo/sorted.go#L80)):*
> I'd say avoid using PlainUintLists -- they encourage using slices instead of encoded FB lists directly. I have some suggestions that might apply to your implementation which can help avoid this.
</details>
A lot of operations (filter, intersect, union, pagination etc) can be composed with one another. This requires intermediate structures and copying / slices is unavoidable. In this case, it is awkward to use FB instead of a simple []uint64.

In this PR, I turn "Result" into an array of algo.Uint64List which can be either task.UidList or []uint64. If there is no filter and no pagination, it will remain as task.UidList and there will be no copying. The structure algo.Uint64List is very generic and very basic, and most importantly, it doesn't itself support slicing. So folks are still discouraged from copying unless they have to do some work.

For pagination, we have at least two choices. One is to not re-copy and just use algo.Uint64ListSlice, which is a slice into algo.Uint64List, an interface. Alternatively, we can do a little copying and do away with algo.Uint64ListSlice. This can potentially free up a lot of memory. The current choice is to do a little copying, and if you are ok with this, I will delete algo.Uint64ListSlice.

---

*[query/query.go, line 148 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSq-npnYUv4D5Tujbf2:-KTCIaIpZ7FjXSOrXJ4r:b6lwjuj) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/query/query.go#L148)):*
> Why do we need to store it? I guess what you're doing is, looking at the sorted to check if the result should be shown or not. But, the absence of the UID in the SG children can just as well be used to realize that, I think.
</details>
As discussed, there are nodes with no children but with a field like `_uid_`. In this case, there is no children to store these UIDs.

---

*[query/query.go, line 244 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSq1farRF-49lZINKOz:-KTCIn0ZF4bBqjb-9AcJ:b-8olwa6) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/query/query.go#L244)):*
> I think an absence could be used:
> 
> if ival, present := cResult[uid]; !present {
>   // ignore. instead of append.
> }
</details>
As discussed, this does not work when there is no children at all but we are looking to get `_uid_`.

---

*[query/query.go, line 657 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSq-RwMI1WSA7ISmDBz:-KTCOLATqXg4CI_15Mg7:b-oxd21q) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/query/query.go#L657)):*
> Why is terms an argument? That seems specific to term matching -- which is just one of the operations available in filters. How would this extend later, when we have say date ranges or other things?
> 
> Have a think about it. Might be that this is the best way to build right now, which would be fine, as long as we have an eye towards how the design would handle all the various searches that could be done (that we know about right now).
</details>
Thanks for pointing out. It might be good to replace both uids, terms with "keys" in the future.

---

*[query/query.go, line 803 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSq2oGzwrgNos2UdhDU:-KSu7Y-wc-2MY50dyWkk:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/query/query.go#L803)):*
> runFilter is probably a better name for this function.
</details>
Done.

---

*[query/query.go, line 811 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSq67Ad9h8zexsmA25O:-KTCOidrjnC0ZEbe-Hu_:bg95te3) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/query/query.go#L811)):*
> If you sent `sorted` to ProcessGraph via the query, then the intersection can happen directly on the server. This would result in potentially lot fewer results being sent back.
</details>
Done. Good point. (Eventually, we might want to check which is smaller in order to decide the optimal execution plan.)

---

*[query/query.go, line 812 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSq3GQKWAA45lq7hZus:-KSu7fL9nVj5iAIyM5wn:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/query/query.go#L812)):*
> Have at least a buffer size of 1, otherwise, the input and output would block for each other.
</details>
Done.

---

*[query/query.go, line 827 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSq5qa7WFliIv5r2idi:-KTCOww9hoIFY34OwMYO:buio9ao) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/query/query.go#L827)):*
> It looks like a nice implementation. But, if we can avoid converting the FB uid results to slices, that could save a lot of memory. In fact, if you send `sorted` in the query to the server itself, then the result is already an intersection and hence you can just pick it up straight away.
</details>
As discussed, sg.Sorted (now renamed as destUIDs) is the union of all "branches". We still need to intersect it with each "branch" and it is more convenient to keep it as a []uint64. If you like, we can use algo.Uint64 here as well, to discourage copying.

---

*[query/query.go, line 832 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSpzkm0VNjXSWtctyMe:-KTCPQUfaMfZeqEejF9W:b-y8e977) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/query/query.go#L832)):*
> This feels strange -- query package is depending upon a lexer state within gql. I think using strings is better.
</details>
I moved it to gql. I hope this is fine. "query" depends on "gql", so if you define in "query", it can't be used in gql and you would need to do another translation of enum from gql to query.

---

*[query/query.go, line 838 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSq-3yQCyOMcCEhgusf:-KTCPvpLZged-6r1H9HN:bh7xq9n) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/query/query.go#L838)):*
> Don't you get results encoded in Flatbuffer list?
</details>
The general case is that filter can be a bunch of ANDs and ORs composed with one another, which would require calls to algo.MergeSorted or algo.IntersectSorted, which all return []uint64 not flatbuffers. We could certainly allow these functions to write out a FB. Is that what you prefer?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/245)*

jchiu0:
In this major update, we got rid of task.Query, task.Result in SubGraph. We also renamed sg.Sorted as sg.destUIDs.

As discussed, we need to have sg.destUIDs for the case where there is no child and the user is looking for `_uid_`.

It is also reduced in size as much as possible before we fan out in a `processGraph` call.

Since sg.destUIDs is there, we might as well reuse it in child nodes. Instead of `task.Query`, child nodes now store sg.srcUIDs which are copies of parentSg.destUIDs. There is no additional memory allocated.

Here are four combinations (there can be many more). But let's just consider these four now.

No pagination, no filter: There is no extra copying. Everything remains as flatbuffers.

No pagination, has filter: After filtering is done on `sg.destUIDs` (aka `sg.sorted`), there is no more additional copying for the current node. During post-traversal and pre-traversal, we check against `sg.destUIDs` to perform the intersection. This saves an additional copy.

Has pagination, no filter: This will cause `sg.Result` (previously `sg.Result.UIDMatrix`) to be rebuilt and `sg.destUIDs` to be rebuilt based on the UIDMatrix. We might want to re-think this case. Perhaps, when there is no filtering, we can do pagination at worker level like we do before. But of course, code is going to get complex.

Has pagination, has filter: This will cause `sg.Result` to be rebuilt after pagination. In this case, there is no choice. The filtering is done on `sg.destUIDs` which is the union of all branches. Before pagination can be performed per branch, we have to intersect each branch with the filtered `sg.destUIDs`. We might as well save a copy.

---

Review status: 3 of 26 files reviewed at latest revision, 49 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-:-KTCQiLFcV5z_od5fPKq:bcxo9ym)*

manishrjain:
They say reviewing a PR takes half the time as it takes to write that PR. I doubt it's half, but it's definitely a good proportion :-). Looks much better now, can still improve.

---

Reviewed 2 of 11 files at r3, 4 of 10 files at r4, 7 of 13 files at r5, 2 of 4 files at r6, 1 of 5 files at r8, 1 of 2 files at r9, 7 of 7 files at r11, 1 of 1 files at r12.
Review status: all files reviewed at latest revision, 69 unresolved discussions, some commit checks failed.

---

*[task.fbs, line 34 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTI8YCBi_YY9inYghog:-KTI8YCCuA2mub6Q-yVO:b-qvz2yb) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/task.fbs#L34)):*
> ```
> 
> table CountList {
> 	count:[ulong];
> ```

count doesn't need to be ulong. Can be uint (32bit). Would bring the payload to half with this change.

---

*[algo/sorted.go, line 177 at r8](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTCVMKandHKxmpDo0z7:-KTCVMKandHKxmpDo0z8:bhlbrqg) ([raw file](https://github.com/dgraph-io/dgraph/blob/5b4e0ead7d3837a8b5ba47889ea6b7671b838542/algo/sorted.go#L177)):*
> ```Go
> 
> // IntersectSorted returns intersection of uint64 lists.
> // TODO(jchiu): If one list is a lot smaller than the others, it might be much
> ```

Stop worrying about really small lists, they'd be fast anyway. Worry about very long lists.

---

*[algo/uidlist.go, line 52 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTI98lbIzk5iJWOazB2:-KTI98lbIzk5iJWOazB3:bozt0gi) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L52)):*
> ```Go
> func (u *UIDList) Size() int {
> 	if u == nil {
> 		return 0
> ```

Shouldn't this be a log fatal? If we're doing this, something is wrong with our code, no?

---

*[algo/uidlist.go, line 61 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTI9G3L6D3RVT2gmpf3:-KTI9G3L6D3RVT2gmpf4:b-blvops) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L61)):*
> ```Go
> 
> // Reslice selects a slice of the data.
> func (u *UIDList) ApplyFilterFn(f func(uid uint64) bool) {
> ```

Just u.ApplyFilter(func()..). Otherwise, it reads, "apply filter func func".

---

*[algo/uidlist.go, line 62 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTI9UFZYbfOVYCEBWGz:-KTI9UFZYbfOVYCEBWH-:b-rgui7p) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L62)):*
> ```Go
> // Reslice selects a slice of the data.
> func (u *UIDList) ApplyFilterFn(f func(uid uint64) bool) {
> 	u.convertToUints()
> ```

No, no, don't do that. Filter can potentially greatly reduce the number of uints. You can have 2 code paths here. One for uints, which is how your code looks right now, and one for task.uidList, where you create a new slice.

---

*[algo/uidlist.go, line 95 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIAHB5b8g7R3ceSEHU:-KTIAHB5b8g7R3ceSEHV:bwz6nz) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L95)):*
> ```Go
> 
> // Intersect intersects with another list and updates this list.
> // TODO(jchiu): Consider binary search if n>>m or n<<m.
> ```

Seems like an unnecessary TODO. We can do this if this shows up in profiling. Otherwise, I doubt there'll be enough motivation to actually complete this TODO.

I think of TODOs as these dirty spots in the code. The fewer the better.

---

*[algo/uidlist.go, line 97 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIAjketzGztnxo1QuD:-KTIAjketzGztnxo1QuE:b-jq354d) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L97)):*
> ```Go
> // TODO(jchiu): Consider binary search if n>>m or n<<m.
> func (u *UIDList) Intersect(v *UIDList) {
> 	u.convertToUints()
> ```

No, please don't. You can do this instead:

var final []uint64
if u.list != nil {
 final = make([]uint64, 0, 10)
} else {
 final = u.uints[:0]
}

Then append to final.
And in the end:

u.list = nil
u.uints = final

---

*[algo/uidlist.go, line 116 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTI9pLgkBNA8WhqoaYv:-KTI9pLhx1Y1vo0D7kuJ:b59iwqc) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L116)):*
> ```Go
> // convertToUints converts internal representation to uints which is easier to
> // work with than flatbuffers.
> func (u *UIDList) convertToUints() {
> ```

Get rid of this. Doing this conversion is probably unnecessary and expensive (also sort of defies the whole point of having this class).

---

*[algo/uidlist.go, line 131 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIBzEViKKujrRm67eL:-KTIBzEViKKujrRm67eM:bosiq3h) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L131)):*
> ```Go
> func IntersectLists(lists []*UIDList) *UIDList {
> 	if len(lists) == 0 {
> 		output := new(UIDList)
> ```

This should be a log fatal, no? Maybe not, just checking.

---

*[algo/uidlist.go, line 166 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTICawlHmfwOGL9X4YE:-KTICawlHmfwOGL9X4YF:b4wbsea) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L166)):*
> ```Go
> 
> 		var skip bool                     // Should we skip val in output?
> 		for j := 0; j < len(lists); j++ { // For each other list in lists.
> ```

So, essentially we're doing this in O(n*m), where we have n lists and m average size of each list.

This is better than Calling u.Intersect(l) on each list l, because that'd be (Lu*n + (n-1)*m), where Lu = length of smallest list. Is that right? If so, write a comment to explain that. Because the question is why write another intersection logic when you already have the Intersect function.

---

*[algo/uidlist.go, line 172 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIE91XbzLUUnIv8neH:-KTIE91XbzLUUnIv8neI:b-8pp0j8) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L172)):*
> ```Go
> 			}
> 			l := lists[j]
> 			k := idx[j]
> ```

l := lists[j]
lidx := idx[j]

---

*[algo/uidlist.go, line 173 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIE7Joc3M5YGY6c3FF:-KTIE7JpqYRa-MAUC47l:b-hjncsx) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L173)):*
> ```Go
> 			l := lists[j]
> 			k := idx[j]
> 			n := l.Size()
> ```

lsz := l.Size()

---

*[algo/uidlist.go, line 250 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIEe9J_1_XAMtk5j3x:-KTIEe9KXSNLNijigC80:b-g7fxpo) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L250)):*
> ```Go
> 
> // UidlistOffset adds a UidList to buffer and returns the offset.
> func (u *UIDList) UidlistOffset(b *flatbuffers.Builder) flatbuffers.UOffsetT {
> ```

call it: AddTo(b *flatbuffers.Builder)

---

*[gql/parser.go, line 91 at r8](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTCVg-wZgDrQ2mWsUho:-KTCVg-wZgDrQ2mWsUhp:b-o248hs) ([raw file](https://github.com/dgraph-io/dgraph/blob/5b4e0ead7d3837a8b5ba47889ea6b7671b838542/gql/parser.go#L91)):*
> ```Go
> const (
> 	FilterOpAnd       = iota
> 	FilterOpOr        = iota
> ```

You don't need to set `= iota` again, I think.

---

*[query/query.go, line 832 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSpzkm0VNjXSWtctyMe:-KTIL9NIVsuO7fc032Aw:b-ba6nle) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/query/query.go#L832)):*
> I moved it to gql. I hope this is fine. "query" depends on "gql", so if you define in "query", it can't be used in gql and you would need to do another translation of enum from gql to query.
</details>
I think query is the generic module, which executes internal format of instructions. GQL is just one language -- there would be others later. Still not convinced that query should depend on GQL. You could literally just have two strings, AND and OR, and that's all you really need to pass info from GQL to Query. So, it's pretty straightfwd to remove this dep.

---

*[query/query.go, line 694 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIIpF6GBbfbXqo9u_W:-KTIIpF6GBbfbXqo9u_X:b-nrwy2e) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/query/query.go#L694)):*
> ```Go
> 		buf := b.FinishedBytes()
> 		tl := new(task.UidList)
> 		x.ParseUidList(tl, buf)
> ```

With the updates in FB for Go, you can now use:
tl := task.GetRootAsUidList(buf, 0)

---

*[query/query.go, line 785 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIJQH_rvAnc6hWZARr:-KTIJQH_rvAnc6hWZARs:b-ce7e3h) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/query/query.go#L785)):*
> ```Go
> 
> 		r := new(task.Result)
> 		x.ParseTaskResult(r, resultBuf)
> ```

r := task.GetRootAsResult(resultBuf, 0)

---

*[query/query.go, line 848 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIKHELUP2emi7CRUI2:-KTIKHELUP2emi7CRUI3:bremhoe) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/query/query.go#L848)):*
> ```Go
> 		child := sg.Children[i]
> 		child.srcUIDs = sg.destUIDs // Make the connection.
> 		if child.Params.AttrType == nil || child.Params.AttrType.IsScalar() {
> ```

What is this checking for? Why?

---

*[query/query.go, line 853 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIKA3OrtlwgAfsTuUP:-KTIKA3OrtlwgAfsTuUQ:b-aqumlr) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/query/query.go#L853)):*
> ```Go
> 			go ProcessGraph(ctx, child, taskQuery, childChan)
> 		} else {
> 			leftToProcess = append(leftToProcess, child)
> ```

What's the logic behind leftToProcess? Add a comment.

---

*[query/query.go, line 970 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIKd2SApM3WHASDNj8:-KTIKd2SApM3WHASDNj9:b-hmbzks) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/query/query.go#L970)):*
> ```Go
> // runFilter traverses filter tree and produce a filtered list of UIDs.
> // Input "destUIDs" is the very original list of destination UIDs of a SubGraph.
> func runFilter(ctx context.Context, destUIDs *algo.UIDList, filter *gql.FilterTree) (*algo.UIDList, error) {
> ```

This goes beyond 100 chars.

---

*[query/query.go, line 1033 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIM37VQzbjxokR102a:-KTIM37VQzbjxokR102b:b-je7cs3) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/query/query.go#L1033)):*
> ```Go
> 	if p.Count < 0 {
> 		// Items from the back of the array, like Python arrays.
> 		return n + p.Count, n
> ```

n + p.Count could still be < 0.

---

*[worker/task.go, line 144 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIMdv2mdJgpXViGJjc:-KTIMdv2mdJgpXViGJjd:bbo009s) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/worker/task.go#L144)):*
> ```Go
> 			// Get taskQuery.Intersect field.
> 			taskList := new(task.UidList)
> 			if q.Intersect(taskList) != nil {
> ```

You're intersecting with taskList which is empty? What?

---

*[x/serialize.go, line 28 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTINBuhF8nREhBDEGNg:-KTINBuhF8nREhBDEGNh:b-viiui8) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/x/serialize.go#L28)):*
> ```Go
> // ParseValueList parses bytes into task.ValueList.
> func ParseValueList(q *task.ValueList, b []byte) {
> 	q.Init(b, flatbuffers.GetUOffsetT(b))
> ```

You don't need any of these now, with the new task.GetRootAs...

---

*[x/serialize.go, line 52 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTINJWqDnlAcN6MwDpU:-KTINJWqDnlAcN6MwDpV:bbdkaub) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/x/serialize.go#L52)):*
> ```Go
> 
> // UidlistOffset adds a UidList to buffer and returns the offset.
> func UidlistOffset(b *flatbuffers.Builder, sorted []uint64) flatbuffers.UOffsetT {
> ```

This is repetition of another function in UIDList.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-:-KTINSPV6ny4v94hWZFi:b-uij207)*

jchiu0:
Review status: 13 of 26 files reviewed at latest revision, 69 unresolved discussions, some commit checks failed.

---

*[task.fbs, line 34 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTI8YCBi_YY9inYghog:-KTJp6ndE7uO8yLbVaBY:b-qu6n94) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/task.fbs#L34)):*
> count doesn't need to be ulong. Can be uint (32bit). Would bring the payload to half with this change.
</details>
Good point. Done.

---

*[algo/sorted.go, line 177 at r8](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTCVMKandHKxmpDo0z7:-KTJoqysqAN-4WE8G_z8:b-cehkop) ([raw file](https://github.com/dgraph-io/dgraph/blob/5b4e0ead7d3837a8b5ba47889ea6b7671b838542/algo/sorted.go#L177)):*
> Stop worrying about really small lists, they'd be fast anyway. Worry about very long lists.
</details>
Ack. Sorry.

---

*[algo/uidlist.go, line 13 at r12](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTDWqGQh21mAHwVdZpB-r12-13:-KTJoq--bBZT8ClBjhB6:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/c051b320ffeec4db76a0b47449d7cfb5efbca766/algo/uidlist.go#L13)):*
> exported type UIDList should have comment or be unexported
</details>
Done.

---

*[algo/uidlist.go, line 34 at r12](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTDWqGQh21mAHwVdZpB-r12-34:-KTJopJy8MCFOTCKLCsl:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/c051b320ffeec4db76a0b47449d7cfb5efbca766/algo/uidlist.go#L34)):*
> comment on exported method UIDList.FromTask should be of the form "FromTask ..."
</details>
Done.

---

*[algo/uidlist.go, line 52 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTI98lbIzk5iJWOazB2:-KTJolwVVXR4zh3QaEWt:b-o9mdwh) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L52)):*
> Shouldn't this be a log fatal? If we're doing this, something is wrong with our code, no?
</details>
Agreed. I added a few more asserts.

---

*[algo/uidlist.go, line 61 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTI9G3L6D3RVT2gmpf3:-KTJoY2vt_hmJgfSxE_Z:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L61)):*
> Just u.ApplyFilter(func()..). Otherwise, it reads, "apply filter func func".
</details>
Done.

---

*[algo/uidlist.go, line 62 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTI9UFZYbfOVYCEBWGz:-KTJoVa24LQsX_dfiVZr:b93pe4l) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L62)):*
> No, no, don't do that. Filter can potentially greatly reduce the number of uints. You can have 2 code paths here. One for uints, which is how your code looks right now, and one for task.uidList, where you create a new slice.
</details>
Can use the slice trick like in Intersect. Done.

---

*[algo/uidlist.go, line 95 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIAHB5b8g7R3ceSEHU:-KTJoQFtQLzXxsn4i0A8:b-5dndzd) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L95)):*
> Seems like an unnecessary TODO. We can do this if this shows up in profiling. Otherwise, I doubt there'll be enough motivation to actually complete this TODO.
> 
> I think of TODOs as these dirty spots in the code. The fewer the better.
</details>
Agreed. I have the bad habit of adding TODOs. Will refrain from that.

---

*[algo/uidlist.go, line 97 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIAjketzGztnxo1QuD:-KTJaiRZVBwMIEljZrID:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L97)):*
> No, please don't. You can do this instead:
> 
> var final []uint64
> if u.list != nil {
>  final = make([]uint64, 0, 10)
> } else {
>  final = u.uints[:0]
> }
> 
> Then append to final.
> And in the end:
> 
> u.list = nil
> u.uints = final
> 
</details>
Done.

---

*[algo/uidlist.go, line 116 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTI9pLgkBNA8WhqoaYv:-KTJarBu-Adl9oBIZnKr:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L116)):*
> Get rid of this. Doing this conversion is probably unnecessary and expensive (also sort of defies the whole point of having this class).
</details>
Done.

---

*[algo/uidlist.go, line 131 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIBzEViKKujrRm67eL:-KTJatjVAJ3yHvNeOVAF:b-a2ddl7) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L131)):*
> This should be a log fatal, no? Maybe not, just checking.
</details>
Just checking is fine.

---

*[algo/uidlist.go, line 166 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTICawlHmfwOGL9X4YE:-KTJLWgsmpMz_Q52wsGd:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L166)):*
> So, essentially we're doing this in O(n*m), where we have n lists and m average size of each list.
> 
> This is better than Calling u.Intersect(l) on each list l, because that'd be (Lu*n + (n-1)*m), where Lu = length of smallest list. Is that right? If so, write a comment to explain that. Because the question is why write another intersection logic when you already have the Intersect function.
</details>
Done.

---

*[algo/uidlist.go, line 172 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIE91XbzLUUnIv8neH:-KTJLqa90tsPk9hmnhTh:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L172)):*
> l := lists[j]
> lidx := idx[j]
</details>
Done.

---

*[algo/uidlist.go, line 173 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIE7Joc3M5YGY6c3FF:-KTJLxAHAuXSvD0TaZfW:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L173)):*
> lsz := l.Size()
</details>
Done.

---

*[algo/uidlist.go, line 250 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIEe9J_1_XAMtk5j3x:-KTJLy7B28ghOpzrMmPX:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/algo/uidlist.go#L250)):*
> call it: AddTo(b *flatbuffers.Builder)
</details>
Done.

---

*[gql/parser.go, line 90 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSfglDsxq4v7ONp0ASz-r3-90:-KTDXK7GcfronJ6vhPQU:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/d0a2ef6e9d2a8d39db01c0cfb842a0a7850c7afb/gql/parser.go#L90)):*
> exported const FilterOpAnd should have comment (or a comment on this block) or be unexported
</details>
Done.

---

*[gql/parser.go, line 86 at r5](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSfglDsxq4v7ONp0ASz-r5-86:-KTDX4-2JsVWGKtWFQV7:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/fbe36a7c80fc350976240ff1bb7bf3d7bcacb612/gql/parser.go#L86)):*
> comment on exported type FilterOp should be of the form "FilterOp ..." (with optional leading article)
</details>
Done.

---

*[gql/parser.go, line 91 at r8](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTCVg-wZgDrQ2mWsUho:-KTJoDnWNJMj2XB5FR4m:b-vav7ve) ([raw file](https://github.com/dgraph-io/dgraph/blob/5b4e0ead7d3837a8b5ba47889ea6b7671b838542/gql/parser.go#L91)):*
> You don't need to set `= iota` again, I think.
</details>
NA. Enums removed.

---

*[posting/list.go, line 110 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KT8ZgI6r5Nj3ffNqSH9-r3-110:-KTJoCsrgRGOz3wD9kLa:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/d0a2ef6e9d2a8d39db01c0cfb842a0a7850c7afb/posting/list.go#L110)):*
> struct field AfterUid should be AfterUID
</details>
Done.

---

*[query/query.go, line 832 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSpzkm0VNjXSWtctyMe:-KTJo7klS8ibCHb1YTuS:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/query/query.go#L832)):*
> I think query is the generic module, which executes internal format of instructions. GQL is just one language -- there would be others later. Still not convinced that query should depend on GQL. You could literally just have two strings, AND and OR, and that's all you really need to pass info from GQL to Query. So, it's pretty straightfwd to remove this dep.
</details>
Done.

---

*[query/query.go, line 694 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIIpF6GBbfbXqo9u_W:-KTJoAGn1PFHeE2M523E:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/query/query.go#L694)):*
> With the updates in FB for Go, you can now use:
> tl := task.GetRootAsUidList(buf, 0)
</details>
Done.

---

*[query/query.go, line 785 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIJQH_rvAnc6hWZARr:-KTJcHzEyaS2GLfcaPB-:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/query/query.go#L785)):*
> r := task.GetRootAsResult(resultBuf, 0)
</details>
Done.

---

*[query/query.go, line 848 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIKHELUP2emi7CRUI2:-KTJcKxfQVjK3iGBC_wx:baa0m4h) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/query/query.go#L848)):*
> What is this checking for? Why?
</details>
Got this from master. I think you have to ask Kostub.

---

*[query/query.go, line 853 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIKA3OrtlwgAfsTuUP:-KTJcRF_O7eIKkOIQzWk:baa0m4h) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/query/query.go#L853)):*
> What's the logic behind leftToProcess? Add a comment.
</details>
Got this from master. I think you have to ask Kostub.

---

*[query/query.go, line 970 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIKd2SApM3WHASDNj8:-KTJcXVg5s93SpKGPsU_:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/query/query.go#L970)):*
> This goes beyond 100 chars.
</details>
Done.

---

*[query/query.go, line 1033 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIM37VQzbjxokR102a:-KTJdWwncZcVqKs-u0ee:b-jsjcc) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/query/query.go#L1033)):*
> n + p.Count could still be < 0.
</details>
Have replaced with a positive mod. Maybe you prefer an assert?

---

*[worker/task.go, line 144 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTIMdv2mdJgpXViGJjc:-KTJmieZq2WpEdosQbX2:b-5xdqeu) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/worker/task.go#L144)):*
> You're intersecting with taskList which is empty? What?
</details>
This is confusing. "Intersect" is a field inside task.Query. This is just checking if the field is populated. I have renamed this field as "ToIntersect". Maybe you can think of a better name?

---

*[x/serialize.go, line 32 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KT9dBEnQeJANTQFN__E-r4-32:-KTJldPjCEcGHV0in9uA:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/253307b32331a4044eb7e1dcf940cb8b6ccbaf83/x/serialize.go#L32)):*
> func ParseUidList should be ParseUIDList
</details>
Done.

---

*[x/serialize.go, line 28 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTINBuhF8nREhBDEGNg:-KTJjwiu4wGB33dT9KbK:b-p3grg) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/x/serialize.go#L28)):*
> You don't need any of these now, with the new task.GetRootAs...
</details>
File removed. I like it.

---

*[x/serialize.go, line 52 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KTINJWqDnlAcN6MwDpU:-KTJlVShIi_tXCswUlCb:bbyxfnl) ([raw file](https://github.com/dgraph-io/dgraph/blob/daa50f3a09e2a70f2c0c5cb1e6059676374847a2/x/serialize.go#L52)):*
> This is repetition of another function in UIDList.
</details>
Now deleted. This is good.

---

*[gql/state.go, line 64 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-KSq0P6YTTDNO09OFPEV:-KTJlbLBj1ExsZHApjzD:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/2cd7031daff1c562c7b2ca6fc67fd3be96000607/gql/state.go#L64)):*
> These are very specific to lexer state. I don't think these should be exposed.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/245)*

manishrjain:
:lgtm: Thanks for being persistent through all the iterations! Feel free to submit.

---

Reviewed 13 of 13 files at r15.
Review status: all files reviewed at latest revision, 45 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-:-KTNbEZBIU9T6TinxPwv:b-890yoe)*

jchiu0:
Thanks for the thorough review too! 

---

Review status: all files reviewed at latest revision, 45 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/245#-:-KTNi31fDSNlMAtoMUqi:bt7ucj9)*

PR #251, opened 2016-09-29
Add pbToJson function

pawanrawal:
Attempt to add a function which can convert the protocol buffer response to a JSON response. This would avoid maintaining both ToJSON and ToProtocolBuffer.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/251)


pawanrawal:
After this change the benchmarks for ToJSON aren't that good https://discuss.dgraph.io/t/adding-a-pbtojson-function/938. Hence decided not to go with this for now.

PR #253, opened 2016-09-30
Add tokenizing support using ICU

jchiu0:
We add a new package called "tok" for tokenizing. For now, we disable almost all of "tok" so that the user is not forced to install ICU yet for the build to succeed. However, you can still run the test with `cd tok && go test -tags tok .`

We add two dependencies for unicode normalization: `golang.org/x/text/transform` and `golang.org/x/text/unicode/norm`. They are very standard Go libraries.

We add one dependency (ICU) for the word boundary analysis. We use only a very small subset of the C API. I have tried using cgogen. It needs a config file and I got some errors that might take me a while to fix. On the other hand, the C functions are almost directly usable, so I don't see much gain from using cgogen. Most of the work has to do with adding some additional logic so that we make fewer cgo calls and the Go code doesn't have to deal with malloc, free or UChar arrays which is an internal ICU type for a lot of stuff including boundary analysis. This is very customized and not possible to generate with cgogen.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/253)


manishrjain:
:lgtm:

---

Reviewed 21 of 21 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/253#-:-KTNcVmGvfPnY_TarZsE:bnfp4nl)*

PR #254, opened 2016-09-30
RAFT Mutations: Consensus based mutations

manishrjain:
Also, remove the logic which returns left over mutations. Instead, on an error, the client should just retry the entire mutation, because our mutations are idempotent (or ought to be).

Also, wait for the mutation to be committed, before returning a response back to the client.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/254)


pawanrawal:
Once this is merged and you do the refactoring I can write some tests for the distributed mutations part.

---

Reviewed 5 of 6 files at r1, 1 of 1 files at r2.

---

*[worker/draft.go, line 63 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/254#-KStHtUoRi39bZ4qnXVx:-KStHtUpkM9D8-cHCvGz:bxft1zo) ([raw file](https://github.com/dgraph-io/dgraph/blob/1da84aa336f4cfa204ba23d9f33bafd832d72324/worker/draft.go#L63)):*
> ```Go
> 	}
> 	p.Unlock()
> 	if !has {
> ```

We can move this above, and then if we don't return we just `delete(p.ids, pid)`?

---

*[worker/draft.go, line 66 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/254#-KSttZkbJ0FKJpV0z2_E:-KSttZkbJ0FKJpV0z2_F:br2qmh7) ([raw file](https://github.com/dgraph-io/dgraph/blob/1da84aa336f4cfa204ba23d9f33bafd832d72324/worker/draft.go#L66)):*
> ```Go
> 		return
> 	}
> 	ch <- err
> ```

if err != nil , would it make sense to do ch <- err earlier? So that the blocked call can respond to the user?

---

*[worker/draft.go, line 126 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/254#-KStFK1386StimUBdzUi:-KStFK1386StimUBdzUj:bf8w078) ([raw file](https://github.com/dgraph-io/dgraph/blob/1da84aa336f4cfa204ba23d9f33bafd832d72324/worker/draft.go#L126)):*
> ```Go
> type header struct {
> 	proposalId int64
> 	msgId      uint32
> ```

Could this just be uint8 or uint16 (if you plan to write it directly https://golang.org/pkg/encoding/binary/#ByteOrder).

---

*[worker/draft.go, line 131 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/254#-KStF76lqFjKa0eOtvMq:-KStF76mPWIbQltMbhv9:b-2sfred) ([raw file](https://github.com/dgraph-io/dgraph/blob/1da84aa336f4cfa204ba23d9f33bafd832d72324/worker/draft.go#L131)):*
> ```Go
> func (h *header) Encode() []byte {
> 	buf := new(bytes.Buffer)
> 	x.Check(binary.Write(buf, binary.LittleEndian, h.proposalId))
> ```

Could we directly write the bytes here? According to Kostub's analysis that is 100 times faster than this.

---

*[worker/mutation.go, line 111 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/254#-KStrE8k49XFlfBF3Ypq:-KStrE8k49XFlfBF3Ypr:bbm91oz) ([raw file](https://github.com/dgraph-io/dgraph/blob/1da84aa336f4cfa204ba23d9f33bafd832d72324/worker/mutation.go#L111)):*
> ```Go
> 	}
> 
> 	// Get a connection from the pool and run mutations over the network.
> ```

So the code below this won't be called now. Would we even need this later? As once we propose a mutation, the instances would call their internal mutate function.

---

*[worker/mutation.go, line 148 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/254#-KStoFgd0BB1Mr_a9B30:-KStoFgeaY6KxjBB-rlv:b-7y8iuh) ([raw file](https://github.com/dgraph-io/dgraph/blob/1da84aa336f4cfa204ba23d9f33bafd832d72324/worker/mutation.go#L148)):*
> ```Go
> }
> 
> // MutateOverNetwork checks which instance should be running the mutations
> ```

which group

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/254#-:-KSu1TheesOkc_gSNqJo:b6mx89f)*

manishrjain:
Review status: 5 of 7 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[worker/draft.go, line 63 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/254#-KStHtUoRi39bZ4qnXVx:-KSu5Hx75G2fTgOdAMAI:b-feh1ft) ([raw file](https://github.com/dgraph-io/dgraph/blob/1da84aa336f4cfa204ba23d9f33bafd832d72324/worker/draft.go#L63)):*
> We can move this above, and then if we don't return we just `delete(p.ids, pid)`?
</details>
We don't do a defer here. Because the lock can potentially get stuck on the channel. Therefore return is after unlocking.

---

*[worker/draft.go, line 66 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/254#-KSttZkbJ0FKJpV0z2_E:-KSu5a2qfDZKgT5GTWCc:bppj29x) ([raw file](https://github.com/dgraph-io/dgraph/blob/1da84aa336f4cfa204ba23d9f33bafd832d72324/worker/draft.go#L66)):*
> if err != nil , would it make sense to do ch <- err earlier? So that the blocked call can respond to the user?
</details>
Not sure what you mean.

---

*[worker/draft.go, line 126 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/254#-KStFK1386StimUBdzUi:-KSuABBRqkPhWl4S4JHW:b-z2p3p6) ([raw file](https://github.com/dgraph-io/dgraph/blob/1da84aa336f4cfa204ba23d9f33bafd832d72324/worker/draft.go#L126)):*
> Could this just be uint8 or uint16 (if you plan to write it directly https://golang.org/pkg/encoding/binary/#ByteOrder).
</details>
Done. Not uint8, but uint16 is allowed.

---

*[worker/draft.go, line 131 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/254#-KStF76lqFjKa0eOtvMq:-KSu62gZxiIcSBqxbS9q:b7zwmno) ([raw file](https://github.com/dgraph-io/dgraph/blob/1da84aa336f4cfa204ba23d9f33bafd832d72324/worker/draft.go#L131)):*
> Could we directly write the bytes here? According to Kostub's analysis that is 100 times faster than this.
</details>
DONE. oh.. forgot about that. Good point. Changed.

---

*[worker/mutation.go, line 111 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/254#-KStrE8k49XFlfBF3Ypq:-KSuAU0a9vLfVFgKa4zW:b8trw1) ([raw file](https://github.com/dgraph-io/dgraph/blob/1da84aa336f4cfa204ba23d9f33bafd832d72324/worker/mutation.go#L111)):*
> So the code below this won't be called now. Would we even need this later? As once we propose a mutation, the instances would call their internal mutate function.
</details>
Done. Added a TODO for now.

---

*[worker/mutation.go, line 148 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/254#-KStoFgd0BB1Mr_a9B30:-KSuAW3abx92AxiD7Xs0:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/1da84aa336f4cfa204ba23d9f33bafd832d72324/worker/mutation.go#L148)):*
> which group
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/254)*

pawanrawal:
:lgtm: 

---

Reviewed 2 of 2 files at r3, 1 of 1 files at r4.

---

*[worker/draft.go, line 66 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/254#-KSttZkbJ0FKJpV0z2_E:-KSuEqNk8sRu9r83oqof:b-cik8p0) ([raw file](https://github.com/dgraph-io/dgraph/blob/1da84aa336f4cfa204ba23d9f33bafd832d72324/worker/draft.go#L66)):*
> Not sure what you mean.
</details>
nvm ignore.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/254#-:-KSuEteX5akKQCNsAcIf:bg7igkv)*

PR #273, opened 2016-10-06
Fix filtering

jchiu0:
In ProcessGraph, when sg.destUIDs is empty (but not nil), we will save some work and return early. As a result, sg.child.srcUIDs might remain as nil.

There are two ways to fix.

(1) Every time you access srcUIDs, you need to check if it is nil.
(2) Check for nil in algo.UIDList's Size() method. This function is used to see if there is any work to be done.

I decide to go with (2). It is more in line with the Go proverb "make zero value useful". This check for nil is only in the Size() method. In other methods, we continue to assert for non-nil algo.UIDList pointers.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/273)

### DIFF
--- a/algo/uidlist.go
+++ b/algo/uidlist.go
@@ -46,7 +46,12 @@ func (u *UIDList) Get(i int) uint64 {
 
 // Size returns size of UIDList.
 func (u *UIDList) Size() int {
-	x.Assert(u != nil)
+	if u == nil {
+		// In a subgraph node, in processGraph, sometimes we might fan out to zero
+		// nodes, i.e., sg.destUIDs is empty. In this case, child subgraph might not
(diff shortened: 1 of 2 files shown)


ashwin95r:
:lgtm_strong: 

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/273#-:-KTQ-6IUotkxtH6o9NoJ:b-uwkowd)*

ashwin95r:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/273)*

PR #279, opened 2016-10-12
Feature/types-tokenize

ashwin95r:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/279)


jchiu0:
One minor concern is a seemingly unnecessary marshalText in tokenizedIndexKeys. It seems that the index functions in types/index.go can work with Value interface. In fact, they can probably take in concrete classes like Float. Whoever needs to use these functions (e.g., in query) should parse themselves using those UnmarshalText functions in our type library. Sometimes, we do not even need to parse, e.g., when we need to do a range loop over integers, for ordering / sorting results.

But this is only a very small concern. Hence :lgtm: 

---

Review status: 0 of 90 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/279#-:-KTsO9jKMTi-M4OKPcsB:boiwmd2)*

jchiu0:
You modified types/index.go to ParseInt while keeping MarshalText.. I will probably change that in an upcoming PR. Feel free to submit.

---

Review status: 0 of 90 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/279#-:-KTsr_kXR4-bRPuTE9yp:b-wnt64f)*

kostub:
Adding MarshalText breaks the geo index since it expects the data in binary format. It might be better to pass the Value in directly. Unmarshaling/Marshaling of geo data is pretty expensive.

---

Review status: 0 of 90 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/279#-:-KTvzINNjwXT94QZ1chQ:b-73ck4v)*

ashwin95r:
Addressed the comments. Merged to master.

PR #284, opened 2016-10-15
Feature/geoquery

kostub:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/284)


jchiu0:
Review status: 0 of 10 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[query/geo.go, line 53 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/284#-KUMbpP33MGYI9n-F5YJ:-KUMbpP33MGYI9n-F5YK:b-bbxkhb) ([raw file](https://github.com/dgraph-io/dgraph/blob/aab7fb4dc565dc0a25af12bc9f15227f90795581/query/geo.go#L53)):*
> ```Go
> 
> 	// Filter the values
> 	sg.srcUIDs = filterUIDs(uids, values, data)
> ```

Wonder if all these work can be done in the worker instead? That is the position we have taken for filters currently. We send the UIDs (in our case destUIDs, in your case srcUIDs) to the worker and the worker has all the index information and do all the filtering and return just the filtered results.

---

*[query/geo_test.go, line 278 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/284#-KUMXWGwUZBPNWSySPn4:-KUMXWGwUZBPNWSySPn5:b-1jve7u) ([raw file](https://github.com/dgraph-io/dgraph/blob/aab7fb4dc565dc0a25af12bc9f15227f90795581/query/geo_test.go#L278)):*
> ```Go
> 		map[string]interface{}{"name": "San Carlos Airport"},
> 		map[string]interface{}{"name": "Mountain View"}}
> 	EqualArrays(t, expected, mp)
> ```

I wonder if testify.EqualValues will work here?

---

*[query/query.go, line 274 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/284#-KUM_rg6GN6Tj6tpZO4H:-KUM_rg6GN6Tj6tpZO4I:b-nlgjuo) ([raw file](https://github.com/dgraph-io/dgraph/blob/aab7fb4dc565dc0a25af12bc9f15227f90795581/query/query.go#L274)):*
> ```Go
> 			}
> 			if sg.GeoFilter != nil {
> 				x.Assertf(len(l) == 1, "There should be exactly 1 uid at the top level.")
> ```

The geofilter only works for the root subgraph?

---

*[query/query.go, line 825 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/284#-KUM_zBozAsvNtjYsW5m:-KUM_zBozAsvNtjYsW5n:b-7y1x5d) ([raw file](https://github.com/dgraph-io/dgraph/blob/aab7fb4dc565dc0a25af12bc9f15227f90795581/query/query.go#L825)):*
> ```Go
> 	}
> 
> 	if err = sg.applyGeoQuery(ctx); err != nil {
> ```

How does a geo query look? For filters,  it is applied to destUIDs, whereas here I see that you are applying to srcUIDs?

For the current filters, say on "friends", we will filter the destUIDs for subgraph having attr=friends. Then we apply the filtered results to UIDmatrix. The task.Query for children nodes will have the filtered UIDs, so we query less for children.

If the filter is applied to srcUIDs, the network call for this attribute has already been made..

Maybe a sample geo query in the form of GraphQL would help me understand this.. Thanks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/284)*

ashwin95r:
Reviewed 4 of 22 files at r2.
Review status: 4 of 10 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[query/query.go, line 825 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/284#-KUM_zBozAsvNtjYsW5m:-KUMeX22xGqnh9ftKdTh:b-duy2wk) ([raw file](https://github.com/dgraph-io/dgraph/blob/aab7fb4dc565dc0a25af12bc9f15227f90795581/query/query.go#L825)):*
> How does a geo query look? For filters,  it is applied to destUIDs, whereas here I see that you are applying to srcUIDs?
> 
> For the current filters, say on "friends", we will filter the destUIDs for subgraph having attr=friends. Then we apply the filtered results to UIDmatrix. The task.Query for children nodes will have the filtered UIDs, so we query less for children.
> 
> If the filter is applied to srcUIDs, the network call for this attribute has already been made..
> 
> Maybe a sample geo query in the form of GraphQL would help me understand this.. Thanks.
</details>
Actually, the query part hasn't been decided yet. This PR handles it from the subgraph level.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/284)*

kostub:
Review status: 4 of 10 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[query/geo.go, line 53 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/284#-KUMbpP33MGYI9n-F5YJ:-KUMxmYD0IMoaY4JVRjv:b-4yb4iy) ([raw file](https://github.com/dgraph-io/dgraph/blob/aab7fb4dc565dc0a25af12bc9f15227f90795581/query/geo.go#L53)):*
> Wonder if all these work can be done in the worker instead? That is the position we have taken for filters currently. We send the UIDs (in our case destUIDs, in your case srcUIDs) to the worker and the worker has all the index information and do all the filtering and return just the filtered results.
</details>
It could be done on the worker side when fetching values. We still need two lookups, one for fetching uids from the index and one for fetching values. We'll have to include additional information in the Task, which I had in my original implementation. In our last conversation on hangouts we decided against it and just use tokens for matching.

---

*[query/geo_test.go, line 278 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/284#-KUMXWGwUZBPNWSySPn4:-KUMx6FgRd_qm4rV0vTZ:bpvo2z8) ([raw file](https://github.com/dgraph-io/dgraph/blob/aab7fb4dc565dc0a25af12bc9f15227f90795581/query/geo_test.go#L278)):*
> I wonder if testify.EqualValues will work here?
</details>
Nope I tried it. The problem is that the the array ordering is non-deterministic. Go has some random enumeration of keys in a map.

---

*[query/query.go, line 274 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/284#-KUM_rg6GN6Tj6tpZO4H:-KUMz9rtGE0cFNTqWx12:bmyhvb9) ([raw file](https://github.com/dgraph-io/dgraph/blob/aab7fb4dc565dc0a25af12bc9f15227f90795581/query/query.go#L274)):*
> The geofilter only works for the root subgraph?
</details>
Most likely it will work for other levels too, but I'm not sure what such a query would represent. Regardless, len(l) == 1 for any level of the subgraph with a geo query.

---

*[query/query.go, line 825 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/284#-KUM_zBozAsvNtjYsW5m:-KUMzZBPLoSYWLBB0VhU:brewo1j) ([raw file](https://github.com/dgraph-io/dgraph/blob/aab7fb4dc565dc0a25af12bc9f15227f90795581/query/query.go#L825)):*
> Actually, the query part hasn't been decided yet. This PR handles it from the subgraph level.
</details>
As Ashwin said the specifics of writing geo queries in graphQL hasn't been decided. 

In the current implementation, there there are no srcUIDs which is why the geoquery puts its result in srcUIDs **and** destUIDs. An example geo query (in a subgraph) will look like this:

```
    sg := &SubGraph{
        Attr:      "geometry",
        GeoFilter: &geo.Filter{Data: data, Type: geo.QueryTypeWithin},
        Children:  []*SubGraph{&SubGraph{Attr: "name"}},
    }
```
which is the query to _find all points with the geometry attribute within the region given by data and return their names_.

One _possibility_ to express in graphql might be something like:
```
{
   me(_loc_: [(x1, y1), ... (x_n, y,n)] @filter(within(geometry)) )
   {
       name
   }
}
```

However we should discuss the graphql specifics in a discuss topic rather than in a PR

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/284)*

jchiu0:
I think we can check this in first. We have a decent understanding of the code. If this is not submitted, the code base might drift underneath. Secondly, the change is not really exposed yet and doesn't affect overall efficiency since sg.Geofilter is just nil. Thirdly, we will quickly follow up with the following changes to complete this feature.

(1) finalize graphQL design (2) move some work to worker (3) apply to destUIDs only like other filters (4) work on all levels of subgraph.

:lgtm: 

---

Review status: 4 of 10 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[query/geo.go, line 53 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/284#-KUMbpP33MGYI9n-F5YJ:-KUP_ARGG7L2rkeYUnrL:b-ukpjka) ([raw file](https://github.com/dgraph-io/dgraph/blob/aab7fb4dc565dc0a25af12bc9f15227f90795581/query/geo.go#L53)):*
<details><summary><i>Previously, kostub (Kostub Deshmukh) wrote…</i></summary>
> It could be done on the worker side when fetching values. We still need two lookups, one for fetching uids from the index and one for fetching values. We'll have to include additional information in the Task, which I had in my original implementation. In our last conversation on hangouts we decided against it and just use tokens for matching.
</details>
We will probably change it in the future as the assumption is that index might have a lot more values than the list of UIDs being filtered. Sorry, I should have told you about this. But for now, I think this is fine. It is a good starting point for us.

---

*[query/query.go, line 274 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/284#-KUM_rg6GN6Tj6tpZO4H:-KUPaIOtp1FvlLixVgKx:btgwyt3) ([raw file](https://github.com/dgraph-io/dgraph/blob/aab7fb4dc565dc0a25af12bc9f15227f90795581/query/query.go#L274)):*
<details><summary><i>Previously, kostub (Kostub Deshmukh) wrote…</i></summary>
> Most likely it will work for other levels too, but I'm not sure what such a query would represent. Regardless, len(l) == 1 for any level of the subgraph with a geo query.
</details>
We might have to update this in the future for other levels and move it to destUIDs. But I think what you have here is already very valuable and can probably be checked in first, and someone else can pick it up quickly from there.

---

*[query/query.go, line 825 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/284#-KUM_zBozAsvNtjYsW5m:-KUPatMbkc5Ux__ua_Fs:b-ptq9w5) ([raw file](https://github.com/dgraph-io/dgraph/blob/aab7fb4dc565dc0a25af12bc9f15227f90795581/query/query.go#L825)):*
<details><summary><i>Previously, kostub (Kostub Deshmukh) wrote…</i></summary>
> As Ashwin said the specifics of writing geo queries in graphQL hasn't been decided. 
> 
> In the current implementation, there there are no srcUIDs which is why the geoquery puts its result in srcUIDs **and** destUIDs. An example geo query (in a subgraph) will look like this:
> 
> ```
>     sg := &SubGraph{
>         Attr:      "geometry",
>         GeoFilter: &geo.Filter{Data: data, Type: geo.QueryTypeWithin},
>         Children:  []*SubGraph{&SubGraph{Attr: "name"}},
>     }
> ```
> which is the query to _find all points with the geometry attribute within the region given by data and return their names_.
> 
> One _possibility_ to express in graphql might be something like:
> ```
> {
>    me(_loc_: [(x1, y1), ... (x_n, y,n)] @filter(within(geometry)) )
>    {
>        name
>    }
> }
> ```
> 
> However we should discuss the graphql specifics in a discuss topic rather than in a PR
</details>
Thanks. This is all the information I need.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/284#-:-KUPccaiprHfIWgEudmw:b-uzg9l8)*

PR #292, opened 2016-10-21
Support sorting using index buckets

jchiu0:
1) We added an in-memory table to store bucket keys that have been seen.
2) We added sorting in types package.
3) We added new RPC "sort" that can do both coarse sort (using index) and fine sort (not using index).
4) We added applyOrder in query package to issue these RPC calls and update Subgraph.
5) AddMutationWithIndex now calls MutateOverNetwork instead of AddMutation.
6) (Might have missed something)

TODO: We will tidy up the index init calls soon.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/292)


manishrjain:
Reviewed 12 of 31 files at r1, 2 of 3 files at r2.
Review status: 14 of 32 files reviewed at latest revision, 33 unresolved discussions, some commit checks failed.

---

*[task.fbs, line 83 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/292#-KUtg7W_tFbeCqISJ4AQ:-KUtg7W_tFbeCqISJ4AR:b-38e189) ([raw file](https://github.com/dgraph-io/dgraph/blob/6202230ecf6cac81c1314b66d568834ed6771317/task.fbs#L83)):*
> ```
> 
> table FineSort {	
> 	uid:UidList;
> ```

Why not just use UidList? Why create FineSort?

---

*[task.fbs, line 87 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/292#-KUtgGd7ukgB8RTtfiex:-KUtgGd7ukgB8RTtfiey:bvv7npd) ([raw file](https://github.com/dgraph-io/dgraph/blob/6202230ecf6cac81c1314b66d568834ed6771317/task.fbs#L87)):*
> ```
> 
> table Sort {
> 	coarse:bool;
> ```

This seems redundant, if you're going to only use one of the following two fields. You can just check for nil pointer or zero data, right?

---

*[task.fbs, line 102 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/292#-KUtho3Bvmi2qIQdJm1q:-KUtho3Bvmi2qIQdJm1r:bkm1auq) ([raw file](https://github.com/dgraph-io/dgraph/blob/6202230ecf6cac81c1314b66d568834ed6771317/task.fbs#L102)):*
> ```
> 	// If idx[3] = 5, then the 3rd element in the sorted list is 5th element
> 	// of our input.
> 	idx:[uint];
> ```

How about the case where the values are requested? For e.g. sort by dob, but also return dob in the result? We could potentially just return the values here as well right?

---

*[types/index.go, line 25 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/292#-KUtipqPOcuafzxvBRKa:-KUtipqPOcuafzxvBRKb:b-a36qm0) ([raw file](https://github.com/dgraph-io/dgraph/blob/6202230ecf6cac81c1314b66d568834ed6771317/types/index.go#L25)):*
> ```Go
> 		return false
> 	}
> 	reader := strings.NewReader(key)
> ```

This seems inefficient. You could just do: `return key[0] == indexRune`

---

*[types/index.go, line 45 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/292#-KUtjzJAkWZGNEAp7twI:-KUtjzJAkWZGNEAp7twJ:b-y7mxpi) ([raw file](https://github.com/dgraph-io/dgraph/blob/6202230ecf6cac81c1314b66d568834ed6771317/types/index.go#L45)):*
> ```Go
> 
> // TokenFromKey returns token from an index key.
> func TokenFromKey(key []byte) []byte {
> ```

Is this for debugging? Because typically we should go from token -> key.xx

---

*[types/sort.go, line 118 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/292#-KUtrsOSLhyAWcRHwxF_:-KUtrsOSLhyAWcRHwxFa:b-s7spqq) ([raw file](https://github.com/dgraph-io/dgraph/blob/6202230ecf6cac81c1314b66d568834ed6771317/types/sort.go#L118)):*
> ```Go
> }
> 
> // Sort sorts the given array in-place.
> ```

Well actually it doesn't sort the given array. It just returns the sorted indices. Though I understand your intentions behind it, I feel it's a bit hacky and an unclean API.

---

*[worker/indexing.go, line 16 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/292#-KUtkR5JhnfO4XmjoKrS:-KUtkR5JhnfO4XmjoKrT:b-w57smf) ([raw file](https://github.com/dgraph-io/dgraph/blob/6202230ecf6cac81c1314b66d568834ed6771317/worker/indexing.go#L16)):*
> ```Go
> func processIndexMutations() {
> 	ctx := context.Background()
> 	for m := range index.MutateChan {
> ```

Oh.. don't do it like this. What you want to do is to pick up the mutations and put them into a Mutations struct, which has Set and Del fields asap. Then, every x ms or so, you can send them over to the network in a goroutine, which you ideally should ideally retry for a long time, before giving up; because that would be bad.

Also, you don't want to stop picking up mutations from the channel, because that'd block our normal Mutations. Unless that's the idea -- is to block mutations -- in which case we should think through this a bit more.

---

*[worker/payload.proto, line 16 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/292#-KUtl8xuwmYaINJezUmG:-KUtl8xuwmYaINJezUmH:bnwo4lr) ([raw file](https://github.com/dgraph-io/dgraph/blob/6202230ecf6cac81c1314b66d568834ed6771317/worker/payload.proto#L16)):*
> ```protobuf
> 	rpc ServeTask (Payload)     returns (Payload) {}
> 	rpc PredicateData (Payload) returns (stream Payload) {}
> 	rpc Sort (Payload)    returns (Payload) {}
> ```

Align `returns` with the rest of them.

---

*[worker/sort.go, line 41 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/292#-KUtlST4vSwp20f21F3-:-KUtlST5UKXSAthHhOcg:b-j6xuuz) ([raw file](https://github.com/dgraph-io/dgraph/blob/6202230ecf6cac81c1314b66d568834ed6771317/worker/sort.go#L41)):*
> ```Go
> 
> 	c := NewWorkerClient(conn)
> 	reply, err := c.Sort(ctx, &Payload{Data: qu})
> ```

Do a select { } here, in case Sort takes a long time, and context finishes before that.

---

*[worker/sort.go, line 60 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/292#-KUtloGKGoHBIPy85W9m:-KUtloGKGoHBIPy85W9n:b-vkvc5e) ([raw file](https://github.com/dgraph-io/dgraph/blob/6202230ecf6cac81c1314b66d568834ed6771317/worker/sort.go#L60)):*
> ```Go
> 	s := task.GetRootAsSort(query.Data, 0)
> 	gid := BelongsTo(string(s.Attr()))
> 	//x.Trace(ctx, "Attribute: %q NumUids: %v groupId: %v Sort", q.Attr(), q.UidsLength(), gid)
> ```

Looks like a good x.Trace to have...

---

*[worker/sort.go, line 91 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/292#-KUtnHVwX0Li5-ySmTuk:-KUtnHVwX0Li5-ySmTul:b-y879x) ([raw file](https://github.com/dgraph-io/dgraph/blob/6202230ecf6cac81c1314b66d568834ed6771317/worker/sort.go#L91)):*
> ```Go
> 	cs := s.CoarseSort(nil)
> 	x.Assert(cs != nil)
> 	return newCoarseSorter(string(s.Attr()), cs).run()
> ```

This is a bit of a strange API. run() feels like it's running in a thread which will finish some point later. Definitely not the Go way.

---

*[worker/sort.go, line 130 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/292#-KUtnE-CyLf-sm9j9ZCR:-KUtnE-CyLf-sm9j9ZCS:b4mg98d) ([raw file](https://github.com/dgraph-io/dgraph/blob/6202230ecf6cac81c1314b66d568834ed6771317/worker/sort.go#L130)):*
> ```Go
> 		cs.counts = make([]int, n)
> 		for i := 0; i < n; i++ {
> 			cs.counts[i] = cs.count
> ```

So, we need to pick `count` results, and that's what you're setting this to. But, you can't stop iterating for `count` at the first token. You can only do that at the last token.

Alternatively, you can intersect all the token PLs first, and then enforce the `count` for each PL in UidMatrix.

---

*[worker/sort.go, line 140 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/292#-KUtqIDFgQlBhjAuHc2i:-KUtqIDFgQlBhjAuHc2j:b-9woocc) ([raw file](https://github.com/dgraph-io/dgraph/blob/6202230ecf6cac81c1314b66d568834ed6771317/worker/sort.go#L140)):*
> ```Go
> }
> 
> // Example: Fix one UID list. Intersect with 4 buckets. Say the size of
> ```

How does it work if we just need to pick the first N results for any given sorting order, without a UidMatrix present. For e.g., [first 10 people by dob]

---

*[worker/sort.go, line 149 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/292#-KUtnZ-XgRXNTW-NOPOT:-KUtnZ-XgRXNTW-NOPOU:b-iuj2ts) ([raw file](https://github.com/dgraph-io/dgraph/blob/6202230ecf6cac81c1314b66d568834ed6771317/worker/sort.go#L149)):*
> ```Go
> // runHelper looks up UIDs for one token / bucket and intersect with each
> // UID list in the UID matrix. Returns true if all UID lists are done.
> func (cs *coarseSorter) runHelper(key []byte) bool {
> ```

The function name says nothing about what it's doing. And so does `run`. Probably call it `sort()`, and then `doSort()` for the helper, or something.

---

*[worker/sort.go, line 165 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/292#-KUtoPiVKnIS6ySWTxhB:-KUtoPiVKnIS6ySWTxhC:b-ivtdfz) ([raw file](https://github.com/dgraph-io/dgraph/blob/6202230ecf6cac81c1314b66d568834ed6771317/worker/sort.go#L165)):*
> ```Go
> 		n := result.Size()
> 
> 		if cs.state[i] == 2 {
> ```

What does 2 mean? Use a const. Also, we didn't store the result of the intersection, and yet we're just abandoning work on this index. Shouldn't we abandon the index before doing the intersection?

---

*[worker/sort.go, line 181 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/292#-KUtotkPUaQjZZGjk7zu:-KUtotkPUaQjZZGjk7zv:b6l58md) ([raw file](https://github.com/dgraph-io/dgraph/blob/6202230ecf6cac81c1314b66d568834ed6771317/worker/sort.go#L181)):*
> ```Go
> 				if cs.count > 0 {
> 					// Say intersection is 100 elements. Offset is 3. So, we're
> 					// keeping 97 elements from this bucket. Decrement count by 97.
> ```

How can you discard the first 3 elements from this bucket? The previous result of the sort != sort by uids within the bucket. I'd think you'd need to keep the entire bucket or nothing at all.

---

*[worker/sort.go, line 215 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/292#-KUtpdlbQXLTrjPCE9gD:-KUtpdlbQXLTrjPCE9gE:b-au4y9w) ([raw file](https://github.com/dgraph-io/dgraph/blob/6202230ecf6cac81c1314b66d568834ed6771317/worker/sort.go#L215)):*
> ```Go
> 	kt := index.GetTokensTable(cs.attr)
> 	for token := kt.GetFirst(); len(token) > 0; token = kt.GetNext(token) {
> 		if cs.runHelper(types.IndexKey(cs.attr, token)) {
> ```

This sounds like it won't work. Say, you need N elements for [barack, obama] -- the two tokens. You run the first token, and pick N. Then you run the second token, you'd almost always pick less than N. Even though there might be overall more results, and could make the N requested.

See my comments above as well where you're filling in `cs.count`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/292)*

PR #299, opened 2016-10-26
Feature/backup

ashwin95r:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/299)


jchiu0:
Review status: 0 of 3 files reviewed at latest revision, 13 unresolved discussions, some commit checks broke.

---

*[posting/backup.go, line 34 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8OC24hQVjtPyMKtk1:-KV8OC24hQVjtPyMKtk2:b-flvnrf) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L34)):*
> ```Go
> 
> 	wg.Add(33)
> 	for i := 0; i < 33; i++ {
> ```

Magical number 33?

---

*[posting/backup.go, line 39 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8O6A_lVC7F1Q84Dcp:-KV8O6A_lVC7F1Q84Dcq:b-2mufhp) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L39)):*
> ```Go
> 	go writeToFile(ch)
> 
> 	for it = it; it.Valid(); it.Next() {
> ```

How about "for it.SeekToFirst; it.Valid(); it.Next()" and delete the above line?

---

*[posting/backup.go, line 47 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8NwgOcRJ5C-ztfJXj:-KV8NwgOcRJ5C-ztfJXk:b4050kf) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L47)):*
> ```Go
> 
> 	close(chkv)
> 	wg.Wait()
> ```

Maybe can add comment like "wait for convertToRdf"?

---

*[posting/backup.go, line 55 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8LuW9DJtI-TWBxCJK:-KV8LuW9DJtI-TWBxCJL:b-1nawf9) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L55)):*
> ```Go
> 	var b bytes.Buffer
> 	var rest []byte
> 	if strings.HasPrefix(string(key), ":") || strings.HasPrefix(string(key), "_uid_") {
> ```

Converting key to a string requires a copy. If you have to do it, just do it once. (Actually I was proposing in a separate thread that we should all just stick to strings as far as possible)

Under "bytes" lib, there should be equivalent of string functions, e.g. HasPrefix. You might want to use that to avoid converting to string here.

Use constant indexRune?

---

*[posting/backup.go, line 58 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8NjBe-LlcCU_DNFnt:-KV8NjBe-LlcCU_DNFnu:b-4vgg6n) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L58)):*
> ```Go
> 		return "", "", false
> 	}
> 	fmt.Printf("---%s\n", key)
> ```

This fmt.Printf is for debugging only? Delete?

---

*[posting/backup.go, line 60 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8MKpob9kg7wd3qpK5:-KV8MKppdrLegMqpfO6S:b-avr2ti) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L60)):*
> ```Go
> 	fmt.Printf("---%s\n", key)
> 	for i, ch := range key {
> 		if ch == '|' {
> ```

Use bytes.IndexRune or IndexByte? Also this function is generally useful. I don't think we have this yet. You could move it to a more central location, where Key() is.

---

*[posting/backup.go, line 77 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8MkQpRCfGe7v96okL:-KV8MkQpRCfGe7v96okM:b5jpddx) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L77)):*
> ```Go
> 		return
> 	}
> 	buf.WriteString("<_uid_:")
> ```

Not checking errors? (i know it is painful but x.Check alleviates it a bit)

---

*[posting/backup.go, line 97 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8TkYU1eStY1ZDrE_-:-KV8TkYVYAXXZ7lwA5x7:bcgzz8p) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L97)):*
> ```Go
> 			x.Check(err)
> 
> 			buf2.Write(str)
> ```

Check returned errors? (with x.Check)

---

*[posting/backup.go, line 123 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8NOGmp9staPuMFvUV:-KV8NOGmp9staPuMFvUW:b-ob39ei) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L123)):*
> ```Go
> }
> 
> func writeToFile(ch chan []byte) error {
> ```

You don't seem to return any error?

---

*[posting/backup.go, line 128 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8TaVrGaGwSgyTEwtc:-KV8TaVrGaGwSgyTEwtd:bhlrlxs) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L128)):*
> ```Go
> 	f, err := os.Create(file)
> 	x.Check(err)
> 	//gw := gzip.NewWriter(f)
> ```

Delete comment?

---

*[posting/backup.go, line 136 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8NF42O1QQfcNtu1q8:-KV8NF42O1QQfcNtu1q9:b-ebapz9) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L136)):*
> ```Go
> 	w.Flush()
> 	//gw.Close()
> 	f.Close()
> ```

defer f.Close() in case this function gets more complicated?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299)*

ashwin95r:
Review status: 0 of 5 files reviewed at latest revision, 14 unresolved discussions, some commit checks broke.

---

*[posting/backup.go, line 34 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8OC24hQVjtPyMKtk1:-KV9Ofsm7sGBwY_1c-oM:b-kya0n) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L34)):*
> Magical number 33?
</details>
made a constant

---

*[posting/backup.go, line 39 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8O6A_lVC7F1Q84Dcp:-KV9P1LRZ-PoSIWDmOij:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L39)):*
> How about "for it.SeekToFirst; it.Valid(); it.Next()" and delete the above line?
</details>
Done.

---

*[posting/backup.go, line 47 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8NwgOcRJ5C-ztfJXj:-KV9P7IS72fxcIoCsNAo:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L47)):*
> Maybe can add comment like "wait for convertToRdf"?
</details>
Done.

---

*[posting/backup.go, line 55 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8LuW9DJtI-TWBxCJK:-KV9P8yWB9T5sdh3BKvJ:bybkf6j) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L55)):*
> Converting key to a string requires a copy. If you have to do it, just do it once. (Actually I was proposing in a separate thread that we should all just stick to strings as far as possible)
> 
> Under "bytes" lib, there should be equivalent of string functions, e.g. HasPrefix. You might want to use that to avoid converting to string here.
> 
> Use constant indexRune?
</details>
using bytes throughout. Done.

---

*[posting/backup.go, line 58 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8NjBe-LlcCU_DNFnt:-KV9PTwzO-pVdIA4mhtS:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L58)):*
> This fmt.Printf is for debugging only? Delete?
</details>
Done.

---

*[posting/backup.go, line 60 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8MKpob9kg7wd3qpK5:-KV9Q86IK-EZ_i6C8leB:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L60)):*
> Use bytes.IndexRune or IndexByte? Also this function is generally useful. I don't think we have this yet. You could move it to a more central location, where Key() is.
</details>
Done.

---

*[posting/backup.go, line 77 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8MkQpRCfGe7v96okL:-KV9QKBYD67tYclPAlLD:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L77)):*
> Not checking errors? (i know it is painful but x.Check alleviates it a bit)
</details>
Done.

---

*[posting/backup.go, line 97 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8TkYU1eStY1ZDrE_-:-KV9QMN765AkoE0aA2r8:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L97)):*
> Check returned errors? (with x.Check)
</details>
Done.

---

*[posting/backup.go, line 123 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8NOGmp9staPuMFvUV:-KV9QXQdqvoCsU5mnC3-:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L123)):*
> You don't seem to return any error?
</details>
Done.

---

*[posting/backup.go, line 136 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8NF42O1QQfcNtu1q8:-KV9Z4VL1CIbrplWCWJV:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L136)):*
> defer f.Close() in case this function gets more complicated?
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299)*

ashwin95r:
Review status: 0 of 8 files reviewed at latest revision, 19 unresolved discussions, some commit checks failed.

---

*[posting/backup.go, line 128 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV8TaVrGaGwSgyTEwtc:-KVNn6LuI-ObhfTYr9ts:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L128)):*
> Delete comment?
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299)*

jchiu0:
Just a bunch of minor comments. I am not too sure about the RAFT part..

---

Review status: 0 of 8 files reviewed at latest revision, 14 unresolved discussions, some commit checks broke.

---

*[posting/backup.go, line 61 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV-ioiJmLN5012SvFCy-r2-61:-KVNwWIJ8UjRqyxW3kb0:bm09uvg) ([raw file](https://github.com/dgraph-io/dgraph/blob/6f5ecb80c0a3d7e31887d37639e05f07e13f898d/posting/backup.go#L61)):*
> var srcUid should be srcUID
</details>
Let's follow Go style naming, e.g., toRDF, not that I particularly like it.

---

*[posting/backup.go, line 35 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNwDYVOHFIExkpq6MW:-KVNwDYVOHFIExkpq6MX:b-6o3gzt) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L35)):*
> ```Go
> 
> 	wg.Add(numRoutines)
> 	for i := 0; i < numRoutines; i++ {
> ```

Maybe numBackupRoutines or backupRoutines?

---

*[posting/list.go, line 141 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV9RVlYWQf6Z-cSl_kF-r2-141:-KVNwvs2jOe3WEVexK2K:b-ypro70) ([raw file](https://github.com/dgraph-io/dgraph/blob/6f5ecb80c0a3d7e31887d37639e05f07e13f898d/posting/list.go#L141)):*
> exported function SplitKey should have comment or be unexported
</details>
Please fix and same for other exported functions.

---

*[worker/payload.proto, line 18 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNzNO-Bnh3X8D9mX5W:-KVNzNO-Bnh3X8D9mX5X:b-whgmp1) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/worker/payload.proto#L18)):*
> ```protobuf
> 	rpc RaftMessage (Payload) returns (Payload) {}
> 	rpc JoinCluster (Payload) returns (Payload) {}
>   rpc InitiateBackup (Payload) returns (Payload) {}
> ```

Fix indent (space to tab)

---

*[worker/worker.go, line 64 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNycNxjV_bF2QhmS9H:-KVNycNxjV_bF2QhmS9I:b-svg64g) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/worker/worker.go#L64)):*
> ```Go
> 	for _, v := range allServers {
> 		for _, ser := range v.list {
> 			if ser.Leader {
> ```

Nit..: if !ser.Leader || ser.Addr == .... { continue } to decrease indent.

---

*[worker/worker.go, line 72 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNythP4JKkzmpf3xYb:-KVNythP4JKkzmpf3xYc:b-7j1v13) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/worker/worker.go#L72)):*
> ```Go
> 				conn, err := pl.Get()
> 				x.Check(err)
> 				defer pl.Put(conn)
> ```

This defer will not happen until the function exits.. maybe you can just stick it at the end of the loop or put them in a function.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-:-KVNzQg4PXXHTFUWq4SN:b-kv72rx)*

manishrjain:
Review status: 0 of 8 files reviewed at latest revision, 29 unresolved discussions, some commit checks broke.

---

*[cmd/dgraph/main.go, line 60 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNvvVe4cFdG69gRMLP:-KVNvvVe4cFdG69gRMLQ:b-2vi2tc) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/cmd/dgraph/main.go#L60)):*
> ```Go
> 	nomutations = flag.Bool("nomutations", false, "Don't allow mutations on this server.")
> 	shutdown    = flag.Bool("shutdown", false, "Allow client to send shutdown signal.")
> 	backup      = flag.Bool("backup", false, "Allow client to request a backup.")
> ```

Instead of doing this, and doing shutdown, it's better to have an internal port (worker port), be used. You could call this for e.g.: IP:internal-port/admin/backup, or /admin/shutdown

---

*[posting/backup.go, line 14 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNwXOIenW5oLrW3c1w:-KVNwXOIenW5oLrW3c1x:bkp8mt5) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L14)):*
> ```Go
> 
> 	"github.com/dgraph-io/dgraph/posting/types"
> 	stype "github.com/dgraph-io/dgraph/types"
> ```

This needs merging.

---

*[posting/backup.go, line 29 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNwfO7ztTGFgOl54KG:-KVNwfO7ztTGFgOl54KH:ba3dfmt) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L29)):*
> ```Go
> 	chkv := make(chan kv, 1000)
> 	// Remove.
> 	aggressivelyEvict()
> ```

Remove.

---

*[posting/backup.go, line 40 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNwuF2CDZ4WHjRglzZ:-KVNwuF2CDZ4WHjRglz_:b-n1128t) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L40)):*
> ```Go
> 	go writeToFile(strconv.Itoa(int(gNum)), ch)
> 
> 	for it.SeekToFirst(); it.Valid(); it.Next() {
> ```

You need to check whether this is the right predicate for the group. Otherwise, you'll end up dumping everything.

Looks like you're doing that later. Better to do it here, so you can avoid a key and value copy. In fact, you could even advance your iterator to the end of the predicate (that you deemed is not part of the group), and then continue iteration like that. This would significantly speed up the iteration.

---

*[posting/backup.go, line 68 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNxrcznLvVx59i80Sq:-KVNxrcznLvVx59i80Sr:b280v62) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L68)):*
> ```Go
> 	_, err := buf.WriteString("<_uid_:")
> 	x.Check(err)
> 	_, err = buf.WriteString(srcUid)
> ```

This is ugly. Use fmt.Sprintf, and avoid this buf altogether.

---

*[posting/backup.go, line 82 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNyCqBZUZU6oB-hLrD:-KVNyCqBZUZU6oB-hLrE:b-p5hak0) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L82)):*
> ```Go
> 		if p.Uid() == math.MaxUint64 && !bytes.Equal(p.ValueBytes(), nil) {
> 			// Value posting
> 			buf2 := buf
> ```

buf := bytes.New...(prefix), where prefix is what you generated using fmt.Sprintf before.

---

*[posting/backup.go, line 83 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNySxhmuYhLTmuR-Y-:-KVNySxiHE_Z83JO1aU0:b-xlxr3y) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L83)):*
> ```Go
> 			// Value posting
> 			buf2 := buf
> 			_, err = buf2.WriteRune('"')
> ```

It seems like it might make sense to have another function in x.go, which would do a 2 arg check, using the second arg as error. Then you can reduce all these checks to one line.

---

*[posting/backup.go, line 116 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNyiWxp2WKwIniAAIn:-KVNyiWxp2WKwIniAAIo:b-72tsnb) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L116)):*
> ```Go
> 		_, err = buf2.WriteString("> .\n")
> 		x.Check(err)
> 		ch <- buf2.Bytes()
> ```

Instead of outputting so often, you could only output when buf exceeds a certain size, and when the PL iteration is over.

---

*[posting/backup.go, line 118 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNyzK8tagQLVmddG1y:-KVNyzK8tagQLVmddG1z:bng2u0h) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L118)):*
> ```Go
> 		ch <- buf2.Bytes()
> 	}
> 	return
> ```

This is unnecessary.

---

*[posting/backup.go, line 130 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNz8b8V3NCPHiuFBPo:-KVNz8b8V3NCPHiuFBPp:beex573) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L130)):*
> ```Go
> func writeToFile(str string, ch chan []byte) {
> 	file := fmt.Sprintf("backup/ddata-%s.gz", str)
> 	err := os.MkdirAll("backup", 0700)
> ```

Maybe add a date to the backup.

---

*[posting/backup.go, line 134 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNzbRuMH371Bmiomzu:-KVNzbRuMH371Bmiomzv:b-xc5c9p) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L134)):*
> ```Go
> 	defer f.Close()
> 	x.Check(err)
> 	w := bufio.NewWriter(f)
> ```

Specify a buffer in MB that you're willing to use.

---

*[posting/backup.go, line 142 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNzkcxLY3xGj4WqPK3:-KVNzkcxLY3xGj4WqPK4:b-g3ppod) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L142)):*
> ```Go
> 	gw.Flush()
> 	gw.Close()
> 	w.Flush()
> ```

w.Close?

---

*[posting/list.go, line 142 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNzr2saXjzljG2IAZf:-KVNzr2saXjzljG2IAZg:b4xpkma) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/list.go#L142)):*
> ```Go
> 
> func SplitKey(key []byte) (string, string, bool) {
> 	if bytes.HasPrefix(key, []byte("|")) || bytes.HasPrefix(key, []byte("_uid_")) {
> ```

Why would a key have '|' as prefix? also, why not backup _uid_? That's our mapping to figure out which uids are available, right?

---

*[worker/worker.go, line 65 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVO-bSHdobZxLZ0wuEO:-KVO-bSHdobZxLZ0wuEP:b8saqtj) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/worker/worker.go#L65)):*
> ```Go
> 		for _, ser := range v.list {
> 			if ser.Leader {
> 				if ser.Addr == groups().Leader(groups().num) {
> ```

I think you can simplify this, and send it to all the nodes in the cluster. The nodes can then figure out which group they're a leader of, and run the backup. Note that they should only reply back once the backup is complete.

---

*[worker/worker.go, line 77 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVO-tMV96LXL0kgBVCx:-KVO-tMWlW0sXAI0QYfq:bebkar8) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/worker/worker.go#L77)):*
> ```Go
> 				query := &Payload{}
> 				c := NewWorkerClient(conn)
> 				_, err = c.InitiateBackup(ctx, query)
> ```

You clearly want to do this in a goroutine, don't you?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299)*

ashwin95r:
Review status: 0 of 8 files reviewed at latest revision, 29 unresolved discussions, some commit checks broke.

---

*[cmd/dgraph/main.go, line 60 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNvvVe4cFdG69gRMLP:-KVOZBu_LCzUZXHWNJtH:b-b07jdc) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/cmd/dgraph/main.go#L60)):*
> Instead of doing this, and doing shutdown, it's better to have an internal port (worker port), be used. You could call this for e.g.: IP:internal-port/admin/backup, or /admin/shutdown
</details>
Will do it as a different task. I feel we should have a seperate port and not use the worker port. (we would need to use cmux for it otherwise which wont be great)

---

*[posting/backup.go, line 105 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV-ioiJmLN5012SvFCy-r1-105:-KVOIM9i4zYLjhN96eQM:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/d47ed76a6de65bb4de6692d8a8929af6504444d4/posting/backup.go#L105)):*
> var destUid should be destUID
</details>
Done.

---

*[posting/backup.go, line 61 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV-ioiJmLN5012SvFCy-r2-61:-KVOFu0NHZJm-lpW7jev:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/6f5ecb80c0a3d7e31887d37639e05f07e13f898d/posting/backup.go#L61)):*
> Let's follow Go style naming, e.g., toRDF, not that I particularly like it.
</details>
Done.

---

*[posting/backup.go, line 29 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNwfO7ztTGFgOl54KG:-KVO9Vf-m0CG92JrwHgr:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L29)):*
> Remove.
</details>
Done.

---

*[posting/backup.go, line 35 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNwDYVOHFIExkpq6MW:-KVO9i5oR58j2juyVjfN:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L35)):*
> Maybe numBackupRoutines or backupRoutines?
</details>
Done.

---

*[posting/backup.go, line 40 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNwuF2CDZ4WHjRglzZ:-KVO9w_JvFvzw7fuDyDU:bp2agc8) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L40)):*
> You need to check whether this is the right predicate for the group. Otherwise, you'll end up dumping everything.
> 
> Looks like you're doing that later. Better to do it here, so you can avoid a key and value copy. In fact, you could even advance your iterator to the end of the predicate (that you deemed is not part of the group), and then continue iteration like that. This would significantly speed up the iteration.
</details>
Skipping the keys at this stage.

---

*[posting/backup.go, line 68 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNxrcznLvVx59i80Sq:-KVOHRUWKUPJsGkzd4qO:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L68)):*
> This is ugly. Use fmt.Sprintf, and avoid this buf altogether.
</details>
Done.

---

*[posting/backup.go, line 82 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNyCqBZUZU6oB-hLrD:-KVOHRyljI4bpC3obIBV:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L82)):*
> buf := bytes.New...(prefix), where prefix is what you generated using fmt.Sprintf before.
</details>
Done.

---

*[posting/backup.go, line 83 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNySxhmuYhLTmuR-Y-:-KVOaa7qG5T8TxczfhQp:b-2ubflj) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L83)):*
> It seems like it might make sense to have another function in x.go, which would do a 2 arg check, using the second arg as error. Then you can reduce all these checks to one line.
</details>
I feel this might be more cleaner. something like x.Check2() might confuse people. Reduced the number of checks by using fmt.Sprintf()

---

*[posting/backup.go, line 116 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNyiWxp2WKwIniAAIn:-KVOIQbwQoNbIKlVgLsj:b-syky0t) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L116)):*
> Instead of outputting so often, you could only output when buf exceeds a certain size, and when the PL iteration is over.
</details>
This is one line that'd be written to a file. similar to what we do while reading in loader. Is there something I'm missing?

---

*[posting/backup.go, line 118 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNyzK8tagQLVmddG1y:-KVO2Q3zZa9S7N3hbkC1:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L118)):*
> This is unnecessary.
</details>
Done.

---

*[posting/backup.go, line 130 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNz8b8V3NCPHiuFBPo:-KVO5rG3cmspRDjzzDgH:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L130)):*
> Maybe add a date to the backup.
</details>
Done.

---

*[posting/backup.go, line 134 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNzbRuMH371Bmiomzu:-KVO7EBZzayYQdJOpB16:b9kn2hl) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L134)):*
> Specify a buffer in MB that you're willing to use.
</details>
Done. Using NewWriterSize now.

---

*[posting/backup.go, line 142 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNzkcxLY3xGj4WqPK3:-KVO5uvOIvSMDtW4R-oN:biaatlx) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/backup.go#L142)):*
> w.Close?
</details>
There is no close for this. Only Flush.

---

*[posting/list.go, line 141 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KV9RVlYWQf6Z-cSl_kF-r2-141:-KVO7aDa321BSUAzNRI6:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/6f5ecb80c0a3d7e31887d37639e05f07e13f898d/posting/list.go#L141)):*
> Please fix and same for other exported functions.
</details>
Done.

---

*[posting/list.go, line 142 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNzr2saXjzljG2IAZf:-KVO0r9Zwzx77he9EhQ9:bg8u714) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/posting/list.go#L142)):*
> Why would a key have '|' as prefix? also, why not backup _uid_? That's our mapping to figure out which uids are available, right?
</details>
sorry its a :, not a |. This backup would be run through loader, so won't entries would be recreated while loading?

---

*[worker/payload.proto, line 18 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNzNO-Bnh3X8D9mX5W:-KVOarkfwvwzI4hsHcdP:b-g7q9bk) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/worker/payload.proto#L18)):*
> Fix indent (space to tab)
</details>
My editor makes it spaces I think.

---

*[worker/worker.go, line 60 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVEAkHBlNSTE0QgFJUv-r3-60:-KVOIf_fh5ZslBvWt9a4:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/996b99db4ec393f66469b95c35693858d07e811c/worker/worker.go#L60)):*
> exported function BackupAll should have comment or be unexported
</details>
Done.

---

*[worker/worker.go, line 64 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNycNxjV_bF2QhmS9H:-KVO-LWKYe4PPp92C4gl:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/worker/worker.go#L64)):*
> Nit..: if !ser.Leader || ser.Addr == .... { continue } to decrease indent.
</details>
Done.

---

*[worker/worker.go, line 65 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVO-bSHdobZxLZ0wuEO:-KVOYuMhUOsR2Ov25n8I:bs6dggz) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/worker/worker.go#L65)):*
> I think you can simplify this, and send it to all the nodes in the cluster. The nodes can then figure out which group they're a leader of, and run the backup. Note that they should only reply back once the backup is complete.
</details>
Won't we reduce the number of calls if we request only the leaders? we have the information in this machine anyways right? Modified the logic to reply only after backup is over.

---

*[worker/worker.go, line 72 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNythP4JKkzmpf3xYb:-KVO-MRc9FroYjM1zri6:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/worker/worker.go#L72)):*
> This defer will not happen until the function exits.. maybe you can just stick it at the end of the loop or put them in a function.
</details>
Done.

---

*[worker/worker.go, line 77 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVO-tMV96LXL0kgBVCx:-KVOZAPQWxwcDebQbkWf:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/worker/worker.go#L77)):*
> You clearly want to do this in a goroutine, don't you?
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299)*

manishrjain:
Review status: 0 of 8 files reviewed at latest revision, 27 unresolved discussions, some commit checks broke.

---

*[posting/backup.go, line 26 at r6](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVOw10j1RdQc6BJtQlS:-KVOw10j1RdQc6BJtQlT:b2ni3dp) ([raw file](https://github.com/dgraph-io/dgraph/blob/f1eb14a2ff1a2fa6c19f83dd05923144c32bcd38/posting/backup.go#L26)):*
> ```Go
> 
> // Backup creates a backup of data by exporting it as an RDF gzip.
> func Backup(gNum uint32) error {
> ```

gid

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299)*

ashwin95r:
Review status: 0 of 14 files reviewed at latest revision, 27 unresolved discussions, some commit checks broke.

---

*[posting/backup.go, line 26 at r6](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVOw10j1RdQc6BJtQlS:-KVSuIX5a8DTQRiZayzM:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/f1eb14a2ff1a2fa6c19f83dd05923144c32bcd38/posting/backup.go#L26)):*
> gid
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299)*

ashwin95r:
Review status: 0 of 6 files reviewed at latest revision, 26 unresolved discussions.

---

*[worker/payload.proto, line 18 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNzNO-Bnh3X8D9mX5W:-KVTN1SS-GfDQAlPv9YK:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/worker/payload.proto#L18)):*
> My editor makes it spaces I think.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299)*

manishrjain:
Review status: 0 of 6 files reviewed at latest revision, 31 unresolved discussions, some commit checks failed.

---

*[cmd/dgraph/main.go, line 378 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVSxGEKS5ctyPzm53WJ:-KVSxGEKS5ctyPzm53WK:b-hgqc6g) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c593a93c767bea93507e613c14fea9fb9a1e3c8/cmd/dgraph/main.go#L378)):*
> ```Go
> 	}
> 
> 	if *backup && string(q) == "BACKUP" {
> ```

Just add a TODO here for the /admin/backup thing.

---

*[worker/backup.go, line 49 at r8](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVYI2hsnai-3O5ufAw9:-KVYI2hsnai-3O5ufAwA:b-f7w3ig) ([raw file](https://github.com/dgraph-io/dgraph/blob/cdeda31ab82c19226d8ce5b5b6f126384e32ca91/worker/backup.go#L49)):*
> ```Go
> 			pre = append(pre, '~')
> 			it.Seek(pre)
> 			continue
> ```

Does this work? I feel it shouldn't, because you never advance the iterator in continue. Do you have a test case?

---

*[types/index.go, line 14 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVT9C2_-Ce4-w4O7zn9:-KVT9C2_-Ce4-w4O7znA:b-vys8vh) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c593a93c767bea93507e613c14fea9fb9a1e3c8/types/index.go#L14)):*
> ```Go
> 	// Posting list keys are prefixed with this rune if it is a mutation meant for
> 	// the index.
> 	IndexRune   = ':'
> ```

Are you using it somewhere?

---

*[posting/backup.go, line 49 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVT7gJntGGVaY-UnCli:-KVT7gJntGGVaY-UnClj:b-jsw9xr) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c593a93c767bea93507e613c14fea9fb9a1e3c8/posting/backup.go#L49)):*
> ```Go
> 		}
> 		pred, uid := splitKey(it.Key().Data())
> 		if pred == lastPred && !belongs {
> ```

Make iterator seek.

---

*[posting/backup.go, line 79 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVSxQ5UExprJfQMAc_7:-KVSxQ5UExprJfQMAc_8:b-fb9z15) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c593a93c767bea93507e613c14fea9fb9a1e3c8/posting/backup.go#L79)):*
> ```Go
> }
> 
> func toRDF(item kv, ch chan []byte) {
> ```

This doesn't need to be 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299)*

ashwin95r:
Review status: 0 of 8 files reviewed at latest revision, 31 unresolved discussions, some commit checks failed.

---

*[cmd/dgraph/main.go, line 378 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVSxGEKS5ctyPzm53WJ:-KVYf6gvxdYBysKa0jWl:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c593a93c767bea93507e613c14fea9fb9a1e3c8/cmd/dgraph/main.go#L378)):*
> Just add a TODO here for the /admin/backup thing.
</details>
Done.

---

*[worker/backup.go, line 14 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVNwXOIenW5oLrW3c1w:-KVYf7uH5U8M59xOeecd:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/48615cdd5fda4cf2e3128bf8726f8cd531d7f5e3/worker/backup.go#L14)):*
> This needs merging.
</details>
Done.

---

*[worker/backup.go, line 49 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVT7gJntGGVaY-UnCli:-KVYfGtOocZKgaiN9Jjl:b6vl8b7) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c593a93c767bea93507e613c14fea9fb9a1e3c8/worker/backup.go#L49)):*
> Make iterator seek.
</details>
It is being seeked to ~.

---

*[worker/backup.go, line 79 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVSxQ5UExprJfQMAc_7:-KVYfMJpx41ryko5zP-i:b-ggpohd) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c593a93c767bea93507e613c14fea9fb9a1e3c8/worker/backup.go#L79)):*
> This doesn't need to be 
</details>
You mean RDF to Rdf?

---

*[worker/backup.go, line 49 at r8](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVYI2hsnai-3O5ufAw9:-KVYf939uM7pR338y_P9:b-c9tp5j) ([raw file](https://github.com/dgraph-io/dgraph/blob/cdeda31ab82c19226d8ce5b5b6f126384e32ca91/worker/backup.go#L49)):*
> Does this work? I feel it shouldn't, because you never advance the iterator in continue. Do you have a test case?
</details>
Added test, we seek the iterator to the start of next predicate using ~.

---

*[types/index.go, line 14 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVT9C2_-Ce4-w4O7zn9:-KVYfPk_Xy2KPFVBM8Um:b6nets0) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c593a93c767bea93507e613c14fea9fb9a1e3c8/types/index.go#L14)):*
> Are you using it somewhere?
</details>
Unexported it. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299)*

manishrjain:
Review status: 0 of 8 files reviewed at latest revision, 21 unresolved discussions, some commit checks failed.

---

*[worker/backup.go, line 132 at r10](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVYkOfuzly6eFiPjVpq:-KVYkOfuzly6eFiPjVpr:b-o0h3i1) ([raw file](https://github.com/dgraph-io/dgraph/blob/492f67429346b47ac35de60053b61d5ff78ec3e8/worker/backup.go#L132)):*
> ```Go
> 
> func writeToFile(str string, ch chan []byte, errChan chan error) {
> 	file := fmt.Sprintf("backup/ddata-%s_%s.gz", str,
> ```

ddata -> dgraph. Also avoid underscore. Use dash.

---

*[worker/backup.go, line 144 at r10](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVYkWTTK-x0eUAet8I9:-KVYkWTTK-x0eUAet8IA:bgoorta) ([raw file](https://github.com/dgraph-io/dgraph/blob/492f67429346b47ac35de60053b61d5ff78ec3e8/worker/backup.go#L144)):*
> ```Go
> 	defer f.Close()
> 	x.Check(err)
> 	w := bufio.NewWriterSize(f, 1024)
> ```

1MB write.

---

*[worker/worker.go, line 61 at r10](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVYjoJ6UJL-9zCyRWaV:-KVYjoJ6UJL-9zCyRWaW:b-mq65tg) ([raw file](https://github.com/dgraph-io/dgraph/blob/492f67429346b47ac35de60053b61d5ff78ec3e8/worker/worker.go#L61)):*
> ```Go
> 
> func initiateBackup() error {
> 	for gid, node := range groups().local {
> ```

groups().local should be accessed via mutex lock.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299)*

manishrjain:
Reviewed 7 of 16 files at r8, 2 of 2 files at r9, 1 of 1 files at r10.
Review status: all files reviewed at latest revision, 21 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299)*

ashwin95r:
Review status: 6 of 8 files reviewed at latest revision, 21 unresolved discussions, some commit checks failed.

---

*[worker/backup.go, line 132 at r10](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVYkOfuzly6eFiPjVpq:-KVYqpPahirnSpAGgfuc:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/492f67429346b47ac35de60053b61d5ff78ec3e8/worker/backup.go#L132)):*
> ddata -> dgraph. Also avoid underscore. Use dash.
</details>
Done.

---

*[worker/backup.go, line 144 at r10](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVYkWTTK-x0eUAet8I9:-KVYqvJsgB-ySTviXTiu:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/492f67429346b47ac35de60053b61d5ff78ec3e8/worker/backup.go#L144)):*
> 1MB write.
</details>
Done.

---

*[worker/worker.go, line 61 at r10](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVYjoJ6UJL-9zCyRWaV:-KVYrKwBzHNes-UIB21D:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/492f67429346b47ac35de60053b61d5ff78ec3e8/worker/worker.go#L61)):*
> groups().local should be accessed via mutex lock.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299)*

manishrjain:
Review status: 5 of 8 files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[worker/worker.go, line 37 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVZbz92eCGNbnF_ssfV:-KVZbz92eCGNbnF_ssfW:btfgby2) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/worker/worker.go#L37)):*
> ```Go
> 	workerPort = flag.Int("workerport", 12345,
> 		"Port used by worker for internal communication.")
> 	backupPath = flag.String("backuppath", "backup",
> ```

Just call the flag backup.

---

*[worker/worker.go, line 67 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVZe6nYeyeq-FeUIwsG:-KVZe6nYeyeq-FeUIwsH:b-oyqa34) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/worker/worker.go#L67)):*
> ```Go
> 	for gid, node := range groups().local {
> 		if node.AmLeader() {
> 			err := Backup(gid, *backupPath)
> ```

Holding a read lock for the entire duration of a Backup is a truly bad idea.

---

*[worker/worker.go, line 82 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVZd3jYl8JeCnI3E-fi:-KVZd3jYl8JeCnI3E-fj:b-7qgz0w) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/worker/worker.go#L82)):*
> ```Go
> 	count := 0
> 	// Send backup calls to other workers.
> 	for _, pl := range pools().all {
> ```

This isn't good. You should systematically go through all the group ids, pick their leader, and send it to them. This way, if you happen to not have the address of one of the group ids, at least you can raise an error and quit backup.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299)*

ashwin95r:
Review status: 5 of 8 files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[worker/worker.go, line 37 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVZbz92eCGNbnF_ssfV:-KVZeuuBUjUWuTRQHYIM:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/worker/worker.go#L37)):*
> Just call the flag backup.
</details>
Done.

---

*[worker/worker.go, line 67 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVZe6nYeyeq-FeUIwsG:-KVZgnEMkr4brgYhMTgL:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/worker/worker.go#L67)):*
> Holding a read lock for the entire duration of a Backup is a truly bad idea.
</details>
Done.

---

*[worker/worker.go, line 82 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVZd3jYl8JeCnI3E-fi:-KVZm3mfcQdvrFeyzcif:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/worker/worker.go#L82)):*
> This isn't good. You should systematically go through all the group ids, pick their leader, and send it to them. This way, if you happen to not have the address of one of the group ids, at least you can raise an error and quit backup.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299)*

jchiu0:
(just some minor comments)
:lgtm: LG for BackupAll issuing backup rpc calls to everyone and only letting leader nodes do the backup.

---

Reviewed 1 of 16 files at r8.
Review status: 5 of 8 files reviewed at latest revision, 24 unresolved discussions, some commit checks failed.

---

*[cmd/dgraph/main.go, line 364 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVd7Aphk38eBIrqDLev:-KVd7Aphk38eBIrqDLew:brqshlo) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/cmd/dgraph/main.go#L364)):*
> ```Go
> 	l.Start = time.Now()
> 	defer r.Body.Close()
> 	q, err := ioutil.ReadAll(r.Body)
> ```

let's cast q to a string once and for all, in this function.

---

*[worker/backup.go, line 38 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVcqnTrCDn5gQyktoWk:-KVcqnTsFRJ3Gm2p95ie:b-mi2m97) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/worker/backup.go#L38)):*
> ```Go
> 	wg.Add(numBackupRoutines)
> 	for i := 0; i < numBackupRoutines; i++ {
> 		go convertToRdf(&wg, chkv, ch)
> ```

It seems that convertToRDF is a small function that is used here only? If so, maybe we can inline.

---

*[worker/backup.go, line 117 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVcrHWwG6_Kovez-1om:-KVcrHWwG6_Kovez-1on:bcxjkz7) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/worker/backup.go#L117)):*
> ```Go
> 		x.Check(err)
> 		if buf.Len() >= 50000 {
> 			ch <- buf.Bytes()
> ```

Wonder if we should log an error here or something. (The buffer gets truncated and would miss a "\n" which is not so nice, but I think it is rare enough if ignore.)

---

*[worker/backup.go, line 132 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVcrhjwoKxK4SIwzauT:-KVcrhjwoKxK4SIwzauU:b-tbfvby) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/worker/backup.go#L132)):*
> ```Go
> 
> func writeToFile(str, bpath string, ch chan []byte, errChan chan error) {
> 	file := fmt.Sprintf("%s/dgraph-%s-%s.gz", bpath, str,
> ```

Optional: path.Join could be nicer just in case bpath ends with /.

---

*[worker/backup_test.go, line 119 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVd6Mhi79zWvcrhHltx:-KVd6Mhi79zWvcrhHlty:b-lablpb) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/worker/backup_test.go#L119)):*
> ```Go
> 			// The only objectId we set was uid 5.
> 			if nq.ObjectId != "" {
> 				require.Equal(t, nq.ObjectId, "_uid_:5")
> ```

I learn this recently. We should write require.Equal(t, expectedValue, valueBeingTested) instead, i.e., swap 2nd and 3rd arguments.

---

*[worker/backup_test.go, line 127 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVd6W8nwIyWPyaN_HXJ:-KVd6W8nwIyWPyaN_HXK:b-wy5sf0) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/worker/backup_test.go#L127)):*
> ```Go
> 	}
> 	// This order will bw presereved due to file naming.
> 	require.Equal(t, counts, []int{4, 2})
> ```

Same here. Swap.

---

*[worker/worker.go, line 67 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVZe6nYeyeq-FeUIwsG:-KVd7YrcoTttHuxGtlXl:b-94tkht) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/worker/worker.go#L67)):*
> Done.
</details>
groups.RLock is fine I suppose?

---

*[worker/worker.go, line 76 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVd6vW3WAz-mTuJ7oBF:-KVd6vW4qeK59r8tnXO_:b-14hf5n) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/worker/worker.go#L76)):*
> ```Go
> }
> 
> // BackupAll Creates backup of all nodes in cluster by initiaing backups in all the master
> ```

typo: initiating

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-:-KVd7HXDtI5lJiat_OIP:bpr73gy)*

ashwin95r:
Review status: 5 of 8 files reviewed at latest revision, 24 unresolved discussions, some commit checks failed.

---

*[cmd/dgraph/main.go, line 364 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVd7Aphk38eBIrqDLev:-KVdTOhCX84JFXZbsxdT:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/cmd/dgraph/main.go#L364)):*
> let's cast q to a string once and for all, in this function.
</details>
Done.

---

*[worker/backup.go, line 38 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVcqnTrCDn5gQyktoWk:-KVdU9B7CzKAdMJDw3oD:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/worker/backup.go#L38)):*
> It seems that convertToRDF is a small function that is used here only? If so, maybe we can inline.
</details>
Done.

---

*[worker/backup.go, line 117 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVcrHWwG6_Kovez-1om:-KVdWu_4uhurnbvO87hy:b-122cc) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/worker/backup.go#L117)):*
> Wonder if we should log an error here or something. (The buffer gets truncated and would miss a "\n" which is not so nice, but I think it is rare enough if ignore.)
</details>
Think this is clarified.

---

*[worker/backup.go, line 132 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVcrhjwoKxK4SIwzauT:-KVdUwdeEm7yVtLGKNY3:brgq1k0) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/worker/backup.go#L132)):*
> Optional: path.Join could be nicer just in case bpath ends with /.
</details>
Sure. Done.

---

*[worker/backup_test.go, line 119 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVd6Mhi79zWvcrhHltx:-KVdVKEMYfhwyHyrmh44:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/worker/backup_test.go#L119)):*
> I learn this recently. We should write require.Equal(t, expectedValue, valueBeingTested) instead, i.e., swap 2nd and 3rd arguments.
</details>
Done.

---

*[worker/backup_test.go, line 127 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVd6W8nwIyWPyaN_HXJ:-KVdVPTpCcF5kAfZvVjx:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/worker/backup_test.go#L127)):*
> Same here. Swap.
</details>
Done.

---

*[worker/worker.go, line 67 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVZe6nYeyeq-FeUIwsG:-KVdX3xZ5oYqYGxnJ1qT:bs9ykcc) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/worker/worker.go#L67)):*
> groups.RLock is fine I suppose?
</details>
Might prevent updates to groups() if backup takes too long. so just storing gids before starting the backup.

---

*[worker/worker.go, line 76 at r11](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVd6vW3WAz-mTuJ7oBF:-KVdVgicKkquk9wFxR7o:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/a7149051ee0b0f3263c6601205d26094ca858e15/worker/worker.go#L76)):*
> typo: initiating
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299)*

manishrjain:
Reviewed 1 of 16 files at r8, 4 of 4 files at r12.
Review status: all files reviewed at latest revision, 27 unresolved discussions, some commit checks failed.

---

*[worker/worker.go, line 62 at r12](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVdspAxv-kLdFqAH5Ai:-KVdspAyDHboGebOnDPt:bvzrlg8) ([raw file](https://github.com/dgraph-io/dgraph/blob/c37b226dd95c37594ce75e9ded409bdb56a9821f/worker/worker.go#L62)):*
> ```Go
> 
> func initiateBackup() error {
> 	var grpList []uint32
> ```

groupList

---

*[worker/worker.go, line 93 at r12](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVduSD9OBVqzwNwq27u:-KVduSD9OBVqzwNwq27v:bg0533x) ([raw file](https://github.com/dgraph-io/dgraph/blob/c37b226dd95c37594ce75e9ded409bdb56a9821f/worker/worker.go#L93)):*
> ```Go
> 			// Don't add the current nodes address in this.
> 			if it.Leader {
> 				addrMap[it.Addr] = true
> ```

You can break after this. Because we already found the leader of this group.

---

*[worker/worker.go, line 110 at r12](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVdvQAojSNwVa9DkX6P:-KVdvQAojSNwVa9DkX6Q:b-esq5x5) ([raw file](https://github.com/dgraph-io/dgraph/blob/c37b226dd95c37594ce75e9ded409bdb56a9821f/worker/worker.go#L110)):*
> ```Go
> 			query := &Payload{}
> 			c := NewWorkerClient(conn)
> 			_, err = c.InitiateBackup(ctx, query)
> ```

I think you'll have to send out the group id that you need backed up. Otherwise, you'll end up with the same group id replicated multiple times.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299)*

ashwin95r:
Review status: all files reviewed at latest revision, 27 unresolved discussions, some commit checks failed.

---

*[worker/worker.go, line 62 at r12](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVdspAxv-kLdFqAH5Ai:-KVdyvdo61CplkG7Fcge:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/c37b226dd95c37594ce75e9ded409bdb56a9821f/worker/worker.go#L62)):*
> groupList
</details>
Done.

---

*[worker/worker.go, line 93 at r12](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVduSD9OBVqzwNwq27u:-KVdzCTcVt2j0xKY1oMb:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/c37b226dd95c37594ce75e9ded409bdb56a9821f/worker/worker.go#L93)):*
> You can break after this. Because we already found the leader of this group.
</details>
Done.

---

*[worker/worker.go, line 110 at r12](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVdvQAojSNwVa9DkX6P:-KVe5UJXWWfkr3Yio7vw:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/c37b226dd95c37594ce75e9ded409bdb56a9821f/worker/worker.go#L110)):*
> I think you'll have to send out the group id that you need backed up. Otherwise, you'll end up with the same group id replicated multiple times.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299)*

manishrjain:
:lgtm:

---

Reviewed 1 of 6 files at r15, 1 of 2 files at r16.
Review status: 4 of 8 files reviewed at latest revision, 32 unresolved discussions, some commit checks failed.

---

*[posting/list.go, line 148 at r16](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVi3STVNuJwtqWNqypQ:-KVi3STVNuJwtqWNqypR:bqlw4v8) ([raw file](https://github.com/dgraph-io/dgraph/blob/658576497cdfae9773446a5a841a45ec70de54f8/posting/list.go#L148)):*
> ```Go
> // (Note that it is not applicable to index keys)
> func SplitKey(key []byte) (string, uint64) {
> 	sKeys := bytes.Split(key, []byte("|"))
> ```

If you don't find the `|`, then what do you do? For e.g. for index keys.x

---

*[worker/backup.go, line 40 at r16](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVi402YWHAYoUxcpgwR:-KVi402YWHAYoUxcpgwS:bi6vh7a) ([raw file](https://github.com/dgraph-io/dgraph/blob/658576497cdfae9773446a5a841a45ec70de54f8/worker/backup.go#L40)):*
> ```Go
> 	for i := 0; i < numBackupRoutines; i++ {
> 		go func() {
> 			for it := range chkv {
> ```

This it is the same the it defined outside. Maybe change the name of this one.

---

*[worker/worker.go, line 67 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVe6uQ8gGEPtreif4fl:-KVe6uQ8gGEPtreif4fm:buw4bts) ([raw file](https://github.com/dgraph-io/dgraph/blob/838a46eae4ee7131f13bc0661958d88fa6ba990a/worker/worker.go#L67)):*
> ```Go
> func initiateBackup(gid uint32) error {
> 	groups().RLock()
> 	if !groups().Node(gid).AmLeader() {
> ```

If you're not accessing internal variables directly, don't acquire locks.

---

*[worker/worker.go, line 72 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVe7CRdqpgiQv6NPV-V:-KVe7CRdqpgiQv6NPV-W:bvq9qt9) ([raw file](https://github.com/dgraph-io/dgraph/blob/838a46eae4ee7131f13bc0661958d88fa6ba990a/worker/worker.go#L72)):*
> ```Go
> 	}
> 	groups().RUnlock()
> 	err := Backup(gid, *backupPath)
> ```

return Backup(...), also, backup can be a private method.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-:-KVi4XpJHiRPR6hjZ_Q0:bnfp4nl)*

ashwin95r:
Review status: 4 of 8 files reviewed at latest revision, 32 unresolved discussions, some commit checks failed.

---

*[posting/list.go, line 148 at r16](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVi3STVNuJwtqWNqypQ:-KVi6rMoFrJQgWb-YL3w:b2ysg7) ([raw file](https://github.com/dgraph-io/dgraph/blob/658576497cdfae9773446a5a841a45ec70de54f8/posting/list.go#L148)):*
> If you don't find the `|`, then what do you do? For e.g. for index keys.x
</details>
This won't be applicable to index keys. Mentioned in the comment.

---

*[worker/backup.go, line 40 at r16](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVi402YWHAYoUxcpgwR:-KVi7LlSbetg1XUVOLtD:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/658576497cdfae9773446a5a841a45ec70de54f8/worker/backup.go#L40)):*
> This it is the same the it defined outside. Maybe change the name of this one.
</details>
Done.

---

*[worker/worker.go, line 67 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVe6uQ8gGEPtreif4fl:-KVi7Z7bcyOffm5croEd:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/838a46eae4ee7131f13bc0661958d88fa6ba990a/worker/worker.go#L67)):*
> If you're not accessing internal variables directly, don't acquire locks.
</details>
Done.

---

*[worker/worker.go, line 72 at r13](https://reviewable.io:443/reviews/dgraph-io/dgraph/299#-KVe7CRdqpgiQv6NPV-V:-KVi7kxo_p_98KXR6ANQ:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/838a46eae4ee7131f13bc0661958d88fa6ba990a/worker/worker.go#L72)):*
> return Backup(...), also, backup can be a private method.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/299)*

PR #304, opened 2016-10-29
Improv/sort

manishrjain:
Small refactorings around sort PR. Disable broken geo and worker/predicate tests for now. Fix others. Disallow infinite sort count. Limit the sort count to 1000, to avoid hitting all buckets.

---

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/304)


jchiu0:
LG for UIDList and all other refactoring.
LG for combining offset and UIDList in sorting.
Just some minor comments.

---

Review status: 0 of 16 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[algo/uidlist.go, line 209 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/304#-KVFqsArFE948WMGZD0r:-KVFqsArFE948WMGZD0s:byh4l07) ([raw file](https://github.com/dgraph-io/dgraph/blob/307ef4975dcbe03a8114766288df60e4e3c3f169/algo/uidlist.go#L209)):*
> ```Go
> 		val := shortList.Get(i)
> 		if i > 0 && val == shortList.Get(i-1) {
> 			log.Fatal("We shouldn't have duplicates in UIDLists")
> ```

x.Assert for stack trace

---

*[types/index.go, line 24 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/304#-KVFvKJbvbnF3pkkIPm4:-KVFvKJbvbnF3pkkIPm5:b2s9efc) ([raw file](https://github.com/dgraph-io/dgraph/blob/307ef4975dcbe03a8114766288df60e4e3c3f169/types/index.go#L24)):*
> ```Go
> 	_, err := buf.WriteString(attr)
> 	x.Check(err)
> 	_, err = buf.WriteRune(indexRune)
> ```

Unintentional change?

---

*[worker/groups.go, line 54 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/304#-KVFQWSs5CxTyp_hhl9N-r2-54:-KVFw41eXUJair4El6NW:b-rll82e) ([raw file](https://github.com/dgraph-io/dgraph/blob/7086e1ca3155a962d3b64031e6b1d9a5a45ca48b/worker/groups.go#L54)):*
> exported function StartRaftNodes should have comment or be unexported
</details>
Please fix.

---

*[worker/sort.go, line 153 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/304#-KVFt0S03xYogvyg0S9k:-KVFt0S03xYogvyg0S9l:b-a67kzj) ([raw file](https://github.com/dgraph-io/dgraph/blob/307ef4975dcbe03a8114766288df60e4e3c3f169/worker/sort.go#L153)):*
> ```Go
> func intersectBucket(ts *task.Sort, attr string, token string, out []intersectedList) error {
> 	count := int(ts.Count())
> 	sType := schema.TypeOf(attr)
> ```

"scalar" was intentionally computed out of the loop over posting lists.. It actually does a lookup into typeNameMap which is a map[string]Type, which feels like a not-too-cheap operation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/304#-:-KVFvgKWP3E3_kArkazb:bcdfsqj)*

manishrjain:
Review status: 0 of 16 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[algo/uidlist.go, line 209 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/304#-KVFqsArFE948WMGZD0r:-KVHjAzMCTD7pPkYDSOz:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/307ef4975dcbe03a8114766288df60e4e3c3f169/algo/uidlist.go#L209)):*
> x.Assert for stack trace
</details>
Done.

---

*[types/index.go, line 24 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/304#-KVFvKJbvbnF3pkkIPm4:-KVHk7vORyPClfTjd-1c:b-kigfvf) ([raw file](https://github.com/dgraph-io/dgraph/blob/307ef4975dcbe03a8114766288df60e4e3c3f169/types/index.go#L24)):*
> Unintentional change?
</details>
No. Actually wanted to change this, so index is co-located with the data.

---

*[worker/groups.go, line 54 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/304#-KVFQWSs5CxTyp_hhl9N-r2-54:-KVHlWnG6ZL-8UdaB3zx:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/7086e1ca3155a962d3b64031e6b1d9a5a45ca48b/worker/groups.go#L54)):*
> Please fix.
</details>
Done.

---

*[worker/sort.go, line 153 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/304#-KVFt0S03xYogvyg0S9k:-KVHl_BkNu-rXov91nhy:brl76mi) ([raw file](https://github.com/dgraph-io/dgraph/blob/307ef4975dcbe03a8114766288df60e4e3c3f169/worker/sort.go#L153)):*
> "scalar" was intentionally computed out of the loop over posting lists.. It actually does a lookup into typeNameMap which is a map[string]Type, which feels like a not-too-cheap operation.
</details>
If we were making thousands of calls in a cheap loop rapidly then agree. Otherwise, we're talking about an in-memory O(num-buckers) hash lookups, against O(num-buckers*thousands*thousands) of look-ups in RocksDB, plus more of thousands of lookups to do the sort of individual UIDLists.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/304)*

jchiu0:
Just one comment about TokenFromKey. I can fix it later as well. :lgtm: 

---

Review status: 0 of 16 files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---

*[types/index.go, line 24 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/304#-KVFvKJbvbnF3pkkIPm4:-KVILm_MnrFSgAVeFKXK:b-b2ixa0) ([raw file](https://github.com/dgraph-io/dgraph/blob/307ef4975dcbe03a8114766288df60e4e3c3f169/types/index.go#L24)):*
> No. Actually wanted to change this, so index is co-located with the data.
</details>
Index is currently co-located. When making those RPC calls, we use the attribute not the key to decide which machine to pass to, and the attribute is passed down from query.

But yes, it is good to be consistent and use a simpler key.

There is however, TokenFromKey (just below) which would need to be updated. Please search for a : instead of a |. TokenFromKey is used to build the TokensTable. We iterate over keys in RocksDB to do this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/304#-:-KVIMbvOfI4DeYN705Im:bnlf0k2)*

jchiu0:
Review status: 0 of 16 files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---

*[algo/uidlist.go, line 208 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/304#-KVIP9a5qDzk-rOzvcPI:-KVIP9a6svIy42P_JYC6:b-1g3kb0) ([raw file](https://github.com/dgraph-io/dgraph/blob/0c4be6c3f1cff7aa6030daeb4a0806c24991badd/algo/uidlist.go#L208)):*
> ```Go
> 		val := shortList.Get(i)
> 		if i > 0 && val == shortList.Get(i-1) {
> 			x.Assertf(false, "We shouldn't have duplicates in UIDLists")
> ```

Maybe x.Assertf(i==0 || val != shortList.Get(i-1), ...)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/304)*

manishrjain:
Review status: 0 of 16 files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---

*[types/index.go, line 24 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/304#-KVFvKJbvbnF3pkkIPm4:-KVd7iXeCKHxUgKerhbK:b-vltusj) ([raw file](https://github.com/dgraph-io/dgraph/blob/307ef4975dcbe03a8114766288df60e4e3c3f169/types/index.go#L24)):*
> Index is currently co-located. When making those RPC calls, we use the attribute not the key to decide which machine to pass to, and the attribute is passed down from query.
> 
> But yes, it is good to be consistent and use a simpler key.
> 
> There is however, TokenFromKey (just below) which would need to be updated. Please search for a : instead of a |. TokenFromKey is used to build the TokensTable. We iterate over keys in RocksDB to do this.
</details>
Sorry, just saw this. Once you LGTM, then typically, I expect that the comments that you had last, are the only ones -- so I addressed those and submitted.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/304)*

PR #370, opened 2016-12-08
Some optimizations in intersection of lists.

ashishnegi:
1. Stopping intersection if some list has reached its end.
2. Doing binary search while finding index of equal or greater than element.

Also, 
```
// IntersectWith intersects u with v. The update is made to u. It assumes
// that u, v are sorted, which are not always the case.
func IntersectWith(u, v *task.List) {
```
Will not non-sorted lists give wrong results ?

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/370)

### DIFF
--- a/algo/uidlist.go
+++ b/algo/uidlist.go
@@ -31,6 +31,7 @@ func IntersectWith(u, v *task.List) {
 		k = indexOfEqualOrGreaterVal(v, k, uid)
 		if k < m && v.Uids[k] == uid {
 			out = append(out, uid)
+			k++
 		}
 	}
 	u.Uids = out
--- a/algo/uidlist.go
+++ b/algo/uidlist.go
@@ -19,19 +19,25 @@ func ApplyFilter(u *task.List, f func(uint64, int) bool) {
 	u.Uids = out
 }
 
-// IntersectWith intersects u with v. The update is made to u. It assumes
-// that u, v are sorted, which are not always the case.
+// IntersectWith intersects u with v. The update is made to u.
+// u, v should be sorted.
 func IntersectWith(u, v *task.List) {
 	out := u.Uids[:0]
 	n := len(u.Uids)
 	m := len(v.Uids)
-	var k int
-	for i := 0; i < n && k < m; i++ {
+	for i, k := 0, 0; i < n && k < m; {
 		uid := u.Uids[i]
-		k = indexOfEqualOrGreaterVal(v, k, uid)
-		if k < m && v.Uids[k] == uid {
+		vid := v.Uids[k]
+		if uid > vid {
+			for k = k + 1; k < m && v.Uids[k] < uid; k++ {
+			}
+		} else if uid == vid {
 			out = append(out, uid)
 			k++
+			i++
+		} else {
+			for i = i + 1; i < n && u.Uids[i] < vid; i++ {
+			}
 		}
 	}
 	u.Uids = out
--- a/algo/uidlist.go
+++ b/algo/uidlist.go
@@ -79,15 +79,15 @@ func IntersectSorted(lists []*task.List) *task.List {
 	// lptrs[j] is the element we are looking at for lists[j].
 	lptrs := make([]int, len(lists))
 	shortList := lists[minLenIdx]
-	for i := 0; i < len(shortList.Uids); i++ {
+	elemsLeft := true  // If some list has no elems left, we can't intersect more.
+
+	for i := 0; i < len(shortList.Uids) && elemsLeft; i++ {
 		val := shortList.Uids[i]
 		if i > 0 && val == shortList.Uids[i-1] {
 			x.AssertTruef(false, "We shouldn't have duplicates in UIDLists")
 		}
 
-		var skip bool         // Should we skip val in output?
-		var noElemsLeft bool  // If some list has no elems left, we can't intersect more.
-
+		var skip bool                     // Should we skip val in output?
 		for j := 0; j < len(lists); j++ { // For each other list in lists.
 			if j == minLenIdx {
 				// No point checking yourself.
@@ -101,9 +101,8 @@ func IntersectSorted(lists []*task.List) *task.List {
 			}
 
 			lptrs[j] = ljp
-
 			if ljp >= lsz || lj.Uids[ljp] > val {
-				noElemsLeft = ljp >= lsz
+				elemsLeft = ljp < lsz
 				skip = true
 				break
 			}
@@ -111,9 +110,6 @@ func IntersectSorted(lists []*task.List) *task.List {
 		}
 		if !skip {
 			output = append(output, val)
-		}
-		if noElemsLeft {
-			break
 		}
 	}
 	return &task.List{Uids: output}
--- a/algo/uidlist.go
+++ b/algo/uidlist.go
@@ -97,7 +97,9 @@ func IntersectSorted(lists []*task.List) *task.List {
 			lj := lists[j]
 			ljp := lptrs[j]
 			lsz := len(lj.Uids)
-			ljp = indexOfEqualOrGreaterVal(lj, ljp, val)
+			for ; ljp < lsz && lj.Uids[ljp] < val; ljp++ {
+			}
+
 			lptrs[j] = ljp
 
 			if ljp >= lsz || lj.Uids[ljp] > val {
@@ -167,13 +169,6 @@ func IndexOf(u *task.List, uid uint64) int {
 		return i
 	}
 	return -1
-}
-
-// IndexOf value which is Equal (Eq) Or Greater than (Gt) the uid passed on
-// the Uids slice, starting to look from startIdx (pass 0 for whole range search).
-func indexOfEqualOrGreaterVal(u *task.List, startIdx int, uid uint64) int {
-	return startIdx + sort.Search(len(u.Uids)-startIdx,
-		func(i int) bool { return u.Uids[startIdx+i] >= uid })
 }
 
 // ToUintsListForTest converts to list of uints for testing purpose only.
--- a/algo/uidlist.go
+++ b/algo/uidlist.go
@@ -26,10 +26,9 @@ func IntersectWith(u, v *task.List) {
 	n := len(u.Uids)
 	m := len(v.Uids)
 	var k int
-	for i := 0; i < n; i++ {
+	for i := 0; i < n && k < m; i++ {
 		uid := u.Uids[i]
-		for ; k < m && v.Uids[k] < uid; k++ {
-		}
+		k = indexOfEqualOrGreaterVal(v, k, uid)
 		if k < m && v.Uids[k] == uid {
 			out = append(out, uid)
 		}
@@ -79,7 +78,9 @@ func IntersectSorted(lists []*task.List) *task.List {
 			x.AssertTruef(false, "We shouldn't have duplicates in UIDLists")
 		}
 
-		var skip bool                     // Should we skip val in output?
+		var skip bool         // Should we skip val in output?
+		var noElemsLeft bool  // If some list has no elems left, we can't intersect more.
+
 		for j := 0; j < len(lists); j++ { // For each other list in lists.
 			if j == minLenIdx {
 				// No point checking yourself.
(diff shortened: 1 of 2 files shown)


manishrjain:
Thanks for the PR, Ashish! This is pretty awesome.

Lists are always going to be sorted. That comment seems wrong. Maybe check with Jiawei, I think he wrote it.

---

Reviewed 1 of 1 files at r1.

---

*[algo/uidlist.go, line 93 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/370#-KYWbLUwyMWQU0L1JOo5:-KYWbLUwyMWQU0L1JOo6:b-pifl79) ([raw file](https://github.com/dgraph-io/dgraph/blob/20863e8a72c69416c30b0fea274d88cf5eee70bd/algo/uidlist.go#L93)):*
> ```Go
> 			ljp := lptrs[j]
> 			lsz := len(lj.Uids)
> 			ljp = indexOfEqualOrGreaterVal(lj, ljp, val)
> ```

It's not clear that using a binary search is going to be faster here than what we do now. Can you add some benchmarks, using various scenarios; ranging from two lists being exactly equal to two lists having no intersection at all, and levels in between.

---

*[algo/uidlist.go, line 167 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/370#-KYWb9R_mUztfmf2tgGu:-KYWb9R_mUztfmf2tgGv:brdqpoc) ([raw file](https://github.com/dgraph-io/dgraph/blob/20863e8a72c69416c30b0fea274d88cf5eee70bd/algo/uidlist.go#L167)):*
> ```Go
> // IndexOf value which is Equal (Eq) Or Greater than (Gt) the uid passed on
> // the Uids slice, starting to look from startIdx (pass 0 for whole range search).
> func indexOfEqualOrGreaterVal(u *task.List, startIdx int, uid uint64) int {
> ```

This could be called geIndex.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/370#-:-KYWdzyzap5iujY1GDwj:b-58cmew)*

ashishnegi:
yes, benchmarking would be helpful.

Also, i also need to see how `sort.Search` works. If it does binary search starting from middle of list and so on, then `indexOfGreaterOrEqualVal` would be slower for lists which are very similar. If it moves forward in steps like 1,2,4,8.. then this can be faster on general cases.
I will do the benchmarking.

ashishnegi:
I benchmarked `IntersectWith` fn that takes 2 lists.
This current PR will make the functions slower.
After debugging, i found that go is `not inlining fn calls`.
I then put code in between `binary search (binmin)`. That is again very slow (>5 times). I think that is because of random access rather than sequential run of the tight for loop. But i do not understand this completely.

I then tried with increasing the steps of sequential run of loop in each iteration as : `k, k+1, k+3, k+7,k+15...`
I use `binary forward (binfor)` name for this in the benchmarks.

`Important`: i had to tweak code to reduce extra memory allocations for the tests, as `IntersectWith` mutates its inputs. I commented that line for all algos. Before, each function invocation, we now just copy back original list on the input list.

Data set : Length of list varies from 100 to 1000000 in factor of 10.
Uniform data : one list has `0,1,2,3...` (factor 1) another list has `0,2,4...` (if factor is 2)

Reading benchmarks : name: `BenchmarkListIntersectUniDist/2:1:binfor:size=100-8`
means `uniform distribution` with first list (u) has data in factor of 2, second has in factor of 1.

[Uniform data results](https://github.com/ashishnegi/dgraph/blob/algo-intersection-bench/algo/benchmarks/benchmarks-f3d232d0db99253160ea6eeb3a34b1d6946d880c.txt) There are also some results with random data at end of this file.

[Non uniform data results](https://github.com/ashishnegi/dgraph/blob/algo-intersection-bench/algo/benchmarks/benchmarks-fc0bfe1c726bfee0657be2f327605288464fa1c3.txt) benchmarks for lists with first/last N % data similar.

Complete details and findings are listed in : [Readme](https://github.com/ashishnegi/dgraph/blob/algo-intersection-bench/algo/benchmarks/readme.md)

Findings :

    Binary forward (binfor) is ~10 % slower when similarity is greater than 10 %.
    If similarity in lists is less than 10 %, it can be 2-3 times faster. 
    This is expected of binary search.

    We take first list (u) as reference and move over other list (v) :
    current implementation : if u is 0,1,2,.. and v is 0,100,200.. i.e. u is denser,
                                             it can be more than 3 times slower than if we have u and v switched.
    Looking at code, this (now) makes sense
    as in first case, we will loop over u completely, while stuck at some element of v
    in second case, we will bail out faster (condition: k < m) and inner loop checks less conditions.

Benchmarks are as good as the benchmarking code. if someone can look over the [code](https://github.com/ashishnegi/dgraph/blob/algo-intersection-bench/algo/uidlist.go#L22) and [benchmarking code](https://github.com/ashishnegi/dgraph/blob/algo-intersection-bench/algo/uidlist_test.go#L307), it will be helpful. I might have overlooked something, since this is my first few days with `go`.

I run benchmark with `go test -bench=.`
Every benchmark file is with `benchmarks-<git-commit-hash>.txt` You can also checkout to that hash to see the results.

ashishnegi:
Looking at the data results again : 

~~`Non uniform data: (more real world ?) : results are somewhat better for binfor.`~~
`Non uniform data: (more real world ?) : binfor is twice as fast than actual.` :)

However, does this kind of data and data length rightly represent our use case ?

manishrjain:
Let's find some time tomorrow to talk this over.

ashishnegi:
Yes, this turned out to be more involved then i thought.
Adding some observations, in case i forget it.
Based on the observation that if `IntersectWith` with `u` and `v` takes `t` time, if we change the order of arguments, it takes time `~2-3 t`. This happens when `u` is denser like `0,1,2...` and `v` is `0,100,200..`.

It is because we are moving over every element of `u` and stuck somewhere in `v` most of the time. When `v` is first argument, we bail out as we reach end of `u` faster.

I added looping over `u` as well. [code](https://github.com/ashishnegi/dgraph/blob/algo-intersection-bench/algo/uidlist.go#L162). Now, in both cases we are getting similar timings. `actua2` is this tweak. [benchmarks](https://github.com/ashishnegi/dgraph/blob/algo-intersection-bench/algo/benchmarks/benchmarks-2f1a20a3bbdb83967b5d2adab98463a434654b79.txt)
This loop in first case adds ~10 % overhead in most cases but save over ~100/200 % in case of arguments order changed.

Encouraged i added this tweak to `binfor` : name `binfo2` in [benchmarks](https://github.com/ashishnegi/dgraph/blob/algo-intersection-bench/algo/benchmarks/benchmarks-f151dfbf40713992bcea96914477bb8a154ecc6a.txt)

ashishnegi:
With some inputs from @manishrjain , we are able to finally optimize on all different kinds of inputs. Improvement of order of 2-3 times is common and sometimes reaching to 10 times. [benchmarks](https://github.com/ashishnegi/dgraph/blob/algo-intersection-bench/algo/benchmarks/benchmarks-final-17038bd26257ccaf636b1ada5d02f9f260fc5768.txt). Also, see difference between adding `binary forward search` to final [actual3gelbin vs actual3gel](https://github.com/ashishnegi/dgraph/blob/algo-intersection-bench/algo/benchmarks/benchmarks-8b5f4b09be9e7df0568bb2880ca3bd1bb668008d.txt)

Code for `IntersectWith3gelBin` is [search for IntersectWith3gelBin](https://github.com/ashishnegi/dgraph/blob/algo-intersection-bench/algo/uidlist.go#L299). It looks a little intimidating, so i abstracted out the next index calculation in a function `binaryForward`. I [benchmarked : actual3gelbinB](https://github.com/ashishnegi/dgraph/blob/algo-intersection-bench/algo/benchmarks/benchmarks-beautiful-vs-inoneplace.txt) to see if it has any negative effects. The negative effects were visible when i generated random data.

`Important` : Finally i also generated some random data. For size `n say 100` i choose different factors `f` as `3,10,100,1000,10000` where generated data is within range `[0 - f * n]`.

Each result file has some sample data in the starting for size `100`.
Results :  by factor [3](https://github.com/ashishnegi/dgraph/blob/algo-intersection-bench/algo/benchmarks/benchmarks-3-random-gelbinfn-gelbin-gel-actual.txt) , [10](https://github.com/ashishnegi/dgraph/blob/algo-intersection-bench/algo/benchmarks/benchmarks-10-random-gelbinfn-gelbin-gel-actual.txt), [100](https://github.com/ashishnegi/dgraph/blob/algo-intersection-bench/algo/benchmarks/benchmarks-100-random-gelbinfn-gelbin-gel-actual.txt), [1000](https://github.com/ashishnegi/dgraph/blob/algo-intersection-bench/algo/benchmarks/benchmarks-1000-random-gelbinfn-gelbin-gel-actual.txt), [10000](https://github.com/ashishnegi/dgraph/blob/algo-intersection-bench/algo/benchmarks/benchmarks-10000-random-gelbinfn-gelbin-gel-actual.txt)

The final decision is to choose whether to add binary search or not in final optimized intersection fn. If we have data where one list overlaps the other, adding binary would be better. If we have data where lists are intermingled, then we should not add binary search.

manishrjain:
:lgtm: Some comments. Congrats on your first PR!

---

Reviewed 2 of 2 files at r3.

---

*[algo/uidlist.go, line 89 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/370#-KYqKM2o2aB2tf-hn30M:-KYqKM2o2aB2tf-hn30N:bkgo50r) ([raw file](https://github.com/dgraph-io/dgraph/blob/d49678c21ee0ef543b2c8538fb08026fea98cbb8/algo/uidlist.go#L89)):*
> ```Go
> 
> 		var skip bool         // Should we skip val in output?
> 		var noElemsLeft bool  // If some list has no elems left, we can't intersect more.
> ```

elemsLeft = true

---

*[algo/uidlist.go, line 91 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/370#-KYqKJEGY236OgUDqJMm:-KYqKJEHRT8sDPRxu1SS:bdbkbol) ([raw file](https://github.com/dgraph-io/dgraph/blob/d49678c21ee0ef543b2c8538fb08026fea98cbb8/algo/uidlist.go#L91)):*
> ```Go
> 		var noElemsLeft bool  // If some list has no elems left, we can't intersect more.
> 
> 		for j := 0; j < len(lists); j++ { // For each other list in lists.
> ```

j < len(lists) && elemsLeft

---

*[algo/uidlist.go, line 106 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/370#-KYqKNlIK0i8LFez4Qaz:-KYqKNlIK0i8LFez4Qb-:b-hchgak) ([raw file](https://github.com/dgraph-io/dgraph/blob/d49678c21ee0ef543b2c8538fb08026fea98cbb8/algo/uidlist.go#L106)):*
> ```Go
> 
> 			if ljp >= lsz || lj.Uids[ljp] > val {
> 				noElemsLeft = ljp >= lsz
> ```

elemsLeft = ljp < lsz

---

*[algo/uidlist.go, line 115 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/370#-KYqKQoCDgYPiuqiXdtf:-KYqKQoCDgYPiuqiXdtg:b-6ccf72) ([raw file](https://github.com/dgraph-io/dgraph/blob/d49678c21ee0ef543b2c8538fb08026fea98cbb8/algo/uidlist.go#L115)):*
> ```Go
> 			output = append(output, val)
> 		}
> 		if noElemsLeft {
> ```

No need for this special line. Make this part of the for cond.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/370#-:-KYqKVYSHjiiDKAAPnm3:b-mrpi7b)*

ashishnegi:

---

*[algo/uidlist.go, line 91 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/370#-KYqKJEGY236OgUDqJMm:-KYqVYs5zzBVSvBJXAFN:bz3jg5h) ([raw file](https://github.com/dgraph-io/dgraph/blob/d49678c21ee0ef543b2c8538fb08026fea98cbb8/algo/uidlist.go#L91)):*

j < len(lists) && elemsLeft
</blockquote></details>

shouldn't `elemsLeft` check be on  outer loop

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/370)*

manishrjain:
:lgtm:

---

Reviewed 2 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/370#-:-KYq_IRcKobeRih_lP56:bnfp4nl)*

PR #438, opened 2017-01-02
fix/icu

jchiu0:
For embedded mode, -icu datafile doesn't seem to work.

We're switching to exporting the var ICU_DATA.

The init in tok package no longer tries to load the data file. Instead it will try to tokenize something simple. If it fails, it will set disableICU=true. This will check fail only when we try to use the tokenizer, say for indexing.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/438)

### DIFF
--- a/tok/embedflag.go
+++ b/tok/embedflag.go
@@ -30,6 +30,8 @@ func init() {
 		if disableICU {
 			x.Printf("Disabling ICU because tokenizer fails simple test case: %v",
 				tokens)
+			return
 		}
+		x.Printf("ICU is working fine!")
 	})
 }
--- a/tok/embedflag.go
+++ b/tok/embedflag.go
@@ -2,40 +2,34 @@
 
 package tok
 
-// govendor get  github.com/dgraph-io/goicu/icuembed
-// govendor get  github.com/dgraph-io/goicu/icuembed/unicode
-
 // #cgo CPPFLAGS: -I../vendor/github.com/dgraph-io/goicu/icuembed
 // #cgo darwin LDFLAGS: -Wl,-undefined -Wl,dynamic_lookup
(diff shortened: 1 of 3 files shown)


manishrjain:
Reviewed 1 of 1 files at r1.

---

*[tok/embedflag.go, line 20 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/438#-K_TL6DYCRgk8IFPCIsg:-K_TL6DZsL7bYwKs4woW:b-dd35uz) ([raw file](https://github.com/dgraph-io/dgraph/blob/395825375b0d6f155426352c29ce0c94ffc6840a/tok/embedflag.go#L20)):*
> ```Go
> 		if err != nil {
> 			disableICU = true
> 			x.Printf("Disabling ICU because we fail to create tokenizer: %v", err)
> ```

Think you're missing newlines in all of these printfs.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/438)*

jchiu0:
Thanks @pawan for verifying. Submitting.

---

Review status: 1 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[tok/embedflag.go, line 20 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/438#-K_TL6DYCRgk8IFPCIsg:-K_VJYjrvpsxdXwM8GXz:bvlghcw) ([raw file](https://github.com/dgraph-io/dgraph/blob/395825375b0d6f155426352c29ce0c94ffc6840a/tok/embedflag.go#L20)):*

Think you're missing newlines in all of these printfs.
</blockquote></details>

x.Printf does a log.Printf, it doesn't require a \n.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/438#-:-K_VJfCdAsCyhcRF0xhP:bb7i7j3)*

PR #460, opened 2017-01-09
Improve/tokenstable

jchiu0:
Removing TokensTable. Using a simple iterator on RocksDB.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/460)


manishrjain:
Reviewed 6 of 6 files at r1.

---

*[posting/index.go, line 182 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/460#-KaArmxaCBstfDBpAcgf:-KaArmxbj_P_LZ5hXOuv:bao3y83) ([raw file](https://github.com/dgraph-io/dgraph/blob/6a89c483a15fa00e8f04a5c7862f65716f437055/posting/index.go#L182)):*
> ```Go
> }
> 
> // TokensForTest returns keys for a table. This is just for testing / debugging.
> ```

Shouldn't this belong to test file then?

---

*[posting/lists.go, line 389 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/460#-KaAtplK34znvGWmr830:-KaAtplK34znvGWmr831:b-wajptr) ([raw file](https://github.com/dgraph-io/dgraph/blob/6a89c483a15fa00e8f04a5c7862f65716f437055/posting/lists.go#L389)):*
> ```Go
> 	lp.water = marks.Get(group)
> 
> 	// If it is index, commit to RocksDB.
> ```

Couple of things:

- Do this in a goroutine. If the key is already present in lhmap; then we don't need to do it at all.
- Do this via a transaction. Check if the key is already present in rocksdb, and only if it's not, should we set it to empty value. Otherwise, we might end up overwriting a valid value, due to a race condition.

---

*[x/keys.go, line 65 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/460#-KaAug0cwnD966E5jEYa:-KaAug0cwnD966E5jEYb:b-cjjs0n) ([raw file](https://github.com/dgraph-io/dgraph/blob/6a89c483a15fa00e8f04a5c7862f65716f437055/x/keys.go#L65)):*
> ```Go
> 
> // IndexParsedKey is like IndexedKey but returns a ParsedKey object.
> func IndexParsedKey(attr, term string) ParsedKey {
> ```

This is a creator function. As per our discussions in the past, we shouldn't be building these; unless there's no other way.

I think you're using this, and then calling ParsedKey.IndexPrefix or something. In that case, you can just create a IndexPrefix(attr) function, which would internally created ParsedKey and call IndexPrefix.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/460)*

jchiu0:
Review status: 1 of 7 files reviewed at latest revision, 3 unresolved discussions.

---

*[posting/index.go, line 182 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/460#-KaArmxaCBstfDBpAcgf:-KaBH9xpKca5_qwvrwhR:b-umawbe) ([raw file](https://github.com/dgraph-io/dgraph/blob/6a89c483a15fa00e8f04a5c7862f65716f437055/posting/index.go#L182)):*

Shouldn't this belong to test file then?
</blockquote></details>

Yes. (Was using it for some debugging in the past)

---

*[posting/lists.go, line 389 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/460#-KaAtplK34znvGWmr830:-KaBlgGbfVTpVCLf8eZ8:b-83330m) ([raw file](https://github.com/dgraph-io/dgraph/blob/6a89c483a15fa00e8f04a5c7862f65716f437055/posting/lists.go#L389)):*

Couple of things:

- Do this in a goroutine. If the key is already present in lhmap; then we don't need to do it at all.
- Do this via a transaction. Check if the key is already present in rocksdb, and only if it's not, should we set it to empty value. Otherwise, we might end up overwriting a valid value, due to a race condition.
</blockquote></details>

PTAL.

---

*[x/keys.go, line 65 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/460#-KaAug0cwnD966E5jEYa:-KaBKs76t5taWmwCxJka:b5jg00o) ([raw file](https://github.com/dgraph-io/dgraph/blob/6a89c483a15fa00e8f04a5c7862f65716f437055/x/keys.go#L65)):*

This is a creator function. As per our discussions in the past, we shouldn't be building these; unless there's no other way.

I think you're using this, and then calling ParsedKey.IndexPrefix or something. In that case, you can just create a IndexPrefix(attr) function, which would internally created ParsedKey and call IndexPrefix.
</blockquote></details>

Done. (There was a place in worker/sort.go where I needed to get both the parsedKey and the indexPrefix, not just IndexPrefix, and I was trying to avoid some unnecessary parsing. But this is really not important, so it's fine.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/460)*

manishrjain:
:lgtm: Just a couple of comments. Address before submitting.

---

Reviewed 6 of 6 files at r2.

---

*[x/keys.go, line 65 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/460#-KaAug0cwnD966E5jEYa:-KaK5cNsl8nHyTNDswcc:b-lyr5f8) ([raw file](https://github.com/dgraph-io/dgraph/blob/6a89c483a15fa00e8f04a5c7862f65716f437055/x/keys.go#L65)):*
<details><summary><i>Previously, jchiu0 (Jay Chiu) wrote…</i></summary><blockquote>

Done. (There was a place in worker/sort.go where I needed to get both the parsedKey and the indexPrefix, not just IndexPrefix, and I was trying to avoid some unnecessary parsing. But this is really not important, so it's fine.)
</blockquote></details>

Is this removed? Still shows here.

---

*[x/keys.go, line 98 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/460#-KaK5yQWDg__HbaRCEro:-KaK5yQXWQiicdAjLBms:bau3xlr) ([raw file](https://github.com/dgraph-io/dgraph/blob/bf42c76edb08ceddd95ae670f43d14478c1156b9/x/keys.go#L98)):*
> ```Go
> 	AssertTrue(len(k) == 2)
> 	k[0] = p.byteType
> 	k[1] = 0xFF // Why not k[0] = p.byteType + 1?
> ```

hmm... that should work too. Maybe change it to that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/460#-:-KaK6689_C7sV202ndzm:b-u5n8ii)*

jchiu0:
Review status: 4 of 7 files reviewed at latest revision, 2 unresolved discussions.

---

*[x/keys.go, line 65 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/460#-KaAug0cwnD966E5jEYa:-KaL0u7RXOOb8COg1nFO:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/6a89c483a15fa00e8f04a5c7862f65716f437055/x/keys.go#L65)):*

Is this removed? Still shows here.
</blockquote></details>

Done.

---

*[x/keys.go, line 98 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/460#-KaK5yQWDg__HbaRCEro:-KaL0vCx6tpXkGzHzFRu:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/bf42c76edb08ceddd95ae670f43d14478c1156b9/x/keys.go#L98)):*

hmm... that should work too. Maybe change it to that.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/460)*

PR #462, opened 2017-01-09
Bugfix: Timeout on invalid query

manishrjain:
We were not using the context timeout when parsing the query. Use a strict deadline so we can return back instead of hanging to the client forever. Note that this still leaves the a parser goroutine in an expensive infinite loop; so this is not a full solution. We'll have to make our parser robust.

Also refactor the gql.Parse() function to return back a result, instead of 2 results: query and mutation. Then painfully propagate that refactoring throughout the system.

[maybe] This should pave the way for a query caching system, which can cache such hanging queries and propagate them through the system.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/462)

### DIFF
--- a/cmd/dgraph/main.go
+++ b/cmd/dgraph/main.go
@@ -390,7 +390,6 @@ func parseQueryAndMutation(ctx context.Context, query string) (res gql.Result, e
 }
 
 func queryHandler(w http.ResponseWriter, r *http.Request) {
-	fmt.Println("Query handler called")
 	// Add a limit on how many pending queries can be run in the system.
 	pendingQueries <- struct{}{}
 	defer func() { <-pendingQueries }()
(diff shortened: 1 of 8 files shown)


ashwin95r:
:lgtm:

---

Reviewed 7 of 8 files at r1.
Review status: 7 of 8 files reviewed at latest revision, 1 unresolved discussion.

---

*[cmd/dgraph/main.go, line 393 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/462#-Ka55avLeP1e27zY3wcH:-Ka55avMU2VGNZ07P3jx:b-v0wiqc) ([raw file](https://github.com/dgraph-io/dgraph/blob/7b2f2a77556ebd9bce0099b9a42c83a68e5c530c/cmd/dgraph/main.go#L393)):*
> ```Go
> 
> func queryHandler(w http.ResponseWriter, r *http.Request) {
> 	fmt.Println("Query handler called")
> ```

x.Print()?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/462#-:-Ka59GknyeTVK8-YlGK4:bnfp4nl)*

pawanrawal:
:lgtm:

---

Review status: 7 of 8 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/462#-:-Ka5AKdgbUqnGaSaKXsj:bnfp4nl)*

PR #496, opened 2017-01-16
Change variable names. Add an assert.

manishrjain:
Some minor changes to make the code more readable.
The assert fails in tests. Why is that the case? We shouldn't be setting UID twice for the same node.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/496)


ashishnegi:
Fixes the bug of calling multiple setuid on same node.
Also one more change, see commit : https://github.com/dgraph-io/dgraph/pull/496/commits/806b510a6413409e38790581cd71f07e15b5bf1e

ashishnegi:
Merged to master.

PR #502, opened 2017-01-17
Adding sync pool for sharing the bytebuffers.

ashishnegi:
~~Using buffer pool in `toFastJson` for sharing buffers between requests..~~
1. We have to make copy of buffer bytes before returning to the pool, as buffer.Bytes() are valid till next read/write on buffer.
On thinking, this seems to be fine with me; as this does not matter when we have less queries; on load, we are returning buffer faster so now new buffers will be created. So, it seems to be the same memory footprint.

Now writing directly to HttpResponseWriter.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/502)

### DIFF
--- a/cmd/dgraph/main.go
+++ b/cmd/dgraph/main.go
@@ -524,6 +524,7 @@ func queryHandler(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	w.Header().Set("Content-Type", "application/json")
 	err = query.ToJson(&l, sgl, w)
 	if err != nil {
 		x.TraceError(ctx, x.Wrapf(err, "Error while converting to Json"))
@@ -532,8 +533,6 @@ func queryHandler(w http.ResponseWriter, r *http.Request) {
 	}
 	x.Trace(ctx, "Latencies: Total: %v Parsing: %v Process: %v Json: %v",
 		time.Since(l.Start), l.Parsing, l.Processing, l.Json)
-
-	w.Header().Set("Content-Type", "application/json")
 }
 
 // storeStatsHandler outputs some basic stats for data store.
--- a/cmd/dgraph/main.go
+++ b/cmd/dgraph/main.go
@@ -524,7 +524,7 @@ func queryHandler(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	js, err := query.ToJson(&l, sgl)
+	err = query.ToJson(&l, sgl, w)
 	if err != nil {
 		x.TraceError(ctx, x.Wrapf(err, "Error while converting to Json"))
 		x.SetStatus(w, x.Error, err.Error())
@@ -534,7 +534,6 @@ func queryHandler(w http.ResponseWriter, r *http.Request) {
 		time.Since(l.Start), l.Parsing, l.Processing, l.Json)
 
 	w.Header().Set("Content-Type", "application/json")
-	w.Write(js)
 }
 
 // storeStatsHandler outputs some basic stats for data store.
--- a/cmd/dgraph/main.go
+++ b/cmd/dgraph/main.go
@@ -527,6 +527,8 @@ func queryHandler(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
 	err = query.ToJson(&l, sgl, w)
 	if err != nil {
+		// since we performed w.Write in ToJson above,
+		// calling WriteHeader with 500 code will be ignored.
 		x.TraceError(ctx, x.Wrapf(err, "Error while converting to Json"))
 		x.SetStatus(w, x.Error, err.Error())
 		return
--- a/query/geo_test.go
+++ b/query/geo_test.go
@@ -17,6 +17,7 @@
 package query
 
 import (
+	"bytes"
 	"context"
 	"io/ioutil"
 	"os"
@@ -105,9 +106,10 @@ func runQuery(t *testing.T, gq *gql.GraphQuery) string {
 	require.NoError(t, err)
 
 	var l Latency
-	js, err := sg.ToFastJSON(&l)
+	var buf bytes.Buffer
+	err = sg.ToFastJSON(&l, &buf)
 	require.NoError(t, err)
-	return string(js)
+	return string(buf.Bytes())
 }
 
 func TestWithinPoint(t *testing.T) {
--- a/query/outputnode.go
+++ b/query/outputnode.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"sort"
 	"strconv"
+	"sync"
 	"time"
 
 	geom "github.com/twpayne/go-geom"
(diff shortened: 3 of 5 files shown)


manishrjain:
Reviewed 1 of 1 files at r1.

---

*[query/outputnode.go, line 375 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/502#-KagGwrVJRx6yPZTkKzT:-KagGwrWvcxaQvKhkoBi:b-fr786u) ([raw file](https://github.com/dgraph-io/dgraph/blob/58185432dfe40467f5363794441d25bd057eb421/query/outputnode.go#L375)):*
> ```Go
> 	jsBuf.Reset()
> 	n.(*fastJsonNode).encode(&jsBuf)
> 	jsBs := make([]byte, jsBuf.Len())
> ```

Memory allocation is an expensive thing -- never take it lightly. This would take away any potential benefits you might get from putting the buffer back into pool.

Now think about how you could make a change so that you won't have to do this copy. A quick look at the callers would show you that there's only one place it's called, and that's right before we write this out to http.ResponseWriter. So, maybe, we're just better off passing the responsewriter here.

Now that you're passing responsewriter to this function, consider this: Would it be possible to use the responsewriter directly instead of writing to buffer first? If we can, it would surely help. We won't need to create and grow bytes.Buffers at all. We remove the intermediary entirely.

https://golang.org/pkg/bufio/#Writer would help provide all the same functions as what you use, but flush directly to response writer. Now, that'd be wonderful, isn't it.

See, there's so much you could derive and improve by refusing to put a hack in place -- and challenging the existing interfaces.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/502)*

ashishnegi:

---

*[query/outputnode.go, line 375 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/502#-KagGwrVJRx6yPZTkKzT:-KajKy-UuyX1LVukqpKx:bnnkts1) ([raw file](https://github.com/dgraph-io/dgraph/blob/58185432dfe40467f5363794441d25bd057eb421/query/outputnode.go#L375)):*

Memory allocation is an expensive thing -- never take it lightly. This would take away any potential benefits you might get from putting the buffer back into pool.

Now think about how you could make a change so that you won't have to do this copy. A quick look at the callers would show you that there's only one place it's called, and that's right before we write this out to http.ResponseWriter. So, maybe, we're just better off passing the responsewriter here.

Now that you're passing responsewriter to this function, consider this: Would it be possible to use the responsewriter directly instead of writing to buffer first? If we can, it would surely help. We won't need to create and grow bytes.Buffers at all. We remove the intermediary entirely.

https://golang.org/pkg/bufio/#Writer would help provide all the same functions as what you use, but flush directly to response writer. Now, that'd be wonderful, isn't it.

See, there's so much you could derive and improve by refusing to put a hack in place -- and challenging the existing interfaces.
</blockquote></details>

Definitely an important lesson, i learned.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/502)*

ashishnegi:
The only problem with writing directly to the final resource is :
1. if after we write some lines, we find some err. Now we can not undo what we wrote.

Luckily, this is not the problem for us, as we check for all errors beforehand.

manishrjain:
Reviewed 3 of 5 files at r2, 2 of 2 files at r3.

---

*[cmd/dgraph/main.go, line 530 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/502#-KajotOdymmrN1SyEbJB:-KajotOdymmrN1SyEbJC:ba59ucs) ([raw file](https://github.com/dgraph-io/dgraph/blob/da7e1ab3beaa9561d287162c75c50160d2fef0e0/cmd/dgraph/main.go#L530)):*
> ```Go
> 	if err != nil {
> 		x.TraceError(ctx, x.Wrapf(err, "Error while converting to Json"))
> 		x.SetStatus(w, x.Error, err.Error())
> ```

Set http status to internal server error.

---

*[query/outputnode.go, line 278 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/502#-KajlK77NcZ8XeAW-oGG:-KajlK78_rmv5H1C_lwE:b-axu1vm) ([raw file](https://github.com/dgraph-io/dgraph/blob/da7e1ab3beaa9561d287162c75c50160d2fef0e0/query/outputnode.go#L278)):*
> ```Go
> }
> 
> func (fj *fastJsonNode) encode(bufW *bufio.Writer) {
> ```

Never end with a capital case. Say bufw instead.

---

*[query/outputnode.go, line 288 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/502#-KajoSdWmeoVDSCjtOdy:-KajoSdWmeoVDSCjtOdz:b-kwq61q) ([raw file](https://github.com/dgraph-io/dgraph/blob/da7e1ab3beaa9561d287162c75c50160d2fef0e0/query/outputnode.go#L288)):*
> ```Go
> 	sort.Strings(allKeys)
> 
> 	bufW.WriteRune('{')
> ```

Handle all the errors.

---

*[query/outputnode.go, line 372 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/502#-KajoaEj9KXV-VQR6ZGb:-KajoaEj9KXV-VQR6ZGc:b-80xfyz) ([raw file](https://github.com/dgraph-io/dgraph/blob/da7e1ab3beaa9561d287162c75c50160d2fef0e0/query/outputnode.go#L372)):*
> ```Go
> 
> 	bufW := bufio.NewWriter(w)
> 	n.(*fastJsonNode).encode(bufW)
> ```

Return error from encode.

---

*[query/outputnode.go, line 373 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/502#-KajoW-OrfbYdZP2L2i2:-KajoW-OrfbYdZP2L2i3:b-td8922) ([raw file](https://github.com/dgraph-io/dgraph/blob/da7e1ab3beaa9561d287162c75c50160d2fef0e0/query/outputnode.go#L373)):*
> ```Go
> 	bufW := bufio.NewWriter(w)
> 	n.(*fastJsonNode).encode(bufW)
> 	bufW.Flush()
> ```

return bufw.Flush()

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/502)*

manishrjain:
:lgtm: Small comments. Address then submit.

---

Reviewed 2 of 2 files at r4.

---

*[query/outputnode.go, line 278 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/502#-KajlK77NcZ8XeAW-oGG:-Kak5gefbZ38Jv74WsYN:b-dvbqbn) ([raw file](https://github.com/dgraph-io/dgraph/blob/da7e1ab3beaa9561d287162c75c50160d2fef0e0/query/outputnode.go#L278)):*

Never end with a capital case. Say bufw instead.
</blockquote></details>

Handle this.

---

*[query/outputnode.go, line 288 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/502#-KajoSdWmeoVDSCjtOdy:-Kak5fTASzcqZTDFO_a1:b32w0x) ([raw file](https://github.com/dgraph-io/dgraph/blob/da7e1ab3beaa9561d287162c75c50160d2fef0e0/query/outputnode.go#L288)):*

Handle all the errors.
</blockquote></details>

OK

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/502#-:-Kak5oDp7JASQ3C4dKON:bfgw98t)*

ashishnegi:
Review status: 3 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

*[cmd/dgraph/main.go, line 530 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/502#-KajotOdymmrN1SyEbJB:-Kak7NHW9sDv3iW8svYb:b8zf1jp) ([raw file](https://github.com/dgraph-io/dgraph/blob/da7e1ab3beaa9561d287162c75c50160d2fef0e0/cmd/dgraph/main.go#L530)):*

Set http status to internal server error.
</blockquote></details>

wrote comment about why it can not be done.

---

*[query/outputnode.go, line 278 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/502#-KajlK77NcZ8XeAW-oGG:-Kak7RaNz0B9uznt9hwa:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/da7e1ab3beaa9561d287162c75c50160d2fef0e0/query/outputnode.go#L278)):*

Handle this.
</blockquote></details>

Done.

---

*[query/outputnode.go, line 372 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/502#-KajoaEj9KXV-VQR6ZGb:-Kak7Vdj6kAt73OUbcLZ:bzhm7a0) ([raw file](https://github.com/dgraph-io/dgraph/blob/da7e1ab3beaa9561d287162c75c50160d2fef0e0/query/outputnode.go#L372)):*

Return error from encode.
</blockquote></details>

bufw contains err in itself. caller can call buf.Flush() to check it.

---

*[query/outputnode.go, line 373 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/502#-KajoW-OrfbYdZP2L2i2:-Kak7VCmlYVkUFI2geLo:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/da7e1ab3beaa9561d287162c75c50160d2fef0e0/query/outputnode.go#L373)):*

return bufw.Flush()
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/502)*

PR #512, opened 2017-01-19
Feature/tok

jchiu0:
Allow users to register tokenizers.

Allow users to specify what tokenizer they want in the schema.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/512)


manishrjain:
Looks good mostly. Just simplify the lexer. Lexer shouldn't be doing parser's job.

---

Reviewed 24 of 24 files at r1.

---

*[schema/state.go, line 48 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/512#-KauleO8DXYXDAi14gIl:-KauleO8DXYXDAi14gIm:bqvsv2e) ([raw file](https://github.com/dgraph-io/dgraph/blob/fa1d0f086195730ebb7729adc077f10e9ef92ed2/schema/state.go#L48)):*
> ```Go
> 	itemIndex
> 	itemReverse
> 	itemDirectiveArg
> ```

I think the lexer should only return left bracket, right bracket and a var. The parser should be the one which determines that it is a directive argument.

@ashwin95r did a big refactoring in gql/lexer to that effect. We should do the same for schema, though that's out of scope for this change.

For this change, remove this itemDirectiveArg, and do the deduction in parser.

---

*[tok/tok.go, line 180 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/512#-KauXlpwcRgeegMEUuoJ:-KauXlpwcRgeegMEUuoK:b-bgz9t2) ([raw file](https://github.com/dgraph-io/dgraph/blob/fa1d0f086195730ebb7729adc077f10e9ef92ed2/tok/tok.go#L180)):*
> ```Go
> func encodeInt(val int32) ([]string, error) {
> 	buf := make([]byte, 5)
> 	binary.BigEndian.PutUint32(buf[1:], uint32(val))
> ```

A negative value would overflow `uint32(val)`. You need to convert the value to positive first before doing the conversion. Also, can you please add a test for this?

---

*[tok/tok_test.go, line 35 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/512#-KauY5e_eGGjaeru47b6:-KauY5e_eGGjaeru47b7:b-usyxi0) ([raw file](https://github.com/dgraph-io/dgraph/blob/fa1d0f086195730ebb7729adc077f10e9ef92ed2/tok/tok_test.go#L35)):*
> ```Go
> 
> func (o byEnc) Less(i, j int) bool {
> 	return o.ints[i] < o.ints[j]
> ```

Can go in one line.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/512#-:-KauoCIjuvf75mnm2IM2:blzvttw)*

ashwin95r:

---

*[schema/testfiles/test_schema_index4, line 1 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/512#-KbCuj38BOy5fckm_fz0:-KbCuj38BOy5fckm_fz1:bc7y67a) ([raw file](https://github.com/dgraph-io/dgraph/blob/fa1d0f086195730ebb7729adc077f10e9ef92ed2/schema/testfiles/test_schema_index4#L1)):*
> ```
> scalar age:int @index(int)
> ```

Is it compulsory to specify the type of tokenizer? Or should we have a default for each type?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/512)*

jchiu0:

---

*[schema/state.go, line 48 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/512#-KauleO8DXYXDAi14gIl:-KbN_tIe8F--YrLwfA0s:bbfmpas) ([raw file](https://github.com/dgraph-io/dgraph/blob/fa1d0f086195730ebb7729adc077f10e9ef92ed2/schema/state.go#L48)):*

I think the lexer should only return left bracket, right bracket and a var. The parser should be the one which determines that it is a directive argument.

@ashwin95r did a big refactoring in gql/lexer to that effect. We should do the same for schema, though that's out of scope for this change.

For this change, remove this itemDirectiveArg, and do the deduction in parser.
</blockquote></details>

Totally agree. Refactoring can be left to another PR.

Removed itemDirectiveArg.

---

*[schema/testfiles/test_schema_index4, line 1 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/512#-KbCuj38BOy5fckm_fz0:-KbNaBWjW659EqI20sYt:bv3sxo4) ([raw file](https://github.com/dgraph-io/dgraph/blob/fa1d0f086195730ebb7729adc077f10e9ef92ed2/schema/testfiles/test_schema_index4#L1)):*

Is it compulsory to specify the type of tokenizer? Or should we have a default for each type?
</blockquote></details>

No it is not compulsory. The defaults are set in tok/tok.go.

---

*[tok/tok.go, line 180 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/512#-KauXlpwcRgeegMEUuoJ:-KbNcOLKuKbzO2A-YIeW:b-3n2b14) ([raw file](https://github.com/dgraph-io/dgraph/blob/fa1d0f086195730ebb7729adc077f10e9ef92ed2/tok/tok.go#L180)):*

A negative value would overflow `uint32(val)`. You need to convert the value to positive first before doing the conversion. Also, can you please add a test for this?
</blockquote></details>

-128 < -1. If you just encode "-val", then -128 is going to be bigger than -1 when encoded. There is already a test (tok_test.go) to check this.

What you want to do is to keep the original bits, i.e., ignore the overflow. In fact, the proper way to do this I think is to just flip the most significant bit (and not waste one byte for storing the sign).

Example: say we're dealing with 8 bit instead of 32 bits, for simplicity.

-128 or 0x80 would be encoded as 0x00
-1 or 0xFF would become 0x7F
1 or 0x01 would become 0x81
127 or 0x7F would become 0xFF.

The advantage of using one additional byte for the sign I think is convenience. You can use all the big endian encodes, decodes directly without having to flip bits.

For this PR, no change is required. Trying to encode -val will fail the test. As for the change I propose, for saving one byte (which might be more significant due to memory alignment), let's do that in a separate PR.

---

*[tok/tok_test.go, line 35 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/512#-KauY5e_eGGjaeru47b6:-KbNfcQagMeb_qRS0Ca2:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/fa1d0f086195730ebb7729adc077f10e9ef92ed2/tok/tok_test.go#L35)):*

Can go in one line.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/512)*

manishrjain:
:lgtm: The code in lexer needs to be moved to parser.

---

Reviewed 1 of 24 files at r1, 6 of 6 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/512#-:-KbhSQnUltBkE2u3v65A:b55mx7w)*

PR #522, opened 2017-01-23
Seektolast if seek to reverse key fails

### LINKED ISSUE #499: Orderdesc doesn't result returns but order does.
To reproduce try loading this https://github.com/dgraph-io/benchmarks/blob/master/data/neo4j/neo.rdf.gz with the goldendata.schema using the loader and then perform this query.
```
curl localhost:8080/query -XPOST -d $'{
                  director(allof("type.object.name.en", "steven spielberg")) {
                      type.object.name.en
                      film.director.film (orderdesc: film.film.initial_release_date) {
                          type.object.name.en
                          film.film.initial_release_date
                      }
                  }
              }'
```
This returns results for order but not orderdesc.

pawanrawal:
If the database didn't have any other keys after the predicate index keys, it didn't send back any results. 

We now check if the seek is valid, if not then we seek to last.

This fixes #499

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/522)

### DIFF
--- a/query/query_test.go
+++ b/query/query_test.go
@@ -241,6 +241,11 @@ func populateGraph(t *testing.T) {
 	})
 	addGeoData(t, ps, 5106, poly, "San Carlos")
 
+	addEdgeToValue(t, "film.film.initial_release_date", 23, "1900-01-02")
+	addEdgeToValue(t, "film.film.initial_release_date", 24, "1909-05-05")
+	addEdgeToValue(t, "film.film.initial_release_date", 25, "1929-01-10")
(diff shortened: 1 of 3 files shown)


manishrjain:
:lgtm:

---

Reviewed 2 of 2 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/522#-:-Kb9cocbmHBeHUWN4x0U:bnfp4nl)*

PR #559, opened 2017-02-01
Switch task.List to a block list

ashwin95r:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/559)


manishrjain:
Review status: 0 of 14 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[algo/uidlist.go, line 25 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-Kbs9BKY6riGen1e8Fjh:-Kbs9BKY6riGen1e8Fji:bb62055) ([raw file](https://github.com/dgraph-io/dgraph/blob/c17b2f2c18038db26de1967cc59b84a5a3567c8b/algo/uidlist.go#L25)):*
> ```Go
> 
> func (l *ListIterator) Valid() bool {
> 	if l == nil || l.list.Blocks == nil || len(l.list.Blocks) == 0 {
> ```

return l.err == nil

---

*[algo/uidlist.go, line 43 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-Kbs9rKB8WQIaRMtbY5E:-Kbs9rKB8WQIaRMtbY5F:baf00zv) ([raw file](https://github.com/dgraph-io/dgraph/blob/c17b2f2c18038db26de1967cc59b84a5a3567c8b/algo/uidlist.go#L43)):*
> ```Go
> 	return l.list.Blocks[l.bIdx].List[l.idx]
> }
> ```

Also add a Seek()

---

*[algo/uidlist.go, line 44 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-Kbs9NXVNxYPVLDyeY1i:-Kbs9NXVNxYPVLDyeY1j:b-uf8c7b) ([raw file](https://github.com/dgraph-io/dgraph/blob/c17b2f2c18038db26de1967cc59b84a5a3567c8b/algo/uidlist.go#L44)):*
> ```Go
> }
> 
> func (l *ListIterator) Next() bool {
> ```

No need to return bool. The api:

for l.Init(); l.Valid(); l.Next() { l.Val() }

---

*[algo/uidlist.go, line 52 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-Kbs9_W_jTU8fxbxvgy1:-Kbs9_W_jTU8fxbxvgy2:b-b01t3x) ([raw file](https://github.com/dgraph-io/dgraph/blob/c17b2f2c18038db26de1967cc59b84a5a3567c8b/algo/uidlist.go#L52)):*
> ```Go
> 		l.idx = 0
> 		l.bIdx++
> 		if l.bIdx >= len(l.list.Blocks) {
> ```

Move this out of the if.

---

*[algo/uidlist.go, line 53 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-Kbs9HGYse2zKo4pcoZ0:-Kbs9HGYse2zKo4pcoZ1:b-22k17t) ([raw file](https://github.com/dgraph-io/dgraph/blob/c17b2f2c18038db26de1967cc59b84a5a3567c8b/algo/uidlist.go#L53)):*
> ```Go
> 		l.bIdx++
> 		if l.bIdx >= len(l.list.Blocks) {
> 			return false
> ```

l.err = EOF

---

*[algo/uidlist.go, line 83 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbsBJ-Y1jLCpFGhzZ-F:-KbsBJ-Y1jLCpFGhzZ-G:bpmo13) ([raw file](https://github.com/dgraph-io/dgraph/blob/c17b2f2c18038db26de1967cc59b84a5a3567c8b/algo/uidlist.go#L83)):*
> ```Go
> }
> 
> func BlockToList(b *task.List) []uint64 {
> ```

Move it to tests.

---

*[algo/uidlist.go, line 98 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbsBP0-RSIw3IL__onC:-KbsBP0-RSIw3IL__onD:buysiic) ([raw file](https://github.com/dgraph-io/dgraph/blob/c17b2f2c18038db26de1967cc59b84a5a3567c8b/algo/uidlist.go#L98)):*
> ```Go
> 		return 0
> 	}
> ```

if blocks are always packed, then this should be (len(l.Blocks) - 1)*blockSize + len(l.Blocks[last].List)

---

*[algo/uidlist.go, line 105 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbsAKLJaSWuj8saRaC-:-KbsAKLJaSWuj8saRaC0:b9dgbl3) ([raw file](https://github.com/dgraph-io/dgraph/blob/c17b2f2c18038db26de1967cc59b84a5a3567c8b/algo/uidlist.go#L105)):*
> ```Go
> }
> 
> func IntersectWith(u, v *task.List) {
> ```

Use iterators.

---

*[algo/uidlist.go, line 432 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbsC0lUbXWK764Xplx7:-KbsC0lUbXWK764Xplx8:b-ridan7) ([raw file](https://github.com/dgraph-io/dgraph/blob/c17b2f2c18038db26de1967cc59b84a5a3567c8b/algo/uidlist.go#L432)):*
> ```Go
> }
> 
> func Ridx(ul *task.List, i int) (int, int) {
> ```

Move these to Swap.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/559)*

ashwin95r:
Review status: 0 of 14 files reviewed at latest revision, 9 unresolved discussions.

---

*[algo/uidlist.go, line 25 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-Kbs9BKY6riGen1e8Fjh:-KbtiuFovhz4cIgTUuPO:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/c17b2f2c18038db26de1967cc59b84a5a3567c8b/algo/uidlist.go#L25)):*

return l.err == nil
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 43 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-Kbs9rKB8WQIaRMtbY5E:-Kbtiwgd88XrIIfwqPBy:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/c17b2f2c18038db26de1967cc59b84a5a3567c8b/algo/uidlist.go#L43)):*

Also add a Seek()
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 44 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-Kbs9NXVNxYPVLDyeY1i:-KbtixTinDJmkTJAAgLa:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/c17b2f2c18038db26de1967cc59b84a5a3567c8b/algo/uidlist.go#L44)):*

No need to return bool. The api:

for l.Init(); l.Valid(); l.Next() { l.Val() }
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 52 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-Kbs9_W_jTU8fxbxvgy1:-Kbtiy_ISFc85BlHSGzj:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/c17b2f2c18038db26de1967cc59b84a5a3567c8b/algo/uidlist.go#L52)):*

Move this out of the if.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 53 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-Kbs9HGYse2zKo4pcoZ0:-Kbtiyz7Kn8NLoMA_s8U:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/c17b2f2c18038db26de1967cc59b84a5a3567c8b/algo/uidlist.go#L53)):*

l.err = EOF
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 105 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbsAKLJaSWuj8saRaC-:-Kbtj2jVRDM7qWZOVUyy:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/c17b2f2c18038db26de1967cc59b84a5a3567c8b/algo/uidlist.go#L105)):*

Use iterators.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 432 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbsC0lUbXWK764Xplx7:-Kbtj3qpMHlSJhXRd8Mg:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/c17b2f2c18038db26de1967cc59b84a5a3567c8b/algo/uidlist.go#L432)):*

Move these to Swap.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/559)*

ashwin95r:
Review status: 0 of 14 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[algo/uidlist.go, line 98 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbsBP0-RSIw3IL__onC:-KbwPFtQ1JanViA6VDLQ:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/c17b2f2c18038db26de1967cc59b84a5a3567c8b/algo/uidlist.go#L98)):*

if blocks are always packed, then this should be (len(l.Blocks) - 1)*blockSize + len(l.Blocks[last].List)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/559)*

manishrjain:
Reviewed 5 of 14 files at r1, 9 of 9 files at r2.
Review status: all files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[algo/uidlist.go, line 83 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbsBJ-Y1jLCpFGhzZ-F:-KbxG2wdZLx49Q30xyfx:blg466l) ([raw file](https://github.com/dgraph-io/dgraph/blob/c17b2f2c18038db26de1967cc59b84a5a3567c8b/algo/uidlist.go#L83)):*

Move it to tests.
</blockquote></details>

This is still here.

---

*[algo/uidlist.go, line 16 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxCXKNyx9zYdGMhzbv:-KbxCXKO47KtQLn4Ilfk:bb7k1ue) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L16)):*
> ```Go
> 	list  *task.List
> 	bIdx  int  // Block index
> 	idx   int  // List index
> ```

lidx

---

*[algo/uidlist.go, line 41 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxCoz09qjvlzaEFYxz:-KbxCoz09qjvlzaEFYy-:byafn4e) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L41)):*
> ```Go
> 	}
> 
> 	i := len(l.Blocks) - 1
> ```

Shouldn't this be zero and zero?

---

*[algo/uidlist.go, line 53 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxD6EmW4RpwF0do0nc:-KbxD6EmW4RpwF0do0nd:bs5i07e) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L53)):*
> ```Go
> // Append appends an UID to the end of task.List following the blockSize specified.
> func (l *WriteIterator) Append(uid uint64) {
> 	l.list.Blocks[l.bIdx].List = append(l.list.Blocks[l.bIdx].List, uid)
> ```

Really? What if it's overwriting the list? You're not using l.idx.

---

*[algo/uidlist.go, line 58 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxDL1wTXnReCVdGmwd:-KbxDL1wTXnReCVdGmwe:boouf1m) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L58)):*
> ```Go
> 		l.list.Blocks[l.bIdx].MaxInt = l.list.Blocks[l.bIdx].List[blockSize-1]
> 		l.idx = 0
> 		l.list.Blocks = append(l.list.Blocks, &task.Block{List: make([]uint64, 0, blockSize)})
> ```

What is the block is already there? This code needs to be rewritten to actually use it like an iterator.

---

*[algo/uidlist.go, line 67 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxDklwOkuJF5B3xyVp:-KbxDklxwmv05vYn-z-L:b-ff7ps1) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L67)):*
> ```Go
> 	lenLast := len(l.list.Blocks[l.bIdx].List)
> 	if lenLast == 0 {
> 		// We didn't add any element to the list
> ```

You should be removing this block from block slice then, right?

---

*[algo/uidlist.go, line 74 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxGUARqjWnni7pOpQ7:-KbxGUASJpA-b60GATIV:b66iqsj) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L74)):*
> ```Go
> 
> // NewListIterator returns a read iterator for the list passed to it.
> func NewListIterator(l *task.List) ListIterator {
> ```

This function causes ListIterator to be allocated on heap, when it could have been on stack. Convert this function to Init() or make it be called via Next(). We don't do constructor functions.

---

*[algo/uidlist.go, line 114 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxGnxpduy5JUBRu6EP:-KbxGnxq1952TogJpKbj:b2yvrjw) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L114)):*
> ```Go
> // Seek seeks to the index whose value is greater than or equal to the given UID.
> // It uses binary search to move around.
> func (l *ListIterator) Seek(uid uint64) {
> ```

This function should take a whence, where 0 means from origin, and 1 means from current position. If we're seeking from current pos, then we can optimize our seeks by iterating the existing list first, then going over the blocks.

---

*[algo/uidlist.go, line 137 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxEtHvOxZDh6UpMKCf:-KbxEtHvOxZDh6UpMKCg:br0zn4a) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L137)):*
> ```Go
> // Val returns the value pointed to by the iterator.
> func (l *ListIterator) Val() uint64 {
> 	if !l.Valid() {
> ```

You don't need to check this here. Caller should have done it.

---

*[algo/uidlist.go, line 140 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxFFQz5HCVIioWgToR:-KbxFFQz5HCVIioWgToS:b-982xvj) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L140)):*
> ```Go
> 		return 0
> 	}
> 	return l.list.Blocks[l.bIdx].List[l.idx]
> ```

Avoid doing this. Let Next() or Prev() fill this up.

---

*[algo/uidlist.go, line 146 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxF-1xSqs8YuNiDHcz:-KbxF-1xSqs8YuNiDHd-:bptf4te) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L146)):*
> ```Go
> // is consumed already.
> func (l *ListIterator) Next() {
> 	if !l.Valid() {
> ```

You can just check the var. No need to call a func. Note that you could be dealing with very long lists.

---

*[algo/uidlist.go, line 166 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxFbuWGT_GOxY0CvHh:-KbxFbuWGT_GOxY0CvHi:b-x482it) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L166)):*
> ```Go
> // the list passed to it.
> func Slice(ul *task.List, start, end int) *task.List {
> 	o := new(task.List)
> ```

Actually, we could be writing back to ul via the WriteIterator.

---

*[algo/uidlist.go, line 180 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxG-oSEg8VZfVSAqEP:-KbxG-oSEg8VZfVSAqEQ:b-w6n6f7) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L180)):*
> ```Go
> 	return o
> }
> func SortedListToBlock(l []uint64) *task.List {
> ```

Where are we using this?

---

*[algo/uidlist.go, line 217 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxGBu1HM5469eK032Q:-KbxGBu1HM5469eK032R:b-fuqfn) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L217)):*
> ```Go
> func IntersectWith(u, v *task.List) {
> 	o := new(task.List)
> 	out := NewWriteIterator(o)
> ```

This can be written back to u directly via writeiterator.

---

*[algo/uidlist.go, line 234 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxHC1tDyzA2JIybMpl:-KbxHC1ugTdjfCDSu0Ka:b-q64pkt) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L234)):*
> ```Go
> 	}
> 	out.End()
> 	u.Blocks = o.Blocks
> ```

See above.

---

*[algo/uidlist.go, line 322 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxHVgx2uFtOi2WXzHc:-KbxHVgx2uFtOi2WXzHd:b3df3h) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L322)):*
> ```Go
> func Difference(u, v *task.List) {
> 	o := new(task.List)
> 	out := NewWriteIterator(o)
> ```

Again can be written to u.

---

*[algo/uidlist.go, line 408 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxESRFKfqQdnHU8WGz:-KbxESRFKfqQdnHU8WH-:b-ijh8rq) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L408)):*
> ```Go
> }
> 
> func ridx(ul *task.List, i int) (int, int) {
> ```

Can't we ensure that all the blocks but the last one is fully filled? That would make this function fast.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/559)*

gacharya255:
Which test is failing?

---

Review status: all files reviewed at latest revision, 28 unresolved discussions, some commit checks failed.

---

*[algo/uidlist.go, line 23 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-Kbx7nY94xzo9r0BBKs1:-Kbx7nY94xzo9r0BBKs2:b-43uau0) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L23)):*
> ```Go
> type WriteIterator struct {
> 	list *task.List
> 	bIdx int
> ```

Can comment that bIdx indexes into l.Blocks and idx indexes into block.List here?

---

*[algo/uidlist.go, line 55 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-Kbx8vt9N6FEwmECnAGG:-Kbx8vt9N6FEwmECnAGH:b-6ywpx5) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L55)):*
> ```Go
> 	l.list.Blocks[l.bIdx].List = append(l.list.Blocks[l.bIdx].List, uid)
> 	l.idx++
> 	if l.idx == blockSize {
> ```

Maybe better to do allocation before writing instead of after writing, as the allocated memory might not be used, though this might make little difference for larger block size.

---

*[algo/uidlist.go, line 117 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-Kbx9h1K-iIctk5--0n5:-Kbx9h1K-iIctk5--0n6:bo8a8qo) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L117)):*
> ```Go
> 	u := l.list
> 	i := sort.Search(len(u.Blocks), func(i int) bool { return u.Blocks[i].MaxInt >= uid })
> 	if i < len(u.Blocks) {
> ```

Better to treat uncommon case first and reduce indent.

if i >= len(u.Blocks) { l.isEnd=true; return }
j := sort.Search(...)

---

*[algo/uidlist.go, line 132 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-Kbx9raOJsTkpyXZC_88:-Kbx9raOJsTkpyXZC_89:bcbl43l) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L132)):*
> ```Go
> // Valid returns true if we haven't reached the end of the list.
> func (l *ListIterator) Valid() bool {
> 	return !l.isEnd
> ```

Maybe we can do without l.isEnd. Maintaining it requires extra care and work. Valid() can just do a simple check on the indices?

---

*[algo/uidlist.go, line 163 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxAcWTx7hGRXK9c-ao:-KbxAcWTx7hGRXK9c-ap:b-976ta4) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L163)):*
> ```Go
> }
> 
> // Slice returns a new task.List with the elements between start and end of
> ```

Maybe mention that start, end are indices not UIDs.

---

*[algo/uidlist.go, line 206 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxBFBgiXMRoun1vJdD:-KbxBFBhA3se_9z7ecAb:b2lazyc) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L206)):*
> ```Go
> 
> func ListLen(l *task.List) int {
> 	if l == nil || l.Blocks == nil {
> ```

len(l.Blocks)==0 instead of l.Blocks==nil just in case there is an empty array which will mess up n below.

---

*[algo/uidlist.go, line 248 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxBZrq-93Ecr49haXa:-KbxBZrq-93Ecr49haXb:bckj2b4) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L248)):*
> ```Go
> 			}
> 		}
> 		out[i].List = out[i].List[:kk]
> ```

After filtering, inner block might not be of blockSize? Might want to be careful with assumptions elsewhere, such as total size function.

---

*[algo/uidlist.go, line 429 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxKuL3-uSU7HgFGrIJ:-KbxKuL3-uSU7HgFGrIK:b9vwyt7) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L429)):*
> ```Go
> 		res += len(ul.Blocks[k].List)
> 	}
> 	return res + j
> ```

Is it possible that j >= len(block.List)?

You might want to use a bit-indexed tree here if blocks can have different sizes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-:-KbxLTdgpS8zQbEvAppS:bz2d1m3)*

ashwin95r:
Review status: 5 of 14 files reviewed at latest revision, 28 unresolved discussions, some commit checks failed.

---

*[algo/uidlist.go, line 83 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbsBJ-Y1jLCpFGhzZ-F:-Kc1vgVUd5PRiUCXv9v2:b-7yrhbi) ([raw file](https://github.com/dgraph-io/dgraph/blob/c17b2f2c18038db26de1967cc59b84a5a3567c8b/algo/uidlist.go#L83)):*

This is still here.
</blockquote></details>

We need it in createTaskQuery which need a uint64 list.

---

*[algo/uidlist.go, line 16 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxCXKNyx9zYdGMhzbv:-KbxaQVEtYtcbVWLcAEo:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L16)):*

lidx
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 23 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-Kbx7nY94xzo9r0BBKs1:-KbxaZGl5iDWv2Sflb3E:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L23)):*

Can comment that bIdx indexes into l.Blocks and idx indexes into block.List here?
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 41 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxCoz09qjvlzaEFYxz:-Kc0no4bkuSai1ntLVs8:b-xybvh9) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L41)):*

Shouldn't this be zero and zero?
</blockquote></details>

If we want to append to an existing list, this would help in doing it. But, sure we can always start from 0 and add the corresponding elements.

---

*[algo/uidlist.go, line 53 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxD6EmW4RpwF0do0nc:-Kc0o3aVg0TEvR-plr7n:b3x24xe) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L53)):*

Really? What if it's overwriting the list? You're not using l.idx.
</blockquote></details>

It wouldn't overwrite as we start with new lists.

---

*[algo/uidlist.go, line 55 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-Kbx8vt9N6FEwmECnAGG:-Kc0NeDYDeuRtfFlwRSU:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L55)):*

Maybe better to do allocation before writing instead of after writing, as the allocated memory might not be used, though this might make little difference for larger block size.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 58 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxDL1wTXnReCVdGmwd:-Kc0ng9Z6FgXzVj8eUiR:b-2s1qqu) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L58)):*

What is the block is already there? This code needs to be rewritten to actually use it like an iterator.
</blockquote></details>

Changed it to allocate only if we need it.

---

*[algo/uidlist.go, line 67 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxDklwOkuJF5B3xyVp:-Kc0PBLBvtYRhEJctjRP:bs7wumx) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L67)):*

You should be removing this block from block slice then, right?
</blockquote></details>

Done. 

---

*[algo/uidlist.go, line 74 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxGUARqjWnni7pOpQ7:-Kc1mvhmE4uqK-RLkxss:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L74)):*

This function causes ListIterator to be allocated on heap, when it could have been on stack. Convert this function to Init() or make it be called via Next(). We don't do constructor functions.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 114 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxGnxpduy5JUBRu6EP:-Kc0TtaBxbsDXvjgdPPK:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L114)):*

This function should take a whence, where 0 means from origin, and 1 means from current position. If we're seeking from current pos, then we can optimize our seeks by iterating the existing list first, then going over the blocks.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 117 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-Kbx9h1K-iIctk5--0n5:-Kc0Pj2_MfUI__F1nvyK:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L117)):*

Better to treat uncommon case first and reduce indent.

if i >= len(u.Blocks) { l.isEnd=true; return }
j := sort.Search(...)
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 132 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-Kbx9raOJsTkpyXZC_88:-Kc0PoFCtw3oUkHtfeSw:b-moehc) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L132)):*

Maybe we can do without l.isEnd. Maintaining it requires extra care and work. Valid() can just do a simple check on the indices?
</blockquote></details>

The check would contain 4 to 5 conditions, so thought this would be better.

---

*[algo/uidlist.go, line 137 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxEtHvOxZDh6UpMKCf:-KbxZ2dofYzos6qdWpqL:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L137)):*

You don't need to check this here. Caller should have done it.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 140 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxFFQz5HCVIioWgToR:-Kc1vuACiMNbehC0TC1q:bsx6htp) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L140)):*

Avoid doing this. Let Next() or Prev() fill this up.
</blockquote></details>

This is just returning a val, am I missing something?

---

*[algo/uidlist.go, line 146 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxF-1xSqs8YuNiDHcz:-KbxZ5h730tt1Z9FM1hH:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L146)):*

You can just check the var. No need to call a func. Note that you could be dealing with very long lists.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 163 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxAcWTx7hGRXK9c-ao:-KbxX3fOMYBjL7bC3pye:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L163)):*

Maybe mention that start, end are indices not UIDs.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 166 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxFbuWGT_GOxY0CvHh:-Kc1pktc-Qzp3SehXRIl:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L166)):*

Actually, we could be writing back to ul via the WriteIterator.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 180 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxG-oSEg8VZfVSAqEP:-Kc1vcMHbzpcqkV0pd-U:bxjeakc) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L180)):*

Where are we using this?
</blockquote></details>

To convert user passed list to task.List

---

*[algo/uidlist.go, line 206 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxBFBgiXMRoun1vJdD:-KbxXiE2OqWKy3Q7RMvu:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L206)):*

len(l.Blocks)==0 instead of l.Blocks==nil just in case there is an empty array which will mess up n below.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 217 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxGBu1HM5469eK032Q:-Kc1pqLXXa1NqjLRHeIL:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L217)):*

This can be written back to u directly via writeiterator.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 234 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxHC1tDyzA2JIybMpl:-Kc1pqkdZbwthsy8k_Xi:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L234)):*

See above.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 248 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxBZrq-93Ecr49haXa:-KbxXm2pHjRLUh1tsA39:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L248)):*

After filtering, inner block might not be of blockSize? Might want to be careful with assumptions elsewhere, such as total size function.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 322 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxHVgx2uFtOi2WXzHc:-Kc1pmS183UNFpj389N2:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L322)):*

Again can be written to u.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 408 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxESRFKfqQdnHU8WGz:-KbxYozIb0IW-ROC9P3L:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L408)):*

Can't we ensure that all the blocks but the last one is fully filled? That would make this function fast.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 429 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxKuL3-uSU7HgFGrIJ:-Kbx_27_LcSiOOKnWdlJ:b53waiw) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L429)):*

Is it possible that j >= len(block.List)?

You might want to use a bit-indexed tree here if blocks can have different sizes.
</blockquote></details>

I changed it so that the blocks are always of `blockSize`, unless it's the last block.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/559)*

manishrjain:
:lgtm: Address the comments before submitting. Also, the Seek needs work, we need to run some benchmarks on the seek.

---

Reviewed 9 of 9 files at r3.
Review status: all files reviewed at latest revision, 30 unresolved discussions, some commit checks failed.

---

*[algo/uidlist.go, line 83 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbsBJ-Y1jLCpFGhzZ-F:-KcQVLaP48oNl-zPMT6a:bxmo55j) ([raw file](https://github.com/dgraph-io/dgraph/blob/c17b2f2c18038db26de1967cc59b84a5a3567c8b/algo/uidlist.go#L83)):*

We need it in createTaskQuery which need a uint64 list.
</blockquote></details>

It can use blocks as well, no? We don't need this.

---

*[algo/uidlist.go, line 180 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxG-oSEg8VZfVSAqEP:-KcQVG5Cd55QYqagAyDE:b-5ema1d) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L180)):*

To convert user passed list to task.List
</blockquote></details>

Not convinced we need this.

---

*[algo/uidlist.go, line 45 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQOFrwdy9QT5V9F2GY:-KcQOFrwdy9QT5V9F2GZ:b6q9nk3) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L45)):*
> ```Go
> 		l.list.Blocks = append(l.list.Blocks, &task.Block{List: make([]uint64, blockSize)})
> 	}
> 	if l.lidx == blockSize {
> ```

If you move this condition before the above cond, you won't have to check the size of l.bidx == len(..) again.

---

*[algo/uidlist.go, line 54 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQPDwC8QJ-EPEb4kjE:-KcQPDwDlHU1HFQA8nCe:b-6wbeu5) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L54)):*
> ```Go
> 		}
> 	}
> 	l.list.Blocks[l.bidx].List[l.lidx] = uid // = append(l.list.Blocks[l.bidx].List, uid)
> ```

it would be wise to have the writeIterator keep the current block in a variable. Micro-optimization, but given how often this would be called -- it would help.

---

*[algo/uidlist.go, line 65 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQQZwk2lH9OepiGbtg:-KcQQZwl0lHb_THeGlDx:b-8e71u0) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L65)):*
> ```Go
> 		return
> 	}
> 	l.list.Blocks[l.bidx].List = l.list.Blocks[l.bidx].List[:l.lidx]
> ```

Both these statements would be more readable if you use a var to reference block.

---

*[algo/uidlist.go, line 70 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQQrHLwAmGzwCx5V1s:-KcQQrHLwAmGzwCx5V1t:b-2dlc1a) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L70)):*
> ```Go
> 
> // NewListIterator returns a read iterator for the list passed to it.
> func (r *ListIterator) Init(l *task.List) {
> ```

ListIterator variable is l below.

---

*[algo/uidlist.go, line 98 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQRAlMeIG6w66p2x6u:-KcQRAlMeIG6w66p2x6v:ba4tdmg) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L98)):*
> ```Go
> // Seek seeks to the index whose value is greater than or equal to the given UID.
> // It uses binary search to move around.
> func (l *ListIterator) Seek(typ int, uid uint64) {
> ```

Call it whence, it's a common variable name.
https://golang.org/pkg/os/#File.Seek

And make it the second arg.

---

*[algo/uidlist.go, line 101 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQRMb7aMlzd67tSdJq:-KcQRMb7aMlzd67tSdJr:b-bgc8pf) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L101)):*
> ```Go
> 	if typ == 1 {
> 		// Seek the current list first.
> 		for l.lidx < len(l.list.Blocks[l.bidx].List) && l.list.Blocks[l.bidx].List[l.lidx] < uid {
> ```

Again, use a variable for referencing the current block. Would make the code more readable.

---

*[algo/uidlist.go, line 107 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQSmPlZ29jvcL4A86L:-KcQSmPlZ29jvcL4A86M:b-n1scq1) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L107)):*
> ```Go
> 			return
> 		}
> 	}
> ```

Add a TODO here to do benchmarking once this code is in place. I think if we're seeking from the current position, we should just do a linear scan over the blocks, instead of doing a binary search. Bin search should only be done if we're seeking from zero.

---

*[algo/uidlist.go, line 130 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQTIAB30cvJT3uGD5C:-KcQTIAB30cvJT3uGD5D:b-4ipnlt) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L130)):*
> ```Go
> // Val returns the value pointed to by the iterator.
> func (l *ListIterator) Val() uint64 {
> 	return l.list.Blocks[l.bidx].List[l.lidx]
> ```

Again, use a var for current block.

---

*[algo/uidlist.go, line 203 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQWCeeWVEKIzc_llA7:-KcQWCeeWVEKIzc_llA8:b-uke4fu) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L203)):*
> ```Go
> 
> 	n := len(l.Blocks) - 1
> 	len := n*blockSize + len(l.Blocks[n].List)
> ```

you're overwriting an internal function here locally.

---

*[algo/uidlist.go, line 210 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQWZvQZpAMCQJYZ38N:-KcQWZvR81zocL_u32eK:b-7brltv) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L210)):*
> ```Go
> 	var itu ListIterator
> 	itu.Init(u)
> 	var itv ListIterator
> ```

NewListIterator and NewWriteIterator was a better api.

---

*[algo/uidlist.go, line 232 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQY7Qtf8RZZ9vybFjf:-KcQY7Qu0tVAjtgGoqdj:b-83e5mc) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L232)):*
> ```Go
> // ApplyFilter applies a filter to our UIDList.
> func ApplyFilter(u *task.List, f func(uint64, int) bool) {
> 	var out WriteIterator
> ```

Better to have: out := NewWriteIterator(u)

---

*[algo/uidlist.go, line 236 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQYPaL_M3zdLuQNUR3:-KcQYPaMAAwENVwCV4EU:b-fdjtjm) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L236)):*
> ```Go
> 	for i, block := range u.Blocks {
> 		for j, uid := range block.List {
> 			if f(uid, Idx(u, i, j)) {
> ```

Why do you need a function for Idx here? Isn't the idx obvious from i and j?

---

*[algo/uidlist.go, line 281 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQZDgf_zTqgWWNSlML:-KcQZDgf_zTqgWWNSlMM:b-6y0w1a) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L281)):*
> ```Go
> 	lptrs := make([]ListIterator, len(lists))
> 	for i, l := range lists {
> 		var it ListIterator
> ```

The prev api was better: lptrs[i] = NewListIterator(l)

---

*[algo/uidlist.go, line 411 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQZkNR79lUOcYxdnJ4:-KcQZkNR79lUOcYxdnJ5:b-nip1o5) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L411)):*
> ```Go
> 	}
> 
> 	return i / blockSize, i - (i / blockSize * blockSize)
> ```

return i/blockSize, i % blockSize

---

*[algo/uidlist.go, line 414 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQZsEgR66_wuRXCkUO:-KcQZsEgR66_wuRXCkUP:b-74fp9b) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L414)):*
> ```Go
> }
> 
> func Idx(ul *task.List, i, j int) int {
> ```

We don't need this function. This can be done directly by the caller.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-:-KcQ_0346OP5W9ntqomM:b-p78hhl)*

ashwin95r:
Review status: 2 of 15 files reviewed at latest revision, 30 unresolved discussions.

---

*[algo/uidlist.go, line 83 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbsBJ-Y1jLCpFGhzZ-F:-KcW2UCOynJ2t32ODfCx:b-eh7cit) ([raw file](https://github.com/dgraph-io/dgraph/blob/c17b2f2c18038db26de1967cc59b84a5a3567c8b/algo/uidlist.go#L83)):*

It can use blocks as well, no? We don't need this.
</blockquote></details>

Only used by tests (Of other packages as well)

---

*[algo/uidlist.go, line 180 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KbxG-oSEg8VZfVSAqEP:-KcW1eU5wr9zYOf514n9:b-eh7cit) ([raw file](https://github.com/dgraph-io/dgraph/blob/37a9f3998a656ec5d47ddc4db2b4d21415d2c715/algo/uidlist.go#L180)):*

Not convinced we need this.
</blockquote></details>

Only used by tests (Of other packages as well)

---

*[algo/uidlist.go, line 45 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQOFrwdy9QT5V9F2GY:-KcVFC2BumlEuq7LZlnt:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L45)):*

If you move this condition before the above cond, you won't have to check the size of l.bidx == len(..) again.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 54 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQPDwC8QJ-EPEb4kjE:-KcWGjmukF-2LytFLkPL:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L54)):*

it would be wise to have the writeIterator keep the current block in a variable. Micro-optimization, but given how often this would be called -- it would help.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 65 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQQZwk2lH9OepiGbtg:-KcWDEWxCrhTELAqsh5c:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L65)):*

Both these statements would be more readable if you use a var to reference block.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 70 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQQrHLwAmGzwCx5V1s:-KcVPgyT1EoVaBoSEeVZ:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L70)):*

ListIterator variable is l below.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 98 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQRAlMeIG6w66p2x6u:-KcVFqsxNvMinkmt1SRV:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L98)):*

Call it whence, it's a common variable name.
https://golang.org/pkg/os/#File.Seek

And make it the second arg.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 101 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQRMb7aMlzd67tSdJq:-KcWGgzKxm0VIk0HaCEh:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L101)):*

Again, use a variable for referencing the current block. Would make the code more readable.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 107 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQSmPlZ29jvcL4A86L:-KcVG96Dn-zjtIDuArHW:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L107)):*

Add a TODO here to do benchmarking once this code is in place. I think if we're seeking from the current position, we should just do a linear scan over the blocks, instead of doing a binary search. Bin search should only be done if we're seeking from zero.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 130 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQTIAB30cvJT3uGD5C:-KcWGi7fIATwY1gL9NAg:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L130)):*

Again, use a var for current block.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 203 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQWCeeWVEKIzc_llA7:-KcVGVXYOaUvFrevBGly:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L203)):*

you're overwriting an internal function here locally.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 210 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQWZvQZpAMCQJYZ38N:-KcVeU4ZoWKM0fAsZIIb:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L210)):*

NewListIterator and NewWriteIterator was a better api.
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 232 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQY7Qtf8RZZ9vybFjf:-KcVeVO6GXNlJxlI2fy6:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L232)):*

Better to have: out := NewWriteIterator(u)
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 236 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQYPaL_M3zdLuQNUR3:-KcVGXtV6XEn5_-ojygn:bclwees) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L236)):*

Why do you need a function for Idx here? Isn't the idx obvious from i and j?
</blockquote></details>

As  discussed, we'd need to expose blockSize otherwise.

---

*[algo/uidlist.go, line 281 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQZDgf_zTqgWWNSlML:-KcVPNX5qLPY7O49Lgkv:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L281)):*

The prev api was better: lptrs[i] = NewListIterator(l)
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 411 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQZkNR79lUOcYxdnJ4:-KcVGgy3-3uyYLR_wU3z:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L411)):*

return i/blockSize, i % blockSize
</blockquote></details>

Done.

---

*[algo/uidlist.go, line 414 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/559#-KcQZsEgR66_wuRXCkUO:-KcVGocyThiBZD3aucNN:b-pisj36) ([raw file](https://github.com/dgraph-io/dgraph/blob/e2e08d0a714743eccfbddc3c0e817608f3a412f6/algo/uidlist.go#L414)):*

We don't need this function. This can be done directly by the caller.
</blockquote></details>

We'd need to expose blockSize. Anyways go should inline this and avoid function calls.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/559)*

ashwin95r:
Merged.

PR #571, opened 2017-02-07
Improve/ui

pawanrawal:
1. Render only first level nodes and their children. Expand/collapse nodes on click.
2. Nodes are now fixed after dragging and shouldn't move back to their initial position.
3. Production config to bundle the assets so that UI runs of port 8080 with Dgraph.
4. Ability to view properties of a node on click.
5. Don't resend query if it hasn't changed after last run.
6. Add shortcut `Ctrl + Enter` to run a query.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/571)


pawanrawal:
Merged

PR #585, opened 2017-02-14
Dynamic Schema Generation

janardhan1993:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/585)


manishrjain:
Reviewed 15 of 15 files at r1.
Review status: all files reviewed at latest revision, 22 unresolved discussions.

---

*[schema/parse.go, line 113 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcuhxe1v6-FttsWx2QB:-Kcuhxe1v6-FttsWx2QC:b-j2702h) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/parse.go#L113)):*
> ```Go
> 			}
> 		} else {
> 			str.setType(name, t)
> ```

What's this `str`?

---

*[schema/parse_test.go, line 32 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcui0mLVMR14aU6gb06:-Kcui0mLVMR14aU6gb07:brczty3) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/parse_test.go#L32)):*
> ```Go
> type nameType struct {
> 	name string
> 	typ  *types.SchemaDescription
> ```

types.Schema

---

*[schema/parse_test.go, line 78 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcui7k40DPRwJqUrlNy:-Kcui7k40DPRwJqUrlNz:b-mgh1v6) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/parse_test.go#L78)):*
> ```Go
> 	dir, ps := initSchemaStore(t)
> 	defer os.RemoveAll(dir)
> 	defer ps.Close()
> ```

Do this via a Main, so all the package level Init() happens in Main. So, ps.Close() happens after all the tests have been run. You can see how we do this in query tests.

---

*[schema/schema.go, line 29 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcuiPk6-v9ujvqmM14c:-KcuiPk6-v9ujvqmM14d:bdj9fkv) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L29)):*
> ```Go
> )
> 
> type schemaInformation struct {
> ```

You're already in package schema. So, just info should have been sufficient. Actually name it state or something. Info is too generic.

---

*[schema/schema.go, line 32 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcuijNjkb-e_Ikt4oYB:-KcuijNjkb-e_Ikt4oYC:b-bof3ta) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L32)):*
> ```Go
> 	x.SafeMutex
> 	// Map containing predicate to type information.
> 	sm map[string]*types.SchemaDescription
> ```

SchemaDescription is too long a name for a struct. Use Schema or something.

---

*[schema/schema.go, line 106 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcujri4xWLxeGTN4_50:-Kcujri4xWLxeGTN4_51:b-oodxd7) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L106)):*
> ```Go
> 		}
> 	}
> 	if err := LoadFromDb(); err != nil {
> ```

Does LoadFromDb need to be a public func?

---

*[schema/schema.go, line 117 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcuk1H073Zc_HW-53se:-Kcuk1H073Zc_HW-53sf:b2jq1ht) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L117)):*
> ```Go
> }
> 
> func getTypeValue(sd *types.SchemaDescription) []byte {
> ```

getTypeValue is really a Marshal. Do you need a function for this? You use it exactly once.

---

*[schema/schema.go, line 126 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcukH_YAGBNsuszr5mQ:-KcukH_YAGBNsuszr5mR:bgsw3qp) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L126)):*
> ```Go
> func LoadFromDb() error {
> 	prefix := x.SchemaPrefix()
> 	idxIt := pstore.NewIterator()
> ```

itr := pstore.NewIterator()

---

*[schema/schema.go, line 133 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcukPnh-vyKFtv_Mdy0:-KcukPnh-vyKFtv_Mdy1:b-ihbuak) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L133)):*
> ```Go
> 		attr := x.Parse(key).Attr
> 		slice := idxIt.Value()
> 		defer slice.Free() // there won't be too many slices
> ```

Still, it's a better code to free the slice once you've used it, from within the loop.

---

*[schema/schema.go, line 143 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcukhUV5lfL7YTt64XB:-KcukhUV5lfL7YTt64XC:b-uunama) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L143)):*
> ```Go
> 
> func reset() {
> 	str = new(schemaInformation)
> ```

str is a bad name. It sounds like string. Also, better to have a getter for this, like `Schema()`, so it's clear what you're doing, i.e. getting a package level object.

---

*[schema/schema.go, line 144 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcul2fQqg_naubW3y1V:-Kcul2fQqg_naubW3y1W:b-fm2243) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L144)):*
> ```Go
> func reset() {
> 	str = new(schemaInformation)
> 	indexedFields = make(map[string]tok.Tokenizer)
> ```

What's this for?

---

*[schema/schema.go, line 145 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcul3iSGVmqVKHF5XjP:-Kcul3iSGVmqVKHF5XjQ:b-fm2243) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L145)):*
> ```Go
> 	str = new(schemaInformation)
> 	indexedFields = make(map[string]tok.Tokenizer)
> 	reversedFields = make(map[string]bool)
> ```

What's this for?

---

*[schema/schema.go, line 169 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcul8BflTF5Da6DA6FV:-Kcul8BflTF5Da6DA6FW:b-2ljr4s) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L169)):*
> ```Go
> 	str.RLock()
> 	defer str.RUnlock()
> 	return str.getTypeOf(pred)
> ```

This lock acquisition could have been done by getTypeOf function right? Also, don't start with `get`. It should be `typeOf`.

---

*[schema/schema.go, line 217 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KculVhdv7h6xuByT7ck:-KculVheBilCrXsGo4cE:bry06zt) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L217)):*
> ```Go
> 				for _, e := range entries {
> 					if e.Water != nil {
> 						e.Water.Ch <- x.Mark{Index: e.Index, Done: true}
> ```

Instead of looping here, send a single entry to `e.Water.Ch`, with all the pending uids. Create a slice if you need to.

---

*[worker/mutation.go, line 45 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcum-p1DGwpVfPA_bdW:-Kcum-p1DGwpVfPA_bdX:b30fyyd) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/worker/mutation.go#L45)):*
> ```Go
> 		var group uint32
> 		var rv x.RaftValue
> 		if rv, ok := ctx.Value("raft").(x.RaftValue); ok {
> ```

This rv here, won't be the same as the one you declared above. So, rv above would never be initialized. Remember that when you use a `:=`, those are new variables which would temporarily shadow the ones active in that scope.

You're better off doing this:

```
rv, ok := ...
if ok { .. }
```
But, then if you have an active `rv`, you don't really need a separate group variable.

---

*[worker/mutation.go, line 94 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcumw1vYmtzDmaEATge:-Kcumw1wumjtQS87jQjP:byn2f0s) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/worker/mutation.go#L94)):*
> ```Go
> 		Water:             posting.SyncMarkFor(rv.Group),
> 	}
> 	if _, err := schema.UpdateIfMissing(&ce); err != nil {
> ```

If schema is not missing, and you don't update it, would you still send Index to be marked as done? Ensure you do that, otherwise, sync watermark would get stuck. In fact, be very careful whenever you deal with watermarks.

---

*[worker/mutation.go, line 103 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcun8skzV_JlFzr8H0f:-Kcun8skzV_JlFzr8H0g:bibxg3s) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/worker/mutation.go#L103)):*
> ```Go
> 	storageType := getType(edge)
> 
> 	if !schemaType.IsScalar() {
> ```

Instead of nesting this, you could write this as:

```
if !schemaType.IsScalar() && !storageType.IsScalar() { .. }
else if !schemaType.IsScalar() { ... }
else if !storageType.IsScalar() { ... }
else { ... }
```
This is a lot more readable.

---

*[worker/mutation.go, line 113 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcunTIdbKIQcsjebfJX:-KcunTIdbKIQcsjebfJY:b-di3ybo) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/worker/mutation.go#L113)):*
> ```Go
> 	}
> 
> 	if storageType != schemaType {
> ```

Explain what you intend to do in this case. Use comments.

Remind me, we're going to store this as storageType right, if specified. So, the precedence is like this:

1. Storage Type.
2. Schema Type.
3. String (or Binary).

In fact, explain how you achieve this.

---

*[worker/predicate.go, line 87 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcunyeu2A7jUUb7K1f9:-Kcunyeu2A7jUUb7K1fA:b-777u02) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/worker/predicate.go#L87)):*
> ```Go
> 			continue
> 		}
> 		// No need to send KC for schema keys
> ```

And why not. Explain that as well. We do need schema info there, right?

---

*[worker/predicate.go, line 241 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcuo6RFGmovquuWMXob:-Kcuo6RFGmovquuWMXoc:b-m9r4oa) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/worker/predicate.go#L241)):*
> ```Go
> 
> 		// No checksum check for schema keys
> 		if !pk.IsSchema() {
> ```

It's always better to decrease indentation and nesting whenever you can. I think you should be able to do that here.

---

*[x/keys.go, line 15 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcuoDFSKh8xFtRkVM4N:-KcuoDFSKh8xFtRkVM4O:b-oubxek) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/x/keys.go#L15)):*
> ```Go
> 	// same prefix for data, index and reverse keys so that relative order of data doesn't change
> 	// keys of same attributes are located together
> 	defaultPrefix = byte(0x00)
> ```

defaultPrefix = byteData; so it's clear what you're doing.

---

*[x/keys.go, line 105 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcup1WlEr9577X85MQn:-Kcup1WlEr9577X85MQo:bk4h1lx) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/x/keys.go#L105)):*
> ```Go
> func (p ParsedKey) SkipPredicate() []byte {
> 	buf := make([]byte, 2+len(p.Attr)+2)
> 	if p.byteType == byteSchema {
> ```

Use a function here to deal with this logic. Here and elsewhere.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/585)*

ajr1993:
Review status: 8 of 16 files reviewed at latest revision, 22 unresolved discussions.

---

*[schema/parse.go, line 113 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcuhxe1v6-FttsWx2QB:-KcvCYW4eLlbMn4NO1B9:b5t4gxs) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/parse.go#L113)):*

What's this `str`?
</blockquote></details>

Renamed it

---

*[schema/parse_test.go, line 32 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcui0mLVMR14aU6gb06:-KcvEiQrF-mxrl2_xnC4:bb6un1j) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/parse_test.go#L32)):*

types.Schema
</blockquote></details>

Done

---

*[schema/parse_test.go, line 78 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcui7k40DPRwJqUrlNy:-KcvEjNDOfBJqm3qBjQw:bb6un1j) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/parse_test.go#L78)):*

Do this via a Main, so all the package level Init() happens in Main. So, ps.Close() happens after all the tests have been run. You can see how we do this in query tests.
</blockquote></details>

Done

---

*[schema/schema.go, line 29 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcuiPk6-v9ujvqmM14c:-KcvEkz2Lwopi7htO9OL:bb6un1j) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L29)):*

You're already in package schema. So, just info should have been sufficient. Actually name it state or something. Info is too generic.
</blockquote></details>

Done

---

*[schema/schema.go, line 32 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcuijNjkb-e_Ikt4oYB:-KcvElvKBmf0xkbBS1zu:bb6un1j) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L32)):*

SchemaDescription is too long a name for a struct. Use Schema or something.
</blockquote></details>

Done

---

*[schema/schema.go, line 106 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcujri4xWLxeGTN4_50:-KcvCbGSQGHsqq-lh4hy:bjdkwhx) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L106)):*

Does LoadFromDb need to be a public func?
</blockquote></details>

This is called after a new node gets the snapshot

---

*[schema/schema.go, line 117 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcuk1H073Zc_HW-53se:-KcvEn4WnOc94v6m1h9I:bxur9jg) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L117)):*

getTypeValue is really a Marshal. Do you need a function for this? You use it exactly once.
</blockquote></details>

Removed the function

---

*[schema/schema.go, line 126 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcukH_YAGBNsuszr5mQ:-KcvEo_7EAy-ZxIE_EQz:b-b9owae) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L126)):*

itr := pstore.NewIterator()
</blockquote></details>

Renamed it

---

*[schema/schema.go, line 133 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcukPnh-vyKFtv_Mdy0:-KcvEpkHO5jNARxbDaSm:bb6un1j) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L133)):*

Still, it's a better code to free the slice once you've used it, from within the loop.
</blockquote></details>

Done

---

*[schema/schema.go, line 143 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcukhUV5lfL7YTt64XB:-KcvErHRPyCQ4zPmClFt:b-b9owae) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L143)):*

str is a bad name. It sounds like string. Also, better to have a getter for this, like `Schema()`, so it's clear what you're doing, i.e. getting a package level object.
</blockquote></details>

Renamed it

---

*[schema/schema.go, line 144 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcul2fQqg_naubW3y1V:-KcvDok-WFFVu6EXfmcU:b-bk0y50) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L144)):*

What's this for?
</blockquote></details>

Moved it inside struct now

---

*[schema/schema.go, line 145 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcul3iSGVmqVKHF5XjP:-KcvDqOVqhXTIG7-0TmP:b4uyn0s) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L145)):*

What's this for?
</blockquote></details>

moved it inside struct now

---

*[schema/schema.go, line 169 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcul8BflTF5Da6DA6FV:-KcvDyK5s6ETwp0nl2MW:bd20zay) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L169)):*

This lock acquisition could have been done by getTypeOf function right? Also, don't start with `get`. It should be `typeOf`.
</blockquote></details>

getTypeOf  can be used during schema loading, since init runs in a single thread. I have moved the lock inside.

---

*[schema/schema.go, line 217 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KculVhdv7h6xuByT7ck:-KcvENxhhz_oXXU3gK3I:b-lj9q3o) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L217)):*

Instead of looping here, send a single entry to `e.Water.Ch`, with all the pending uids. Create a slice if you need to.
</blockquote></details>

But they can belong to different raft groups, should i group watermarks per raft group and then send it in batch ?

---

*[worker/mutation.go, line 45 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcum-p1DGwpVfPA_bdW:-KcvEThdtAdVPX4qh9Y6:b-1ti77a) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/worker/mutation.go#L45)):*

This rv here, won't be the same as the one you declared above. So, rv above would never be initialized. Remember that when you use a `:=`, those are new variables which would temporarily shadow the ones active in that scope.

You're better off doing this:

```
rv, ok := ...
if ok { .. }
```
But, then if you have an active `rv`, you don't really need a separate group variable.
</blockquote></details>

:= doesn't shadow always, if there is a variable declared in the scope it resues that. Atleast one new variable should be there before :=

---

*[worker/mutation.go, line 94 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcumw1vYmtzDmaEATge:-KcvEaLyWNPZenQZAnSF:b-634dsg) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/worker/mutation.go#L94)):*

If schema is not missing, and you don't update it, would you still send Index to be marked as done? Ensure you do that, otherwise, sync watermark would get stuck. In fact, be very careful whenever you deal with watermarks.
</blockquote></details>

Sending watermark in case of schema update fails

---

*[worker/mutation.go, line 103 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcun8skzV_JlFzr8H0f:-KcvFIA6ZvINYtVMs_dg:bb6un1j) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/worker/mutation.go#L103)):*

Instead of nesting this, you could write this as:

```
if !schemaType.IsScalar() && !storageType.IsScalar() { .. }
else if !schemaType.IsScalar() { ... }
else if !storageType.IsScalar() { ... }
else { ... }
```
This is a lot more readable.
</blockquote></details>

Done

---

*[worker/mutation.go, line 113 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcunTIdbKIQcsjebfJX:-KcvFmiBfcjO2OL59L_R:b-agi6p3) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/worker/mutation.go#L113)):*

Explain what you intend to do in this case. Use comments.

Remind me, we're going to store this as storageType right, if specified. So, the precedence is like this:

1. Storage Type.
2. Schema Type.
3. String (or Binary).

In fact, explain how you achieve this.
</blockquote></details>

Added Comments

---

*[worker/predicate.go, line 87 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcunyeu2A7jUUb7K1f9:-KcvGcnDEFgryFVuCOAa:bvih3tr) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/worker/predicate.go#L87)):*

And why not. Explain that as well. We do need schema info there, right?
</blockquote></details>

For pl we send checksum so that if there is no change, then they are not sent over wire. In case of schema we won't save anything with checksum.

---

*[worker/predicate.go, line 241 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcuo6RFGmovquuWMXob:-KcvPxeth016CWybbD6n:b-i3xyql) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/worker/predicate.go#L241)):*

It's always better to decrease indentation and nesting whenever you can. I think you should be able to do that here.
</blockquote></details>

For pl we need checksum checking, so it should be nested

---

*[x/keys.go, line 15 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcuoDFSKh8xFtRkVM4N:-KcvHrM7nj9B22n1esUO:b-iwaos5) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/x/keys.go#L15)):*

defaultPrefix = byteData; so it's clear what you're doing.
</blockquote></details>

This is not byteData,  i have added same prefix to data, reverse and index keys so that their relative sorted order doesn't change. We needed a way to iterate over schema keys, so there is different prefix for that.

---

*[x/keys.go, line 105 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcup1WlEr9577X85MQn:-KcvIqpfU9CVWakiKntL:bb6un1j) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/x/keys.go#L105)):*

Use a function here to deal with this logic. Here and elsewhere.
</blockquote></details>

Done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/585)*

gacharya255:
Review status: 8 of 17 files reviewed at latest revision, 27 unresolved discussions, some commit checks failed.

---

*[posting/list.go, line 336 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcvbLe7sY1nwtQCTpyP:-KcvbLe7sY1nwtQCTpyQ:b-l4hgot) ([raw file](https://github.com/dgraph-io/dgraph/blob/5cd1e19f31bbe2e0db193b502103a5202b54e18d/posting/list.go#L336)):*
> ```Go
> }
> 
> func EdgeTypeID(edge *task.DirectedEdge) types.TypeID {
> ```

EdgeTypeID is type ID for destination vertex? Maybe add a comment for this exported function.

---

*[schema/parse_test.go, line 27 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcvjmXZsulUsY1wf3a2:-KcvjmXZsulUsY1wf3a3:b-vv4k68) ([raw file](https://github.com/dgraph-io/dgraph/blob/5cd1e19f31bbe2e0db193b502103a5202b54e18d/schema/parse_test.go#L27)):*
> ```Go
> 	"github.com/dgraph-io/dgraph/types"
> 	"github.com/dgraph-io/dgraph/x"
> 	"github.com/stretchr/testify/require"
> ```

Our convention is that external github deps are in a separate group. The extra line in between is intentional.

---

*[schema/schema.go, line 133 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcukPnh-vyKFtv_Mdy0:-Kcvm17Xxh4fk34Io5te:b-jtm01b) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L133)):*
<details><summary><i>Previously, ajr1993 (Janardhan Reddy) wrote…</i></summary><blockquote>

Done
</blockquote></details>

I think you shouldn't free.. This value buffer belongs to the RocksDB iterator and is reused across iterations. It should be cleared when you close the iterator.

(You can dig through rdbc.cc in our codebase and RocksDB https://github.com/facebook/rocksdb/blob/master/db/db_iter.cc#L174)

---

*[schema/schema.go, line 129 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcvqCqdFF-wgtdccxkz:-KcvqCqf4h6O9_rI6o84:b7drvs2) ([raw file](https://github.com/dgraph-io/dgraph/blob/5cd1e19f31bbe2e0db193b502103a5202b54e18d/schema/schema.go#L129)):*
> ```Go
> }
> 
> func (s *state) IsReversed(pred string) bool {
> ```

Exported functions should have comments.

---

*[schema/schema.go, line 145 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcvbjFR_zwIhIaJqb8R:-KcvbjFS2pvD52qPQfmH:b-z8l8th) ([raw file](https://github.com/dgraph-io/dgraph/blob/5cd1e19f31bbe2e0db193b502103a5202b54e18d/schema/schema.go#L145)):*
> ```Go
> 
> func stateInfo() *state {
> 	return predicateInfo
> ```

Maybe we can remove this function as it is not exported either.

---

*[x/keys.go, line 32 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcvneEuflMpWZP0UNYD:-KcvneEuflMpWZP0UNYE:bk5zpsk) ([raw file](https://github.com/dgraph-io/dgraph/blob/5cd1e19f31bbe2e0db193b502103a5202b54e18d/x/keys.go#L32)):*
> ```Go
> func SchemaKey(attr string) []byte {
> 	buf := make([]byte, 2+len(attr)+2)
> 	buf[0] = byteSchema
> ```

Wonder if you can save one byte here and just use the current first byte like index, reverse to indicate a schema key?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/585)*

ajr1993:
Review status: 8 of 17 files reviewed at latest revision, 27 unresolved discussions, some commit checks failed.

---

*[posting/list.go, line 336 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcvbLe7sY1nwtQCTpyP:-Kcw0s7LghhIQlNyiu-z:bb7f2mf) ([raw file](https://github.com/dgraph-io/dgraph/blob/5cd1e19f31bbe2e0db193b502103a5202b54e18d/posting/list.go#L336)):*

EdgeTypeID is type ID for destination vertex? Maybe add a comment for this exported function.
</blockquote></details>

done

---

*[schema/parse_test.go, line 27 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcvjmXZsulUsY1wf3a2:-Kcw0cMnVZIA8oRpVIqm:boleko6) ([raw file](https://github.com/dgraph-io/dgraph/blob/5cd1e19f31bbe2e0db193b502103a5202b54e18d/schema/parse_test.go#L27)):*

Our convention is that external github deps are in a separate group. The extra line in between is intentional.
</blockquote></details>

Done. There is no space now, https://github.com/dgraph-io/dgraph/blob/dynamic-schema-generation/schema/parse_test.

---

*[schema/schema.go, line 133 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcukPnh-vyKFtv_Mdy0:-Kcw5rQKSubP7tCUQ5al:bi13bgj) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L133)):*

I think you shouldn't free.. This value buffer belongs to the RocksDB iterator and is reused across iterations. It should be cleared when you close the iterator.

(You can dig through rdbc.cc in our codebase and RocksDB https://github.com/facebook/rocksdb/blob/master/db/db_iter.cc#L174)
</blockquote></details>

fixed, no need to free

---

*[schema/schema.go, line 217 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KculVhdv7h6xuByT7ck:-KcwS8KhheIBA9YOkZUy:b-896fjx) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/schema/schema.go#L217)):*
<details><summary><i>Previously, ajr1993 (Janardhan Reddy) wrote…</i></summary><blockquote>

But they can belong to different raft groups, should i group watermarks per raft group and then send it in batch ?
</blockquote></details>

Done

---

*[schema/schema.go, line 129 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcvqCqdFF-wgtdccxkz:-Kcw0d_mGP5yLbhtxswb:b-9p02nr) ([raw file](https://github.com/dgraph-io/dgraph/blob/5cd1e19f31bbe2e0db193b502103a5202b54e18d/schema/schema.go#L129)):*

Exported functions should have comments.
</blockquote></details>

Comments are already there

---

*[schema/schema.go, line 145 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcvbjFR_zwIhIaJqb8R:-Kcw0fpTCPDzLE2g9iRR:btzj2qo) ([raw file](https://github.com/dgraph-io/dgraph/blob/5cd1e19f31bbe2e0db193b502103a5202b54e18d/schema/schema.go#L145)):*

Maybe we can remove this function as it is not exported either.
</blockquote></details>

Added this function to denote we are accessing package level object as suggested by manish

---

*[worker/mutation.go, line 94 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kcumw1vYmtzDmaEATge:-KcwSB8kyEfAOCAhUExp:b-srr47g) ([raw file](https://github.com/dgraph-io/dgraph/blob/edea5634c970e41d14d729bf94cf30b03a8684b4/worker/mutation.go#L94)):*
<details><summary><i>Previously, ajr1993 (Janardhan Reddy) wrote…</i></summary><blockquote>

Sending watermark in case of schema update fails
</blockquote></details>

Ignore the earlier comment, watermark is sent only if schema update is successful, 

---

*[x/keys.go, line 32 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcvneEuflMpWZP0UNYD:-Kcw3zaOtrVfb4uRUBXT:b9bozim) ([raw file](https://github.com/dgraph-io/dgraph/blob/5cd1e19f31bbe2e0db193b502103a5202b54e18d/x/keys.go#L32)):*

Wonder if you can save one byte here and just use the current first byte like index, reverse to indicate a schema key?
</blockquote></details>

the first two bytes currently denotes the length of the attribute, so the first byte is covering all possible byteprefix space.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/585)*

gacharya255:
:lgtm:

---

Review status: 7 of 17 files reviewed at latest revision, 23 unresolved discussions, some commit checks failed.

---

*[worker/mutation.go, line 102 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kd-_jAMBOlUUwn5hTXl:-Kd-_jAMBOlUUwn5hTXm:becz9s0) ([raw file](https://github.com/dgraph-io/dgraph/blob/2f922e3ac6db5e075de3cb2fdc7d235f279746d6/worker/mutation.go#L102)):*
> ```Go
> 		if dst, err := types.Convert(src, schemaType); err != nil {
> 			return err
> 		} else {
> ```

Minor: You can drop this else and reduce indent.

---

*[worker/mutation.go, line 107 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kd-_oPpbIebQi2O48-2:-Kd-_oPpbIebQi2O48-3:byb5u5a) ([raw file](https://github.com/dgraph-io/dgraph/blob/2f922e3ac6db5e075de3cb2fdc7d235f279746d6/worker/mutation.go#L107)):*
> ```Go
> 			if err = types.Marshal(dst, &b); err != nil {
> 				return err
> 			} else {
> ```

Minor: Can drop this else.

---

*[x/keys.go, line 32 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcvneEuflMpWZP0UNYD:-Kd-XQE0dec1OqOHihjW:b-7v05ay) ([raw file](https://github.com/dgraph-io/dgraph/blob/5cd1e19f31bbe2e0db193b502103a5202b54e18d/x/keys.go#L32)):*
<details><summary><i>Previously, ajr1993 (Janardhan Reddy) wrote…</i></summary><blockquote>

the first two bytes currently denotes the length of the attribute, so the first byte is covering all possible byteprefix space.
</blockquote></details>

Mainly I am concerned about dataKey increasing by one byte. But it is probably a separate problem. We have been doing <len><attr><keyType><otherStuff>. Maybe <keyType><len><attr><otherStuff> would have been better. But we can discuss this separately.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-:-Kd-_yYyzbiDVCvl_27c:bnfp4nl)*

ajr1993:
Review status: 7 of 17 files reviewed at latest revision, 23 unresolved discussions, some commit checks failed.

---

*[worker/mutation.go, line 102 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kd-_jAMBOlUUwn5hTXl:-Kd-b23kwvBARde1KMU7:bb6un1j) ([raw file](https://github.com/dgraph-io/dgraph/blob/2f922e3ac6db5e075de3cb2fdc7d235f279746d6/worker/mutation.go#L102)):*

Minor: You can drop this else and reduce indent.
</blockquote></details>

Done

---

*[worker/mutation.go, line 107 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kd-_oPpbIebQi2O48-2:-Kd-b2oY51nwrqqnH1AD:b-v95ovz) ([raw file](https://github.com/dgraph-io/dgraph/blob/2f922e3ac6db5e075de3cb2fdc7d235f279746d6/worker/mutation.go#L107)):*
> Reply...
Done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/585)*

manishrjain:
Review status: 7 of 17 files reviewed at latest revision, 23 unresolved discussions, some commit checks failed.

---

*[schema/schema.go, line 32 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcwVb2rEEKY0lY8Wawa:-KcwVb2rEEKY0lY8Wawb:b-m4bwcg) ([raw file](https://github.com/dgraph-io/dgraph/blob/9485eb51cf75985542b9f08fd0e859b30de3f256/schema/schema.go#L32)):*
> ```Go
> 	x.SafeMutex
> 	// Map containing predicate to type information.
> 	info map[string]*types.Schema
> ```

I ban you from naming anything 'info' or 'information'. Think about what this map does, then name accordingly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/585)*

ajr1993:
Review status: 7 of 17 files reviewed at latest revision, 23 unresolved discussions, some commit checks failed.

---

*[schema/schema.go, line 32 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcwVb2rEEKY0lY8Wawa:-Kd4i5ok3dyhOfKAJnZk:b-7rl0kd) ([raw file](https://github.com/dgraph-io/dgraph/blob/9485eb51cf75985542b9f08fd0e859b30de3f256/schema/schema.go#L32)):*

I ban you from naming anything 'info' or 'information'. Think about what this map does, then name accordingly.
</blockquote></details>

done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/585)*

manishrjain:
Reviewed 3 of 9 files at r2, 1 of 4 files at r4, 2 of 6 files at r5, 11 of 11 files at r7.
Review status: 21 of 23 files reviewed at latest revision, 22 unresolved discussions, some commit checks failed.

---

*[posting/list.go, line 337 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kd56ngGyqbssG1ARQfp:-Kd56ngGyqbssG1ARQfq:b-z2y0fa) ([raw file](https://github.com/dgraph-io/dgraph/blob/4c2b8095330292ac0a70b043c179beff2193b85c/posting/list.go#L337)):*
> ```Go
> 
> // EdgeTypeID returns the typeid of destiantion vertex
> func EdgeTypeID(edge *task.DirectedEdge) types.TypeID {
> ```

TypeId

---

*[schema/parse.go, line 152 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kd57DvVOzW7HHgZtKe4:-Kd57DvVOzW7HHgZtKe5:b-itqen0) ([raw file](https://github.com/dgraph-io/dgraph/blob/4c2b8095330292ac0a70b043c179beff2193b85c/schema/parse.go#L152)):*
> ```Go
> 
> // processIndexDirective works on "@index" or "@index(customtokenizer)".
> func processIndexDirective(it *lex.ItemIterator, name string, typ types.TypeID) error {
> ```

s/name/pred

---

*[schema/parse_test.go, line 24 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kd57VSQiuqEjS9jHxjC:-Kd57VSQiuqEjS9jHxjD:b-aubp6w) ([raw file](https://github.com/dgraph-io/dgraph/blob/4c2b8095330292ac0a70b043c179beff2193b85c/schema/parse_test.go#L24)):*
> ```Go
> 	"testing"
> 
> 	"github.com/stretchr/testify/require"
> ```

Run this test with `-race`, and ensure that it doesn't have any race cond.

---

*[schema/schema.go, line 145 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KcvbjFR_zwIhIaJqb8R:-Kd58PmSImOHb0PNN1yM:b-13uje4) ([raw file](https://github.com/dgraph-io/dgraph/blob/5cd1e19f31bbe2e0db193b502103a5202b54e18d/schema/schema.go#L145)):*
<details><summary><i>Previously, ajr1993 (Janardhan Reddy) wrote…</i></summary><blockquote>

Added this function to denote we are accessing package level object as suggested by manish
</blockquote></details>

And now exported it as well, removing the need for public wrapper functions.

---

*[schema/schema.go, line 48 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kd5AvygMxhXoJ1i_KiB:-Kd5AvygMxhXoJ1i_KiC:b-seaio8) ([raw file](https://github.com/dgraph-io/dgraph/blob/4c2b8095330292ac0a70b043c179beff2193b85c/schema/schema.go#L48)):*
> ```Go
> // UpdateIfMissing checks sets the schema if it doesn't exist
> // already for given predicate
> func (s *state) UpdateIfMissing(se *SyncEntry) (types.TypeID, error) {
> ```

Don't send types.TypeID. Only return a bool to indicate if updated.

---

*[schema/schema.go, line 53 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kd59TgjQ1TN5PZOU6i0:-Kd59TgjQ1TN5PZOU6i1:b-mq0bg9) ([raw file](https://github.com/dgraph-io/dgraph/blob/4c2b8095330292ac0a70b043c179beff2193b85c/schema/schema.go#L53)):*
> ```Go
> 
> 	if oldVal, ok := s.predicate[se.Attr]; ok && oldVal.ValueType != se.SchemaDescription.ValueType {
> 		return types.TypeID(oldVal.ValueType), x.Errorf("Schema for attr %s already set to %d", se.Attr, se.SchemaDescription.ValueType)
> ```

Over 100 chars.

---

*[schema/schema.go, line 81 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kd581QEFWnByuQ4GNxP:-Kd581QEFWnByuQ4GNxQ:b-bci2z8) ([raw file](https://github.com/dgraph-io/dgraph/blob/4c2b8095330292ac0a70b043c179beff2193b85c/schema/schema.go#L81)):*
> ```Go
> 	s.Lock()
> 	defer s.Unlock()
> 	if schema, ok := s.predicate[attr]; ok {
> ```

Ensure that type is either not specified or set to UID. Can't be anything else. In fact, set it to UID if type is not there.

---

*[schema/schema.go, line 174 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kd58_PuMrb6JQ0PRODw:-Kd58_PuMrb6JQ0PRODx:b-pw5z7l) ([raw file](https://github.com/dgraph-io/dgraph/blob/4c2b8095330292ac0a70b043c179beff2193b85c/schema/schema.go#L174)):*
> ```Go
> 
> // LoadFromDb reads schema information from db
> // and stores it in memory
> ```

This can fit in one line. 100 chars.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/585)*

ajr1993:
Review status: 12 of 32 files reviewed at latest revision, 22 unresolved discussions, some commit checks failed.

---

*[posting/list.go, line 337 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kd56ngGyqbssG1ARQfp:-KdAPS5148oZgvoiXsxC:bb7f2mf) ([raw file](https://github.com/dgraph-io/dgraph/blob/4c2b8095330292ac0a70b043c179beff2193b85c/posting/list.go#L337)):*

TypeId
</blockquote></details>

done

---

*[schema/parse.go, line 152 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kd57DvVOzW7HHgZtKe4:-KdAQUeCmiG-SSrKrbWt:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/4c2b8095330292ac0a70b043c179beff2193b85c/schema/parse.go#L152)):*

s/name/pred
</blockquote></details>

Done.

---

*[schema/parse_test.go, line 24 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kd57VSQiuqEjS9jHxjC:-KdAPB2HOjSqil4_BeEs:b-98evp5) ([raw file](https://github.com/dgraph-io/dgraph/blob/4c2b8095330292ac0a70b043c179beff2193b85c/schema/parse_test.go#L24)):*

Run this test with `-race`, and ensure that it doesn't have any race cond.
</blockquote></details>

Didn't find any race conditions in schema

---

*[schema/schema.go, line 48 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kd5AvygMxhXoJ1i_KiB:-KdAPoKdGhGKMY-lDm4N:bb7f2mf) ([raw file](https://github.com/dgraph-io/dgraph/blob/4c2b8095330292ac0a70b043c179beff2193b85c/schema/schema.go#L48)):*

Don't send types.TypeID. Only return a bool to indicate if updated.
</blockquote></details>

done

---

*[schema/schema.go, line 53 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kd59TgjQ1TN5PZOU6i0:-KdAQY9XdiwKwm8ylJ-X:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/4c2b8095330292ac0a70b043c179beff2193b85c/schema/schema.go#L53)):*

Over 100 chars.
</blockquote></details>

Done.

---

*[schema/schema.go, line 81 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kd581QEFWnByuQ4GNxP:-KdAPrVz8gblCBvtbKfT:bb7f2mf) ([raw file](https://github.com/dgraph-io/dgraph/blob/4c2b8095330292ac0a70b043c179beff2193b85c/schema/schema.go#L81)):*

Ensure that type is either not specified or set to UID. Can't be anything else. In fact, set it to UID if type is not there.
</blockquote></details>

done

---

*[schema/schema.go, line 174 at r7](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-Kd58_PuMrb6JQ0PRODw:-KdAPxxSkDSVd1mzCYcw:bb7f2mf) ([raw file](https://github.com/dgraph-io/dgraph/blob/4c2b8095330292ac0a70b043c179beff2193b85c/schema/schema.go#L174)):*

This can fit in one line. 100 chars.
</blockquote></details>

done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/585)*

manishrjain:
Review status: 12 of 32 files reviewed at latest revision, 24 unresolved discussions, some commit checks failed.

---

*[cmd/dgraph/main.go, line 253 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdAi3kG6RxQP2yxO9_Q:-KdAi3kG6RxQP2yxO9_R:bm58iqd) ([raw file](https://github.com/dgraph-io/dgraph/blob/fd6db92ed74ae6a0263a4f3b288be9a3ce111680/cmd/dgraph/main.go#L253)):*
> ```Go
> 	var err error
> 
> 	if err = validateTypes(mu.Set); err != nil {
> ```

We should validate here based on the existing schema.

---

*[query/graph/graphresponse.proto, line 34 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdAilSo6FgqVyTh4FwY:-KdAilSo6FgqVyTh4FwZ:bufpndf) ([raw file](https://github.com/dgraph-io/dgraph/blob/f03cc62b33e46fca5d0538df13520b0914453eca/query/graph/graphresponse.proto#L34)):*
> ```protobuf
>         bytes date_val = 7;
>         bytes datetime_val = 8;
>         string default_val = 15;
> ```

set it to zero.

---

*[schema/schema.go, line 30 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdAj8hAckmYLLfRUqDT:-KdAj8hAckmYLLfRUqDU:bmky42v) ([raw file](https://github.com/dgraph-io/dgraph/blob/f03cc62b33e46fca5d0538df13520b0914453eca/schema/schema.go#L30)):*
> ```Go
> 
> var (
> 	pState *state
> ```

inconsistent with pstore.

---

*[schema/schema.go, line 48 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdAk8ypVp0f1O7B2sWA:-KdAk8ypVp0f1O7B2sWB:b-6z8776) ([raw file](https://github.com/dgraph-io/dgraph/blob/f03cc62b33e46fca5d0538df13520b0914453eca/schema/schema.go#L48)):*
> ```Go
> // Update updates the schema in memory and sends an entry to syncCh so that it can be
> // comitted later
> func (s *state) Update(se *SyncEntry) {
> ```

Return the bool about whether it updated.

---

*[schema/schema.go, line 50 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdAkMOg2aMWt09F7bg5:-KdAkMOg2aMWt09F7bg6:bf5io3k) ([raw file](https://github.com/dgraph-io/dgraph/blob/f03cc62b33e46fca5d0538df13520b0914453eca/schema/schema.go#L50)):*
> ```Go
> func (s *state) Update(se *SyncEntry) {
> 	s.Lock()
> 	defer s.Unlock()
> ```

UpdateIfMissing would have a read lock.

---

*[schema/schema.go, line 53 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdAk0eGAfRkuyIJD0rt:-KdAk0eIFbeutLloYSq1:b1gfabq) ([raw file](https://github.com/dgraph-io/dgraph/blob/f03cc62b33e46fca5d0538df13520b0914453eca/schema/schema.go#L53)):*
> ```Go
> 
> 	_, ok := s.predicate[se.Attr]
> 	x.AssertTruef(!ok, "Schema already exists for predicate %s", se.Attr)
> ```

Don't assert. There function gets called via goroutines.

---

*[schema/schema.go, line 121 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdAkaXicDSmDJYDxs1E:-KdAkaXicDSmDJYDxs1F:b9xc4j3) ([raw file](https://github.com/dgraph-io/dgraph/blob/f03cc62b33e46fca5d0538df13520b0914453eca/schema/schema.go#L121)):*
> ```Go
> 	defer s.RUnlock()
> 	if schema, ok := s.predicate[pred]; ok {
> 		return schema.Tokenizer != ""
> ```

return len(schema.Tokenizer) > 0

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/585)*

janardhan1993:
Review status: 5 of 102 files reviewed at latest revision, 24 unresolved discussions, some commit checks failed.

---

*[cmd/dgraph/main.go, line 253 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdAi3kG6RxQP2yxO9_Q:-KdBHhsnnY_W1BlR-zfi:bb7f2mf) ([raw file](https://github.com/dgraph-io/dgraph/blob/fd6db92ed74ae6a0263a4f3b288be9a3ce111680/cmd/dgraph/main.go#L253)):*

We should validate here based on the existing schema.
</blockquote></details>

done

---

*[query/graph/graphresponse.proto, line 34 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdAilSo6FgqVyTh4FwY:-KdBHj-WQpdUhF8-DlBG:bb7f2mf) ([raw file](https://github.com/dgraph-io/dgraph/blob/f03cc62b33e46fca5d0538df13520b0914453eca/query/graph/graphresponse.proto#L34)):*

set it to zero.
</blockquote></details>

done

---

*[schema/schema.go, line 30 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdAj8hAckmYLLfRUqDT:-KdBHjuHz3dlipIhIK2m:bb7f2mf) ([raw file](https://github.com/dgraph-io/dgraph/blob/f03cc62b33e46fca5d0538df13520b0914453eca/schema/schema.go#L30)):*

inconsistent with pstore.
</blockquote></details>

done

---

*[schema/schema.go, line 48 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdAk8ypVp0f1O7B2sWA:-KdBHkljhvQTvK6B0DRv:bb7f2mf) ([raw file](https://github.com/dgraph-io/dgraph/blob/f03cc62b33e46fca5d0538df13520b0914453eca/schema/schema.go#L48)):*

Return the bool about whether it updated.
</blockquote></details>

done

---

*[schema/schema.go, line 50 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdAkMOg2aMWt09F7bg5:-KdBHlbXUf_5FEz7ZbWN:bb7f2mf) ([raw file](https://github.com/dgraph-io/dgraph/blob/f03cc62b33e46fca5d0538df13520b0914453eca/schema/schema.go#L50)):*

UpdateIfMissing would have a read lock.
</blockquote></details>

done

---

*[schema/schema.go, line 53 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdAk0eGAfRkuyIJD0rt:-KdBHmR3cKjfQFnWI2Ah:bb7f2mf) ([raw file](https://github.com/dgraph-io/dgraph/blob/f03cc62b33e46fca5d0538df13520b0914453eca/schema/schema.go#L53)):*

Don't assert. There function gets called via goroutines.
</blockquote></details>

done

---

*[schema/schema.go, line 121 at r9](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdAkaXicDSmDJYDxs1E:-KdBHn9-BsE1LfemqBtA:bb7f2mf) ([raw file](https://github.com/dgraph-io/dgraph/blob/f03cc62b33e46fca5d0538df13520b0914453eca/schema/schema.go#L121)):*

return len(schema.Tokenizer) > 0
</blockquote></details>

done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/585)*

manishrjain:
:lgtm: Amazing job! Really like the changes here.

---

Reviewed 4 of 19 files at r9, 16 of 92 files at r11, 6 of 9 files at r12, 2 of 2 files at r14.
Review status: all files reviewed at latest revision, 23 unresolved discussions, some commit checks failed.

---

*[schema/schema.go, line 36 at r14](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdEkf0yyipD6RKqAKpH:-KdEkf0yyipD6RKqAKpI:b-ef9yp1) ([raw file](https://github.com/dgraph-io/dgraph/blob/d5d8dc8cc1afe314450c14792cede8fe3d01e86d/schema/schema.go#L36)):*
> ```Go
> 
> type state struct {
> 	// Can have fine grained locking later if necessary, per group or predicate
> ```

You can have an eventlog (look at watermark), and then print any info to that.

---

*[schema/schema.go, line 49 at r14](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdEk44mHFYEjV2esDsL:-KdEk44mHFYEjV2esDsM:b-6nzv6v) ([raw file](https://github.com/dgraph-io/dgraph/blob/d5d8dc8cc1afe314450c14792cede8fe3d01e86d/schema/schema.go#L49)):*
> ```Go
> // Update updates the schema in memory if it doesn't exist already. And sends an
> // entry to syncCh so that it can be comitted later
> func (s *state) UpdateIfMissing(se *SyncEntry) (types.TypeID, bool) {
> ```

Remove this.

---

*[schema/schema.go, line 84 at r14](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdEk8gvFwElLBYf_kxN:-KdEk8gvFwElLBYf_kxO:bcr7q8r) ([raw file](https://github.com/dgraph-io/dgraph/blob/d5d8dc8cc1afe314450c14792cede8fe3d01e86d/schema/schema.go#L84)):*
> ```Go
> 	s.predicate[se.Attr] = &schema
> 	se.Water.Ch <- x.Mark{Index: se.Index, Done: false}
> ```

Remove vertical space.

---

*[schema/schema.go, line 87 at r14](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdEk9m0M2-qPy9ZBkHu:-KdEk9m0M2-qPy9ZBkHv:b7xm7xe) ([raw file](https://github.com/dgraph-io/dgraph/blob/d5d8dc8cc1afe314450c14792cede8fe3d01e86d/schema/schema.go#L87)):*
> ```Go
> 	syncCh <- se
> 
> 	fmt.Printf("Setting schema for attr %s: %v\n", se.Attr, se.Schema.ValueType)
> ```

Remove this fmt. This can take a context, and you can write to context using x.Trace(..)

---

*[schema/schema.go, line 118 at r14](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdEkV6W5EQSoTeDVysf:-KdEkV6W5EQSoTeDVysg:b20yzhy) ([raw file](https://github.com/dgraph-io/dgraph/blob/d5d8dc8cc1afe314450c14792cede8fe3d01e86d/schema/schema.go#L118)):*
> ```Go
> // SetIndex sets the tokenizer for given predicate
> func (s *state) SetIndex(pred string, tokenizer string) {
> 	fmt.Println("Setting tokenizer", pred, tokenizer)
> ```

Remove all fmt.Print

---

*[schema/schema.go, line 131 at r14](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdEkZDxYapuOq49bNlG:-KdEkZDxYapuOq49bNlH:ba3dfmt) ([raw file](https://github.com/dgraph-io/dgraph/blob/d5d8dc8cc1afe314450c14792cede8fe3d01e86d/schema/schema.go#L131)):*
> ```Go
> 	defer s.Unlock()
> 	s.predicate[pred] = schema
> 	fmt.Printf("Setting schema for attr %s: %v\n", pred, schema.ValueType)
> ```

Remove.

---

*[schema/schema.go, line 197 at r14](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdElc0aqdy9NbABqJ3_:-KdElc0bbd4xcXis1oZY:b-8co3jw) ([raw file](https://github.com/dgraph-io/dgraph/blob/d5d8dc8cc1afe314450c14792cede8fe3d01e86d/schema/schema.go#L197)):*
> ```Go
> 
> // ReloadData loads schema from file and then later from db
> func ReloadData(file string) error {
> ```

Move this function to test file; and also make it private.

---

*[worker/draft.go, line 464 at r14](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdEn1GQ-82rs9WpJMmP:-KdEn1GRPt85ik-pR41l:ba3dfmt) ([raw file](https://github.com/dgraph-io/dgraph/blob/d5d8dc8cc1afe314450c14792cede8fe3d01e86d/worker/draft.go#L464)):*
> ```Go
> 		// schema here without any locking
> 		var proposal task.Proposal
> 		x.AssertTrue(len(e.Data) > 0)
> ```

Remove.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-:-KdEnhQXDDpuiTVjxuBe:b-dvfwss)*

janardhan1993:
Review status: 30 of 32 files reviewed at latest revision, 23 unresolved discussions, some commit checks failed.

---

*[schema/schema.go, line 36 at r14](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdEkf0yyipD6RKqAKpH:-KdEurbvm6yhELj8mWz7:bb6un1j) ([raw file](https://github.com/dgraph-io/dgraph/blob/d5d8dc8cc1afe314450c14792cede8fe3d01e86d/schema/schema.go#L36)):*

You can have an eventlog (look at watermark), and then print any info to that.
</blockquote></details>

Done

---

*[schema/schema.go, line 49 at r14](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdEk44mHFYEjV2esDsL:-KdEuefVwABR5_zcm9ss:bb6un1j) ([raw file](https://github.com/dgraph-io/dgraph/blob/d5d8dc8cc1afe314450c14792cede8fe3d01e86d/schema/schema.go#L49)):*

Remove this.
</blockquote></details>

Done

---

*[schema/schema.go, line 84 at r14](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdEk8gvFwElLBYf_kxN:-KdEufwWFTgrpvWA76t-:bb6un1j) ([raw file](https://github.com/dgraph-io/dgraph/blob/d5d8dc8cc1afe314450c14792cede8fe3d01e86d/schema/schema.go#L84)):*

Remove vertical space.
</blockquote></details>

Done

---

*[schema/schema.go, line 87 at r14](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdEk9m0M2-qPy9ZBkHu:-KdEuhBlmNsVnf2y5LYA:bb6un1j) ([raw file](https://github.com/dgraph-io/dgraph/blob/d5d8dc8cc1afe314450c14792cede8fe3d01e86d/schema/schema.go#L87)):*

Remove this fmt. This can take a context, and you can write to context using x.Trace(..)
</blockquote></details>

Done

---

*[schema/schema.go, line 118 at r14](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdEkV6W5EQSoTeDVysf:-KdEuiVIeqVnlwvTapm5:bb6un1j) ([raw file](https://github.com/dgraph-io/dgraph/blob/d5d8dc8cc1afe314450c14792cede8fe3d01e86d/schema/schema.go#L118)):*

Remove all fmt.Print
</blockquote></details>

Done

---

*[schema/schema.go, line 131 at r14](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdEkZDxYapuOq49bNlG:-KdEujfpsUQ-O3RYYfTH:bb6un1j) ([raw file](https://github.com/dgraph-io/dgraph/blob/d5d8dc8cc1afe314450c14792cede8fe3d01e86d/schema/schema.go#L131)):*

Remove.
</blockquote></details>

Done

---

*[schema/schema.go, line 197 at r14](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdElc0aqdy9NbABqJ3_:-KdEul95j9rfxXTc-86y:bb6un1j) ([raw file](https://github.com/dgraph-io/dgraph/blob/d5d8dc8cc1afe314450c14792cede8fe3d01e86d/schema/schema.go#L197)):*

Move this function to test file; and also make it private.
</blockquote></details>

Done

---

*[worker/draft.go, line 464 at r14](https://reviewable.io:443/reviews/dgraph-io/dgraph/585#-KdEn1GQ-82rs9WpJMmP:-KdEumJlEQJv0U4RkODy:bb6un1j) ([raw file](https://github.com/dgraph-io/dgraph/blob/d5d8dc8cc1afe314450c14792cede8fe3d01e86d/worker/draft.go#L464)):*

Remove.
</blockquote></details>

Done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/585)*

janardhan1993:
Merged Manually

PR #591, opened 2017-02-15
Add TLS support between nodes and client

Motakjuq:
Add additional flags to configure the server to use TLS
- tls: boolean, use a tls listener on server, default true
- cert: string, certificate path, default blank
- cert_key: certificate key path, default blank

For client (dgraphloader)
- tls: boolean, use tls to connect with the server, default true
- insecure: boolean, skip certificate validation (useful for self-signed certificates), default false

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/591)


Motakjuq:
Please close this PR, I had misconfigured an environment and I used a wrong user

PR #593, opened 2017-02-15
Add TLS configuration for client and server

Motakjuq:
Add additional flags to configure the server to use TLS
- tls: boolean, use a tls listener on server, default false
- cert: string, certificate path, default blank
- cert_key: certificate key path, default blank

For client (dgraphloader)
- tls: boolean, use tls to connect with the server, default false
- insecure: boolean, skip certificate validation (useful for self-signed certificates), default false

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/593)

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -19,6 +19,8 @@ package client
 import (
 	"context"
 	"fmt"
+	"log"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -171,7 +173,12 @@ func (batch *BatchMutation) request(req *Req) {
 RETRY:
 	_, err := batch.dc.Run(context.Background(), &req.gr)
 	if err != nil {
-		fmt.Printf("Retrying req: %d. Error: %v\n", counter, err)
+		errString := err.Error()
+		// Invalid certificate, irrecoverable
+		if strings.Contains(errString, "x509") {
+			log.Fatal(errString)
+		}
+		fmt.Printf("Retrying req: %d. Error: %v\n", counter, errString)
 		time.Sleep(5 * time.Millisecond)
 		goto RETRY
 	}
--- a/client/client.go
+++ b/client/client.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 
 	"github.com/dgraph-io/dgraph/query/graph"
 )
@@ -174,8 +175,8 @@ RETRY:
 	_, err := batch.dc.Run(context.Background(), &req.gr)
 	if err != nil {
 		errString := err.Error()
-		// Invalid certificate, irrecoverable
-		if strings.Contains(errString, "x509") {
+		// Irrecoverable
+		if strings.Contains(errString, "x509") || grpc.Code(err) == codes.Internal {
 			log.Fatal(errString)
 		}
 		fmt.Printf("Retrying req: %d. Error: %v\n", counter, errString)
--- a/cmd/dgraph/main.go
+++ b/cmd/dgraph/main.go
@@ -82,11 +82,11 @@ var (
 	tlsCert          = flag.String("tls.cert", "", "Certificate file path.")
 	tlsKey           = flag.String("tls.cert_key", "", "Certificate key file path.")
 	tlsKeyPass       = flag.String("tls.cert_key_passphrase", "", "Certificate key passphrase.")
-	tlsClientAuth    = flag.String("tls.client_auth", "", "Enable TLS client auth")
-	tlsClientCACerts = flag.String("tls.client_ca_certs", "", "Client CA Certs file path.")
-	tlsSystemCACerts = flag.Bool("tls.use_system_ca", false, "Include System CA into Clients CA Certs.")
-	tlsMinVersion    = flag.String("tls.min_version", "TLS11", "TLS min versions.")
-	tlsMaxVersion    = flag.String("tls.max_version", "TLS12", "TLS max versions.")
+	tlsClientAuth    = flag.String("tls.client_auth", "", "Enable TLS client authentication")
+	tlsClientCACerts = flag.String("tls.ca_certs", "", "CA Certs file path.")
+	tlsSystemCACerts = flag.Bool("tls.use_system_ca", false, "Include System CA into CA Certs.")
+	tlsMinVersion    = flag.String("tls.min_version", "TLS11", "TLS min version.")
+	tlsMaxVersion    = flag.String("tls.max_version", "TLS12", "TLS max version.")
 )
 
 type mutationResult struct {
--- a/cmd/dgraph/main.go
+++ b/cmd/dgraph/main.go
@@ -67,7 +67,7 @@ var (
 	port       = flag.Int("port", 8080, "Port to run server on.")
 	bindall    = flag.Bool("bindall", false,
 		"Use 0.0.0.0 instead of localhost to bind to all addresses on local machine.")
-	tlsEnabled     = flag.Bool("tls", true, "Use TLS connections with clients")
+	tlsEnabled     = flag.Bool("tls", false, "Use TLS connections with clients")
 	cert           = flag.String("cert", "", "Certificate file path")
 	key            = flag.String("cert_key", "", "Certificate key file key")
 	nomutations    = flag.Bool("nomutations", false, "Don't allow mutations on this server.")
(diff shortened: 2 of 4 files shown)


tzdybal:
Thanks for submission @Motakjuq.

I think that for production environment more options are required:

- server name
- CA certificates paths

Optionally, following configuration could be useful:

- password for encrypted keys
- [client authentication](https://golang.org/pkg/crypto/tls/#ClientAuthType) 
- protocol version (currently it's hardcoded)
- cipher suites
- curves preferences

Motakjuq:
Hi @tzdybal, I can make some changes but I need some feedback.

- CA: I wasn't sure to add it, but it could be useful, so I'm gonna add a "ca" param to the server.
- Password for encrypted key: I'm not a fan of this, but I'm gonna add a "cert_key_password" param to the server.
- Client auth: This is other topic, I think it should be added with the user authentication feature.

Feedback required:
- Server name: I can add this param into the client. I don't think it is useful for the server, because AFAIK server_name is used it only for the client, unless the server supports virtual host, but it's not the case. Correct me if I'm wrong.
- Protocol version: Currently support tls1.1 **and** tls1.2, if you need to define a specific version, it could be added a "tls_version" param which accept tls1.1 **and/or** tls1.2. I do not think it's a good idea include support for tls1.0 and ssl3. Do I add this param?, should it need to support tls1.0 and ssl3?
- Cipher suites and curves preferences: It could be added but it's too specific, if the server needs it, I can add them, but I'm going to need a list of which are going to be supported. Should I add this configuration? and, Should it be configurable or hard-coded?

manishrjain:
Thanks for doing this @motakjuq! I don't know much about TLS, so I'll leave it to @tzdybal to guide you through this. But, this functionality has been on top of my mind for a while. Glad that you're working on it.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/593#-:-KdBbf61x5WVmEr0HVqw:bu33jj)*

tzdybal:
@Motakjuq
- CA certs configuration is definitely very important. I think, that single PEM file with list/chain of CA certs is a reasonable way of providing CA certs - and it's easy to implement using [x509.CertPool](https://golang.org/pkg/crypto/x509/#CertPool.AppendCertsFromPEM). Another option is to configure CA cert path, where certs are located in separate files, but I think it's more complicated solution. In my opinion, `ca_certs` is a more precise name for this configuration option.
- I totally agree that storing a private key password in the configuration file is by no means more secure than storing unencrypted private key. It's all about file access control. In my experience, encrypted private keys are very popular. And if passwords are supported, both scenarios are possible (because `cert_key_pass` shouldn't be required).
- Client authentication may also be used only for mutual authentication on presentation layer (TLS). Later this may be used for application layer user authentication.
- Server name is also used for server verification (server name has to match the certificate). If `insecure` is set to false, server name is required (see [tls.Client](https://golang.org/pkg/crypto/tls/#Client)).
- Protocol version (min and max), cypher suites and curves - I see two use cases for this options. 1 - hardening, 2 - usage of clients written in different technologies which may support only a limited subset of TLS. Those options should directly correspond to `crypto/tls` configuration options for consistency (e.g. `TLS12` for protocol version, `TLS_RSA_WITH_AES_128_CBC_SHA256` for cypher suite, `X25519` for curve). IMHO support for SSL and TLS 1.0 shouldn't be implemented.

To summarize:
 -  `cert`, `cert_key`, `ca_certs`, `server_name` (client only) are necessary for proper server certificate validation which is required for production environment. 
 - private key password, protocol version, cypher suites, curves and client authentication are not necessary - feel free to implement as many as you like :wink:

tzdybal:
There is a lot of TLS related configuration options, so you may also think about grouping them as YAML structure, like:

    tls:
      ca_certs:  ca.pem
      cert:      server.crt
      key:       server.key
      min_proto: TLS12

Motakjuq:
I updated the params used by the client and server, now the following features are supported

Server:
- TLS connection (tls.on: default false)
- Use of certificates with and without encrypted key (tls.cert: default empty ; tls.cert_key: default empty ; tls.cert_key_passphrase: default empty)
- Client Auth using client certificate using value [request | requireany | verifyifgiven | requireandverify] (tls.client_auth default empty)
- Configuration of CA for client auth, also it can be used the CAs stored in the server (tls.client_ca_certs: default empty ; tls.use_system_ca: default false)
- Configurable Min and Max TLS version, currently the supported version are tls11 and tls12. (tls.min_version: default "TLS11" ; tls.max_version: default "TLS12")

Client:
- TLS connection (tls.on: default false)
- Insecure (tls.insecure: default false)
- Server Name if insecure is false (tls.server_name: default empty)
- Use of certificates with and without encrypted key (only used if server auth is enabled) (tls.cert: default empty ; tls.cert_key: default empty ; tls.cert_key_passphrase: default empty)
- Configuration of CA certs (ca uses to validate the server), also it can be used the CAs stored in the client (tls.root_ca_certs: default empty ; tls.use_system_ca: default false)
- Configurable Min and Max TLS version, currently the supported version are tls11 and tls12. (tls.min_version: default "TLS11" ; tls.max_version: default "TLS12")

tzdybal:
Nice job! Thanks for submission. I have only minor comments.

---

Reviewed 4 of 4 files at r2.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[cmd/dgraph/main.go, line 85 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/593#-KdUfsXrT3bmw4YOe2bZ:-KdUfsXtJFeDwdWwfDP8:bay9367) ([raw file](https://github.com/dgraph-io/dgraph/blob/38435187c2c7a72c2f7a8b578ad175ac580e50fe/cmd/dgraph/main.go#L85)):*
> ```Go
> 	tlsKey           = flag.String("tls.cert_key", "", "Certificate key file path.")
> 	tlsKeyPass       = flag.String("tls.cert_key_passphrase", "", "Certificate key passphrase.")
> 	tlsClientAuth    = flag.String("tls.client_auth", "", "Enable TLS client auth")
> ```

I think that "authentication" instead of "auth" should be used in option description.

---

*[cmd/dgraph/main.go, line 86 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/593#-KdUjQSwQd9F7FVJflzB:-KdUjQSximsQz6jpxPJE:b-ip5b6f) ([raw file](https://github.com/dgraph-io/dgraph/blob/38435187c2c7a72c2f7a8b578ad175ac580e50fe/cmd/dgraph/main.go#L86)):*
> client_ca_certs
 
Maybe just "tls.ca_certs" for simplicity?

---

*[cmd/dgraph/main.go, line 88 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/593#-KdUge1-hzRoPz-VUBTy:-KdUge10fkWnda6zGr2j:bx6j7xc) ([raw file](https://github.com/dgraph-io/dgraph/blob/38435187c2c7a72c2f7a8b578ad175ac580e50fe/cmd/dgraph/main.go#L88)):*
> ```Go
> 	tlsClientCACerts = flag.String("tls.client_ca_certs", "", "Client CA Certs file path.")
> 	tlsSystemCACerts = flag.Bool("tls.use_system_ca", false, "Include System CA into Clients CA Certs.")
> 	tlsMinVersion    = flag.String("tls.min_version", "TLS11", "TLS min versions.")
> ```

Why version**s**, not version?

---

*[cmd/dgraphloader/main.go, line 36 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/593#-KdUindasJnO0EKVvAcT:-KdUindbui7XaeraGaR5:by99uxc) ([raw file](https://github.com/dgraph-io/dgraph/blob/38435187c2c7a72c2f7a8b578ad175ac580e50fe/cmd/dgraphloader/main.go#L36)):*
> Skip insecure cert validation

I would rather say: Skip certificate validation (insecure)

---

*[cmd/dgraphloader/main.go, line 41 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/593#-KdUjrI2vzNKuxUHPxf0:-KdUjrI3JcsbzDkNgtGl:b391sfl) ([raw file](https://github.com/dgraph-io/dgraph/blob/38435187c2c7a72c2f7a8b578ad175ac580e50fe/cmd/dgraphloader/main.go#L41)):*
> tls.root_ca_certs

Maybe just "tls.ca_certs" for simplicity?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/593#-:-KdUoG7EEsgGup2qIk5d:b-ff22vm)*

Motakjuq:
All changes were applied

tzdybal:
:lgtm:

---

Reviewed 2 of 2 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/593#-:-Kdeg6QjrWTNv_CMpoDK:bnfp4nl)*

PR #638, opened 2017-02-27
Facets/todos

ashishnegi:
1. Optimize filtering based on facets : pre process facets filter tree to calculate tokens and facets only once per node.
2. Store facets data in binary form.
3. Refactor task.go : move big switch into another fn.
4.  Store tokens in facets. No need to tokenize again.

Todo Problem :

Since we store tokens in facets itself, to create a facet in `FacetFor`, i need `tok` package.
`tok` needs `types` package.
`types` need `facets` package for `types.proto` having `facets` in PL.
leading to cyclic dependency.
~~I have currenly created `utils/facets.go` in `root dgraph`.~~
Better package management tips required.

Other way is to move `tokenize` fn out of `tok` into separate `package` as it only requires `string` rather than `types.Val`.
Then use same package in `tok` and `facets`.

I have currently created directory `facets/utils` rather than at root folder.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/638)


ashishnegi:
Review status: 0 of 12 files reviewed at latest revision, 1 unresolved discussion.

---

*[utils/facets.go, line 1 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-KdxrfnQkVJOvqekQ5vS:-KdxrfnQkVJOvqekQ5vT:bqg64ox) ([raw file](https://github.com/dgraph-io/dgraph/blob/f2b05129196d4ecb748d809c8bcf1ff29f784845/utils/facets.go#L1)):*
> ```Go
> package utils
> ```

Problem :
Since we store tokens in facets itself, to create a facet in FacetFor, i need tok package.
tok needs types package.
types need facets package for types.proto having facets in PL.
leading to cyclic dependency.
I have currenly created utils/facets.go in root dgraph.
Better package management tips required.

Other way is to move tokenize fn out of tok into separate package as it only requires string rather than types.Val.
Then use same package in tok and facets.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/638)*

gacharya255:
LG for FacetsFor to convert to binary.
LG for preprocessFilter and applyFacetsTree.
LG for parseQuery refactoring.
Regarding the dependencies, I would recommend moving the protos out, but let's wait to hear from Manish.

---

Review status: 0 of 12 files reviewed at latest revision, 4 unresolved discussions.

---

*[utils/facets.go, line 1 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-KdxrfnQkVJOvqekQ5vS:-Ke7NmImKXYFf1yKRaW2:b-m907y9) ([raw file](https://github.com/dgraph-io/dgraph/blob/f2b05129196d4ecb748d809c8bcf1ff29f784845/utils/facets.go#L1)):*
<details><summary><i>Previously, ashishnegi (Ashish Negi) wrote…</i></summary><blockquote>

Problem :
Since we store tokens in facets itself, to create a facet in FacetFor, i need tok package.
tok needs types package.
types need facets package for types.proto having facets in PL.
leading to cyclic dependency.
I have currenly created utils/facets.go in root dgraph.
Better package management tips required.

Other way is to move tokenize fn out of tok into separate package as it only requires string rather than types.Val.
Then use same package in tok and facets.
</blockquote></details>

I would think types is more basic than facets. If possible, types.proto should not use facets.proto and we can try to move Posting from types.proto to facets.proto. Also I think it is good for protos to be in a subdirectory on their own so that they are decoupled from our code.

---

*[worker/task.go, line 348 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-Ke7RwkGhK8s9_u9NW7d:-Ke7RwkHwIik13h80PI6:bme0h6n) ([raw file](https://github.com/dgraph-io/dgraph/blob/cfd23c5ef24cbd043f5e6d555eb7e9cf5049b723/worker/task.go#L348)):*
> ```Go
> 	tokens         []string
> 	geoQuery       *types.GeoQueryData
> 	err            error
> ```

I don't see res.err being used?

---

*[worker/task.go, line 358 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-Ke7RbO-aS0eB5lVB9ew:-Ke7RbO-aS0eB5lVB9ex:bh5ih0a) ([raw file](https://github.com/dgraph-io/dgraph/blob/cfd23c5ef24cbd043f5e6d555eb7e9cf5049b723/worker/task.go#L358)):*
> ```Go
> }
> 
> func parseQuery(q *task.Query) (*resultParseQuery, error) {
> ```

Side comment: It is a bit strange that we have to parse the query after it is supposed to be parsed. But I see why you want to move this out. It does look much cleaner. Maybe this is "getQueryContext".

---

*[worker/task.go, line 633 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-Ke7Qfv3FrfRaUAYikII:-Ke7Qfv3FrfRaUAYikIJ:b-8ouws1) ([raw file](https://github.com/dgraph-io/dgraph/blob/cfd23c5ef24cbd043f5e6d555eb7e9cf5049b723/worker/task.go#L633)):*
> ```Go
> }
> 
> func preProcessFilter(tree *facets.FilterTree) (*facetsTree, error) {
> ```

Maybe preprocessFilter instead of preProcessFilter?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-:-Ke7SxzdKl8RhgJB2HXY:b737w8x)*

ashishnegi:
Review status: 0 of 12 files reviewed at latest revision, 4 unresolved discussions.

---

*[utils/facets.go, line 1 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-KdxrfnQkVJOvqekQ5vS:-Ke8IZ9K36iNnIiGEQOM:bqxq0zc) ([raw file](https://github.com/dgraph-io/dgraph/blob/f2b05129196d4ecb748d809c8bcf1ff29f784845/utils/facets.go#L1)):*

I would think types is more basic than facets. If possible, types.proto should not use facets.proto and we can try to move Posting from types.proto to facets.proto. Also I think it is good for protos to be in a subdirectory on their own so that they are decoupled from our code.
</blockquote></details>

actually, facets is more basic.
i am saying this because , types proto contains posting  and postinglist only. Posting has facets. so, facet is more basic. 
Moving posting out will solve this as postinglist uses posting.
Also, currently `facets.proto` is also in a subdirectory under types `types/facets`.

`types -> facets` we can not change this.
`tok -> types`  : the tok is designed around types ; so not possible.
`facets -> tok` because of tokenizer. (The part we need is actually an algorithm that can take string, so i can move that algorithm to separate package.)

so, to break `tok -> types -> facets -> tok` chain, we have to 
1. either have `tok -> types -> facets` and `utils ->facets` and `utils->tok`; where `FacetFor` is in utils.  Which is current implementation.
2. OR `tok -> types -> facets` and `tok -> utils` and `facets -> utils` where algo of tokenizer is in `utils`.

---

*[worker/task.go, line 348 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-Ke7RwkGhK8s9_u9NW7d:-Ke8HgaCFK29PxMot6r8:b-ev7von) ([raw file](https://github.com/dgraph-io/dgraph/blob/cfd23c5ef24cbd043f5e6d555eb7e9cf5049b723/worker/task.go#L348)):*

I don't see res.err being used?
</blockquote></details>

Done. removed.. 

---

*[worker/task.go, line 358 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-Ke7RbO-aS0eB5lVB9ew:-Ke8I7Vu3Hs2eh0DKLBb:b-6vl1mk) ([raw file](https://github.com/dgraph-io/dgraph/blob/cfd23c5ef24cbd043f5e6d555eb7e9cf5049b723/worker/task.go#L358)):*

Side comment: It is a bit strange that we have to parse the query after it is supposed to be parsed. But I see why you want to move this out. It does look much cleaner. Maybe this is "getQueryContext".
</blockquote></details>

we are mostly working over the `query.SrcFunc`in this fn .. so now calling it `getSrcFnContext`

---

*[worker/task.go, line 633 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-Ke7Qfv3FrfRaUAYikII:-Ke8IJJcud8OZki5y7H3:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/cfd23c5ef24cbd043f5e6d555eb7e9cf5049b723/worker/task.go#L633)):*

Maybe preprocessFilter instead of preProcessFilter?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/638)*

ashishnegi:
Review status: 0 of 14 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[utils/facets.go, line 1 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-KdxrfnQkVJOvqekQ5vS:-Ke8bkAjWfwVq6iF-e9w:b-rw6sr9) ([raw file](https://github.com/dgraph-io/dgraph/blob/f2b05129196d4ecb748d809c8bcf1ff29f784845/utils/facets.go#L1)):*
<details><summary><i>Previously, ashishnegi (Ashish Negi) wrote…</i></summary><blockquote>

actually, facets is more basic.
i am saying this because , types proto contains posting  and postinglist only. Posting has facets. so, facet is more basic. 
Moving posting out will solve this as postinglist uses posting.
Also, currently `facets.proto` is also in a subdirectory under types `types/facets`.

`types -> facets` we can not change this.
`tok -> types`  : the tok is designed around types ; so not possible.
`facets -> tok` because of tokenizer. (The part we need is actually an algorithm that can take string, so i can move that algorithm to separate package.)

so, to break `tok -> types -> facets -> tok` chain, we have to 
1. either have `tok -> types -> facets` and `utils ->facets` and `utils->tok`; where `FacetFor` is in utils.  Which is current implementation.
2. OR `tok -> types -> facets` and `tok -> utils` and `facets -> utils` where algo of tokenizer is in `utils`.
</blockquote></details>

Now, i have created `types/facets/utils` directory. This is better than creating root level folder.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/638)*

gacharya255:
Review status: 0 of 14 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[utils/facets.go, line 1 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-KdxrfnQkVJOvqekQ5vS:-Ke9GvHW6Z9q3s6Ymu-n:bsu6vv4) ([raw file](https://github.com/dgraph-io/dgraph/blob/f2b05129196d4ecb748d809c8bcf1ff29f784845/utils/facets.go#L1)):*
<details><summary><i>Previously, ashishnegi (Ashish Negi) wrote…</i></summary><blockquote>

Now, i have created `types/facets/utils` directory. This is better than creating root level folder.
</blockquote></details>

Thanks for putting a lot of thought into this. When I thought of "types" I was thinking of the types logic instead of the proto. Unfortunately, types also contains the posting list proto which is a rather high level proto. I think we could just move all these protos into one directory on its own, so posting, facets protos all go into one directory, one package. Then both the current types and facets packages can depend on this proto package.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/638)*

gacharya255:
Review status: 0 of 14 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[utils/facets.go, line 1 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-KdxrfnQkVJOvqekQ5vS:-Ke9JPxlUaCuX8vPOt0b:b-2j6c5b) ([raw file](https://github.com/dgraph-io/dgraph/blob/f2b05129196d4ecb748d809c8bcf1ff29f784845/utils/facets.go#L1)):*

Thanks for putting a lot of thought into this. When I thought of "types" I was thinking of the types logic instead of the proto. Unfortunately, types also contains the posting list proto which is a rather high level proto. I think we could just move all these protos into one directory on its own, so posting, facets protos all go into one directory, one package. Then both the current types and facets packages can depend on this proto package.
</blockquote></details>

To elaborate a bit, I feel that the "types" package should just handle very basic scalar type operations, like conversions between the different data types. Token should depend on types because it is more high level. It has to know "types" before it can do any tokenizing. Facets should depend on types and also token because it is even more high level. It requires tokenization of some data.

Now, posting lists is a very high level data structure. But it shouldn't have to cause types package (a very low level package) to depend on higher level packages. Neither should it cause some arm-twisting, and movement of actual logic around. It is just a proto. All the protos should not depend on anything other than other protos. I think they should all go into just one or two packages. I don't see a need to split them up into many files or packages.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/638)*

ashishnegi:
Review status: 0 of 14 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[utils/facets.go, line 1 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-KdxrfnQkVJOvqekQ5vS:-KeCF72aIGBUc0nr0agO:b34bve8) ([raw file](https://github.com/dgraph-io/dgraph/blob/f2b05129196d4ecb748d809c8bcf1ff29f784845/utils/facets.go#L1)):*

To elaborate a bit, I feel that the "types" package should just handle very basic scalar type operations, like conversions between the different data types. Token should depend on types because it is more high level. It has to know "types" before it can do any tokenizing. Facets should depend on types and also token because it is even more high level. It requires tokenization of some data.

Now, posting lists is a very high level data structure. But it shouldn't have to cause types package (a very low level package) to depend on higher level packages. Neither should it cause some arm-twisting, and movement of actual logic around. It is just a proto. All the protos should not depend on anything other than other protos. I think they should all go into just one or two packages. I don't see a need to split them up into many files or packages.
</blockquote></details>

Thanks this sounds like a good solution.
I will add a task for this. This PR looks big enough already.
Lets have @mrjn Manish comment on this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/638)*

manishrjain:
Yeah, moving the protos out sounds like the right thing to do.

---

Review status: 0 of 14 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[tok/tokens.go, line 9 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-KeXgzwzD1d9ktwmFPPU:-KeXgzwzD1d9ktwmFPPV:b-eca34l) ([raw file](https://github.com/dgraph-io/dgraph/blob/c0a7393cce79fbbbd425dd2702914df41c73c500/tok/tokens.go#L9)):*
> ```Go
> 
> //  Might want to allow user to replace this.
> var termTokenizer TermTokenizer
> ```

Tomasz has changed this, so ICU tokenizer is no longer being used. Might make sense to sync his changes, and base this upon them.

---

*[types/facets/facet_types.go, line 68 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-KeXhOBCQp0MRryyTJZo:-KeXhOBCQp0MRryyTJZp:b-fcibp8) ([raw file](https://github.com/dgraph-io/dgraph/blob/c0a7393cce79fbbbd425dd2702914df41c73c500/types/facets/facet_types.go#L68)):*
> ```Go
> 	a.Facets[i], a.Facets[j] = a.Facets[j], a.Facets[i]
> }
> func (a Facets) Less(i, j int) bool {
> ```

Can use the new sort.Slice potentially.

---

*[types/facets/utils/utils.go, line 103 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-KeXiZ5QqTSWdHntE1UM:-KeXiZ5QqTSWdHntE1UN:bfh5er) ([raw file](https://github.com/dgraph-io/dgraph/blob/c0a7393cce79fbbbd425dd2702914df41c73c500/types/facets/utils/utils.go#L103)):*
> ```Go
> 		if (a[i].Key != b[i].Key) ||
> 			!bytes.Equal(a[i].Value, b[i].Value) ||
> 			(a[i].ValType != b[i].ValType) {
> ```

check valtype first, that's a cheaper comparison; then key; then value.

---

*[types/facets/utils/utils.go, line 111 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-KeXiNtmwx5DyjPVbttz:-KeXiNtmwx5DyjPVbtu-:biimhut) ([raw file](https://github.com/dgraph-io/dgraph/blob/c0a7393cce79fbbbd425dd2702914df41c73c500/types/facets/utils/utils.go#L111)):*
> ```Go
> 
> // valAndValType returns interface val and valtype for facet.
> func valAndValType(val string) (interface{}, facets.Facet_ValType, error) {
> ```

Isn't there a struct you could use for the result? Wrapping interface{} and Facet_ValType?

---

*[types/facets/utils/utils.go, line 111 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-KeXiTZjxHB_L5LZWmM-:-KeXiTZjxHB_L5LZWmM0:bp6a1ze) ([raw file](https://github.com/dgraph-io/dgraph/blob/c0a7393cce79fbbbd425dd2702914df41c73c500/types/facets/utils/utils.go#L111)):*
> ```Go
> 
> // valAndValType returns interface val and valtype for facet.
> func valAndValType(val string) (interface{}, facets.Facet_ValType, error) {
> ```

facetValType(..)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-:-KeXgpb8QR3FUgfOVaY1:bsnl5z6)*

manishrjain:
Reviewed 7 of 12 files at r1, 8 of 8 files at r3.
Review status: all files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[types/facets/utils/utils.go, line 66 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-Keb-BLwRoht1FEnEFK-:-Keb-BLwRoht1FEnEFK0:b-vj3e9c) ([raw file](https://github.com/dgraph-io/dgraph/blob/c0a7393cce79fbbbd425dd2702914df41c73c500/types/facets/utils/utils.go#L66)):*
> ```Go
> // FacetFor returns Facet for given key and val.
> func FacetFor(key, val string) (*facets.Facet, error) {
> 	v, vt, err := valAndValType(val)
> ```

move valandvaltype above this func.

---

*[types/facets/utils/utils.go, line 112 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-Keb-LaH8WsJhaZlgwYN:-Keb-LaH8WsJhaZlgwYO:b-mqyiji) ([raw file](https://github.com/dgraph-io/dgraph/blob/c0a7393cce79fbbbd425dd2702914df41c73c500/types/facets/utils/utils.go#L112)):*
> ```Go
> // valAndValType returns interface val and valtype for facet.
> func valAndValType(val string) (interface{}, facets.Facet_ValType, error) {
> 	if fint, err := strconv.ParseInt(val, 10, 32); err == nil {
> ```

intVal, err := . Why 10? Could be hex encoded.

---

*[types/facets/utils/utils.go, line 114 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-Keb-P9P5HVrIp70LKUL:-Keb-P9P5HVrIp70LKUM:b4a513b) ([raw file](https://github.com/dgraph-io/dgraph/blob/c0a7393cce79fbbbd425dd2702914df41c73c500/types/facets/utils/utils.go#L114)):*
> ```Go
> 	if fint, err := strconv.ParseInt(val, 10, 32); err == nil {
> 		return int32(fint), facets.Facet_INT32, nil
> 	} else if nume := err.(*strconv.NumError); nume.Err == strconv.ErrRange {
> ```

numErr, numErr == strconv.ErrRange?

---

*[types/facets/utils/utils.go, line 128 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-Keb-lUTYvRJX5167KHG:-Keb-lUTYvRJX5167KHH:b-d0liyt) ([raw file](https://github.com/dgraph-io/dgraph/blob/c0a7393cce79fbbbd425dd2702914df41c73c500/types/facets/utils/utils.go#L128)):*
> ```Go
> 		}
> 	}
> 	if ffloat, err := strconv.ParseFloat(val, 64); err == nil {
> ```

floatVal; and others similarly.

---

*[worker/task.go, line 135 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-Keb0DEFrX2XB5A58t_g:-Keb0DEFrX2XB5A58t_h:b-dyfdto) ([raw file](https://github.com/dgraph-io/dgraph/blob/c0a7393cce79fbbbd425dd2702914df41c73c500/worker/task.go#L135)):*
> ```Go
> func processTask(q *task.Query, gid uint32) (*task.Result, error) {
> 	attr := q.Attr
> 	resParseFn, err := getSrcFnContext(q)
> ```

src, err := parseSrcFunc

---

*[worker/task.go, line 142 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-Keb0Ti3IJWaCvE8tUFJ:-Keb0Ti4E3nShjOOGDaP:b-kh7xod) ([raw file](https://github.com/dgraph-io/dgraph/blob/c0a7393cce79fbbbd425dd2702914df41c73c500/worker/task.go#L142)):*
> ```Go
> 	fnType, f := resParseFn.fnType, resParseFn.fname
> 	tokens := resParseFn.tokens
> 	geoQuery := resParseFn.geoQuery
> ```

Don't do these referencing. Use the src.Variable directly.

---

*[worker/task.go, line 357 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-Keb0s86lUDi-9Bu_Cbv:-Keb0s86lUDi-9Bu_Cbw:bz3oyze) ([raw file](https://github.com/dgraph-io/dgraph/blob/c0a7393cce79fbbbd425dd2702914df41c73c500/worker/task.go#L357)):*
> ```Go
> }
> 
> func getSrcFnContext(q *task.Query) (*functionContext, error) {
> ```

parseSrcFunc

---

*[worker/task.go, line 360 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-Keb0zmNmqzoVDixBd2L:-Keb0zmNmqzoVDixBd2M:b4uvidu) ([raw file](https://github.com/dgraph-io/dgraph/blob/c0a7393cce79fbbbd425dd2702914df41c73c500/worker/task.go#L360)):*
> ```Go
> 	fnType, f := parseFuncType(q.SrcFunc)
> 	attr := q.Attr
> 	res := &functionContext{fnType: fnType, fname: f}
> ```

fc := &functionContext{..}

---

*[worker/task.go, line 675 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-Keb1kGTQdtIWDgFFH0I:-Keb1kGTQdtIWDgFFH0J:b-647dkr) ([raw file](https://github.com/dgraph-io/dgraph/blob/c0a7393cce79fbbbd425dd2702914df41c73c500/worker/task.go#L675)):*
> ```Go
> 	numChild := len(tree.Children)
> 	switch strings.ToLower(tree.Op) {
> 	case "not":
> ```

Add a comment that we have already checked for ... 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/638)*

manishrjain:
:lgtm:

---

Review status: all files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/638#-:-Keb2mdC5ySgMeNH2hIe:bnfp4nl)*

PR #639, opened 2017-02-27
Refactor/ui

pawanrawal:
This PR doesn't add much new functionality but is a major refactoring which improves how the code is structured. Earlier most of it is was in one file. It has been broken into different components and hence different files which is more in the way of how things should be in React.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/639)


pawanrawal:
Merged

PR #666, opened 2017-03-06
Add option to filter/delete previous queries and improve properties table layout.

pawanrawal:
1. Previous queries can now be filtered using a regex match. There is also a delete button which allows deletion of the queries.

2. Properties table now uses Flex and the number of columns in it change dynamically based on screen width.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/666)


tzdybal:
:lgtm:

---

Reviewed 16 of 16 files at r1.

---

*[dashboard/src/components/App.js, line 437 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/666#-KeZXpXQXre05u6vl_P9:-KeZXpXRMo7d7zL6GfTv:bcc12bb) ([raw file](https://github.com/dgraph-io/dgraph/blob/832856efb04191d34188313ab3c6123d67467656/dashboard/src/components/App.js#L437)):*
> ```JavaScript
>     }
> 
>     // TODO - Abstract out, get, put delete so that state is updated both for react
> ```

I totally agree with the TODO ;)

I guess that this is bigger change, not in scope of current work?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/666#-:-KeZYuCuXJ2TjEk304L4:bnfp4nl)*

pawanrawal:

---

*[dashboard/src/components/App.js, line 437 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/666#-KeZXpXQXre05u6vl_P9:-KeaJUqkjwm7Yk8_TDeK:bp0lmri) ([raw file](https://github.com/dgraph-io/dgraph/blob/832856efb04191d34188313ab3c6123d67467656/dashboard/src/components/App.js#L437)):*

I totally agree with the TODO ;)

I guess that this is bigger change, not in scope of current work?
</blockquote></details>

Yes, I guess React Redux (https://github.com/reactjs/react-redux) might give it to us out of the box. This would need a refactoring.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/666)*

pawanrawal:
Have merged it to master, but will wait for your comments @sungwoncho and fix anything you think needs fixing.

dashboard/src/components/Navbar.js
```diff
@@ -13,10 +13,13 @@ class NavBar extends Component {
...
                         </a>
                     </Navbar.Brand>
+                    <Navbar.Toggle />
                 </Navbar.Header>
-                <Nav>
-                    <NavItem eventKey={1} href="#">Visualization</NavItem>
-                </Nav>
+                <Navbar.Collapse>
+                    <Nav>
+                        <NavItem eventKey={1} href="#">Visualization</NavItem>
```
sungwoncho: This link doesn't seem to do anything. Do we need it?

dashboard/src/components/PreviousQueryList.js
```diff
@@ -0,0 +1,70 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      filterText: "",
+    };
+  }
+
+  filterQueries = function() {
+    var that = this;
+    return this.props.queries.filter(function(item) {
```
sungwoncho: if we use fat arrow function in the callback, can we stop using `var that = this`?

dashboard/src/components/PreviousQueryList.js
```diff
@@ -0,0 +1,70 @@
...
+                    update={this.props.update}
+                    delete={this.props.delete}
+                    key={i}
+                    idx={i}
+                    lastRun={query.lastRun}
+                    unique={i}
+                  />
+                );
+              },
+              this,
```
sungwoncho: maybe we could make this more readable by using fat arrow for the callback for `map`. Then we won't have to supply `this` context here.

dashboard/src/components/Properties.js
```diff
@@ -4,70 +4,49 @@ import React, { Component } from "react";
...
+                {!isEdge &&
+                    <div className="Properties">
+                        {Object.keys(attrs).map(function(key, idx) {
+                            return (
+                                <div className="Properties-key-val" key={idx}>
+                                    <div className="Properties-key">
+                                        {key}:&nbsp;
+                                    </div>
+                                    <div className="Properties-val">
+                                        {String(attrs[key])}
```
sungwoncho: Is `String` necessary?

dashboard/src/index.js
```diff
@@ -1,11 +1,15 @@
...
 
-ReactDOM.render(
-  <App />,
-  document.getElementById('root')
-);
+ReactDOM.render(<App />, document.getElementById("root"));
+
+if (module.hot) {
+    module.hot.accept("./components/App", () => {
+        const NextApp = require("./components/App").default;
```
sungwoncho: any reason we use `require` here, rather than `import`ing it at the top?

sungwoncho:
> Properties table now uses Flex and the number of columns in it change dynamically based on screen width.

Can you explain this? Is this referring to responsive grid system?

pawanrawal:
No, its not for the reponsive Grid system. Its for the Properties component which is a table that needs to adjust the number of columns according to the amount of content available.

---

Review status: 14 of 18 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[dashboard/src/index.js, line 12 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/666#-KebbWeBwry1lHDTaJzN-r2-12:-KepjnOoj1fD56UHqXcw:b-k8nsn7) ([raw file](https://github.com/dgraph-io/dgraph/blob/8f39ce973a1b99171440a2f9d818acfb4ef35f69/dashboard/src/index.js#L12)):*
<details><summary><i>Previously, sungwoncho (Sung Won Cho) wrote…</i></summary><blockquote>

any reason we use `require` here, rather than `import`ing it at the top?
</blockquote></details>

Hey, this is for Hot Module Replacement (for JS files, CSS is already included in Create React App) on the fly. So if any component is updated, the whole page shouldn't be refreshed and only outdated parts should be replaced. Helps in development.

So I am requiring the new contents here. Also, its not possible to use `import` anywhere apart from the top of the file.

Reference - https://github.com/facebookincubator/create-react-app/issues/93.

---

*[dashboard/src/components/Navbar.js, line 20 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/666#-KeX2K-AGlQ40N3MR7hl-r2-20:-KepkI3wd3miVkMRbYCE:b-a5zfp4) ([raw file](https://github.com/dgraph-io/dgraph/blob/8f39ce973a1b99171440a2f9d818acfb4ef35f69/dashboard/src/components/Navbar.js#L20)):*
<details><summary><i>Previously, sungwoncho (Sung Won Cho) wrote…</i></summary><blockquote>

This link doesn't seem to do anything. Do we need it?
</blockquote></details>

Could get rid of it for now, would become relevant when we have different routes , `/admin` etc.

---

*[dashboard/src/components/PreviousQueryList.js, line 19 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/666#-KeX2K-AGlQ40N3MR7hq-r2-19:-Kepm43ig8rniYmbY9Ck:b19ya0b) ([raw file](https://github.com/dgraph-io/dgraph/blob/8f39ce973a1b99171440a2f9d818acfb4ef35f69/dashboard/src/components/PreviousQueryList.js#L19)):*
<details><summary><i>Previously, sungwoncho (Sung Won Cho) wrote…</i></summary><blockquote>

if we use fat arrow function in the callback, can we stop using `var that = this`?
</blockquote></details>

Yes, got rid of it in the new PR.

---

*[dashboard/src/components/PreviousQueryList.js, line 61 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/666#-KeX2K-AGlQ40N3MR7hq-r2-61:-Kepp1vtFCubIF-ke11E:bfs27t2) ([raw file](https://github.com/dgraph-io/dgraph/blob/8f39ce973a1b99171440a2f9d818acfb4ef35f69/dashboard/src/components/PreviousQueryList.js#L61)):*
<details><summary><i>Previously, sungwoncho (Sung Won Cho) wrote…</i></summary><blockquote>

maybe we could make this more readable by using fat arrow for the callback for `map`. Then we won't have to supply `this` context here.
</blockquote></details>

fixed it in the new PR.

---

*[dashboard/src/components/Properties.js, line 27 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/666#-KeXd7vYxfR3kX7qaAA0-r2-27:-KepmQb_81nlOaUS1bhF:bvy5uqe) ([raw file](https://github.com/dgraph-io/dgraph/blob/8f39ce973a1b99171440a2f9d818acfb4ef35f69/dashboard/src/components/Properties.js#L27)):*
<details><summary><i>Previously, sungwoncho (Sung Won Cho) wrote…</i></summary><blockquote>

Is `String` necessary?
</blockquote></details>

Yeah, because we can get boolean values back too and they are not rendered.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/666#-:-Kepp6cXnWe9Wyh3Et1D:b-t02gtc)*

PR #679, opened 2017-03-08
WIP: Improvement/raft

janardhan1993:
Ignore heartbeat not meant for that node.
If a new node starts at same port and address with different id, remove older node.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/679)


manishrjain:
Reviewed 4 of 4 files at r1.

---

*[worker/draft.go, line 468 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/679#-KelMTBPKa8DwaNMgd5u:-KelMTBPKa8DwaNMgd5v:b-xdlxfl) ([raw file](https://github.com/dgraph-io/dgraph/blob/b414332fff0ed6e9dda194572ba95123293e56f0/worker/draft.go#L468)):*
> ```Go
> 			cc.Unmarshal(e.Data)
> 
> 			if cc.Type != raftpb.ConfChangeRemoveNode && len(cc.Context) > 0 {
> ```

if cc.Type == ... { } else if len(cc.Context) > 0 { .. }

---

*[worker/groups.go, line 52 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/679#-KelMi_q0EWgxFYMYocZ:-KelMi_q0EWgxFYMYoc_:b-pnfan2) ([raw file](https://github.com/dgraph-io/dgraph/blob/b414332fff0ed6e9dda194572ba95123293e56f0/worker/groups.go#L52)):*
> ```Go
> 	all map[uint32]*servers
> 	// removed stores the groupId to servers map for removed servers
> 	removed    map[uint32]*servers
> ```

Shouldn't this just be the id of the server (/node) instead of a groupid -> ids. group shouldn't make any difference here. Because once a node is out, it's out of all the groups.

---

*[worker/groups.go, line 196 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/679#-KelNLUyB9x13FRlv3Gb:-KelNLUyB9x13FRlv3Gc:bpac6oe) ([raw file](https://github.com/dgraph-io/dgraph/blob/b414332fff0ed6e9dda194572ba95123293e56f0/worker/groups.go#L196)):*
> ```Go
> }
> 
> func (g *groupi) isRemoved(groupId uint32, id uint64) bool {
> ```

Should just be removed(id uint64) I think.

---

*[worker/groups.go, line 413 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/679#-KelOUTLbXQvx21ltE-Q:-KelOUTLbXQvx21ltE-R:b-f7hti1) ([raw file](https://github.com/dgraph-io/dgraph/blob/b414332fff0ed6e9dda194572ba95123293e56f0/worker/groups.go#L413)):*
> ```Go
> 		}
> 
> 		go func(groupId uint32, nodeId uint64, addr string) {
> ```

This seems like duplicate code w.r.t. syncMembership. We already have a way to apply or propagate membership. So, why not just use that? You can modify syncMembership to do what you need to do here.

---

*[worker/groups.go, line 550 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/679#-KelPJ-Z73wE9L4IqWoA:-KelPJ-Z73wE9L4IqWoB:bkrqkox) ([raw file](https://github.com/dgraph-io/dgraph/blob/b414332fff0ed6e9dda194572ba95123293e56f0/worker/groups.go#L550)):*
> ```Go
> 		// Print out the entire list.
> 		for gid, sl := range g.all {
> 			fmt.Printf("Group: %v. Active List: %+v\n", gid, sl.list)
> ```

Use event logs to track all these membership updates. If we have none, then create an event log for this purpose.

---

*[worker/groups.go, line 631 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/679#-KelPR69mf8AFbcEjsus:-KelPR69mf8AFbcEjsut:bh0pid7) ([raw file](https://github.com/dgraph-io/dgraph/blob/b414332fff0ed6e9dda194572ba95123293e56f0/worker/groups.go#L631)):*
> ```Go
> // RemoveMembership is the RPC call for removing membership for servers
> // which don't serve group zero.
> func (w *grpcWorker) RemoveMembership(ctx context.Context,
> ```

Again, this doesn't need to be new code. You can just modify the existing one.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/679)*

gacharya255:

---

*[worker/draft.go, line 386 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/679#-KfMBjQYyAjlvrSvu2N9:-KfMBjQYyAjlvrSvu2NA:bw95ybc) ([raw file](https://github.com/dgraph-io/dgraph/blob/b414332fff0ed6e9dda194572ba95123293e56f0/worker/draft.go#L386)):*
> ```Go
> 		return
> 	case err := <-ch:
> 		if grpc.ErrorDesc(err) == ErrorInvalidToNodeId {
> ```

Wonder if you tried testing this part? I am slightly worried that x.Errorf might do something unexpected such that the error string here might not be what you expect.

---

*[worker/groups.go, line 199 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/679#-KfMCyO7p7irPY2XbtRk:-KfMCyO7p7irPY2XbtRl:b-oiqc25) ([raw file](https://github.com/dgraph-io/dgraph/blob/b414332fff0ed6e9dda194572ba95123293e56f0/worker/groups.go#L199)):*
> ```Go
> 	g.RLock()
> 	defer g.RUnlock()
> 	if g.removed == nil {
> ```

Optional: One strange thing about Go is that nil map will return the default value and found = false. In this case, if g.removed is nil, g.removed[groupId] will be a nil map. And a range over a nil map is valid. It will not crash. I believe you can skip the check of g.removed==nil and also sl == nil.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/679)*

janardhan1993:
Have sent a mail to rafte group. Will work on it after i hear back from them.

janardhan1993:
@manishrjain So upon discussion with xiang li, i think it's best if we keep removed servers list in application layer and ban them if we the node can come back later.

1. We had a node with id 3 running, then we stopped it and restarted with different id 4. In this case we can safely remove node with id 3 from quorum and all map in group zero.
2. We removed a node and at that time the node was available, so when the remove configuration reaches that server it removes it self(other raft nodes will also remove it) and removes itself from the all map in group zero.
3. We remove a node the node was not reachable(may be lost in network partition). On removing the node, the quorum of nodes would remove it. But later say we add a new node with same id and then then the old node came back(network partition was removed). It would screw up the membership information in group zero. So in this case we should store list of banned id's when the server we were going to remove was not reachable, so that a new node can't be started with that id. (Similar to etcd)

We are exposing two apis
1. Remove (node id, list of group ids(optional)) - if groups id's are not specified we remove all groups served by that server
2. Force remove, say we had only single node serving group 5 and it crashed. Every other node which we start would try to connect to the dead node and we can't bring back the group 5 ever. In this case the user could use force remove option(there would be data loss though) to remove that dead node serving group 5. This would be useful when we don't have quorum to remove the node or there was a single node for that group and it crashed. In this case also we will ban that node id.

janardhan1993:
closing this and raising a new PR. Master has gone too far ahead, will rebase and raise another PR(otherwise reviewable will show a huge diff).

PR #681, opened 2017-03-09
Integrate/redux

pawanrawal:
Integrated Redux

1. Makes persisting to localstorage easier. We just need to manage Redux state and it handles persistence for us.
2. Components have been broken down into Presentation components and Container components. There is still something more I can do but that's for another time.
3. Redux offers superior debugging because of time travel available in redux tools.
4. We are passing less state and props around. 

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/681)


tzdybal:
Lots of code, this makes review hard. I have some minor questions.

---

Reviewed 28 of 30 files at r1, 12 of 12 files at r2.

---

*[dashboard/src/actions/index.js, line 136 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/681#-KemWnpiYzs-0MreeXNV:-KemWnpkCMJKi8C6ewj1:b-5wboy0) ([raw file](https://github.com/dgraph-io/dgraph/blob/98deb550f848bd9a3857161ea67004081eb5324e/dashboard/src/actions/index.js#L136)):*
> ```JavaScript
>                         let mantainSortOrder = showTreeView(query);
>                         dispatch(saveSuccessResponse(null, result));
>                         renderGraph(query, result, mantainSortOrder)(dispatch);
> ```

Maybe it's newbie question, but why are you using this notation, instead of `dispatch(renderGraph(query, result, mantainSortOrder))` (my understanding is that `thunk` should handle this)?

---

*[dashboard/src/containers/Helpers.js, line 14 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/681#-KeoU3waWCk-gA89zK7M:-KeoU3wbUg7O-vEktIi6:b5tlll5) ([raw file](https://github.com/dgraph-io/dgraph/blob/626b80b0688e230a9fb052e34a86d7331b0b950c/dashboard/src/containers/Helpers.js#L14)):*
> ```JavaScript
> }
> 
> export function parseJSON(response) {
> ```

Why is this helper any better than direct call to response.json()? 

---

*[dashboard/src/containers/Helpers.js, line 31 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/681#-KeoULaBpsqQkgIUZkrt:-KeoULaBpsqQkgIUZkru:bnwol5n) ([raw file](https://github.com/dgraph-io/dgraph/blob/626b80b0688e230a9fb052e34a86d7331b0b950c/dashboard/src/containers/Helpers.js#L31)):*
> ```JavaScript
> 
> export function aggregationPrefix(properties) {
>   let aggTerms = ["count", "max(", "min(", "sum("];
> ```

Why there is no '(' in `"count"`? Why count is processed differently than other terms?

---

*[dashboard/src/containers/Helpers.js, line 216 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/681#-KeoXCrsgLJJmzjhHkEi:-KeoXCrsgLJJmzjhHkEj:b-1o2g9g) ([raw file](https://github.com/dgraph-io/dgraph/blob/626b80b0688e230a9fb052e34a86d7331b0b950c/dashboard/src/containers/Helpers.js#L216)):*
> ```JavaScript
> }
> 
> export function processGraph(
> ```

This function is over 200 LOC. Maybe it could be split for clarity?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/681#-:-KeoT9QAbMBFEj2F73yW:bmolocw)*

pawanrawal:
Review status: 26 of 33 files reviewed at latest revision, 4 unresolved discussions.

---

*[dashboard/src/actions/index.js, line 136 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/681#-KemWnpiYzs-0MreeXNV:-KeppqrvQEHEEaw2DhKX:bjv0cdu) ([raw file](https://github.com/dgraph-io/dgraph/blob/98deb550f848bd9a3857161ea67004081eb5324e/dashboard/src/actions/index.js#L136)):*

Maybe it's newbie question, but why are you using this notation, instead of `dispatch(renderGraph(query, result, mantainSortOrder))` (my understanding is that `thunk` should handle this)?
</blockquote></details>

Yes that should work too and is cleaner. Good catch.

---

*[dashboard/src/containers/Helpers.js, line 14 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/681#-KeoU3waWCk-gA89zK7M:-KeppvZH_a1FRrD9sVYD:brj6fre) ([raw file](https://github.com/dgraph-io/dgraph/blob/626b80b0688e230a9fb052e34a86d7331b0b950c/dashboard/src/containers/Helpers.js#L14)):*

Why is this helper any better than direct call to response.json()? 
</blockquote></details>

isn't really any better. replaced it.

---

*[dashboard/src/containers/Helpers.js, line 31 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/681#-KeoULaBpsqQkgIUZkrt:-KepqMH3KlS3wqyDmBD7:b-a6c2ak) ([raw file](https://github.com/dgraph-io/dgraph/blob/626b80b0688e230a9fb052e34a86d7331b0b950c/dashboard/src/containers/Helpers.js#L31)):*

Why there is no '(' in `"count"`? Why count is processed differently than other terms?
</blockquote></details>

Thats because of the output of count vs the output of other fields. https://wiki.dgraph.io/Query_Language#Count. It comes as a child of the predicate. Min, max need to return a key which tells on what child were those aggregations performed.

---

*[dashboard/src/containers/Helpers.js, line 216 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/681#-KeoXCrsgLJJmzjhHkEi:-KeprlcH36y6T9Upwmv7:bk8mvl) ([raw file](https://github.com/dgraph-io/dgraph/blob/626b80b0688e230a9fb052e34a86d7331b0b950c/dashboard/src/containers/Helpers.js#L216)):*

This function is over 200 LOC. Maybe it could be split for clarity?
</blockquote></details>

refactored it a bit and added more comments. Actually I realised its big because of a lot of variable declaration which I believe can't be reduced. Apart from that the actual logic isn't very big. Will keep on refactoring it as and when I realize that it can be done. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/681)*

PR #730, opened 2017-03-21
Support aggregate values in variables

ashwin95r:
Support queries like: 
```
{
 b as var(func: anyofterms(name, "spielberg nolan tarantino")) @filter(gt(count(director.film), 1)) {
  a as count(director.film) 
 }

 order(id: var(b), orderdesc: var(a)) {
  name 
  count(director.film)
 }
}


{
 b as var(func: anyofterms(name, "spielberg nolan tarantino")) @filter(gt(count(director.film), 1)) {
  director.film {
    a as min(initial_release_date)
  }
 }

 order(id: var(b), orderasc: var(a)) {
  name 
  director.film { min(initial_release_date)}
 }
}
```

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/730)

### DIFF
--- a/gql/parser.go
+++ b/gql/parser.go
@@ -1531,6 +1531,7 @@ func getRoot(it *lex.ItemIterator) (gq *GraphQuery, rerr error) {
 // godeep constructs the subgraph from the lexed items and a GraphQuery node.
 func godeep(it *lex.ItemIterator, gq *GraphQuery) error {
 	var isCount uint16
+	varName := ""
 	curp := gq // Used to track current node, for nesting.
 	for it.Next() {
 		item := it.Item()
@@ -1557,17 +1558,19 @@ func godeep(it *lex.ItemIterator, gq *GraphQuery) error {
 			}
 
 			if peekIt[0].Typ == itemName && strings.ToLower(peekIt[0].Val) == "as" {
-				varName := item.Val
+				varName = item.Val
 				it.Next() // "As" was checked before.
-				it.Next()
-				pred := it.Item()
-				child := &GraphQuery{
-					Args: make(map[string]string),
-					Attr: pred.Val,
-					Var:  varName,
-				}
-				gq.Children = append(gq.Children, child)
-				curp = child
+				/*
+					it.Next()
+						pred := it.Item()
+						child := &GraphQuery{
+							Args: make(map[string]string),
+							Attr: pred.Val,
+							Var:  varName,
+						}
+						gq.Children = append(gq.Children, child)
+						curp = child
+				*/
 				continue
 			}
 
@@ -1576,7 +1579,9 @@ func godeep(it *lex.ItemIterator, gq *GraphQuery) error {
 				item.Val = val
 				child := &GraphQuery{
 					Args: make(map[string]string),
+					Var:  varName,
 				}
+				varName = ""
 				it.Prev()
 				if child.Func, err = parseFunction(it); err != nil {
 					return err
@@ -1609,7 +1614,9 @@ func godeep(it *lex.ItemIterator, gq *GraphQuery) error {
 				Args:    make(map[string]string),
 				Attr:    item.Val,
 				IsCount: isCount == 1,
+				Var:     varName,
 			}
+			varName = ""
 			gq.Children = append(gq.Children, child)
 			curp = child
 			if isCount == 1 {
--- a/gql/parser.go
+++ b/gql/parser.go
@@ -1560,17 +1560,6 @@ func godeep(it *lex.ItemIterator, gq *GraphQuery) error {
 			if peekIt[0].Typ == itemName && strings.ToLower(peekIt[0].Val) == "as" {
 				varName = item.Val
 				it.Next() // "As" was checked before.
-				/*
-					it.Next()
-						pred := it.Item()
-						child := &GraphQuery{
-							Args: make(map[string]string),
-							Attr: pred.Val,
-							Var:  varName,
-						}
-						gq.Children = append(gq.Children, child)
-						curp = child
-				*/
 				continue
 			}
 
--- a/gql/parser_test.go
+++ b/gql/parser_test.go
@@ -91,11 +91,11 @@ func TestParseQueryWithVarValAgg(t *testing.T) {
 	require.Equal(t, 2, len(res.Query))
 	require.Equal(t, "L", res.Query[0].NeedsVar[0])
 	require.Equal(t, "n", res.Query[0].NeedsVar[1])
+	require.Equal(t, "n", res.Query[0].Args["orderasc"])
+	require.Equal(t, "name", res.Query[0].Children[0].Attr)
 	require.Equal(t, "L", res.Query[1].Children[0].Var)
 	require.Equal(t, "n", res.Query[1].Children[0].Children[0].Var)
 	require.Equal(t, "min", res.Query[1].Children[0].Children[0].Func.Name)
-	require.Equal(t, "n", res.Query[0].Args["orderasc"])
-	require.Equal(t, "name", res.Query[0].Children[0].Attr)
 }
 
 func TestParseQueryWithVarValCount(t *testing.T) {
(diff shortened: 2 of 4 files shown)


janardhan1993:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[query/query.go, line 975 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/730#-KfkaGU5Z_9zLXrG-LyE:-KfkaGU5Z_9zLXrG-LyF:bb38e8k) ([raw file](https://github.com/dgraph-io/dgraph/blob/6bf41b634a1b8ba8a5932baee4867979d242c2f1/query/query.go#L975)):*
> ```Go
> 			// If the length of child UID list is zero and it has no valid value, then the
> 			// current UID should be removed from this level.
> 			if len(child.values[i].Val) == 0 && (len(child.counts) < i) && len(child.uidMatrix[i].Uids) == 0 {
> ```

1. Why are we filtering out uid's that don't have atleast one uid in every child, shouldn't it be atleast one uid in atleast one child.

Example: 
<alice1> <name> "alice"
<alice2> <name> "alice"
<alice3> <name> "alice"
<alice1> <age> "1"

{
               f as var(func: anyofterms(name, "alice alice2 alice3")) {
			name
			age
		}
		
		persons1(id: var(f)) {
			name
			age
		}
		
}

would return only alice1
but one would expect all three to be returned

2. I am not that familiar with this code, what does len(child.counts) < i check for? if a child has DoCount set then isn't count populated for all uids ? If no edge is present then it could be zero right ?

---

*[query/query.go, line 1375 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/730#-Kfkea5EOfhtlCU62VVX:-Kfkea5EOfhtlCU62VVY:b8j1wnq) ([raw file](https://github.com/dgraph-io/dgraph/blob/6bf41b634a1b8ba8a5932baee4867979d242c2f1/query/query.go#L1375)):*
> ```Go
> 		typ := values[0].Tid
> 		types.Sort(typ, values, ul, sg.Params.OrderDesc)
> 		sg.uidMatrix[i] = ul
> ```

If a value for a predicate is not present for a node it's not present in output if we are sorting on that predicate since we use index. I think we should have similar behaviour here.

Value can be nil bytes also in case of min, max or even normal predicate. types.Sort would cause the server to crash if nil byte is passed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/730)*

ashwin95r:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[query/query.go, line 975 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/730#-KfkaGU5Z_9zLXrG-LyE:-KfnMOOj3bCFS3Bl4YQH:b5i25hk) ([raw file](https://github.com/dgraph-io/dgraph/blob/6bf41b634a1b8ba8a5932baee4867979d242c2f1/query/query.go#L975)):*

1. Why are we filtering out uid's that don't have atleast one uid in every child, shouldn't it be atleast one uid in atleast one child.

Example: 
<alice1> <name> "alice"
<alice2> <name> "alice"
<alice3> <name> "alice"
<alice1> <age> "1"

{
               f as var(func: anyofterms(name, "alice alice2 alice3")) {
			name
			age
		}
		
		persons1(id: var(f)) {
			name
			age
		}
		
}

would return only alice1
but one would expect all three to be returned

2. I am not that familiar with this code, what does len(child.counts) < i check for? if a child has DoCount set then isn't count populated for all uids ? If no edge is present then it could be zero right ?

</blockquote></details>

1. For variables, if the user specifies some predicate, we treat it as required compulsorily. If not we remove that UID from this level. Hence the checks.
2. Yeah the count can be zero. This is just a sanity check to make sure we have a count populated for that UID at index i. (should be <= btw, fixed that)

---

*[query/query.go, line 1375 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/730#-Kfkea5EOfhtlCU62VVX:-KfnR5ejx9ls8gEDWlZt:b-4toc4a) ([raw file](https://github.com/dgraph-io/dgraph/blob/6bf41b634a1b8ba8a5932baee4867979d242c2f1/query/query.go#L1375)):*

If a value for a predicate is not present for a node it's not present in output if we are sorting on that predicate since we use index. I think we should have similar behaviour here.

Value can be nil bytes also in case of min, max or even normal predicate. types.Sort would cause the server to crash if nil byte is passed.
</blockquote></details>

The nil values won't be populated in the map. So, skipping the UIDs which dont have a value.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/730)*

janardhan1993:
:lgtm:

---

Reviewed 1 of 1 files at r2.
Review status: 1 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[query/query.go, line 975 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/730#-KfkaGU5Z_9zLXrG-LyE:-Kfo-iePLSDKjEGdcOvn:b-4fcutv) ([raw file](https://github.com/dgraph-io/dgraph/blob/6bf41b634a1b8ba8a5932baee4867979d242c2f1/query/query.go#L975)):*

1. For variables, if the user specifies some predicate, we treat it as required compulsorily. If not we remove that UID from this level. Hence the checks.
2. Yeah the count can be zero. This is just a sanity check to make sure we have a count populated for that UID at index i. (should be <= btw, fixed that)
</blockquote></details>

It would be good to document our assumption in wiki. Sometimes user's intent could be any one of the predicate. Ex:- He wants to populate the id of people having a contact(phone or email) living in bangalore.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/730#-:-Kfo03ol5Dx8OAHUZ5M8:bnfp4nl)*

manishrjain:
LGTM

---

Reviewed 3 of 4 files at r1, 1 of 1 files at r2.

---

*[gql/parser_test.go, line 97 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/730#-KfoEnNlm5yd72xKSMok:-KfoEnNm_vY2DyOt-U_Z:bu9z0bk) ([raw file](https://github.com/dgraph-io/dgraph/blob/558bc01935bcbf370e685b2cac8cb204724c5e03/gql/parser_test.go#L97)):*
> ```Go
> 	require.Equal(t, "n", res.Query[1].Children[0].Children[0].Var)
> 	require.Equal(t, "min", res.Query[1].Children[0].Children[0].Func.Name)
> 	require.Equal(t, "n", res.Query[0].Args["orderasc"])
> ```

group Query[0], and Query[1] separately.

---

*[query/query.go, line 975 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/730#-KfoEsUSCV1UY7iO77-W:-KfoEsUSCV1UY7iO77-X:b-1ex4r) ([raw file](https://github.com/dgraph-io/dgraph/blob/558bc01935bcbf370e685b2cac8cb204724c5e03/query/query.go#L975)):*
> ```Go
> 			// If the length of child UID list is zero and it has no valid value, then the
> 			// current UID should be removed from this level.
> 			if len(child.values[i].Val) == 0 && (len(child.counts) <= i) && len(child.uidMatrix[i].Uids) == 0 {
> ```

100 chars.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/730#-:-KfoFFDQ3nUqD2CAioEB:bb6ywmx)*

ashwin95r:

---

*[query/query.go, line 975 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/730#-KfkaGU5Z_9zLXrG-LyE:-KfoFxlfbAbaQjAYlv8W:bg03ll7) ([raw file](https://github.com/dgraph-io/dgraph/blob/6bf41b634a1b8ba8a5932baee4867979d242c2f1/query/query.go#L975)):*

It would be good to document our assumption in wiki. Sometimes user's intent could be any one of the predicate. Ex:- He wants to populate the id of people having a contact(phone or email) living in bangalore.
</blockquote></details>

Sure, noted it in my wiki task.

---

*[query/query.go, line 975 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/730#-KfoEsUSCV1UY7iO77-W:-KfoFvVVxy0KQwU9-B9i:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/558bc01935bcbf370e685b2cac8cb204724c5e03/query/query.go#L975)):*

100 chars.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/730)*

ashwin95r:

---

*[gql/parser_test.go, line 97 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/730#-KfoEnNlm5yd72xKSMok:-KfoGQgnCbK4uijGNiUW:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/558bc01935bcbf370e685b2cac8cb204724c5e03/gql/parser_test.go#L97)):*

group Query[0], and Query[1] separately.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/730)*

ashwin95r:
Merged.

PR #782, opened 2017-03-29
Have a cascade directive

ashwin95r:
This query
```
{
  HP(func: allofterms(name, "Harry Potter")) @cascade {
    name
    starring{
        performance.character {
          name@ta
        }
        performance.actor @filter(allofterms(name, "Warwick")){
            name@ta
         }
    }
  }
}
```
Without the cascade directive, would return all the performance character. But if we want to see only the filtered (on sibling) nodes, the only way is to use a variable which would force a cascade and use it in another block. So it's better to provide a `@cascade` directive at root which can help achieve this.

So, if any of the specified field is missing, that enitity would be discarded.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/782)

### DIFF
--- a/gql/parser.go
+++ b/gql/parser.go
@@ -48,6 +48,7 @@ type GraphQuery struct {
 	Children     []*GraphQuery
 	Filter       *FilterTree
 	Normalize    bool
+	Cascade      bool
 	Facets       *Facets
 	FacetsFilter *FilterTree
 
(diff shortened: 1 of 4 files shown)


janardhan1993:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/782#-:-KgNolx35GGbNzwE__lr:bnfp4nl)*

ashwin95r:
Merged.

PR #798, opened 2017-04-03
Show JSON response in tab

pawanrawal:
Show JSON response in a tab in the UI.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/798)


dashboard/src/actions/index.js
```diff
@@ -137,19 +139,18 @@ export const runQuery = query => {
                             // We display the response from server.
                         } else {
                             dispatch(addQuery(query));
-                            dispatch(
-                                saveSuccessResponse(JSON.stringify(result))
-                            );
+                            dispatch(saveSuccessResponse(null, result, true));
```
sungwoncho: Passing empty string might be more readable instead of explicitly passing `null`

dashboard/src/actions/index.js
```diff
@@ -48,15 +48,17 @@ const fetchedResponse = () => ({
     fetching: false
 });
 
-const saveSuccessResponse = (text, data) => ({
+const saveSuccessResponse = (text, data, mutation) => ({
```
sungwoncho: `isMutation`?

dashboard/src/actions/index.js
```diff
@@ -137,19 +139,18 @@ export const runQuery = query => {
...
-                        dispatch(saveSuccessResponse(null, result));
+                        dispatch(saveSuccessResponse(null, result, false));
                         dispatch(renderGraph(query, result, mantainSortOrder));
                     } else {
                         dispatch(
                             saveErrorResponse(
-                                "Your query did not return any results."
+                                "Your query did not return any results.",
+                                result
                             )
```
sungwoncho: is there a reason we are saving error messages in redux store? I think they might belong in react component states because they do not need to be shared application wide.

dashboard/src/components/Graph.js
```diff
@@ -1,14 +1,29 @@
...
+        json,
+        selectTab,
+        selectedTab
+    } = props,
+        graphClass = fs ? "Graph-fs" : "Graph-s",
+        bgColor,
+        hourglass = isFetching ? "Graph-hourglass" : "",
+        graphHeight = fs ? "Graph-full-height" : "Graph-fixed-height",
+        showGraph = selectedTab === "1" ? "" : "Graph-hide",
+        showJSON = selectedTab === "2" ? "" : "Graph-hide";
```
sungwoncho: Using https://github.com/JedWatson/classnames, we can simplify the optional class name.

    className={classnames({'Graph-hide': !showJSON});

If handling optional class names this way proves to be a pain, we can start using that package.

pawanrawal:
Review status: 0 of 11 files reviewed at latest revision, 4 unresolved discussions.

---

*[dashboard/src/actions/index.js, line 51 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/798#-Kgmj_piS2UGyQ5BX6xx-r1-51:-Kgrpk9atZW7ACr7SATL:bb74ce2) ([raw file](https://github.com/dgraph-io/dgraph/blob/543b03859af4d79b80395a56659306d4df15813b/dashboard/src/actions/index.js#L51)):*
<details><summary><i>Previously, sungwoncho (Sung Won Cho) wrote…</i></summary><blockquote>

IDEA: `isMutation`?
</blockquote></details>

Sure

---

*[dashboard/src/actions/index.js, line 142 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/798#-Kgmj_piS2UGyQ5BX6xx-r1-142:-KgrqHjgLMzSdUtVjf-C:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/543b03859af4d79b80395a56659306d4df15813b/dashboard/src/actions/index.js#L142)):*
<details><summary><i>Previously, sungwoncho (Sung Won Cho) wrote…</i></summary><blockquote>

Passing empty string might be more readable instead of explicitly passing `null`, because saveSuccessResponse takes `(text, data, mutation)` as args
</blockquote></details>

Done.

---

*[dashboard/src/actions/index.js, line 154 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/798#-Kgmj_piS2UGyQ5BX6xx-r1-154:-KgrqXVcTTvXNM17Niff:b-jdd83k) ([raw file](https://github.com/dgraph-io/dgraph/blob/543b03859af4d79b80395a56659306d4df15813b/dashboard/src/actions/index.js#L154)):*
<details><summary><i>Previously, sungwoncho (Sung Won Cho) wrote…</i></summary><blockquote>

is there a reason we are saving error messages in redux store? I think they might belong in react component states because they do not need to be shared application wide.
</blockquote></details>

Thats correct, though the issue is that it is triggered by clicking the Run button, whereas some other component, the Response box has to get this. So even if I were to store it the state, it would go to the App component, and then I pass it down the chain.
Right now, I just have a container which listens on changes to this.

---

*[dashboard/src/components/Graph.js, line 25 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/798#-Kgmj_piS2UGyQ5BX6xz-r1-25:-Kgrqs05O_OgGtbgEDET:binvfw5) ([raw file](https://github.com/dgraph-io/dgraph/blob/543b03859af4d79b80395a56659306d4df15813b/dashboard/src/components/Graph.js#L25)):*
<details><summary><i>Previously, sungwoncho (Sung Won Cho) wrote…</i></summary><blockquote>

Using https://github.com/JedWatson/classnames, we can simplify the optional class name.

    className={classnames({'Graph-hide': !showJSON});

If handling optional class names this way proves to be a pain, we can start using that package.
</blockquote></details>

Thanks, will check that out.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/798)*

pawanrawal:
Merged

PR #872, opened 2017-04-23
Case insensitive regexp + new syntax

tzdybal:
This PR contains following changes:

1. Introduction of new syntax for regular expressions - `regexp(attr, /expression/flags)`.
1. Support of ignore case flag - `i`.
1. Bug fixes in trigram query processing. 

The case insensitivity should work without those changes, because regex library supports [modifiers](http://www.regular-expressions.info/modifiers.html) (`i` flag itself is implemented with `(?i)` modifier). New syntax was introduced for convenience.

Tokenization is not affected by the changes. Adding lower-cased trigrams to the index would make it impossible to search for lower-cased expressions in case-sensitive manner (as there is no way to distinguish lower-cased and normal tokens).

The `(?i)` modifier, forces generation of trigrams in all possible case combinations. This greatly improves the number of generated trigrams and may affect performance. Every trigram produces at least 8 trigrams, like `/abc/i/` results in following trigrams: `abc`, `abC`, `aBc`, `Abc`, `aBC`, `AbC`, `ABc`, `ABC` (uppercase <-> lowercase mapping [is not always 1 to 1](http://unicode.org/faq/casemap_charprop.html#7a), even more trigrams may be generated from single source trigram).

### DIFF
--- a/gql/parser.go
+++ b/gql/parser.go
@@ -364,7 +364,7 @@ func checkValueType(vm varMap) error {
 }
 
 func substituteVar(f string, res *string, vmap varMap) error {
-	if f[0] == '$' {
+	if len(f) > 0 && f[0] == '$' {
 		va, ok := vmap[f]
 		if !ok {
 			return x.Errorf("Variable not defined %v", f)
@@ -1307,6 +1307,18 @@ L:
 						return nil, x.Errorf("Invalid use of $ in func args")
 					}
 					isDollar = true
+					continue
+				} else if itemInFunc.Typ == itemRegex {
+					end := strings.LastIndex(itemInFunc.Val, "/")
+					x.AssertTrue(end >= 0)
+					expr := strings.Replace(itemInFunc.Val[1:end], "\\/", "/", -1)
+					flags := ""
+					if end+1 < len(itemInFunc.Val) {
+						flags = itemInFunc.Val[end+1:]
+					}
+
+					g.Args = append(g.Args, expr, flags)
+					expectArg = false
 					continue
 				} else if itemInFunc.Typ != itemName {
 					return nil, x.Errorf("Expected arg after func [%s], but got item %v",
--- a/gql/parser_test.go
+++ b/gql/parser_test.go
@@ -2841,7 +2841,7 @@ func TestParseQueryWithAttrLang(t *testing.T) {
 func TestParseQueryWithAttrLang2(t *testing.T) {
 	query := `
 	{
-	  me(func:regexp(name, "^[a-zA-z]*[^Kk ]?[Nn]ight"), orderasc: name@en, first:5) {
+	  me(func:regexp(name, /^[a-zA-z]*[^Kk ]?[Nn]ight/), orderasc: name@en, first:5) {
 		name@en
 		name@de
 		name@it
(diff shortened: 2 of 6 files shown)


ashwin95r:
Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[gql/state.go, line 610 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/872#-KiS7AoYbmZkiLIm-AIi:-KiS7AoYbmZkiLIm-AIj:b2lx45u) ([raw file](https://github.com/dgraph-io/dgraph/blob/1d8845c8baeaae4a2bd73b9506a1fd5e79df7786/gql/state.go#L610)):*
> ```Go
> }
> 
> func isRegexFlag(r rune) bool {
> ```

Just checeking my understanding, do all characters have a valid meaning?  I have some across `g` and `i`

---

*[query/query_test.go, line 2745 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/872#-KiS7xgpiRQFjNIw8r1z:-KiS7xgpiRQFjNIw8r2-:b-jnghsb) ([raw file](https://github.com/dgraph-io/dgraph/blob/1d8845c8baeaae4a2bd73b9506a1fd5e79df7786/query/query_test.go#L2745)):*
> ```Go
>     {
> 	  me(id:0x1234) {
> 		pattern @filter(regexp(value, /SUB[cm]/i)) {
> ```

Is `/<something-valid-regex>/abcd` a valid case? Also, do we support `g` flag?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/872)*

ashwin95r:
Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/872)*

tzdybal:

---

*[gql/state.go, line 610 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/872#-KiS7AoYbmZkiLIm-AIi:-KiSAONPyZz0Ne3HXGmA:b-mk9dnf) ([raw file](https://github.com/dgraph-io/dgraph/blob/1d8845c8baeaae4a2bd73b9506a1fd5e79df7786/gql/state.go#L610)):*

Just checeking my understanding, do all characters have a valid meaning?  I have some across `g` and `i`
</blockquote></details>

No. Currently only `i` is supported. Everything else is just ignored. I don't want to verify flag correctness in lexer.
This check is very general, to enable adding more flags.

---

*[query/query_test.go, line 2745 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/872#-KiS7xgpiRQFjNIw8r1z:-KiSBJgg5Q8tYT5YgBFy:b-a3d9a3) ([raw file](https://github.com/dgraph-io/dgraph/blob/1d8845c8baeaae4a2bd73b9506a1fd5e79df7786/query/query_test.go#L2745)):*

Is `/<something-valid-regex>/abcd` a valid case? Also, do we support `g` flag?
</blockquote></details>

`abcd` currently will be silently ignored. I can add checking logic and return error.
What do you mean by `g` flag? Typically (for example in Perl and JS) it's 'global match', meaning match all occurrences. I think it's not applicable for us, because we're only interested if _any_ match is possible.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/872)*

ashwin95r:
:lgtm:

---

---

*[query/query_test.go, line 2745 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/872#-KiS7xgpiRQFjNIw8r1z:-KiSDCyte7Cd08RSi0Wy:b4ocptv) ([raw file](https://github.com/dgraph-io/dgraph/blob/1d8845c8baeaae4a2bd73b9506a1fd5e79df7786/query/query_test.go#L2745)):*

`abcd` currently will be silently ignored. I can add checking logic and return error.
What do you mean by `g` flag? Typically (for example in Perl and JS) it's 'global match', meaning match all occurrences. I think it's not applicable for us, because we're only interested if _any_ match is possible.
</blockquote></details>

Makes sense, if you think ignoring is better it's alright

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/872#-:-KiSDrF6et_uMKDH4GcX:bnfp4nl)*

manishrjain:
:lgtm: small comments.

---

Reviewed 4 of 6 files at r1, 3 of 3 files at r2.

---

*[gql/parser.go, line 1313 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/872#-KiXb-J6gWt6gay4oIQS:-KiXb-J6gWt6gay4oIQT:bneosl9) ([raw file](https://github.com/dgraph-io/dgraph/blob/067af1e1be7e78aca5259d4c7e6008d6f7422bc8/gql/parser.go#L1313)):*
> ```Go
> 					continue
> 				} else if itemInFunc.Typ == itemRegex {
> 					end := strings.LastIndex(itemInFunc.Val, "/")
> ```

What if the expression ends with `/something\/`? This case doesn't handle that.

---

*[query/query_test.go, line 2745 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/872#-KiS7xgpiRQFjNIw8r1z:-KiXcOelNB5HrKshfW8i:b-o3l9s7) ([raw file](https://github.com/dgraph-io/dgraph/blob/1d8845c8baeaae4a2bd73b9506a1fd5e79df7786/query/query_test.go#L2745)):*

Makes sense, if you think ignoring is better it's alright
</blockquote></details>

We should return an error saying invalid regexp flags. We shouldn't ignore anything that can be rectified by the user.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/872#-:-KiXcgHdt03D9p5iTSMb:b1kjqmq)*

tzdybal:
Review status: 4 of 7 files reviewed at latest revision, 2 unresolved discussions.

---

*[gql/parser.go, line 1313 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/872#-KiXb-J6gWt6gay4oIQS:-KiXiJVBr6SrGlLC462T:b-vkndqt) ([raw file](https://github.com/dgraph-io/dgraph/blob/067af1e1be7e78aca5259d4c7e6008d6f7422bc8/gql/parser.go#L1313)):*

What if the expression ends with `/something\/`? This case doesn't handle that.
</blockquote></details>

Such expression causes parsing error. Created a test for such case.

---

*[query/query_test.go, line 2745 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/872#-KiS7xgpiRQFjNIw8r1z:-KiXmhHPtJGeR4Jd4iVB:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/1d8845c8baeaae4a2bd73b9506a1fd5e79df7786/query/query_test.go#L2745)):*

We should return an error saying invalid regexp flags. We shouldn't ignore anything that can be rectified by the user.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/872)*

PR #905, opened 2017-05-08
Experiment/loader

janardhan1993:
Prior to this change, whenever we were loading data. Before doing index mutation, the index key was being persisted to db(which takes around 6-9ms on my machine). So mutation waits on the lock and hence loading was slow.(This coupled with the fact that mutations within a request are executed sequentially was causing the data loading to be slow).

The index key would be persisted eventually now(immediately). After this change data loads 4 times faster on my machine. If write to db fails then server would crash and 
1. if synced watermark reaches the index, the index would be persisted to db so on restart there is no need to write index key
2. otherwise the raft log would be replayed and the index key would be written again.

The only downside would be that the mutation would succeed, but we would need to wait for 10-20ms for filtering or sorting.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/905)

### DIFF
--- a/posting/index_test.go
+++ b/posting/index_test.go
@@ -138,6 +138,7 @@ func TestTokensTable(t *testing.T) {
 		Entity: 157,
 	}
 	addMutationWithIndex(t, l, edge, Set)
+	time.Sleep(20 * time.Millisecond)
 
 	key = x.IndexKey("name", "david")
 	slice, err := ps.Get(key)
--- a/posting/lists.go
+++ b/posting/lists.go
@@ -443,9 +443,8 @@ func GetOrCreate(key []byte, group uint32) (rlist *List, decr func()) {
 	// index key to the data store, with essentially an empty value. We just need
 	// the keys for filtering / sorting.
 	if l == lp && pk.IsIndex() {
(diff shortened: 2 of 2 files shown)


manishrjain:
Reviewed 2 of 2 files at r1.

---

*[posting/lists.go, line 447 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/905#-Kjaf0NJZnzfykwOIBx6:-Kjaf0NJZnzfykwOIBx7:bk0jgv0) ([raw file](https://github.com/dgraph-io/dgraph/blob/beec17b91df634e57fa42d85bbd2a94d4a1607b3/posting/lists.go#L447)):*
> ```Go
> 	if l == lp && pk.IsIndex() {
> 		go func(key []byte) {
> 			l.Lock()
> ```

Locking has to be done before entering goroutine. Otherwise, there would be a race condition between another `GetOrCreate`, which could end up overwriting what's already written.

In Badger, we could improve this, by providing a `Touch` function, so a key is generated in the KV store quickly. So, for now, best to avoid doing any changes. Just add a TODO for me.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/905)*

Issue #912, opened 2017-05-09
Returning uid when no further edge specified (debug mode) 


MichaelJCompton:
Currently 

```
{
	me(id: 0x443e596719408209) {
		director.film 
	}
}
```

won't return the uid of each film in debug mode.  But that seems a clear intent of such a query.

One can fake it with
```
{
	me(id: 0x443e596719408209) {
		director.film { thisIsNotAnEdge }
	}
}
```
or with a real edge.  But these things won't always work if there is an `@cascade` directive used as well.

It would be nice to be able to query for triple patterns like this without having to find an extra edge.

MichaelJCompton:
Actually, I'm not sure if debug mode is the right thing - it seems that wanting `uid`'s in results is more generally applicable than just debugging.  For example, to update some nodes, one would query for the right nodes with `uid`, write out triples and batch an update.

Maybe there could be a directive in the query that switches on `uid` returns and also gives a behaviour like this issue?

Motakjuq:
You can use `_uid_` in the query to get the uid when you are not using debug

MichaelJCompton:
Looks like this also works

```
{
  q(func: allofterms(name, "Peter Jackson")) {
    director.film { _uid_ }
  }
}
```

Thanks to @Motakjuq

pawanrawal:
Now
```
{
	me(id: 0x443e596719408209) {
		director.film 
	}
}
```
Would return all film ids in debug mode.

PR #930, opened 2017-05-11
Support for alias for _uid_

### LINKED ISSUE #861: Alias support for ID
Aliases do not currently work for UID, and quite possibly, XID. That is, the following query:

```
{
    me(id:0x01) {
        id:_uid_
        MyName:name
        gender
        alive
        Buddies:friend {
            BudName:name
        }
    }
}
```

Will not return a result with `id`, but `_uid_` instead, assuming we are using the JSON output. 

From the looks of it, there doesn't seem to be a nice quick fix for this (happy to be told otherwise!) since:

1. The proto models are pretty much set on `Uid`, and `Xid`, so changing the `SetUID(..)` interface is probably not an option
2. But even disregarding the proto node, it seems like the codebase relies heavily on the _knowledge_ of `_uid_`, and there are cases where `SetUID(..)` is called without knowledge of the alias, yet

That being said, it does seem pretty trivial to alter the code to set an additional ID field using the alias without breaking the existing tests. But, that will just result in 2 fields / attributes containing the ID when an alias is used. That is potentially confusing, and inconsistent behaviour, so it is probably not an option (unless the old `_uid_` is deleted).

I think this would be a great feature / fix to have since it is something that works in most _official_ GraphQL implementations.

---

For now, I have somewhat achieved it by doing:

```go
// outputnode.go
type uidAlias interface {
	SetUIDAlias(uid uint64, alias string)
}

func (fj *fastJsonNode) setUID(uid uint64, alias string) {
        // `setUID` is just the original `SetUID`, but without the `_uid_` hardcoding
        // i.e. `fj.attrs["_uid_"]` -> `fj.attrs[alias]`
}

func (fj *fastJsonNode) SetUID(uid uint64) {
	fj.setUID(uid, "_uid_")
}

func (fj *fastJsonNode) SetUIDAlias(uid uint64, alias string) {
	fj.setUID(uid, alias)
	if _, found := fj.attrs["_uid_"]; found {
		delete(fj.attrs, "_uid_")
	}
}
```

And:

```go
// query.go
} else if pc.Attr == "_uid_" {
    if !uidAlreadySet {
        uidAlreadySet = true
        dst.SetUID(uid)
    }
    if ua, ok := dst.(uidAlias); ok && fieldName != "_uid_" {
        ua.SetUIDAlias(uid, fieldName)
    }
}
```

It passes all the tests (even ones that test for the UID). But, I get the feeling that there will be potential edge cases where the attributes will conflict. Needless to say, it does not stop someone from already causing a conflict like:

```
{
    me(id:0x01) {
        _uid_:name
        gender
        _uid_
        alive
        Buddies:friend {
            BudName:name
        }
    }
}
```

pawanrawal:
Fixes #861 and #912.

For proto we have made XID and UID as part of properties so that they are treated like other properties and can have alias. Also remove the IsDebug parameter from SubGraph since GetUid did the same thing.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/930)


pawanrawal:
Review status: 0 of 10 files reviewed at latest revision, 1 unresolved discussion.

---

*[query/query.go, line 294 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/930#-KjtksTxASBVFjOyai5z:-KjtksTyr5aieNxy8Wur:b-h8mmrd) ([raw file](https://github.com/dgraph-io/dgraph/blob/6e85a0244ba1adad412d86608574ddc6e0b872e6/query/query.go#L294)):*
> ```Go
> 
> 	if sg.Params.GetUid {
> 		dst.SetUID(uid, "_uid_")
> ```

Only issue with this would be that for debug endpoint if the user asks for id explicitly with an alias
```
{
  me(id: m.06pj8) {
    n: name
    id: _uid_
  }
}
```
Then we would return both `id` and `_uid_` which is ok because even the UI depends on `_uid_`. If the user just asks for `_uid_` then the entry in the map would be overwritten  if its same or else server would crash because of assert.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/930)*

ashwin95r:
:lgtm:

---

Reviewed 10 of 10 files at r1.

---

*[query/query.go, line 409 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/930#-KjuMNgeEbeymxkYsSlf:-KjuMNgfOORReRshw57T:b-ommbpc) ([raw file](https://github.com/dgraph-io/dgraph/blob/6e85a0244ba1adad412d86608574ddc6e0b872e6/query/query.go#L409)):*
> ```Go
> 			}
> 			for childIdx, childUID := range ul.Uids {
> 				if invalidUids[childUID] || fieldName == "" {
> ```

What's this fieldName check for?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/930#-:-KjuMVET1HZGFNbxCoE1:bnfp4nl)*

ashwin95r:

---

*[query/query.go, line 409 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/930#-KjuMNgeEbeymxkYsSlf:-KjuO1TV_1nAu0vevFPR:b-7mgj8l) ([raw file](https://github.com/dgraph-io/dgraph/blob/6e85a0244ba1adad412d86608574ddc6e0b872e6/query/query.go#L409)):*

What's this fieldName check for?
</blockquote></details>

I don't think we need this `fieldName` check.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/930)*

ashwin95r:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[query/shortest.go, line 415 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/930#-KjucgCZ43I4kIrPDeCQ:-KjucgCZ43I4kIrPDeCR:bsese5o) ([raw file](https://github.com/dgraph-io/dgraph/blob/6e85a0244ba1adad412d86608574ddc6e0b872e6/query/shortest.go#L415)):*
> ```Go
> 	}
> 
> 	node := new(SubGraph)
> ```

Something like this has to be added here: 
```
if nodeInfo.facet != nil {
    // For consistent later processing.
    node.Params.Facet = &protos.Param{}
  }
  uid := result[len(result)-1]
  nodeInfo := dist[uid]
```'
```


---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/930)*

pawanrawal:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[query/query.go, line 409 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/930#-KjuMNgeEbeymxkYsSlf:-Kk8vCJehpT5MP1hyeei:bz5oy3s) ([raw file](https://github.com/dgraph-io/dgraph/blob/6e85a0244ba1adad412d86608574ddc6e0b872e6/query/query.go#L409)):*

I don't think we need this `fieldName` check.
</blockquote></details>

Keeping this or else shortest path tests fail. Will file a bug after merging this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/930)*

janardhan1993:
:lgtm:

---

Reviewed 10 of 10 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[query/query.go, line 294 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/930#-KjtksTxASBVFjOyai5z:-Kk9Zq2GPWrpVr3UOWgn:b-4ijwqj) ([raw file](https://github.com/dgraph-io/dgraph/blob/6e85a0244ba1adad412d86608574ddc6e0b872e6/query/query.go#L294)):*

Only issue with this would be that for debug endpoint if the user asks for id explicitly with an alias
```
{
  me(id: m.06pj8) {
    n: name
    id: _uid_
  }
}
```
Then we would return both `id` and `_uid_` which is ok because even the UI depends on `_uid_`. If the user just asks for `_uid_` then the entry in the map would be overwritten  if its same or else server would crash because of assert.
</blockquote></details>

Instead of crashing if user asks for _uid_ explicity and sets debug, we should throw error may be

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/930#-:-Kk9ZOIyiSBTS6J5334P:bnfp4nl)*

pawanrawal:
Merged

Issue #1071, opened 2017-06-19
Fail in stop and restart cluster


MichaelJCompton:
When I stop individual nodes in a cluster and restart, all seems well.  When I stop all nodes and try to restart the first node brought up crashes.

E.G.

groups.conf
```
default: fp % 3 + 1
```

start servers with

```
~/go/src/github.com/dgraph-io/dgraph/cmd/dgraph/dgraph --group_conf ~/dgraph/groups.conf --groups "0,1" --idx 1 --port 8081 --workerport 12345 --my "127.0.0.1:12345" --debugmode=true --p ~/dgraph/svr1/p --w ~/dgraph/svr1/w > ~/dgraph/svr1/dgraph.log 2>&1 &
```

```
~/go/src/github.com/dgraph-io/dgraph/cmd/dgraph/dgraph --group_conf ~/dgraph/groups.conf --groups "1,2" --idx 2 --port 8082 --workerport 12346 --my "127.0.0.1:12346" --peer "127.0.0.1:12345" --debugmode=true --p ~/dgraph/svr2/p --w ~/dgraph/svr2/w > ~/dgraph/svr2/dgraph.log 2>&1 &
```

```
~/go/src/github.com/dgraph-io/dgraph/cmd/dgraph/dgraph --group_conf ~/dgraph/groups.conf --groups "2,3" --idx 3 --port 8083 --workerport 12347 --my "127.0.0.1:12347" --peer "127.0.0.1:12346" --debugmode=true --p ~/dgraph/svr3/p --w ~/dgraph/svr3/w > ~/dgraph/svr3/dgraph.log 2>&1 &
```

load some data

```
~/go/src/github.com/dgraph-io/dgraph/cmd/dgraphloader/dgraphloader --d 127.0.0.1:8082 --s ~/dgraph/21million.schema --r ~/dgraph/1million.rdf.gz
```

I then waited for the writes on all servers to finish and brought down all the servers with
```
curl -s localhost:8081/admin/shutdown
curl -s localhost:8082/admin/shutdown
curl -s localhost:8083/admin/shutdown
```

When I try to restart, the first node up crashes with

```
2017/06/19 16:38:47 error.go:45: No connection exists

github.com/dgraph-io/dgraph/x.Wrap
	/Users/Michael/go/src/github.com/dgraph-io/dgraph/x/error.go:85
github.com/dgraph-io/dgraph/x.Check
[...]

PR #1112, opened 2017-07-04
Avoid 1701 crash

srh:
This makes us stop crashing the way we do in #1071.  There are still a lot other places where we'll willfully exit the process because a network connection failed -- and shouldn't.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1112)

### DIFF
--- a/worker/conn.go
+++ b/worker/conn.go
@@ -97,11 +97,13 @@ func (p *poolsi) connect(addr string) {
 	resp, err := c.Echo(context.Background(), query)
 	if err != nil {
 		log.Printf("While trying to connect to %q, got error: %v\n", addr, err)
-		return
+		// Don't return -- let's still put the empty pool in the map.  Its users
+		// have to handle errors later anyway.
+	} else {
+		x.AssertTrue(bytes.Equal(resp.Data, query.Data))
+		x.Check(pool.Put(conn))
+		fmt.Printf("Connection with %q successful.\n", addr)
 	}
-	x.AssertTrue(bytes.Equal(resp.Data, query.Data))
-	x.Check(pool.Put(conn))
-	fmt.Printf("Connection with %q successful.\n", addr)
 
 	p.Lock()
 	defer p.Unlock()


srh:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[worker/conn.go, line 104 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1112#-KoAtBX9bU6Dsy4SE6gf:-KoAtBXAlCQFxcZxTQIF:b8qhz0a) ([raw file](https://github.com/dgraph-io/dgraph/blob/9cb151fba6b479cd5eafce30dfc7839c8c783474/worker/conn.go#L104)):*
> ```Go
> 	} else {
> 		x.AssertTrue(bytes.Equal(resp.Data, query.Data))
> 		x.Check(pool.Put(conn))
> ```

You'll note that we call pool.Put(conn) here.  But we don't in the then branch (where before we just returned).

Should the conn be explicitly closed in the former case?  I assumed the existing code knew what it's doing and didn't.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1112)*

srh:
Let me amend my previous comment: While the code *looks* like it willfully exits the process because a network connection (for example, when newPool does a log.Fatal in worker/conn.go) it seems like grpc doesn't really return an error there, even if connecting to the wrong IP or port, or if it can't resolve the hostname.

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1112#-:-KoB1KSdyOMtoAuSLCQZ:b8yba34)*

manishrjain:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[worker/conn.go, line 104 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1112#-KoAtBX9bU6Dsy4SE6gf:-KoBF0RtdOlWfeh8DSBy:bvd8ii5) ([raw file](https://github.com/dgraph-io/dgraph/blob/9cb151fba6b479cd5eafce30dfc7839c8c783474/worker/conn.go#L104)):*

You'll note that we call pool.Put(conn) here.  But we don't in the then branch (where before we just returned).

Should the conn be explicitly closed in the former case?  I assumed the existing code knew what it's doing and didn't.
</blockquote></details>

I think the idea was that we check the health of this connection, before putting it in the pool.

If we can't establish a good connection, we could two things:

- Put it in pool anyway, hoping that it would improve later. This would achieve the same behavior as when we start with a healthy conn, but then it goes unhealthy later.
- Close the connection. But, I wonder what happens when we try to get a conn pool for the addr; and we can't find it. That might lead to some crashes.

If we go the 1st route, we can just avoid testing Echo here, and do Echos in the background to check health of connections periodically.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1112)*

srh:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[worker/conn.go, line 104 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1112#-KoAtBX9bU6Dsy4SE6gf:-KoBGvR1HElerrxehKDT:bp4xb1t) ([raw file](https://github.com/dgraph-io/dgraph/blob/9cb151fba6b479cd5eafce30dfc7839c8c783474/worker/conn.go#L104)):*
In some places we lazily create a pool.  The reason is, in the config that tripped this error, raft group 0 was assigned to only one node.  We have some places, I think, where we want to verify that any connection is "healthy" and other places where we have a specific use for the connection (at which point we'll implicitly see whether it's healthy).  So it might make sense to mark pools that have been freshly created -- those are the ones which we'll actually want to send an echo to make sure the gRPC connection works.  And then maybe background check health of stale connections.

> Close the connection. But, I wonder what happens when we try to get a conn pool for the addr; and we can't find it. That might lead to some crashes.

We'd still have the conn pool -- it might need to create a new connection, but that's a common occurrence.  So no problem there.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1112)*

manishrjain:
:lgtm:

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[worker/conn.go, line 104 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1112#-KoAtBX9bU6Dsy4SE6gf:-KoBPJg-yNld_oNUIJN_:b-w13ugt) ([raw file](https://github.com/dgraph-io/dgraph/blob/9cb151fba6b479cd5eafce30dfc7839c8c783474/worker/conn.go#L104)):*

In some places we lazily create a pool.  The reason is, in the config that tripped this error, raft group 0 was assigned to only one node.  We have some places, I think, where we want to verify that any connection is "healthy" and other places where we have a specific use for the connection (at which point we'll implicitly see whether it's healthy).  So it might make sense to mark pools that have been freshly created -- those are the ones which we'll actually want to send an echo to make sure the gRPC connection works.  And then maybe background check health of stale connections.

> Close the connection. But, I wonder what happens when we try to get a conn pool for the addr; and we can't find it. That might lead to some crashes.

We'd still have the conn pool -- it might need to create a new connection, but that's a common occurrence.  So no problem there.
</blockquote></details>

Alright. SGTM.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1112#-:-KoBPP550Js69XYWKVLq:bnfp4nl)*

srh:
In master as of 7531a5033798144e2199ef8d04ac6fc0e0ccc112.

Issue #1121, opened 2017-07-05
Improve handling of connection failures


srh:
Right now we don't have a way to track the health of a peer and our connectivity towards it, or use that information.

We want the following:
 - To be able to respond to the health of a peer, instead of (for example) sitting around sending queries to a dead peer.  To track our node's opinion of peer reachability.
 - ~Edit: Adding: Use more than just notions of network availability for querying -- we have to wait for a peer to be up-to-date.~  Update: This is handled by #1344.
 - ~To tighten up the management of connection pools (don't keep around the old connection pool when an address has moved).~  (edit: this is done)

~Also, a microbug: `syncMemberships` uses `defer` in a loop right now.~ (edit: this is done)

srh:
According to info online, the gRPC connection object, `grpc.ClientConn`, can be shared among goroutines.  We need to take advantage of this.

One big problem with our connection pool based approach is that it only performs one gRPC call at a time on a given TCP connection.  We also only have 5 spare connections.  That means individual RPC's will often suffer connection handshake latency, and it also means we might run out of file descriptors under high load.  (That would be the case unless query evaluation logic batches operations somehow.)

So, it looks like the right thing to do is:
- Drop connection pooling, and use a single `grpc.ClientConn` concurrently.
- If having only one becomes a problem, have more `grpc.ClientConn`'s that are used concurrently.

I'm going to try having just one `grpc.ClientConn` now.

manishrjain:
Sounds good. Just check how many concurrent requests can be done over just one connection. Is there a limit? Would requests suffer from having just one connection? At what throughput, would you want to have more than 1 connection?

srh:
There is no intrinsic limit to the number of concurrent requests, except there is a limit to the total number of requests per gRPC stream.  The reason is, it uses HTTP 2, which limits each end to making 2^30 streams over the lifetime of the connection.  Stream numbers can't repeat, so, you run out of them, and a new connection has to be made.

One sort of pain from having just one connection is when you get TCP packet loss, and then suddenly all the messages are backed up.  Whereas if you could use multiple streams or react to latency somehow (or use SCTP) you'd avoid that.  We might get single-digit numbers of users complaining about that, see https://github.com/rethinkdb/rethinkdb/issues/5852 for example.

I think you don't get more throughput from more streams unless there's a lot of packet loss or maybe (I don't know) if things are relatively exotic, like having more than one network interface.

srh:
[...]

PR #1159, opened 2017-07-10
Connection pool: one connection (each way), close dead connections

srh:
This makes the connection "pool" hold one gRPC connection per request.

The other thing it does is keep a refcount of what's using a given pool, so that we don't keep connection pools for dead servers forever.

This addresses part of #1121.  The part about tracking and using connection health is not addressed.

Connects to #1121.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1159)

### DIFF
--- a/worker/assign.go
+++ b/worker/assign.go
@@ -78,7 +78,7 @@ func AssignUidsOverNetwork(ctx context.Context, num *protos.Num) (*protos.Assign
 		}
 		return &emptyAssignedIds, err
 	}
-	defer pools().put(p)
+	defer pools().release(p)
 	conn := p.Get()
 	if tr, ok := trace.FromContext(ctx); ok {
 		tr.LazyPrintf("Calling AssignUids for group: %d, addr: %s", leaseGid, addr)
--- a/worker/assign.go
+++ b/worker/assign.go
@@ -78,7 +78,7 @@ func AssignUidsOverNetwork(ctx context.Context, num *protos.Num) (*protos.Assign
 		}
 		return &emptyAssignedIds, err
 	}
-	defer pools().release(p)
+	defer pools().decrRef(p)
 	conn := p.Get()
 	if tr, ok := trace.FromContext(ctx); ok {
 		tr.LazyPrintf("Calling AssignUids for group: %d, addr: %s", leaseGid, addr)
--- a/worker/assign.go
+++ b/worker/assign.go
@@ -78,6 +78,7 @@ func AssignUidsOverNetwork(ctx context.Context, num *protos.Num) (*protos.Assign
 		}
 		return &emptyAssignedIds, err
 	}
+	defer pools().put(p)
 	conn := p.Get()
 	if tr, ok := trace.FromContext(ctx); ok {
 		tr.LazyPrintf("Calling AssignUids for group: %d, addr: %s", leaseGid, addr)
--- a/worker/assign.go
+++ b/worker/assign.go
@@ -71,7 +71,13 @@ func AssignUidsOverNetwork(ctx context.Context, num *protos.Num) (*protos.Assign
 	if tr, ok := trace.FromContext(ctx); ok {
 		tr.LazyPrintf("Not leader of group: %d. Sending to: %d", leaseGid, lid)
 	}
-	p := pools().get(addr)
+	p, err := pools().get(addr)
+	if err != nil {
+		if tr, ok := trace.FromContext(ctx); ok {
+			tr.LazyPrintf("Error while retrieving connection: %+v", err)
+		}
+		return &emptyAssignedIds, err
+	}
 	conn, err := p.Get()
 	if err != nil {
 		if tr, ok := trace.FromContext(ctx); ok {
--- a/worker/assign.go
+++ b/worker/assign.go
@@ -79,7 +79,6 @@ func AssignUidsOverNetwork(ctx context.Context, num *protos.Num) (*protos.Assign
 		return &emptyAssignedIds, err
 	}
 	conn := p.Get()
-	defer p.Put(conn)
 	if tr, ok := trace.FromContext(ctx); ok {
 		tr.LazyPrintf("Calling AssignUids for group: %d, addr: %s", leaseGid, addr)
 	}
--- a/worker/assign.go
+++ b/worker/assign.go
@@ -78,13 +78,7 @@ func AssignUidsOverNetwork(ctx context.Context, num *protos.Num) (*protos.Assign
 		}
 		return &emptyAssignedIds, err
 	}
-	conn, err := p.Get()
-	if err != nil {
-		if tr, ok := trace.FromContext(ctx); ok {
-			tr.LazyPrintf("Error while retrieving connection: %+v", err)
-		}
-		return &emptyAssignedIds, err
-	}
+	conn := p.Get()
 	defer p.Put(conn)
 	if tr, ok := trace.FromContext(ctx); ok {
 		tr.LazyPrintf("Calling AssignUids for group: %d, addr: %s", leaseGid, addr)
--- a/worker/conn.go
+++ b/worker/conn.go
@@ -86,7 +86,7 @@ func (p *poolsi) get(addr string) (*pool, error) {
 }
 
 // One of these must be called for each call to get(...).
-func (p *poolsi) put(pl *pool) {
+func (p *poolsi) release(pl *pool) {
 	// We close the conn after unlocking p.
 	newRefcount := atomic.AddInt64(&pl.refcount, -1)
 	if newRefcount == 0 {
@@ -140,7 +140,7 @@ func (p *poolsi) connect(addr string) (*pool, bool) {
 	// No need to block this thread just to print some messages.
 	pool.AddOwner() // matches p.put() in goroutine
 	go func() {
-		defer p.put(pool)
+		defer p.release(pool)
 		err = TestConnection(pool)
 		if err != nil {
 			log.Printf("Connection to %q fails, got error: %v\n", addr, err)
--- a/worker/conn.go
+++ b/worker/conn.go
@@ -70,11 +70,14 @@ func (p *poolsi) any() *pool {
 	return nil
 }
 
-func (p *poolsi) get(addr string) *pool {
+func (p *poolsi) get(addr string) (*pool, error) {
 	p.RLock()
 	defer p.RUnlock()
-	pool, _ := p.all[addr]
-	return pool
+	pool, ok := p.all[addr]
+	if !ok {
+		return nil, errNoConnection
+	}
+	return pool, nil
 }
 
 func (p *poolsi) connect(addr string) {
--- a/worker/conn.go
+++ b/worker/conn.go
@@ -90,6 +90,16 @@ func (p *poolsi) connect(addr string) {
 
 	pool, err := newPool(addr, 5)
 	x.Checkf(err, "Unable to connect to host %s", addr)
+
+	p.Lock()
+	_, has = p.all[addr]
+	if has {
+		p.Unlock()
+		return
+	}
+	p.all[addr] = pool
+	p.Unlock()
+
 	query := new(protos.Payload)
 	query.Data = make([]byte, 10)
 	x.Check2(rand.Read(query.Data))
@@ -108,14 +118,6 @@ func (p *poolsi) connect(addr string) {
 		x.Check(pool.Put(conn))
 		fmt.Printf("Connection with %q successful.\n", addr)
 	}
-
-	p.Lock()
-	defer p.Unlock()
-	_, has = p.all[addr]
-	if has {
-		return
-	}
-	p.all[addr] = pool
 }
 
 // NewPool creates a new "pool" with one or more gRPC connections.
--- a/worker/conn.go
+++ b/worker/conn.go
@@ -121,7 +121,7 @@ func (p *poolsi) connect(addr string) (*pool, bool) {
 	}
 	p.RUnlock()
 
-	pool, err := newPool(addr, 5)
+	pool, err := newPool(addr)
 	// TODO: This can get triggered with totally bogus config.
 	x.Checkf(err, "Unable to connect to host %s", addr)
 
@@ -174,7 +174,7 @@ func TestConnection(p *pool) error {
 }
 
 // NewPool creates a new "pool" with one gRPC connection, refcount 0.
-func newPool(addr string, maxCap int) (*pool, error) {
+func newPool(addr string) (*pool, error) {
 	conn, err := grpc.Dial(addr, grpc.WithInsecure())
 	if err != nil {
 		return nil, err
--- a/worker/conn.go
+++ b/worker/conn.go
@@ -44,15 +44,7 @@ type pool struct {
 
 	Addr     string
 	refcount int
-	health   poolHealth
 }
-
-type poolHealth int
-
-const (
-	healthy = iota
-	unhealthy
-)
 
 type poolsi struct {
 	sync.RWMutex
@@ -173,13 +165,11 @@ func TestConnection(p *pool) error {
 	c := protos.NewWorkerClient(conn)
 	resp, err := c.Echo(context.Background(), query)
 	if err != nil {
-		p.health = unhealthy
 		return err
 	}
 	// If a server is sending bad echos, do we have to freak out and die?
 	x.AssertTruef(bytes.Equal(resp.Data, query.Data),
 		"non-matching Echo response value from %v", p.Addr)
-	p.health = healthy
 	return nil
 }
 
@@ -190,14 +180,10 @@ func newPool(addr string, maxCap int) (*pool, error) {
 		return nil, err
 	}
 	// The pool hasn't been added to poolsi yet, so it gets no refcount.
-	return &pool{conn: conn, Addr: addr, refcount: 0, health: healthy}, nil
+	return &pool{conn: conn, Addr: addr, refcount: 0}, nil
 }
 
 // Get returns the connection to use from the pool of connections.
 func (p *pool) Get() *grpc.ClientConn {
 	return p.conn
-}
-
-func (p *pool) Health() poolHealth {
-	return p.health
 }
--- a/worker/conn.go
+++ b/worker/conn.go
@@ -93,11 +93,6 @@ func (p *poolsi) get(addr string) (*pool, error) {
 
 // One of these must be called for each call to get(...).
 func (p *poolsi) put(pl *pool) {
-	p.decrRefcountAndCleanup(pl)
-}
-
-// You must _not_ hold the lock on p.
-func (p *poolsi) decrRefcountAndCleanup(pl *pool) {
 	// We close the conn after unlocking p.
 	p.RLock()
 	pl.refcount--
@@ -119,20 +114,21 @@ func destroyPool(pl *pool) {
 }
 
 // Returns a pool that you should call put() on.
-func (p *poolsi) connect(addr string) *pool {
+func (p *poolsi) connect(addr string) (*pool, bool) {
 	if addr == *myAddr {
-		return nil
+		return nil, false
 	}
 	p.RLock()
 	existingPool, has := p.all[addr]
 	if has {
 		existingPool.refcount++
 		p.RUnlock()
-		return existingPool
+		return existingPool, true
 	}
 	p.RUnlock()
 
 	pool, err := newPool(addr, 5)
+	// TODO: This can get triggered with totally bogus config.
 	x.Checkf(err, "Unable to connect to host %s", addr)
 
 	p.Lock()
@@ -141,7 +137,7 @@ func (p *poolsi) connect(addr string) *pool {
 		p.Unlock()
 		destroyPool(pool)
 		existingPool.refcount++
-		return existingPool
+		return existingPool, true
 	}
 	p.all[addr] = pool
 	// TODO: Callers should decrement this refcount.
@@ -150,14 +146,13 @@ func (p *poolsi) connect(addr string) *pool {
 
 	err = TestConnection(pool)
 	if err != nil {
-
 		log.Printf("Connection to %q fails, got error: %v\n", addr, err)
 		// Don't return -- let's still put the empty pool in the map.  Its users
 		// have to handle errors later anyway.
 	} else {
 		fmt.Printf("Connection with %q healthy.\n", addr)
 	}
-	return pool
+	return pool, true
 }
 
 // TestConnection tests if we can run an Echo query on a connection.
--- a/worker/conn.go
+++ b/worker/conn.go
@@ -132,7 +132,6 @@ func (p *poolsi) connect(addr string) (*pool, bool) {
 		return existingPool, true
 	}
 	p.all[addr] = pool
-	// TODO: Callers should decrement this refcount.
 	pool.refcount++
 	// This refcount gets decremented by the goroutine here
 	pool.refcount++
--- a/worker/conn.go
+++ b/worker/conn.go
@@ -42,8 +42,9 @@ type pool struct {
 	// messages in the same TCP stream.
 	conn *grpc.ClientConn
 
-	Addr   string
-	health poolHealth
+	Addr     string
+	refcount int
+	health   poolHealth
 }
 
 type poolHealth int
@@ -73,6 +74,7 @@ func (p *poolsi) any() (*pool, error) {
 	p.RLock()
 	defer p.RUnlock()
 	for _, pool := range p.all {
+		pool.refcount++
 		return pool, nil
 	}
 	return nil, errNoConnection
@@ -85,25 +87,50 @@ func (p *poolsi) get(addr string) (*pool, error) {
 	if !ok {
 		return nil, errNoConnection
 	}
+	pool.refcount++
 	return pool, nil
 }
 
 // One of these must be called for each call to get(...).
-// TODO: Make put include information about pool health.
-func (p *poolsi) put(_ *pool) {
-	// nothing to do.
+func (p *poolsi) put(pl *pool) {
+	p.decrRefcountAndCleanup(pl)
 }
 
+// You must _not_ hold the lock on p.
+func (p *poolsi) decrRefcountAndCleanup(pl *pool) {
+	// We close the conn after unlocking p.
+	p.RLock()
+	pl.refcount--
+	if pl.refcount == 0 {
+		delete(p.all, pl.Addr)
+	}
+	p.RUnlock()
+
+	if pl.refcount == 0 {
+		destroyPool(pl)
+	}
+}
+
+func destroyPool(pl *pool) {
+	err := pl.conn.Close()
+	if err != nil {
+		log.Printf("Error closing cluster connection: %v\n", err.Error())
+	}
+}
+
+// Returns a pool that you should call put() on.
 func (p *poolsi) connect(addr string) *pool {
 	if addr == *myAddr {
 		return nil
 	}
 	p.RLock()
 	existingPool, has := p.all[addr]
-	p.RUnlock()
 	if has {
+		existingPool.refcount++
+		p.RUnlock()
 		return existingPool
 	}
+	p.RUnlock()
 
 	pool, err := newPool(addr, 5)
 	x.Checkf(err, "Unable to connect to host %s", addr)
@@ -112,9 +139,13 @@ func (p *poolsi) connect(addr string) *pool {
 	existingPool, has = p.all[addr]
 	if has {
 		p.Unlock()
+		destroyPool(pool)
+		existingPool.refcount++
 		return existingPool
 	}
 	p.all[addr] = pool
+	// TODO: Callers should decrement this refcount.
+	pool.refcount++
 	p.Unlock()
 
 	err = TestConnection(pool)
@@ -150,13 +181,14 @@ func TestConnection(p *pool) error {
 	return nil
 }
 
-// NewPool creates a new "pool" with one gRPC connection.
+// NewPool creates a new "pool" with one gRPC connection, refcount 0.
 func newPool(addr string, maxCap int) (*pool, error) {
 	conn, err := grpc.Dial(addr, grpc.WithInsecure())
 	if err != nil {
 		return nil, err
 	}
-	return &pool{conn: conn, Addr: addr}, nil
+	// The pool hasn't been added to poolsi yet, so it gets no refcount.
+	return &pool{conn: conn, Addr: addr, refcount: 0, health: healthy}, nil
 }
 
 // Get returns the connection to use from the pool of connections.
--- a/worker/conn.go
+++ b/worker/conn.go
@@ -42,8 +42,16 @@ type pool struct {
 	// messages in the same TCP stream.
 	conn *grpc.ClientConn
 
-	Addr string
+	Addr   string
+	health poolHealth
 }
+
+type poolHealth int
+
+const (
+	healthy = iota
+	unhealthy
+)
 
 type poolsi struct {
 	sync.RWMutex
@@ -111,6 +119,7 @@ func (p *poolsi) connect(addr string) *pool {
 
 	err = TestConnection(pool)
 	if err != nil {
+
 		log.Printf("Connection to %q fails, got error: %v\n", addr, err)
 		// Don't return -- let's still put the empty pool in the map.  Its users
 		// have to handle errors later anyway.
@@ -131,11 +140,13 @@ func TestConnection(p *pool) error {
 	c := protos.NewWorkerClient(conn)
 	resp, err := c.Echo(context.Background(), query)
 	if err != nil {
+		p.health = unhealthy
 		return err
 	}
-	// If a server is sending bad echos, we do want to freak out and die.
+	// If a server is sending bad echos, do we have to freak out and die?
 	x.AssertTruef(bytes.Equal(resp.Data, query.Data),
 		"non-matching Echo response value from %v", p.Addr)
+	p.health = healthy
 	return nil
 }
 
@@ -151,4 +162,8 @@ func newPool(addr string, maxCap int) (*pool, error) {
 // Get returns the connection to use from the pool of connections.
 func (p *pool) Get() *grpc.ClientConn {
 	return p.conn
+}
+
+func (p *pool) Health() poolHealth {
+	return p.health
 }
--- a/worker/conn.go
+++ b/worker/conn.go
@@ -35,15 +35,13 @@ var (
 	errNoConnection = fmt.Errorf("No connection exists")
 )
 
-// Pool is used to manage the grpc client connections for communicating with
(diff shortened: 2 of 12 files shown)


srh:
I recommend waiting for an update with basic fixes before reviewing.

---

Review status: 0 of 12 files reviewed at latest revision, 10 unresolved discussions.

---

*[worker/conn.go, line 47 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-KoesR6yZgmaF8RMsxMo:-KoesR6yZgmaF8RMsxMp:b-h6o9hj) ([raw file](https://github.com/dgraph-io/dgraph/blob/0fc8c8a6b7c7155ac5279e24553c52df3b7c7e51/worker/conn.go#L47)):*
> ```Go
> 	Addr     string
> 	refcount int
> 	health   poolHealth
> ```

This health field is a stunted health field, just a placeholder.  I should remove it before merging to master.

---

*[worker/conn.go, line 143 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Koet2cElJVk6fzpFWKA:-Koet2cElJVk6fzpFWKB:bk9v7da) ([raw file](https://github.com/dgraph-io/dgraph/blob/0fc8c8a6b7c7155ac5279e24553c52df3b7c7e51/worker/conn.go#L143)):*
> ```Go
> 	}
> 	p.all[addr] = pool
> 	// TODO: Callers should decrement this refcount.
> ```

This TODO is resolved, needs to be removed.

---

*[worker/conn.go, line 176 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-KoetC3LJRhTIkYhZ_xC:-KoetC3LJRhTIkYhZ_xD:b-wbd1xf) ([raw file](https://github.com/dgraph-io/dgraph/blob/0fc8c8a6b7c7155ac5279e24553c52df3b7c7e51/worker/conn.go#L176)):*
> ```Go
> 	resp, err := c.Echo(context.Background(), query)
> 	if err != nil {
> 		p.health = unhealthy
> ```

I mentioned this health field should be removed before these changes are merged into master, but also its access should be in some kind of lock.

---

*[worker/draft.go, line 51 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-KoetRcmV59oXrhp7RRz:-KoetRcmV59oXrhp7RS-:b-mj4llm) ([raw file](https://github.com/dgraph-io/dgraph/blob/0fc8c8a6b7c7155ac5279e24553c52df3b7c7e51/worker/draft.go#L51)):*
> ```Go
> type peerPoolEntry struct {
> 	peer string
> 	pool *pool
> ```

Should comment that this is an owning pointer.

---

*[worker/draft.go, line 226 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Koet_gzWZYRSn3f_dtA:-Koet_gzWZYRSn3f_dtB:bhfvdil) ([raw file](https://github.com/dgraph-io/dgraph/blob/473eddb72ed56356cab8cb7469ad711fa081aa5b/worker/draft.go#L226)):*
> ```Go
> 
> func (n *node) Connect(pid uint64, addr string) {
> 	for n == nil {
> ```

FYI: (Made it so n can't be nil.)

---

*[worker/draft.go, line 241 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Koetdfsa1r_eVPZQVuA:-Koetdfsa1r_eVPZQVuB:bz835n0) ([raw file](https://github.com/dgraph-io/dgraph/blob/0fc8c8a6b7c7155ac5279e24553c52df3b7c7e51/worker/draft.go#L241)):*
> ```Go
> 
> // p can be nil
> func (n *node) SetPeer(pid uint64, addr string, p *pool) {
> ```

Should rename p parameter to pOrNil.

---

*[worker/draft.go, line 263 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-KoetyH0vunIahEgVJfQ:-KoetyH0vunIahEgVJfR:bcq97en) ([raw file](https://github.com/dgraph-io/dgraph/blob/0fc8c8a6b7c7155ac5279e24553c52df3b7c7e51/worker/draft.go#L263)):*
> ```Go
> 	addr, ok := n.GetPeer(pid)
> 	x.AssertTruef(ok, "Unable to find conn pool for peer: %d", pid)
> 	// TODO: Make a peerPool invariant that addrs not empty
> ```

This might be an invariant which has been handled, and the TODO should be removed.

---

*[worker/draft.go, line 474 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Koeu7hoIX4qNLzbDwfD:-Koeu7hoIX4qNLzbDwfE:b-c9dnbi) ([raw file](https://github.com/dgraph-io/dgraph/blob/0fc8c8a6b7c7155ac5279e24553c52df3b7c7e51/worker/draft.go#L474)):*
> ```Go
> 		return
> 	}
> 	// TODO: Make this an invariant.
> ```

With this assertion in multiple places, a runtime assertion should be handled within GetPeer.

---

*[worker/groups.go, line 66 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Koev-0VUWpDmDZ8k9-T:-Koev-0VUWpDmDZ8k9-U:b4g2vw5) ([raw file](https://github.com/dgraph-io/dgraph/blob/0fc8c8a6b7c7155ac5279e24553c52df3b7c7e51/worker/groups.go#L66)):*
> ```Go
> 
> type servers struct {
> 	byNodeID map[uint64]int
> ```

Should add a comment...

---

*[worker/worker.go, line 66 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-KoeuNu-fSOIq6cylkn5:-KoeuNu-fSOIq6cylkn6:biq3j3q) ([raw file](https://github.com/dgraph-io/dgraph/blob/0fc8c8a6b7c7155ac5279e24553c52df3b7c7e51/worker/worker.go#L66)):*
> ```Go
> 		grpc.MaxRecvMsgSize(x.GrpcMaxSize),
> 		grpc.MaxSendMsgSize(x.GrpcMaxSize),
> 		grpc.MaxConcurrentStreams(math.MaxInt32))
> ```

FYI: Here's where we set MaxConcurrentStreams (up from the default of 100).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-:-KoevHTOxKsvwJ70NSzx:b-hhi2d2)*

srh:
Review status: 0 of 12 files reviewed at latest revision, 2 unresolved discussions.

---

*[worker/draft.go, line 474 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Koeu7hoIX4qNLzbDwfD:-Kof2Of4IKIqvOxwM5h_:bw0k9a5) ([raw file](https://github.com/dgraph-io/dgraph/blob/0fc8c8a6b7c7155ac5279e24553c52df3b7c7e51/worker/draft.go#L474)):*

With this assertion in multiple places, a runtime assertion should be handled within GetPeer.
</blockquote></details>

I put it in SetPeer.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159)*

srh:
(Okay, the update has been made.)

---

Review status: 0 of 12 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-:-Kof2a2XGyycU5FcGnCD:bpgguu9)*

manishrjain:
Reviewed 9 of 12 files at r1, 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[worker/assign.go, line 81 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof55tVlGHKW9QHbWiM:-Kof55tVlGHKW9QHbWiN:b-aqdjbq) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/assign.go#L81)):*
> ```Go
> 		return &emptyAssignedIds, err
> 	}
> 	defer pools().put(p)
> ```

So, does pools() hold multiple copies of p for the same address?

---

*[worker/conn.go, line 89 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof82lvrFijk4Lzkvvw:-Kof82lvrFijk4Lzkvvx:b-dj4c8i) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/conn.go#L89)):*
> ```Go
> 
> // One of these must be called for each call to get(...).
> func (p *poolsi) put(pl *pool) {
> ```

Call it `decrRef()`, which is more in sync with what's happening here.

---

*[worker/conn.go, line 99 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof5Us4oaUx4mOV-NLA:-Kof5Us4oaUx4mOV-NLB:bj6imbl) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/conn.go#L99)):*
> ```Go
> 
> 	if newRefcount == 0 {
> 		destroyPool(pl)
> ```

This if block can be merged with above if block. But, why close connection? Why not keep it around?

We might get a query at any point, and if we need to dial a new connection, it would make our query latency bad (also, because when the query is done and you push put the connection back, it would again destroy it).

If we don't close a connection, we also don't need to bother with ref counting.

---

*[worker/conn.go, line 124 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof8MaLvMFNGfVTA_6d:-Kof8MaLvMFNGfVTA_6e:bw5n71) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/conn.go#L124)):*
> ```Go
> 	p.RUnlock()
> 
> 	pool, err := newPool(addr, 5)
> ```

I think we need a refactor of naming. This is actually `newConn(...)`, instead of `newPool`.

---

*[worker/conn.go, line 144 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof7o-DpciEXfxcfROA:-Kof7o-DpciEXfxcfROB:bren1tr) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/conn.go#L144)):*
> ```Go
> 	go func() {
> 		defer p.put(pool)
> 		err = TestConnection(pool)
> ```

Why bother with TestConnection? Its result has no effect on us anyway.

---

*[worker/conn.go, line 158 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof8ZrRhpUcgF26Omf6:-Kof8ZrRhpUcgF26Omf7:b-9kbrix) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/conn.go#L158)):*
> ```Go
> 
> // TestConnection tests if we can run an Echo query on a connection.
> func TestConnection(p *pool) error {
> ```

Based on current usage, I don't see the point of this function. But, maybe we want to do regular pings to see if this connection holds and use that information for health checks etc.

---

*[worker/conn.go, line 177 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof6zN38-Zt6ZzIQPGu:-Kof6zN38-Zt6ZzIQPGv:b-plfkr) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/conn.go#L177)):*
> ```Go
> 
> // NewPool creates a new "pool" with one gRPC connection, refcount 0.
> func newPool(addr string, maxCap int) (*pool, error) {
> ```

Remove the maxCap argument.

---

*[worker/draft.go, line 226 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Koet_gzWZYRSn3f_dtA:-Kof92hR4ADjYg_kMSQY:b2nhktp) ([raw file](https://github.com/dgraph-io/dgraph/blob/473eddb72ed56356cab8cb7469ad711fa081aa5b/worker/draft.go#L226)):*

FYI: (Made it so n can't be nil.)
</blockquote></details>

OK.

---

*[worker/draft.go, line 241 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Koetdfsa1r_eVPZQVuA:-Kof9uAihqmxIRjjKyXB:b32w0x) ([raw file](https://github.com/dgraph-io/dgraph/blob/0fc8c8a6b7c7155ac5279e24553c52df3b7c7e51/worker/draft.go#L241)):*

Should rename p parameter to pOrNil.
</blockquote></details>

OK

---

*[worker/draft.go, line 50 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof8mwqgHg-U2_L_8IO:-Kof8mwqgHg-U2_L_8IP:bthennn) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/draft.go#L50)):*
> ```Go
> 
> type peerPoolEntry struct {
> 	peer string
> ```

peer is already in the struct name. I think this is `addr`.

---

*[worker/draft.go, line 52 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof9SffrOiWLGe5hme9:-Kof9SffrOiWLGe5hmeA:b-qwtnld) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/draft.go#L52)):*
> ```Go
> 	peer string
> 	// An owning reference to a pool for this peer (or nil if self-peer).
> 	poolOrNil *pool
> ```

Why do we have a pointer to the pool? We don't seem to be using it.

In fact, might make sense to have node hold a connection to its peers directly; while the queries use the common pool. That way, mutations and queries won't be blocked on each other, if we exceed the 1000 limit of concurrent requests per connection.

---

*[worker/draft.go, line 62 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof8x5AyVSmwIzSrMk6:-Kof8x5BBhb0OkaEUz8d:b-im3ypx) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/draft.go#L62)):*
> ```Go
> }
> 
> func (p *peerPool) get(id uint64) (string, bool) {
> ```

Do we really want just the address? Most likely we need a connection to that peer.

---

*[worker/draft.go, line 668 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof9NZEvor1ZJb18eAQ:-Kof9NZEvor1ZJb18eAR:bcmyzst) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/draft.go#L668)):*
> ```Go
> 
> func (n *node) retrieveSnapshot(rc protos.RaftContext) {
> 	addr, ok := n.GetPeer(rc.Id)
> ```

Why bother returning address, if we're going to get a pool anyway.

---

*[worker/groups.go, line 66 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Koev-0VUWpDmDZ8k9-T:-KofA5qS2RZ1MZ2kvCzI:b-ij39yc) ([raw file](https://github.com/dgraph-io/dgraph/blob/0fc8c8a6b7c7155ac5279e24553c52df3b7c7e51/worker/groups.go#L66)):*

Should add a comment...
</blockquote></details>

I think we could use `container.List`, which can allow byNodeID to have a direct pointer to the `list.Element`. So,removal of servers from the list, or addition of them in the front, etc., would result in simpler code.

For example, for addToServers:

```
elem := sl.list.PushBack(server)
byNodeID[id] = elem
if update.Leader {
  sl.list.MoveToFront(elem)
}

```

---

*[worker/groups.go, line 92 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-KofBl4FmG1bk8ggzcnO:-KofBl4FmG1bk8ggzcnP:b-ejcg4l) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/groups.go#L92)):*
> ```Go
> }
> 
> func swapServers(sl *servers, i int, j int) {
> ```

Won't need swapServers using container/list.

---

*[worker/groups.go, line 94 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-KofBrxat3sNNWjr5Ayy:-KofBrxat3sNNWjr5Ayz:b-mttafu) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/groups.go#L94)):*
> ```Go
> func swapServers(sl *servers, i int, j int) {
> 	tmp := sl.list[i]
> 	sl.list[i] = sl.list[j]
> ```

Swap can be done in Go like this:

```
sl.list[i], sl.list[j] = sl.list[j], sl.list[i]
```

---

*[worker/groups.go, line 100 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-KofBRLPcDjwrDuvLGbp:-KofBRLPcDjwrDuvLGbq:b-c69djx) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/groups.go#L100)):*
> ```Go
> }
> 
> func removeFromServersIfPresent(sl *servers, nodeID uint64) {
> ```

With container/list:

```
ele, has := sl.byNodeID[nodeID]
if !has {
  return
}
s := sl.list.Remove(ele).(*server)
pools().decrRef(s.PoolOrNil)
```

---

*[worker/groups.go, line 491 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-KofCFsz3bURLLsithLV:-KofCFsz3bURLLsithLW:b10uymn) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/groups.go#L491)):*
> ```Go
> 	var update *protos.MembershipUpdate
> 	for {
> 		func() {
> ```

This is not a defer. So, if you just need a block to restrict the scope of vars, you can do that without creating a function.

---

*[worker/groups.go, line 497 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-KofCQpnLV6HRM9S146m:-KofCQpnLV6HRM9S146n:bck5gmt) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/groups.go#L497)):*
> ```Go
> 
> 			c := protos.NewWorkerClient(conn)
> 			update, err = c.UpdateMembership(g.ctx, &mu)
> ```

call pools().put(pl) here. No need for defer (pretty simple 2 line code).

---

*[worker/worker.go, line 66 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-KoeuNu-fSOIq6cylkn5:-KofCa7OKcURJrg6CEOM:bs2xpgl) ([raw file](https://github.com/dgraph-io/dgraph/blob/0fc8c8a6b7c7155ac5279e24553c52df3b7c7e51/worker/worker.go#L66)):*

FYI: Here's where we set MaxConcurrentStreams (up from the default of 100).
</blockquote></details>

OK. Max int sounds good for internal worker comm.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159)*

srh:
Review status: all files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[worker/assign.go, line 81 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof55tVlGHKW9QHbWiM:-KofUyPmLGRRJaRae_Oi:b1a0h2h) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/assign.go#L81)):*

So, does pools() hold multiple copies of p for the same address?
</blockquote></details>

I'm not sure that get and put have that implication.

---

*[worker/conn.go, line 89 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof82lvrFijk4Lzkvvw:-KofV3f1FgkEvhZgSCxl:b-kz7lid) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/conn.go#L89)):*

Call it `decrRef()`, which is more in sync with what's happening here.
</blockquote></details>

So we have `get`/`decrRef` pairs?  Or maybe that should be `getRef`/`decrRef`.  I think `get`/`release` is best.

---

*[worker/conn.go, line 99 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof5Us4oaUx4mOV-NLA:-KofNed3wBW7SWoDDECg:b-46ngwb) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/conn.go#L99)):*

This if block can be merged with above if block. But, why close connection? Why not keep it around?

We might get a query at any point, and if we need to dial a new connection, it would make our query latency bad (also, because when the query is done and you push put the connection back, it would again destroy it).

If we don't close a connection, we also don't need to bother with ref counting.
</blockquote></details>

As long as a node is in a group a connection will be kept open.  Only if you remove a node from all groups will it be removed.  Are there nodes that aren't part of any group?  (They only run queries?)

---

*[worker/conn.go, line 124 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof8MaLvMFNGfVTA_6d:-KofVPE-JOa1Tlm8To8B:b8fjvup) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/conn.go#L124)):*

I think we need a refactor of naming. This is actually `newConn(...)`, instead of `newPool`.
</blockquote></details>

We could rename `pool` to `connection`.  But I want to minimize code churn here.

---

*[worker/conn.go, line 144 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof7o-DpciEXfxcfROA:-KofNyvE3SK61LGZr_I9:b-tj4dqq) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/conn.go#L144)):*

Why bother with TestConnection? Its result has no effect on us anyway.
</blockquote></details>

I was thinking the same thing.

---

*[worker/draft.go, line 52 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof9SffrOiWLGe5hme9:-KofO_2Ip9odVtpmeHlr:b-tm7b5h) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/draft.go#L52)):*

Why do we have a pointer to the pool? We don't seem to be using it.

In fact, might make sense to have node hold a connection to its peers directly; while the queries use the common pool. That way, mutations and queries won't be blocked on each other, if we exceed the 1000 limit of concurrent requests per connection.
</blockquote></details>

What limit are you referring to?

---

*[worker/draft.go, line 62 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof8x5AyVSmwIzSrMk6:-KofU0rMKnGbLLlI_ePO:b-hdaucs) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/draft.go#L62)):*

Do we really want just the address? Most likely we need a connection to that peer.
</blockquote></details>

Agreed, I thought I was going to do that.

---

*[worker/groups.go, line 66 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Koev-0VUWpDmDZ8k9-T:-KofQrkbXxQU_PICGSIH:b-poulwj) ([raw file](https://github.com/dgraph-io/dgraph/blob/0fc8c8a6b7c7155ac5279e24553c52df3b7c7e51/worker/groups.go#L66)):*

I think we could use `container.List`, which can allow byNodeID to have a direct pointer to the `list.Element`. So,removal of servers from the list, or addition of them in the front, etc., would result in simpler code.

For example, for addToServers:

```
elem := sl.list.PushBack(server)
byNodeID[id] = elem
if update.Leader {
  sl.list.MoveToFront(elem)
}

```
</blockquote></details>

AnyServer, which randomly selects from the list, wouldn't be efficient that way.  Even without that, it's trading having compile-time type info in exchange for the convenience of not implementing swapping.

---

*[worker/groups.go, line 497 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-KofCQpnLV6HRM9S146m:-KofShPCn84SwKtstgaT:bp55t8l) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/groups.go#L497)):*

call pools().put(pl) here. No need for defer (pretty simple 2 line code).
</blockquote></details>

So, what you're saying here is, we can assume that NewWorkerClient and UpdateMembership won't panic?  Is that because they wouldn't panic across a package boundary?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159)*

srh:
Review status: all files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[worker/draft.go, line 52 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof9SffrOiWLGe5hme9:-KofYi3lpP3HWikHBUuA:b-85yziu) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/draft.go#L52)):*

What limit are you referring to?
</blockquote></details>

We have a pointer because it's an owning reference and you have to call put/release/decrRef on it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159)*

srh:
Review status: 2 of 12 files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[worker/conn.go, line 144 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof7o-DpciEXfxcfROA:-KofaQYZfxYn7aWVFm9-:bsi1kgu) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/conn.go#L144)):*

I was thinking the same thing.
</blockquote></details>

I think we should keep it for now.  Is there other more useful output of connection health?

---

*[worker/conn.go, line 177 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof6zN38-Zt6ZzIQPGu:-KofaY-6WAZPkSauxHLG:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/conn.go#L177)):*

Remove the maxCap argument.
</blockquote></details>

Done.

---

*[worker/draft.go, line 50 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof8mwqgHg-U2_L_8IO:-KofaZ-9t1-vlbl7xPwC:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/draft.go#L50)):*

peer is already in the struct name. I think this is `addr`.
</blockquote></details>

Done.

---

*[worker/draft.go, line 62 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof8x5AyVSmwIzSrMk6:-KofabUiEJeyP-YxO2fp:b-e47ozv) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/draft.go#L62)):*

Agreed, I thought I was going to do that.
</blockquote></details>

Okay, there are a couple of reasons.  One is, we do need a method to get the address -- we do get the address in one case to construct a Raft configuration change update.  The other is, or was, some conservatism of the code change here -- the existing code handled the case where a different peer allegedly has the same address as ourselves rather tolerantly, and the result is that peerPoolEntry can have a nil address.  How about I put the fix for that, for which I'd rather have a separate bisectable commit, in a separate PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159)*

manishrjain:
Review status: 2 of 12 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[worker/conn.go, line 47 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-KoesR6yZgmaF8RMsxMo:-Koj9ok-DPxaiExgj40g:b32w0x) ([raw file](https://github.com/dgraph-io/dgraph/blob/0fc8c8a6b7c7155ac5279e24553c52df3b7c7e51/worker/conn.go#L47)):*

This health field is a stunted health field, just a placeholder.  I should remove it before merging to master.
</blockquote></details>

OK

---

*[worker/conn.go, line 143 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Koet2cElJVk6fzpFWKA:-KojA_-EPUMu71wulptM:b2nhktp) ([raw file](https://github.com/dgraph-io/dgraph/blob/0fc8c8a6b7c7155ac5279e24553c52df3b7c7e51/worker/conn.go#L143)):*

This TODO is resolved, needs to be removed.
</blockquote></details>

OK.

---

*[worker/conn.go, line 89 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof82lvrFijk4Lzkvvw:-Koj9s0fhoaItKDfGwq9:b-2g4bjs) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/conn.go#L89)):*

So we have `get`/`decrRef` pairs?  Or maybe that should be `getRef`/`decrRef`.  I think `get`/`release` is best.
</blockquote></details>

get / release SGTM.

---

*[worker/conn.go, line 99 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof5Us4oaUx4mOV-NLA:-KojA52YjunvUbK1akBr:bbwe4u1) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/conn.go#L99)):*

As long as a node is in a group a connection will be kept open.  Only if you remove a node from all groups will it be removed.  Are there nodes that aren't part of any group?  (They only run queries?)
</blockquote></details>

OK.

No. All nodes are part of some group. If a server stops serving a group, that node is closed. I reckon that'd be when the connection would be closed.

---

*[worker/conn.go, line 124 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof8MaLvMFNGfVTA_6d:-KojAGQm0cvXfkDb-C81:b8sjxb4) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/conn.go#L124)):*

We could rename `pool` to `connection`.  But I want to minimize code churn here.
</blockquote></details>

OK. Add a TODO then to do the name refactoring. Once you submit this PR, you can send a refactor.

---

*[worker/conn.go, line 144 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof7o-DpciEXfxcfROA:-KojAcE7-oKJgNCrqk1l:b-h8pnhr) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/conn.go#L144)):*

I think we should keep it for now.  Is there other more useful output of connection health?
</blockquote></details>

The idea for conn health was to avoid sending it queries if it's down or slow. Mutations would still need to be sent.

---

*[worker/draft.go, line 52 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof9SffrOiWLGe5hme9:-KojAz49Wbs_kQWkggVi:b-mvd6hc) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/draft.go#L52)):*

We have a pointer because it's an owning reference and you have to call put/release/decrRef on it.
</blockquote></details>

OK. Re: the limit, for some reason I thought we have a 1000 max per conn. But, that's not the case. We're doing unlimited concurrent requests per conn. So, ignore that comment.

---

*[worker/draft.go, line 62 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof8x5AyVSmwIzSrMk6:-KojBGvA6mzjnXFvk6ve:bg7twah) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/draft.go#L62)):*

Okay, there are a couple of reasons.  One is, we do need a method to get the address -- we do get the address in one case to construct a Raft configuration change update.  The other is, or was, some conservatism of the code change here -- the existing code handled the case where a different peer allegedly has the same address as ourselves rather tolerantly, and the result is that peerPoolEntry can have a nil address.  How about I put the fix for that, for which I'd rather have a separate bisectable commit, in a separate PR.
</blockquote></details>

OK. If it's going to be part of a separate PR, add a TODO and work on it after submitting this.

---

*[worker/draft.go, line 668 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-Kof9NZEvor1ZJb18eAQ:-KojPH1FciLogACD3wzF:b-jijldc) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/draft.go#L668)):*

Why bother returning address, if we're going to get a pool anyway.
</blockquote></details>

In fact, if all we're using the GetPeer() is to get the address (same as before), then node shouldn't be the one holding the connection anyway. We can always retrieve it from the `pools()`. I'm not sure we really need ref counting. If we look carefully at what causes a server to be removed from the cluster, that would help figure out when to disconnect.

---

*[worker/groups.go, line 497 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-KofCQpnLV6HRM9S146m:-KojQJwFgTSKFgREU5JH:bjcn5vm) ([raw file](https://github.com/dgraph-io/dgraph/blob/29971ae79bee33461a5f5943a301b692a6255266/worker/groups.go#L497)):*

So, what you're saying here is, we can assume that NewWorkerClient and UpdateMembership won't panic?  Is that because they wouldn't panic across a package boundary?
</blockquote></details>

If they panic, the server would crash anyway. We don't catch panics. But, they shouldn't, if error handling is done nicely.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159)*

manishrjain:
:lgtm:

---

Reviewed 10 of 10 files at r3.
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1159#-:-KojVKjqhuCp4ic1fBWm:bnfp4nl)*

srh:
In master as of a0d178a512096230f76382805deed1103561de40.

PR #1174, opened 2017-07-11
Add checkpoint support for dgraphloader

pawanrawal:
Our each batch request has RDF's from different files and with non-sequential line numbers. So we need watermarks to keep track of the last line number for each file that the server has processed. 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1174)

### DIFF
--- a/client/checkpoint.go
+++ b/client/checkpoint.go
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Dgraph Labs, Inc.
+ * Copyright 2017 Dgraph Labs, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/client/checkpoint.go
+++ b/client/checkpoint.go
@@ -26,6 +26,10 @@ import (
 	"github.com/dgraph-io/dgraph/x"
 )
 
+// A watermark for each file
+type syncMarks map[string]*x.WaterMark
+
+// Create syncmarks for files and store them in dgraphClient.
 func (d *Dgraph) NewSyncMarks(files []string) {
 	for _, file := range files {
 		bp := filepath.Base(file)
@@ -33,10 +37,13 @@ func (d *Dgraph) NewSyncMarks(files []string) {
 	}
 }
 
+// Get checkpoint for file from Badger.
 func (d *Dgraph) Checkpoint(file string) (uint64, error) {
 	return d.alloc.getFromKV(fmt.Sprintf("checkpoint-%s", file))
 }
 
+// This is called to syncAllMarks. It is useful to find the final line that was processed for each
+// file.
 func (d *Dgraph) syncAllMarks() {
 	for _, mark := range d.marks {
 		for mark.WaitingFor() {
@@ -45,17 +52,19 @@ func (d *Dgraph) syncAllMarks() {
 	}
 }
 
+// Used to write checkpoints to Badger.
 func (d *Dgraph) writeCheckpoint() {
-	wb := make([]*badger.Entry, 0, 5)
+	wb := make([]*badger.Entry, 0, len(d.marks))
 	for file, mark := range d.marks {
 		w := mark.DoneUntil()
 		if w == 0 {
 			continue
 		}
-		buf := make([]byte, 10)
-		n := binary.PutUvarint(buf, w)
+		var buf [10]byte
+		n := binary.PutUvarint(buf[:], w)
 		wb = badger.EntriesSet(wb, []byte(fmt.Sprintf("checkpoint-%s", file)), buf[:n])
 	}
+
 	if err := d.alloc.kv.BatchSet(wb); err != nil {
 		fmt.Printf("Error while writing to disk %v\n", err)
 	}
@@ -66,7 +75,7 @@ func (d *Dgraph) writeCheckpoint() {
 	}
 }
 
-// Used to store checkpoints for various files.
+// Used to store checkpoints for various files periodically.
 func (d *Dgraph) storeCheckpoint() {
 	ticker := time.NewTicker(time.Minute)
 	defer ticker.Stop()
@@ -75,9 +84,6 @@ func (d *Dgraph) storeCheckpoint() {
 		d.writeCheckpoint()
 	}
 }
-
-// A watermark for each file
-type syncMarks map[string]*x.WaterMark
 
 func (g syncMarks) create(file string) *x.WaterMark {
 	if g == nil {
@@ -91,8 +97,4 @@ func (g syncMarks) create(file string) *x.WaterMark {
 	w.Init()
 	g[file] = w
 	return w
-}
-
-func (d *Dgraph) SyncMarkFor(file string) *x.WaterMark {
-	return d.marks[file]
 }
--- a/client/checkpoint.go
+++ b/client/checkpoint.go
@@ -19,7 +19,6 @@ package client
 import (
 	"encoding/binary"
 	"fmt"
-	"path/filepath"
 	"sync"
 	"time"
 
@@ -44,7 +43,7 @@ func (d *Dgraph) storeCheckpoint() {
 		}
 
 		for file, w := range m {
-			var byte [10]buf
+			var buf []byte
 			n := binary.PutUvarint(buf, w)
 			wb = badger.EntriesSet(wb, []byte(fmt.Sprintf("checkpoint-%s", file)), buf[:n])
 		}
@@ -60,11 +59,11 @@ func (d *Dgraph) storeCheckpoint() {
 	}
 }
 
-func NewSyncMarks(files []string) {
-	for _, file := range files {
-		bp := filepath.Base(file)
-	}
-}
+//func NewSyncMarks(files []string) {
+//	for _, file := range files {
+//		bp := filepath.Base(file)
+//	}
+//}
 
 type syncMarks struct {
 	sync.RWMutex
--- a/client/checkpoint.go
+++ b/client/checkpoint.go
@@ -37,7 +37,7 @@ type syncMarks map[string]waterMark
 // Create syncmarks for files and store them in dgraphClient.
 func (d *Dgraph) NewSyncMarks(files []string) error {
 	if d.marks != nil {
-		return fmt.Errof("NewSyncMarks should only be called once.")
+		return fmt.Errorf("NewSyncMarks should only be called once.")
 	}
 
 	for _, file := range files {
--- a/client/checkpoint.go
+++ b/client/checkpoint.go
@@ -33,7 +33,7 @@ func (d *Dgraph) Checkpoint(file string) uint64 {
 
 // Used to store checkpoints for various files.
 func (d *Dgraph) storeCheckpoint() {
-	ticker := time.NewTicker(10 * time.Second)
+	ticker := time.NewTicker(time.Minute)
 	defer ticker.Stop()
 
 	wb := make([]*badger.Entry, 0, 5)
@@ -44,7 +44,6 @@ func (d *Dgraph) storeCheckpoint() {
 		}
 
 		for file, w := range m {
-			fmt.Println("Storing checkpoint for", file, w)
 			buf := make([]byte, 10)
 			n := binary.PutUvarint(buf, w)
 			wb = badger.EntriesSet(wb, []byte(fmt.Sprintf("checkpoint-%s", file)), buf[:n])
--- a/client/checkpoint.go
+++ b/client/checkpoint.go
@@ -36,7 +36,7 @@ type syncMarks map[string]waterMark
 
 // Create syncmarks for files and store them in dgraphClient.
 func (d *Dgraph) NewSyncMarks(files []string) error {
-	if d.marks != nil {
+	if len(d.marks) > 0 {
 		return fmt.Errorf("NewSyncMarks should only be called once.")
 	}
 
--- a/client/checkpoint.go
+++ b/client/checkpoint.go
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2016 Dgraph Labs, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package client
+
+import (
+	"sync"
+
+	"github.com/dgraph-io/dgraph/x"
+)
+
+type syncMarks struct {
+	sync.RWMutex
+	// A watermark for each file.
+	m map[string]*x.WaterMark
+}
+
+func (g *syncMarks) create(file string) *x.WaterMark {
+	g.Lock()
+	defer g.Unlock()
+	if g.m == nil {
+		g.m = make(map[string]*x.WaterMark)
+	}
+
+	if prev, present := g.m[file]; present {
+		return prev
+	}
+	w := &x.WaterMark{Name: file}
+	w.Init()
+	g.m[file] = w
+	return w
+}
+
+func (g *syncMarks) Get(file string) *x.WaterMark {
+	g.RLock()
+	if w, present := g.m[file]; present {
+		g.RUnlock()
+		return w
+	}
+	g.RUnlock()
+	return g.create(file)
+}
+
+func (d *Dgraph) SyncMarkFor(file string) *x.WaterMark {
+	return d.marks.Get(file)
+}
--- a/client/checkpoint.go
+++ b/client/checkpoint.go
@@ -19,6 +19,7 @@ package client
 import (
 	"encoding/binary"
 	"fmt"
+	"path/filepath"
 	"sync"
 	"time"
 
@@ -43,7 +44,7 @@ func (d *Dgraph) storeCheckpoint() {
 		}
 
 		for file, w := range m {
-			buf := make([]byte, 10)
+			var byte [10]buf
 			n := binary.PutUvarint(buf, w)
 			wb = badger.EntriesSet(wb, []byte(fmt.Sprintf("checkpoint-%s", file)), buf[:n])
 		}
@@ -56,6 +57,12 @@ func (d *Dgraph) storeCheckpoint() {
 			}
 		}
 		wb = wb[:0]
+	}
+}
+
+func NewSyncMarks(files []string) {
+	for _, file := range files {
+		bp := filepath.Base(file)
 	}
 }
 
--- a/client/checkpoint.go
+++ b/client/checkpoint.go
@@ -36,6 +36,10 @@ type syncMarks map[string]waterMark
 
 // Create syncmarks for files and store them in dgraphClient.
 func (d *Dgraph) NewSyncMarks(files []string) error {
+	if d.marks != nil {
+		return fmt.Errof("NewSyncMarks should only be called once.")
+	}
+
 	for _, file := range files {
 		ap, err := filepath.Abs(file)
 		if err != nil {
@@ -79,7 +83,6 @@ func (d *Dgraph) writeCheckpoint() {
 		}
 		wm.last = doneUntil
 		d.marks[file] = wm
-		fmt.Printf("d.marks: %+v\n", d.marks)
 		var buf [10]byte
 		n := binary.PutUvarint(buf[:], doneUntil)
 		wb = badger.EntriesSet(wb, []byte(fmt.Sprintf("checkpoint-%s", file)), buf[:n])
--- a/client/checkpoint.go
+++ b/client/checkpoint.go
@@ -39,11 +39,11 @@ func (d *Dgraph) storeCheckpoint() {
 	for range ticker.C {
 		m := d.marks.watermarks()
 		if m == nil {
-			return
+			continue
 		}
 
 		for file, w := range m {
-			var buf []byte
+			buf := make([]byte, 10)
 			n := binary.PutUvarint(buf, w)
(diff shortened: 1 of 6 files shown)


janardhan1993:
:lgtm:

---

Reviewed 5 of 5 files at r1.

---

*[client/mutations.go, line 315 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-KokVG9B76N_snORgxx5:-KokVG9C6Q0JmTlYHXHy:buk5ezn) ([raw file](https://github.com/dgraph-io/dgraph/blob/44926ec1918df266bd2ec7f4632df9ea1665d96c/client/mutations.go#L315)):*
> ```Go
> 		entry.mark.Ch <- x.Mark{Index: entry.line, Done: true}
> 	}
> 	req.entries = req.entries[:0]
> ```

move this inside reset

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-:-KokVk3TCU9ks-BFGODZ:bnfp4nl)*

pawanrawal:
Review status: 2 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[client/mutations.go, line 315 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-KokVG9B76N_snORgxx5:-KoknszhKrslqxYGtnn3:bb7f2mf) ([raw file](https://github.com/dgraph-io/dgraph/blob/44926ec1918df266bd2ec7f4632df9ea1665d96c/client/mutations.go#L315)):*

move this inside reset
</blockquote></details>

done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174)*

manishrjain:
Please don't submit, until profiling is done on this change. Sounds expensive CPU wise.

---

Reviewed 2 of 5 files at r1, 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[client/checkpoint.go, line 46 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-Kot3EpLnfpCBIRAZ79Q:-Kot3EpMJQWGEDywK_yK:b-cb95gk) ([raw file](https://github.com/dgraph-io/dgraph/blob/71ae5842c47d119d3e19b4e59c65db134346c800/client/checkpoint.go#L46)):*
> ```Go
> 
> 		for file, w := range m {
> 			buf := make([]byte, 10)
> ```

var byte[10] buf

---

*[client/checkpoint.go, line 63 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-Kot3n0MlaB395l7zFVA:-Kot3n0MlaB395l7zFVB:b-6ya13m) ([raw file](https://github.com/dgraph-io/dgraph/blob/71ae5842c47d119d3e19b4e59c65db134346c800/client/checkpoint.go#L63)):*
> ```Go
> 
> type syncMarks struct {
> 	sync.RWMutex
> ```

Why not create all the entries upfront, and avoid all this locking around this map here?

---

*[client/checkpoint.go, line 98 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-Kot48DYZqHszW-g82BU:-Kot48DZgO0KEtl1_Kgo:b-srs6cs) ([raw file](https://github.com/dgraph-io/dgraph/blob/71ae5842c47d119d3e19b4e59c65db134346c800/client/checkpoint.go#L98)):*
> ```Go
> }
> 
> func (g *syncMarks) watermarks() map[string]uint64 {
> ```

You don't really need this map. You can just call DoneUntil() directly when iterating over the files.

---

*[client/mutations.go, line 313 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-Kot5PuFqcZ_MQCvS0e_:-Kot5PuFqcZ_MQCvS0ea:b8sr2r4) ([raw file](https://github.com/dgraph-io/dgraph/blob/71ae5842c47d119d3e19b4e59c65db134346c800/client/mutations.go#L313)):*
> ```Go
> 	for _, entry := range req.entries {
> 		// Mark watermarks as done.
> 		entry.mark.Ch <- x.Mark{Index: entry.line, Done: true}
> ```

This might be very slow because it's done per line per file. We should do it at a request level, instead of at a line level. But, that'd require changing how we create a request.

Can you profile how much slower this makes our loader. If it doesn't, then this approach is good. But, I feel this would peg our cpu pretty hard.

---

*[client/mutations.go, line 385 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-Kot6FmX00HXg1COCxVj:-Kot6FmX00HXg1COCxVk:b-cygd3c) ([raw file](https://github.com/dgraph-io/dgraph/blob/71ae5842c47d119d3e19b4e59c65db134346c800/client/mutations.go#L385)):*
> ```Go
> 	}
> 	if len(file) > 0 && line != 0 {
> 		sm.Ch <- x.Mark{Index: line}
> ```

Can you check how much slower this makes our loader?

---

*[cmd/dgraphloader/.gitignore, line 2 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-Kot6OlzwAFoBP9lX11F:-Kot6OlzwAFoBP9lX11G:b-7iwvb8) ([raw file](https://github.com/dgraph-io/dgraph/blob/71ae5842c47d119d3e19b4e59c65db134346c800/cmd/dgraphloader/.gitignore#L2)):*
> ```
> /dgraphloader
> c
> ```

this?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-:-Kot7Xai1hBT9p95R-r6:bmhmsk5)*

pawanrawal:
Benchmarking loader performance on i3 now. Will report the numbers in an hour.

---

Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-:-KoyllnaZK2yz5lKUCLo:b-3wefgd)*

pawanrawal:
Tried loading 42 Million RDF's

Before the change it took 15m45s, after the change it took 17m. Will make other changes suggested and see if that improves the performance.

---

Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-:-KozF-v2PoS-pay0czSM:b-ooprze)*

manishrjain:
What we could do is to break up rpcs, so that each rpc holds updates from only one file. That way, we can then watermark only one line number per rpc. This would be a lot cheaper operation.

---

Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-:-KpIV4BwuAL_pVsG6t0x:bfej9gc)*

pawanrawal:
yeah, that should work. I will make this change, update the numbers and the PR by tomorrow.

---

Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[cmd/dgraphloader/.gitignore, line 2 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-Kot6OlzwAFoBP9lX11F:-KozUXXI3fiQAZ7xSkdA:b-tlv419) ([raw file](https://github.com/dgraph-io/dgraph/blob/71ae5842c47d119d3e19b4e59c65db134346c800/cmd/dgraphloader/.gitignore#L2)):*

this?
</blockquote></details>

It's the directory used by badger.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-:-KpJtMIc6Uz0KPWlQ-u2:b-q3yz1q)*

pawanrawal:
We now batch edges in dgraphloader and send them to the client. This makes sure that a batch only contains edges from one file. Now we can emit just one watermark per request. After this change, the loader doesn't take any more time than it used to previously.

---

Review status: 1 of 5 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[client/checkpoint.go, line 46 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-Kot3EpLnfpCBIRAZ79Q:-KpO_JHDZrQH_eO8ungg:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/71ae5842c47d119d3e19b4e59c65db134346c800/client/checkpoint.go#L46)):*

var byte[10] buf
</blockquote></details>

Done.

---

*[client/checkpoint.go, line 63 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-Kot3n0MlaB395l7zFVA:-KpOW_ITs9b32iHuxh7T:b-6oiydf) ([raw file](https://github.com/dgraph-io/dgraph/blob/71ae5842c47d119d3e19b4e59c65db134346c800/client/checkpoint.go#L63)):*

Why not create all the entries upfront, and avoid all this locking around this map here?
</blockquote></details>

I do so in NewSyncMarks now.

---

*[client/checkpoint.go, line 98 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-Kot48DYZqHszW-g82BU:-KpO_TYh7rxgydO35tnO:ba3dfob) ([raw file](https://github.com/dgraph-io/dgraph/blob/71ae5842c47d119d3e19b4e59c65db134346c800/client/checkpoint.go#L98)):*

You don't really need this map. You can just call DoneUntil() directly when iterating over the files.
</blockquote></details>

Removed

---

*[client/mutations.go, line 313 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-Kot5PuFqcZ_MQCvS0e_:-KpO_W8asqkGghHqGDvh:b-q13778) ([raw file](https://github.com/dgraph-io/dgraph/blob/71ae5842c47d119d3e19b4e59c65db134346c800/client/mutations.go#L313)):*

This might be very slow because it's done per line per file. We should do it at a request level, instead of at a line level. But, that'd require changing how we create a request.

Can you profile how much slower this makes our loader. If it doesn't, then this approach is good. But, I feel this would peg our cpu pretty hard.
</blockquote></details>

It doesn't increase the time for the loader anymore. It finishes 5-10s before/after. Doing it per request now.

---

*[client/mutations.go, line 385 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-Kot6FmX00HXg1COCxVj:-KpO_rC0rPJzYo7tAAw0:b6z2se2) ([raw file](https://github.com/dgraph-io/dgraph/blob/71ae5842c47d119d3e19b4e59c65db134346c800/client/mutations.go#L385)):*

Can you check how much slower this makes our loader?
</blockquote></details>

Not slow anymore.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-:-KpOc8BcUh0yU8onr-p8:b-8o6qfw)*

manishrjain:
Some more comments. Will do a final review once you address.

---

Review status: 1 of 5 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[client/checkpoint.go, line 35 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-KpSMyzurod2S34aCqSV:-KpSMyzurod2S34aCqSW:b-tqkz39) ([raw file](https://github.com/dgraph-io/dgraph/blob/08777ec6c35a5fd9b4f2209d6da9cfa8e18019f1/client/checkpoint.go#L35)):*
> ```Go
> func (d *Dgraph) NewSyncMarks(files []string) {
> 	for _, file := range files {
> 		bp := filepath.Base(file)
> ```

Don't you need to add an extension? Or, ensure that the filename generated is not going to overwrite the current file.

---

*[client/checkpoint.go, line 45 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-KpSNHoeRvxREx_1ZPpG:-KpSNHoeRvxREx_1ZPpH:bc9e5mn) ([raw file](https://github.com/dgraph-io/dgraph/blob/08777ec6c35a5fd9b4f2209d6da9cfa8e18019f1/client/checkpoint.go#L45)):*
> ```Go
> }
> 
> // This is called to syncAllMarks. It is useful to find the final line that was processed for each
> ```

"This is called to syncAllMarks" -- not clear what you're saying.

---

*[client/checkpoint.go, line 49 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-KpSNF2hMlxvAbkHZFf4:-KpSNF2hMlxvAbkHZFf5:bffhzvc) ([raw file](https://github.com/dgraph-io/dgraph/blob/08777ec6c35a5fd9b4f2209d6da9cfa8e18019f1/client/checkpoint.go#L49)):*
> ```Go
> func (d *Dgraph) syncAllMarks() {
> 	for _, mark := range d.marks {
> 		for mark.WaitingFor() {
> ```

Not sure what this loop does.

---

*[client/checkpoint.go, line 60 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-KpSNcOtRvB0WHsCh9Mj:-KpSNcOtRvB0WHsCh9Mk:brgufdn) ([raw file](https://github.com/dgraph-io/dgraph/blob/08777ec6c35a5fd9b4f2209d6da9cfa8e18019f1/client/checkpoint.go#L60)):*
> ```Go
> 	for file, mark := range d.marks {
> 		w := mark.DoneUntil()
> 		if w == 0 {
> ```

You could also track the last one. If it's the same as the last one, no need to write again.

---

*[client/checkpoint.go, line 83 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-KpSJE9ussGbuMwn1o5b:-KpSJE9ussGbuMwn1o5c:b3ps3jk) ([raw file](https://github.com/dgraph-io/dgraph/blob/08777ec6c35a5fd9b4f2209d6da9cfa8e18019f1/client/checkpoint.go#L83)):*
> ```Go
> 	defer ticker.Stop()
> 
> 	for range ticker.C {
> ```

Who will stop this loop?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-:-KpSO7r7l82WibhqzG16:b-mxqvkk)*

pawanrawal:
Review status: 1 of 5 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[client/checkpoint.go, line 35 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-KpSMyzurod2S34aCqSV:-KpSqBQTTa1sUIN_G99I:b89y1ye) ([raw file](https://github.com/dgraph-io/dgraph/blob/08777ec6c35a5fd9b4f2209d6da9cfa8e18019f1/client/checkpoint.go#L35)):*

Don't you need to add an extension? Or, ensure that the filename generated is not going to overwrite the current file.
</blockquote></details>

I am using absolute path now which is with extension.

---

*[client/checkpoint.go, line 45 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-KpSNHoeRvxREx_1ZPpG:-KpSomrqqI2fbUaiOgPG:b3hqk53) ([raw file](https://github.com/dgraph-io/dgraph/blob/08777ec6c35a5fd9b4f2209d6da9cfa8e18019f1/client/checkpoint.go#L45)):*

"This is called to syncAllMarks" -- not clear what you're saying.
</blockquote></details>

Updated the comment to clarify.

---

*[client/checkpoint.go, line 49 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-KpSNF2hMlxvAbkHZFf4:-KpSoht0ALmg1KRTafrp:b-t5s4op) ([raw file](https://github.com/dgraph-io/dgraph/blob/08777ec6c35a5fd9b4f2209d6da9cfa8e18019f1/client/checkpoint.go#L49)):*

Not sure what this loop does.
</blockquote></details>

We are waiting for all watermarks to be processed.

---

*[client/checkpoint.go, line 60 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-KpSNcOtRvB0WHsCh9Mj:-KpSh-VXW86U_JJVzCPe:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/08777ec6c35a5fd9b4f2209d6da9cfa8e18019f1/client/checkpoint.go#L60)):*

You could also track the last one. If it's the same as the last one, no need to write again.
</blockquote></details>

Done.

---

*[client/checkpoint.go, line 83 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-KpSJE9ussGbuMwn1o5b:-KpSh02S1lpRua-2dMRz:b7dmdm7) ([raw file](https://github.com/dgraph-io/dgraph/blob/08777ec6c35a5fd9b4f2209d6da9cfa8e18019f1/client/checkpoint.go#L83)):*

Who will stop this loop?
</blockquote></details>

I should, doing now in Flush.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174)*

manishrjain:
:lgtm: Few more comments.

---

Reviewed 1 of 4 files at r3, 3 of 3 files at r4.

---

*[client/checkpoint.go, line 49 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-KpSNF2hMlxvAbkHZFf4:-KpVfZYfg31mjaeXDbWn:b-ckhid6) ([raw file](https://github.com/dgraph-io/dgraph/blob/08777ec6c35a5fd9b4f2209d6da9cfa8e18019f1/client/checkpoint.go#L49)):*

We are waiting for all watermarks to be processed.
</blockquote></details>

You probably need to do it exactly once. Just put this logic where you need it directly. It would be more clear there.

---

*[client/checkpoint.go, line 102 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-KpVgExQbW4QHHYzpfbC:-KpVgExQbW4QHHYzpfbD:bvw0q22) ([raw file](https://github.com/dgraph-io/dgraph/blob/f9388c235ec476d0333ae874105cc56f85ed4555/client/checkpoint.go#L102)):*
> ```Go
> 
> // Used to store checkpoints for various files periodically.
> func (d *Dgraph) storeCheckpoint(ticker *time.Ticker) {
> ```

writeCheckpoinPeriodically. But, this func is so simple, there's no need to have a separate one. Just use func() in the place where you call it directly.

---

*[client/mutations.go, line 333 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-KpVhaTQ2zJmHR8nujc_:-KpVhaTQ2zJmHR8nujca:b-nnkx76) ([raw file](https://github.com/dgraph-io/dgraph/blob/f9388c235ec476d0333ae874105cc56f85ed4555/client/mutations.go#L333)):*
> ```Go
> 
> func (d *Dgraph) makeSchemaRequests() {
> 	var req Req
> ```

Why this change? This needs to be allocated on heap, because it's crossing function boundaries.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-:-KpViQZT4_6SAU_6RmyT:b6epxqe)*

pawanrawal:

---

*[client/checkpoint.go, line 49 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-KpSNF2hMlxvAbkHZFf4:-KpXWRQ9-xCFmTzCXUdW:b-4q9e2g) ([raw file](https://github.com/dgraph-io/dgraph/blob/08777ec6c35a5fd9b4f2209d6da9cfa8e18019f1/client/checkpoint.go#L49)):*

You probably need to do it exactly once. Just put this logic where you need it directly. It would be more clear there.
</blockquote></details>

Right, needs to be only done once.  Calling it BatchFlush.

---

*[client/checkpoint.go, line 102 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-KpVgExQbW4QHHYzpfbC:-KpXW4wz_QqC-aAnrK_U:b-t76g8u) ([raw file](https://github.com/dgraph-io/dgraph/blob/f9388c235ec476d0333ae874105cc56f85ed4555/client/checkpoint.go#L102)):*

writeCheckpoinPeriodically. But, this func is so simple, there's no need to have a separate one. Just use func() in the place where you call it directly.
</blockquote></details>

Removed this and calling it inline.

---

*[client/mutations.go, line 333 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-KpVhaTQ2zJmHR8nujc_:-KpXZ6TtKo-AZ9MwqBd-:b61xews) ([raw file](https://github.com/dgraph-io/dgraph/blob/f9388c235ec476d0333ae874105cc56f85ed4555/client/mutations.go#L333)):*

Why this change? This needs to be allocated on heap, because it's crossing function boundaries.
</blockquote></details>

I did an escape analysis which showed me it doesn't escape and hence should be allocated on stack.
```
../../client/mutations.go:341: (*Dgraph).makeSchemaRequests req does not escape
../../client/mutations.go:344: (*Dgraph).makeSchemaRequests req does not escape
../../client/mutations.go:355: (*Dgraph).makeSchemaRequests req does not escape
```
As far as I know go passes a copy of the `r` to the channel. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174)*

pawanrawal:

---

*[client/mutations.go, line 333 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174#-KpVhaTQ2zJmHR8nujc_:-KpXe7IDnOyHvEX06ddo:b-da5pxi) ([raw file](https://github.com/dgraph-io/dgraph/blob/f9388c235ec476d0333ae874105cc56f85ed4555/client/mutations.go#L333)):*

I did an escape analysis which showed me it doesn't escape and hence should be allocated on stack.
```
../../client/mutations.go:341: (*Dgraph).makeSchemaRequests req does not escape
../../client/mutations.go:344: (*Dgraph).makeSchemaRequests req does not escape
../../client/mutations.go:355: (*Dgraph).makeSchemaRequests req does not escape
```
As far as I know go passes a copy of the `r` to the channel. 
</blockquote></details>

Ok I see it finally escapes to heap while calling `request`.
```
../../client/mutations.go:297: moved to heap: req
```
Reverting this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1174)*

pawanrawal:
Merged to master.

Issue #1180, opened 2017-07-12
Sending data to new nodes is very slow


srh:
There's a chance this is caused by the same underlying problem as #1168.

(Edit:  Actually, this might not involve sending a "snapshot" -- I haven't verified that it actually updated using the snapshot mechanism.)

Anyway, if you make a 3-node cluster, kill a node, and then load a bunch of data, and then revive the node (as described in https://github.com/dgraph-io/dgraph/issues/1169#issuecomment-314667006), it takes a very long time to bring the new node up to speed.

Something on the order of 1m to 1m30s to load the data initially, and then 15 minutes to catch the new node up.

While it's catching up, the CPU usage on any given core is at most 5%, maybe 10%.  It seems like the slowness is from too much idling.

Another peculiar thing is the high disk usage of the node (of svr3) -- the dataset was 1million.rdf.gz:

```
60M	/home/srh/blah/svr1/w
195M	/home/srh/blah/svr1/p
255M	/home/srh/blah/svr1
60M	/home/srh/blah/svr2/w
181M	/home/srh/blah/svr2/p
241M	/home/srh/blah/svr2
60M	/home/srh/blah/svr3/w
380M	/home/srh/blah/svr3/p
440M	/home/srh/blah/svr3
```

The actual contents:

```
../svr1/p:
total 199048
-rw-r--r-- 1 srh srh 203817056 Jul 12 00:05 000000.vlog
-rw-r--r-- 1 srh srh         0 Jul 11 23:46 clog

../svr1/w:
total 61304
-rw-r--r-- 1 srh srh 62770134 Jul 12 00:05 000000.vlog
-rw-r--r-- 1 srh srh        0 Jul 11 23:46 clog

../svr2/p:
total 184968
-rw-r--r-- 1 srh srh 189401412 Jul 12 00:05 000000.vlog
-rw-r--r-- 1 srh srh         0 Jul 11 23:46 clog

../svr2/w:
total 61304
-rw-r--r-- 1 srh srh 62769560 Jul 12 00:05 000000.vlog
-rw-r--r-- 1 srh srh        0 Jul 11 23:46 clog

../svr3/p:
total 388188
-rw-r--r-- 1 srh srh 397493236 Jul 12 00:05 000000.vlog
-rw-r--r-- 1 srh srh       165 Jul 11 23:46 000001.sst
-rw-r--r-- 1 srh srh         0 Jul 11 23:49 clog

../svr3/w:
total 61304
[...]

PR #1191, opened 2017-07-13
Decrease tick interval to 50 ms.

srh:
This is a stop-gap.  For whatever reason, we aren't replaying log entries any faster than one entry per tick.  This makes replaying the log be about 15x slower than loading the data initially cost (when running locally on a Thinkpad W520).  So, reducing the tick interval by a factor of 20 seems like it'll make the Raft tick frequency not be the bottleneck.

In general this won't be true, for example, if we did the initial writes more slowly, then they'd likewise be forwarded to other nodes more slowly (with fewer writes per raft index).  This generally doesn't fix the overall problem.

Connected to #1180.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1191)

### DIFF
--- a/worker/draft.go
+++ b/worker/draft.go
@@ -238,8 +238,10 @@ func newNode(gid uint32, id uint64, myAddr string) *node {
 		gid:   gid,
 		store: store,
 		cfg: &raft.Config{
-			ID:              id,
-			ElectionTick:    10,
+			ID: id,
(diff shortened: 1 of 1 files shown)


srh:
Closing.  I want to look at this further. It looks like it might be tied to the heartbeat interval rather than the Tick rate.

PR #1208, opened 2017-07-18
Remove needless select statement in doSendMessage

srh:
I want to make sure I'm not missing anything here -- this change should have no effect, right?

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1208)

### DIFF
--- a/worker/draft.go
+++ b/worker/draft.go
@@ -495,20 +495,9 @@ func (n *node) doSendMessage(to uint64, data []byte) {
 	c := protos.NewWorkerClient(conn)
 	p := &protos.Payload{Data: data}
 
-	ch := make(chan error, 1)
-	go func() {
-		_, err = c.RaftMessage(ctx, p)
-		ch <- err
-	}()
-
-	select {
-	case <-ctx.Done():
-		return
-	case <-ch:
-		// We don't need to do anything if we receive any error while sending message.
-		// RAFT would automatically retry.
-		return
-	}
+	// We don't do anything if we receive any error while sending a message.  Raft would
+	// automatically retry.
+	_, _ = c.RaftMessage(ctx, p)
 }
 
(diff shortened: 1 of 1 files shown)


manishrjain:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[worker/draft.go, line 505 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1208#-KpMjmQ0OZ-PUnvl8iRE:-KpMjmQ0OZ-PUnvl8iRF:b-3j90wi) ([raw file](https://github.com/dgraph-io/dgraph/blob/30237a1429debab73eff38fea2f724914ca38b77/worker/draft.go#L505)):*
> ```Go
> 
> 	select {
> 	case <-ctx.Done():
> ```

This is important. Essentially the context could be cancelled. In which case, we don't want to wait for the raft message call, we should return immediately.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1208)*

srh:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[worker/draft.go, line 505 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1208#-KpMjmQ0OZ-PUnvl8iRE:-KpMnbl2ZDjlLny8Jaxo:byi55uv) ([raw file](https://github.com/dgraph-io/dgraph/blob/30237a1429debab73eff38fea2f724914ca38b77/worker/draft.go#L505)):*

This is important. Essentially the context could be cancelled. In which case, we don't want to wait for the raft message call, we should return immediately.
</blockquote></details>

Okay, so it looks like the problem here is, at the lowest level the Write call on the gRPC network connection can actually block and _not_ be canceled by the ctx, right?  It looks like that's the case.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1208)*

PR #1230, opened 2017-07-23
Rework parseFacets, disallow 'a as orderdesc: b'

srh:
A lot of parsing functions are implemented in the following style:

```
for it.Next() {
    if case1 {
        make sure case1 valid
        ...     
    } else if case2 {
        make sure case2 valid
        ...
    } else if case3 {
        make sure case3 valid
        ...
    } ...
}
```

This is really difficult to work with because each parsing case needs to handle any other possible parsing state.  There's basically n^2 stuff to think about.

It also does a poor job of expressing the intent of what is supposed to be parsed.  The reason is, all the code paths end up heading to the same place, and it's hard to tell which behaviors are intentional.  For example, was this supposed to be valid:  `@facets(a as orderasc: b)`  Probably not.  But is `@facets()` supposed to be treated the same as `@facets`, by setting AllKeys to true?  I have no idea -- the code just spills into that case, but there's nothing to indicate whether the possibility was contemplated by the developers.

It's much better to make parsing functions in a manner such that each point in the function body corresponds to a particular kind of place in the parse tree.  It's analogous to what happens when you replace an event loop with a stateful object with a goroutine.

It's also a good idea to make liberal use of helper functions like `tryParseItemType` and `trySkipItemVal`, or others as we might create.

It looks like due to the way the code was originally written, a lot of parsing code is using this `for it.Next()` style.  I think whenever we work on an old parsing function which uses that style and has gotten too complicated, we should consider refactoring it like this PR does, depending on the time/effort/correctness trade-off.

----

So anyway,

This PR has one material change to the behavior of parseFacets.  It disallows `@facets(a as orderasc: b)`.

But there's also the question:  Should `@facets()` still be treated like `@facets`?  Or should it be treated like an empty list of facets?  (My guess is we should change it to act like an empty list of facets, because I don't think users would appreciate having a special case.)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1230)

### DIFF
--- a/gql/parser.go
+++ b/gql/parser.go
@@ -1668,6 +1668,21 @@ func tryParseFacetList(it *lex.ItemIterator) (res facetRes, parseOk bool, err er
 			return res, ok, err
 		}
 
+		// Combine the facetitem with our result.
+		{
+			if facetItem.varName != "" {
+				facetVar[facetItem.facetName] = facetItem.varName
+			}
+			facets.Keys = append(facets.Keys, facetItem.facetName)
+			if facetItem.ordered {
+				if orderkey != "" {
+					return res, false, x.Errorf("Invalid use of orderasc/orderdesc in facets")
+				}
+				orderdesc = facetItem.orderdesc
+				orderkey = facetItem.facetName
+			}
+		}
+
 		// Now what?  Either close-paren or a comma.
 		if _, ok := tryParseItemType(it, itemRightRound); ok {
 			sort.Slice(facets.Keys, func(i, j int) bool {
@@ -1687,27 +1702,8 @@ func tryParseFacetList(it *lex.ItemIterator) (res facetRes, parseOk bool, err er
 			res.f, res.vmap, res.facetOrder, res.orderdesc = &facets, facetVar, orderkey, orderdesc
 			return res, true, nil
 		}
-		if item, ok := tryParseItemType(it, itemComma); !ok {
-			if len(facets.Keys) == 0 {
-				return res, false, nil
-			}
-			return res, false, x.Errorf(
-				"Expected ',' or ')' in facet list, got %q", item.Val)
-		}
-
-		// Combine the facetItem with our result.
-		{
-			if facetItem.varName != "" {
-				facetVar[facetItem.facetName] = facetItem.varName
-			}
-			facets.Keys = append(facets.Keys, facetItem.facetName)
-			if facetItem.ordered {
-				if orderkey != "" {
-					return res, false, x.Errorf("Invalid use of orderasc/orderdesc in facets")
-				}
-				orderdesc = facetItem.orderdesc
-				orderkey = facetItem.facetName
-			}
+		if _, ok := tryParseItemType(it, itemComma); !ok {
+			return res, false, nil
 		}
 	}
 }
--- a/gql/parser.go
+++ b/gql/parser.go
@@ -1668,21 +1668,6 @@ func tryParseFacetList(it *lex.ItemIterator) (res facetRes, parseOk bool, err er
 			return res, ok, err
 		}
 
-		// Combine the facetitem with our result.
-		{
-			if facetItem.varName != "" {
-				facetVar[facetItem.facetName] = facetItem.varName
-			}
-			facets.Keys = append(facets.Keys, facetItem.facetName)
-			if facetItem.ordered {
-				if orderkey != "" {
-					return res, false, x.Errorf("Invalid use of orderasc/orderdesc in facets")
-				}
-				orderdesc = facetItem.orderdesc
-				orderkey = facetItem.facetName
-			}
-		}
-
 		// Now what?  Either close-paren or a comma.
 		if _, ok := tryParseItemType(it, itemRightRound); ok {
 			sort.Slice(facets.Keys, func(i, j int) bool {
@@ -1702,8 +1687,27 @@ func tryParseFacetList(it *lex.ItemIterator) (res facetRes, parseOk bool, err er
 			res.f, res.vmap, res.facetOrder, res.orderdesc = &facets, facetVar, orderkey, orderdesc
 			return res, true, nil
 		}
-		if _, ok := tryParseItemType(it, itemComma); !ok {
-			return res, false, nil
+		if item, ok := tryParseItemType(it, itemComma); !ok {
+			if len(facets.Keys) == 0 {
+				return res, false, nil
+			}
+			return res, false, x.Errorf(
+				"Expected ',' or ')' in facet list, got %q", item.Val)
+		}
+
+		// Combine the facetItem with our result.
+		{
+			if facetItem.varName != "" {
+				facetVar[facetItem.facetName] = facetItem.varName
+			}
+			facets.Keys = append(facets.Keys, facetItem.facetName)
+			if facetItem.ordered {
+				if orderkey != "" {
+					return res, false, x.Errorf("Invalid use of orderasc/orderdesc in facets")
+				}
+				orderdesc = facetItem.orderdesc
+				orderkey = facetItem.facetName
+			}
 		}
 	}
 }
--- a/gql/parser.go
+++ b/gql/parser.go
@@ -1672,6 +1672,10 @@ func tryParseFacetList(it *lex.ItemIterator) (res facetRes, parseOk bool, err er
 		// Combine the facetitem with our result.
 		{
 			if facetItem.varName != "" {
+				if _, has := facetVar[facetItem.facetName]; has {
+					return res, false, x.Errorf("Duplicate variable mappings for facet %v",
+						facetItem.facetName)
+				}
 				facetVar[facetItem.facetName] = facetItem.varName
 			}
 			facets.Keys = append(facets.Keys, facetItem.facetName)
--- a/gql/parser.go
+++ b/gql/parser.go
@@ -1702,8 +1702,13 @@ func tryParseFacetList(it *lex.ItemIterator) (res facetRes, parseOk bool, err er
 			res.f, res.vmap, res.facetOrder, res.orderdesc = &facets, facetVar, orderkey, orderdesc
 			return res, true, nil
 		}
-		if _, ok := tryParseItemType(it, itemComma); !ok {
-			return res, false, nil
+		if item, ok := tryParseItemType(it, itemComma); !ok {
+			if len(facets.Keys) < 2 {
+				return res, false, nil
+			}
+			// We got a comma already, so this is definitely a facet list -- return an error.
+			return res, false, x.Errorf(
+				"Expected ',' or ')' in facet list", item.Val)
 		}
 	}
 }
--- a/gql/parser.go
+++ b/gql/parser.go
@@ -1655,11 +1655,8 @@ func tryParseFacetList(it *lex.ItemIterator) (res facetRes, parseOk bool, err er
 	var orderkey string
 
 	if _, ok := tryParseItemType(it, itemRightRound); ok {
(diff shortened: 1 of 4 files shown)


srh:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[gql/parser.go, line 1654 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230#-Kpl3cmi_uO0lJG6zs2u:-Kpl3cmi_uO0lJG6zs2v:bu30xh) ([raw file](https://github.com/dgraph-io/dgraph/blob/9accf3244951e2aee1f38fa45b589e1207dea812/gql/parser.go#L1654)):*
> ```Go
> 
> 	if _, ok := tryParseItemType(it, itemRightRound); ok {
> 		// TODO: Pre-existing code did all keys on @facets(), instead of no keys.  Might be a
> ```

This TODO should be replaced with a comment, and either facets.AllKeys should be set to true or not.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230)*

pawanrawal:
:lgtm:  Looks a good change. Makes the code more understandable and easier to maintain. Got some questions and comments.

---

Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[gql/parser.go, line 1654 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230#-Kpl3cmi_uO0lJG6zs2u:-KprY2jrcyJOOwuVHBlO:b-xbo27j) ([raw file](https://github.com/dgraph-io/dgraph/blob/9accf3244951e2aee1f38fa45b589e1207dea812/gql/parser.go#L1654)):*

This TODO should be replaced with a comment, and either facets.AllKeys should be set to true or not.
</blockquote></details>

To me sounds like an error or a mistake by the user. We should return an error.

---

*[gql/parser.go, line 1616 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230#-KprX3O54AZ7hU5ev74o:-KprX3O54AZ7hU5ev74p:b-ljy7lx) ([raw file](https://github.com/dgraph-io/dgraph/blob/a0413e65ba63d1385220259a986261cd3a5c5144/gql/parser.go#L1616)):*
> ```Go
> 	// Now try to consume "as".
> 	if !trySkipItemVal(it, "as") {
> 		name1 = collectName(it, name1)
> ```

I think `collectName`  doesn't need the val as the second argument. It can just start with `it.Item().Val`. Maybe add a TODO to check for that. 

---

*[gql/parser.go, line 1706 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230#-KprYzyGVOBzDDIBlPRB:-KprYzyHpS0INlLzfQhZ:b-b1wisk) ([raw file](https://github.com/dgraph-io/dgraph/blob/a0413e65ba63d1385220259a986261cd3a5c5144/gql/parser.go#L1706)):*
> ```Go
> 		}
> 		if item, ok := tryParseItemType(it, itemComma); !ok {
> 			if len(facets.Keys) < 2 {
> ```

Why this check?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230#-:-KprZCZGhMgP8WG0m__T:b5fgmlp)*

srh:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[gql/parser.go, line 1654 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230#-Kpl3cmi_uO0lJG6zs2u:-Kprap0UOuRoh-W9FZBz:b-gguph5) ([raw file](https://github.com/dgraph-io/dgraph/blob/9accf3244951e2aee1f38fa45b589e1207dea812/gql/parser.go#L1654)):*

To me sounds like an error or a mistake by the user. We should return an error.
</blockquote></details>

I think there's the potential that generated queries would produce @facets() when their list of facets is empty.  I think users wouldn't want to have to work around the edge case of an error. What they'd want is consistent behavior, with an empty facets document for that edge.  Otherwise, they'd have to check if their request will retrieve no facets and remove the @facets part entirely, and also add a special case to their handling of the query result.

---

*[gql/parser.go, line 1616 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230#-KprX3O54AZ7hU5ev74o:-KprfdkrB_ssgMfJxQvN:b-n5c22j) ([raw file](https://github.com/dgraph-io/dgraph/blob/a0413e65ba63d1385220259a986261cd3a5c5144/gql/parser.go#L1616)):*

I think `collectName`  doesn't need the val as the second argument. It can just start with `it.Item().Val`. Maybe add a TODO to check for that. 
</blockquote></details>

 We pass an expression val+item.Val in one of its other callers.

---

*[gql/parser.go, line 1706 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230#-KprYzyGVOBzDDIBlPRB:-Kpra9rGso3x-Lr1m1z6:bcchmpf) ([raw file](https://github.com/dgraph-io/dgraph/blob/a0413e65ba63d1385220259a986261cd3a5c5144/gql/parser.go#L1706)):*

Why this check?
</blockquote></details>

The function either returns an error (which means there was definitely a parse error), or it returns that it could not parse but you could try parseFilter instead.  We know it's not a valid parseFilter after we've seen a comma.  I'll add commenting to make this more clear.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230)*

ashwin95r:
Just have one comment. :lgtm: 

---

Review status: 0 of 3 files reviewed at latest revision, 5 unresolved discussions.

---

*[gql/parser.go, line 1674 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230#-KprbA6cujRp8UZXoFp2:-KprbA6cujRp8UZXoFp3:bmbxp9n) ([raw file](https://github.com/dgraph-io/dgraph/blob/a0413e65ba63d1385220259a986261cd3a5c5144/gql/parser.go#L1674)):*
> ```Go
> 		{
> 			if facetItem.varName != "" {
> 				facetVar[facetItem.facetName] = facetItem.varName
> ```

Looks like, this will allow `@facets(a as w1, a as w2)` and `@facets(a as w1, b as w1)`. Both of them should error out. Though I think they'll be caught later. Would be nice to have some tests covering this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230#-:-KpreVOjY017Oo0BV3Ut:btjvcp2)*

srh:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[gql/parser.go, line 1674 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230#-KprbA6cujRp8UZXoFp2:-KprwoTx5CNk7ASEQwFU:b-dh89de) ([raw file](https://github.com/dgraph-io/dgraph/blob/a0413e65ba63d1385220259a986261cd3a5c5144/gql/parser.go#L1674)):*

Looks like, this will allow `@facets(a as w1, a as w2)` and `@facets(a as w1, b as w1)`. Both of them should error out. Though I think they'll be caught later. Would be nice to have some tests covering this.
</blockquote></details>

The former will get caught later, duplicate variables.  But the latter just overwrites the map key.

We allow duplicate facet keys for some reason.  Since we have that, we can disallow multiple variable assignments in the deduplication logic.  Or do we want to assign to both variables?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230)*

ashwin95r:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[gql/parser.go, line 1674 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230#-KprbA6cujRp8UZXoFp2:-Kprzf8u_Fm_ezEmnulf:b-rw4y6e) ([raw file](https://github.com/dgraph-io/dgraph/blob/a0413e65ba63d1385220259a986261cd3a5c5144/gql/parser.go#L1674)):*

The former will get caught later, duplicate variables.  But the latter just overwrites the map key.

We allow duplicate facet keys for some reason.  Since we have that, we can disallow multiple variable assignments in the deduplication logic.  Or do we want to assign to both variables?
</blockquote></details>

I think we should disallow assigning multiple variables to the same key. I can't think of any case where this would be necessary.

And the deduplication I think is to be nicer in case the user gave the same facet twice by mistake.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230)*

pawanrawal:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[gql/parser.go, line 1654 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230#-Kpl3cmi_uO0lJG6zs2u:-KpsBOXbbQWJRUBWpKU5:bigpj1m) ([raw file](https://github.com/dgraph-io/dgraph/blob/9accf3244951e2aee1f38fa45b589e1207dea812/gql/parser.go#L1654)):*

I think there's the potential that generated queries would produce @facets() when their list of facets is empty.  I think users wouldn't want to have to work around the edge case of an error. What they'd want is consistent behavior, with an empty facets document for that edge.  Otherwise, they'd have to check if their request will retrieve no facets and remove the @facets part entirely, and also add a special case to their handling of the query result.
</blockquote></details>

Ok, sounds good.

---

*[gql/parser.go, line 1616 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230#-KprX3O54AZ7hU5ev74o:-KpsBAJO1XlrTLe6AEiy:b32w1t) ([raw file](https://github.com/dgraph-io/dgraph/blob/a0413e65ba63d1385220259a986261cd3a5c5144/gql/parser.go#L1616)):*

 We pass an expression val+item.Val in one of its other callers.
</blockquote></details>

Ok

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230)*

srh:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[gql/parser.go, line 1654 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230#-Kpl3cmi_uO0lJG6zs2u:-KpvvhTep0_0NFMkd4Mx:b-94nrjf) ([raw file](https://github.com/dgraph-io/dgraph/blob/9accf3244951e2aee1f38fa45b589e1207dea812/gql/parser.go#L1654)):*

Ok, sounds good.
</blockquote></details>

If we just change the parsing code, it looks like users will still have to handle the special case of there being no `@facets` document.  It seems we do this sort of thing, removing empty result sets, a lot in general.  So I don't know that I want to return an empty `@facets` document if that's not the sort of thing we usually do.

---

*[gql/parser.go, line 1674 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230#-KprbA6cujRp8UZXoFp2:-KpvxoD83EwETJEzzarV:bn5z8kl) ([raw file](https://github.com/dgraph-io/dgraph/blob/a0413e65ba63d1385220259a986261cd3a5c5144/gql/parser.go#L1674)):*

I think we should disallow assigning multiple variables to the same key. I can't think of any case where this would be necessary.

And the deduplication I think is to be nicer in case the user gave the same facet twice by mistake.
</blockquote></details>

Okay, assigning multiple variable names for a key is disallowed now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1230)*

srh:
In master as of 83b4d749b0ea254ddb53d5885eb801a996383b7c.

manishrjain:
How do we retrieve all the facets now, without knowing the keys?

PR #1244, opened 2017-07-25
Make disk IO and processing concurrent

ashwin95r:
In `processTask` fetching the values and processing it were sequential which was slow when we had huge UID lists. 

For example, this query on SuperUser data (50M rdfs)  has ~350K UIDs which are of the type `Question`
```
{
  var(func: eq(Type, "Question")) {
    a as count(Upvote)
    b as count(Downvote)
    sc as math(cond(b > a/2, a, 0))
  }
    
  contr(func: uid(sc), orderdesc: val(sc), first:100) {
    val(a)
    val(b)
    val(sc)
    Title{
      Text
    }
  }
}
```
The processing time reduces from `~20s` to `~9s` when we make `GetOrCreate` and the processing concurrent.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1244)

### DIFF
--- a/worker/task.go
+++ b/worker/task.go
@@ -220,6 +220,9 @@ type plf struct {
 	decr func()
 }
 
+// fetchPl get the PLs for the specified keys and returns them in order.
+// NOTE: when using this function, never break or return from the channel loop
+// as we MUST decr() the reference for each PL fetched.
 func fetchPl(keyCh chan []byte, och chan plf, stop chan struct{}, gid uint32) {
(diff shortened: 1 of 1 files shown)


ashwin95r:
After testing it out a bit more, just doign the decr() and releasing the PLs at the earliest gives us most of the benefits. So we don't have to use channels to fetch the PLs separately. Continuing this in #1255.

PR #1249, opened 2017-07-25
Stream log messages to peer that's catching up

srh:
Connected to #1180.

This makes us stream Raft log entries to a peer that's fallen behind, instead of sending one raftpb.Message per second, or per heartbeat.  Puts all MsgApps on a gRPC stream for pushback.

This also adds Kick() to etcd so that we can "kick" it into sending more MsgApp messages, instead of one per heartbeat.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1249)


srh:
Closing -- I'm going to send another PR in the morning after another look.

PR #1269, opened 2017-07-27
Decrease tick interval to 20 ms.

srh:
Note that etcd does 5 ms, and Cockroach does 200 ms by default.  Note that Cockroach has complicated usage of the library so things there might not be so simple.  I went with 20 ms because that's low enough to make us catch up a replica on the same order of time that an initial series of writes happened, on a particular piece of hardware with a particular usage pattern.

Connected to #1180.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1269)

### DIFF
--- a/worker/draft.go
+++ b/worker/draft.go
@@ -239,9 +239,9 @@ func newNode(gid uint32, id uint64, myAddr string) *node {
 		store: store,
 		cfg: &raft.Config{
 			ID: id,
-			// 500 ms if we call Tick() every 50 ms.
+			// 200 ms if we call Tick() every 20 ms.
 			ElectionTick: 10,
-			// 50 ms if we call Tick() every 50 ms.
+			// 20 ms if we call Tick() every 20 ms.
 			HeartbeatTick:   1,
 			Storage:         store,
 			MaxSizePerMsg:   4096,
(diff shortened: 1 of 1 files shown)


srh:
Edited the comments, which I should have looked at.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1269#-:-Kq5GqJP7QV2IH7-n35p:bfrdtrb)*

manishrjain:
:lgtm: @janardhan1993 just checked. The pings within Amazon DC are under a millisecond (200 micros). So, we should be good on that account as well.

---

Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---

*[worker/draft.go, line 242 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1269#-Kq5gc72Wjax1L1WpJnF:-Kq5gc72Wjax1L1WpJnG:bkqfc3t) ([raw file](https://github.com/dgraph-io/dgraph/blob/65a75b3c537fb9d44f6e7ca415ae1c6db3b508b8/worker/draft.go#L242)):*
> ```Go
> 		cfg: &raft.Config{
> 			ID: id,
> 			// 200 ms if we call Tick() every 20 ms.
> ```

The comments can fit in the same line as the code.

---

*[worker/draft.go, line 244 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1269#-Kq5gfUiwtMit4e_a0Vi:-Kq5gfUiwtMit4e_a0Vj:b-yq3bu8) ([raw file](https://github.com/dgraph-io/dgraph/blob/65a75b3c537fb9d44f6e7ca415ae1c6db3b508b8/worker/draft.go#L244)):*
> ```Go
> 			// 200 ms if we call Tick() every 20 ms.
> 			ElectionTick: 10,
> 			// 20 ms if we call Tick() every 20 ms.
> ```

and here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1269#-:-Kq5giwvD4PxqsVoZs-J:bn4zqyv)*

srh:
In master as of f1ac05ee66db41d0d1c32c74ac870fbd9661b8a1

PR #1280, opened 2017-07-31
Dump/backup

pawanrawal:
We dump backup for every group that the server is leader of.

TODO - Have a way to import the backup.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1280)


pawanrawal:
We need to have backups for Badger before we can do this in a proper way. Closing this for now.

Issue #1295, opened 2017-08-02
Incorrect function behaviour when language is not specified


tzdybal:
```
mutation {
  schema {
    test: string @index(hash, term) .
  }
  set {
    <_:t1> <test> "India"@en .
  }
}
```
Query:
```
{
  i(func:eq(test, "India")) {
    tae: test@en
    ta: test
  }
}
```

Query returns no results, but according to docs it should.

> Note In functions, language lists and `.` notation are not allowed. In functions, a string edge without a language tag means apply function to all languages, while a single language tag means apply to only the given language.

tzdybal:
Problem is visible only for `hash` index. With `exact` index `eq` works as expected.

tzdybal:
Seems that problem is visible for all functions and index types. Previous tests were to trivial - English was the only language used, therefore it was selected as 'any language'.

manishrjain:
This is how we currently do retrieval:

Retrieval:
test@en => English
test@. => Any language
test => Value without lang tag

We should implement the function in the same manner.

PR #1297, opened 2017-08-02
Correct handling of language in eq with hash index

tzdybal:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1297)

### DIFF
--- a/query/query_test.go
+++ b/query/query_test.go
@@ -301,6 +301,9 @@ func populateGraph(t *testing.T) {
 	addEdgeToLangValue(t, "lossy", 0x1002, "Honey badger", "en", nil)
 	addEdgeToLangValue(t, "lossy", 0x1003, "Honey bee", "en", nil)
 
+	// full_name has hash index, we need following data for bug with eq (#1295)
+	addEdgeToLangValue(t, "full_name", 0x10000, "Her Majesty Elizabeth the Second, by the Grace of God of the United Kingdom of Great Britain and Northern Ireland and of Her other Realms and Territories Queen, Head of the Commonwealth, Defender of the Faith", "en", nil)
+
 	// regex test data
 	// 0x1234 is uid of interest for regex testing
 	addEdgeToValue(t, "name", 0x1234, "Regex Master", nil)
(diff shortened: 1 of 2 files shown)


pawanrawal:
What would happen if name was in multiple languages and we did an eq for name in one of the languages. Since we pick a language at random we might not get a match?

---

Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[query/query_test.go, line 305 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1297#-KqWYKs1PhDwp5yw_aUc:-KqWYKs1PhDwp5yw_aUd:bahgi91) ([raw file](https://github.com/dgraph-io/dgraph/blob/a934554268a158cfa17ac0cc0f1cfdce00068243/query/query_test.go#L305)):*
> ```Go
> 
> 	// full_name has hash index, we need following data for bug with eq (#1295)
> 	addEdgeToLangValue(t, "full_name", 0x10000, "Her Majesty Elizabeth the Second, by the Grace of God of the United Kingdom of Great Britain and Northern Ireland and of Her other Realms and Territories Queen, Head of the Commonwealth, Defender of the Faith", "en", nil)
> ```

Probably doesn't need to be such a long string.

---

*[worker/sort.go, line 436 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1297#-KqWXtuU6A1Y8vnBqgOP:-KqWXtuU6A1Y8vnBqgOQ:bkaiqys) ([raw file](https://github.com/dgraph-io/dgraph/blob/a934554268a158cfa17ac0cc0f1cfdce00068243/worker/sort.go#L436)):*
> ```Go
> 
> // fetchValue gets the value for a given UID.
> func fetchValue(uid uint64, attr string, langs []string, scalar types.TypeID) (types.Val, error) {
> ```

This function is also called by sortByValue above. We don't want to add a `.` language in that case.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1297#-:-KqWbed3x-__7MitEbNC:bbmwk71)*

tzdybal:
Problem is bigger than I expected (as pointed out by @pawanrawal). I'm closing this PR and open new one when ready.

PR #1335, opened 2017-08-15
Backup requests

srh:
This makes read operations create backup requests if the first hasn't responded within 2 milliseconds.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1335)

### DIFF
--- a/raftwal/wal.go
+++ b/raftwal/wal.go
@@ -114,7 +114,7 @@ func (w *Wal) StoreSnapshot(gid uint32, s raftpb.Snapshot) error {
 	return nil
 }
 
-// Store stores the snapshot, hardstate and entries for a given RAFT group.
+// Store stores the hardstate and entries for a given RAFT group.
 func (w *Wal) Store(gid uint32, h raftpb.HardState, es []raftpb.Entry) error {
 	wb := make([]*badger.Entry, 0, 100)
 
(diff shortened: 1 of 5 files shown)


srh:

---

*[worker/task.go, line 1025 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1335#-KrcBuEcgpDP8r7ey-z0:-KrcBuEdlfu1z9YWsTh0:b-soa69f) ([raw file](https://github.com/dgraph-io/dgraph/blob/9f69bb239880169cdbc95788b9a7cb20484bd881/worker/task.go#L1025)):*
> ```Go
> 	}
> 
> 	x.AssertTruef(groups().ServesGroup(gid),
> ```

The change here is unrelated -- in the case of `ctx.Done()`, the pointer reply could have been initialized but have uninitialized fields (on platforms with certain memory-ordering semantics), which I generally don't want to pass around (even if we do return a non-nil `ctx.Err()`).

But come to think of it, this is a grpc call, and everything I've seen avoids returning nil result values in gRPC calls.  The code before this might return nil too.  Is it the case that we must never return a nil `*protos.Result`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1335)*

srh:
This PR is slightly premature, so I recommend holding off on looking at it until I update.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1335#-:-KrcDTUlV6Lm3pmB38lJ:b-iccmnl)*

PR #1338, opened 2017-08-17
dynamic memory_mb via /admin/request endpoint

peterstace:
* Adds the /admin/request endpoint (PUT)
* I made it generic so that we could add things other than `memory_mb` easily later.
* It's a PUT endpoint. Parameters are encoded in the body in URL query parameter encoding.
* Even bad requests return Status OK (seems to be how the other endpoints do things). Is this what we want?

I did some manual testing to make sure everything works:

```
$ curl -d memory_mb=2000 -d foo=bar -X PUT localhost:8080/admin/request | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   165  100   143  100    22  22587   3474 --:--:-- --:--:-- --:--:-- 23833
{
  "errors": [
    {
      "code": "Success",
      "message": "memory_mb set to 2000.000000"
    },
    {
      "code": "ErrorInvalidRequest",
      "message": "Unknown parameter name: foo"
    }
  ]
}
```
```
$ curl -d memory_mb=1000 -d foo=bar -X PUT localhost:8080/admin/request | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   180  100   158  100    22  26870   3741 --:--:-- --:--:-- --:--:-- 31600
{
  "errors": [
    {
      "code": "ErrorInvalidRequest",
      "message": "memory_mb must be at least 1024"
    },
    {
      "code": "ErrorInvalidRequest",
      "message": "Unknown parameter name: foo"
    }
  ]
}
```
```
$ curl -d memory_mb=not_a_float -d foo=bar -X PUT localhost:8080/admin/request | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   190  100   161  100    29   8931   1608 --:--:-- --:--:-- --:--:--  9470
{
  "errors": [
    {
      "code": "ErrorInvalidRequest",
      "message": "Invalid for memory_mb: not_a_float"
    },
    {
      "code": "ErrorInvalidRequest",
      "message": "Unknown parameter name: foo"
    }
  ]
}
```

I haven't done any HTTP/REST stuff before, so please review extra carefully :)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1338)

### DIFF
--- a/cmd/dgraph/main.go
+++ b/cmd/dgraph/main.go
@@ -466,8 +466,8 @@ func storeStatsHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 // handlerInit does some standard checks. Returns false if something is wrong.
-func handlerInit(w http.ResponseWriter, r *http.Request, allowedMethod string) bool {
-	if r.Method != allowedMethod {
+func handlerInit(w http.ResponseWriter, r *http.Request) bool {
+	if r.Method != http.MethodGet {
 		x.SetStatus(w, x.ErrorInvalidMethod, "Invalid method")
 		return false
 	}
@@ -481,7 +481,7 @@ func handlerInit(w http.ResponseWriter, r *http.Request, allowedMethod string) b
 }
 
 func shutDownHandler(w http.ResponseWriter, r *http.Request) {
-	if !handlerInit(w, r, http.MethodGet) {
+	if !handlerInit(w, r) {
 		return
 	}
 
@@ -510,7 +510,7 @@ func shutdownServer() {
 }
 
 func exportHandler(w http.ResponseWriter, r *http.Request) {
-	if !handlerInit(w, r, http.MethodGet) {
+	if !handlerInit(w, r) {
 		return
 	}
 	ctx := context.Background()
--- a/cmd/dgraph/main.go
+++ b/cmd/dgraph/main.go
@@ -522,41 +522,43 @@ func exportHandler(w http.ResponseWriter, r *http.Request) {
 	w.Write([]byte(`{"code": "Success", "message": "Export completed."}`))
 }
 
-func requestHandler(w http.ResponseWriter, r *http.Request) {
-	if !handlerInit(w, r, http.MethodPut) {
+func memoryLimitHandler(w http.ResponseWriter, r *http.Request) {
+	switch r.Method {
+	case http.MethodGet:
+		memoryLimitGetHandler(w, r)
+	case http.MethodPut:
+		memoryLimitPutHandler(w, r)
+	default:
+		w.WriteHeader(http.StatusMethodNotAllowed)
+	}
+}
+
+func memoryLimitPutHandler(w http.ResponseWriter, r *http.Request) {
+	var memoryMB float64
+	if _, err := fmt.Fscanf(r.Body, "%f", &memoryMB); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
-	if err := r.ParseForm(); err != nil {
-		x.SetStatus(w, x.ErrorInvalidRequest, "Could not parse form: "+err.Error())
+	if memoryMB < dgraph.MinAllottedMemory {
+		w.WriteHeader(http.StatusBadRequest)
+		fmt.Fprintf(w, "memory_mb must be at least %.0f\n", dgraph.MinAllottedMemory)
 		return
 	}
-	var res x.QueryRes
-	for k, v := range r.PostForm {
-		if len(v) == 0 {
-			continue
-		}
-		switch k {
-		case "memory_mb":
-			memStr := v[0]
-			memFlt, err := strconv.ParseFloat(memStr, 64)
-			if err != nil {
-				res.AddStatus(x.ErrorInvalidRequest, "Invalid for memory_mb: "+memStr)
-				continue
-			}
-			if memFlt < dgraph.MinAllottedMemory {
-				res.AddStatus(x.ErrorInvalidRequest,
-					fmt.Sprintf("memory_mb must be at least %.0f", dgraph.MinAllottedMemory))
-				continue
-			}
-			posting.Config.Mu.Lock()
-			posting.Config.AllottedMemory = memFlt
-			posting.Config.Mu.Unlock()
-			res.AddStatus("Success", fmt.Sprintf("memory_mb set to %f", memFlt))
-		default:
-			res.AddStatus(x.ErrorInvalidRequest, "Unknown parameter name: "+k)
-		}
+
+	posting.Config.Mu.Lock()
+	posting.Config.AllottedMemory = memoryMB
+	posting.Config.Mu.Unlock()
+	w.WriteHeader(http.StatusOK)
+}
+
+func memoryLimitGetHandler(w http.ResponseWriter, r *http.Request) {
+	posting.Config.Mu.Lock()
+	memoryMB := posting.Config.AllottedMemory
+	posting.Config.Mu.Unlock()
+
+	if _, err := fmt.Fprintln(w, memoryMB); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
 	}
-	x.Reply(w, res)
 }
 
 func hasGraphOps(mu *protos.Mutation) bool {
@@ -695,7 +697,7 @@ func setupServer(che chan error) {
 	http.HandleFunc("/debug/store", storeStatsHandler)
 	http.HandleFunc("/admin/shutdown", shutDownHandler)
 	http.HandleFunc("/admin/export", exportHandler)
-	http.HandleFunc("/admin/request", requestHandler)
+	http.HandleFunc("/admin/config/memory_mb", memoryLimitHandler)
 
 	// UI related API's.
 	// Share urls have a hex string as the shareId. So if
--- a/cmd/dgraph/main.go
+++ b/cmd/dgraph/main.go
@@ -466,8 +466,8 @@ func storeStatsHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 // handlerInit does some standard checks. Returns false if something is wrong.
-func handlerInit(w http.ResponseWriter, r *http.Request) bool {
-	if r.Method != "GET" {
+func handlerInit(w http.ResponseWriter, r *http.Request, allowedMethod string) bool {
+	if r.Method != allowedMethod {
 		x.SetStatus(w, x.ErrorInvalidMethod, "Invalid method")
 		return false
 	}
@@ -481,7 +481,7 @@ func handlerInit(w http.ResponseWriter, r *http.Request) bool {
 }
 
 func shutDownHandler(w http.ResponseWriter, r *http.Request) {
-	if !handlerInit(w, r) {
+	if !handlerInit(w, r, http.MethodGet) {
 		return
 	}
 
@@ -510,7 +510,7 @@ func shutdownServer() {
 }
 
 func exportHandler(w http.ResponseWriter, r *http.Request) {
-	if !handlerInit(w, r) {
+	if !handlerInit(w, r, http.MethodGet) {
 		return
 	}
 	ctx := context.Background()
@@ -520,6 +520,43 @@ func exportHandler(w http.ResponseWriter, r *http.Request) {
 	}
 	w.Header().Set("Content-Type", "application/json")
 	w.Write([]byte(`{"code": "Success", "message": "Export completed."}`))
+}
+
+func requestHandler(w http.ResponseWriter, r *http.Request) {
+	if !handlerInit(w, r, http.MethodPut) {
+		return
+	}
+	if err := r.ParseForm(); err != nil {
+		x.SetStatus(w, x.ErrorInvalidRequest, "Could not parse form: "+err.Error())
+		return
+	}
+	var res x.QueryRes
+	for k, v := range r.PostForm {
+		if len(v) == 0 {
+			continue
+		}
+		switch k {
+		case "memory_mb":
+			memStr := v[0]
+			memFlt, err := strconv.ParseFloat(memStr, 64)
+			if err != nil {
+				res.AddStatus(x.ErrorInvalidRequest, "Invalid for memory_mb: "+memStr)
+				continue
+			}
+			if memFlt < dgraph.MinAllottedMemory {
+				res.AddStatus(x.ErrorInvalidRequest,
+					fmt.Sprintf("memory_mb must be at least %.0f", dgraph.MinAllottedMemory))
+				continue
+			}
+			posting.Config.Mu.Lock()
+			posting.Config.AllottedMemory = memFlt
+			posting.Config.Mu.Unlock()
+			res.AddStatus("Success", fmt.Sprintf("memory_mb set to %f", memFlt))
+		default:
+			res.AddStatus(x.ErrorInvalidRequest, "Unknown parameter name: "+k)
+		}
+	}
+	x.Reply(w, res)
 }
 
 func hasGraphOps(mu *protos.Mutation) bool {
@@ -658,6 +695,7 @@ func setupServer(che chan error) {
 	http.HandleFunc("/debug/store", storeStatsHandler)
 	http.HandleFunc("/admin/shutdown", shutDownHandler)
 	http.HandleFunc("/admin/export", exportHandler)
+	http.HandleFunc("/admin/request", requestHandler)
 
 	// UI related API's.
 	// Share urls have a hex string as the shareId. So if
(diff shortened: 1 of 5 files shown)


manishrjain:
@janardhan1993 , @srh : Can you please review this change?

janardhan1993:

---

*[cmd/dgraph/main.go, line 552 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1338#-Ks7YiQPhG36mD6WuTsp:-Ks7YiQPhG36mD6WuTsq:bdwnxkx) ([raw file](https://github.com/dgraph-io/dgraph/blob/fdba8cc6ace5d72d1215c8aea0e0606b651e53ad/cmd/dgraph/main.go#L552)):*
> ```Go
> 			}
> 			posting.Config.Mu.Lock()
> 			posting.Config.AllottedMemory = memFlt
> ```

we need to change the lru cache size also. Lru cache size is less than allotted memory. What we can do is probably set it so the ratio lru cache capacity/allotted memory remaings same.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1338)*

manishrjain:

---

*[cmd/dgraph/main.go, line 552 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1338#-Ks7YiQPhG36mD6WuTsp:-Ks7fGbATcWXygXv3mNb:b-gjzk8y) ([raw file](https://github.com/dgraph-io/dgraph/blob/fdba8cc6ace5d72d1215c8aea0e0606b651e53ad/cmd/dgraph/main.go#L552)):*

we need to change the lru cache size also. Lru cache size is less than allotted memory. What we can do is probably set it so the ratio lru cache capacity/allotted memory remaings same.
</blockquote></details>

Hey @janardhan1993 -- can we instead do it so that the lru cache adjusts on a continuous basis -- it increases or decreases by a percentage of the current limit -- dependinging upon the memory usage.

This way, LRU cache would be more flexible and wouldn't need to be directly set from here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1338)*

janardhan1993:
I will open another issue for adaptable lru then.

srh:
The query response has a field named "errors" and then the user has to pick through that list to find a result with code "Success".  This is a bad thing to do because users will not expect successful results in an array named "errors".  It feels unpolished to use that name.  If we have a list of general results, they should be named something neutral (like "results", or something).

Also, it's nice for users of an API to have one place to check to see if an error has happened.  Then, often, simple logic will happen like logging the entire response, or reporting back to the user.  The problem here is, the caller has to check all the results in a loop.

I would also name the URL path name something other than "request".  We aren't retrieving a request or writing to a request.  I've never made a serious HTTP endpoint like this so treat my opinion with a grain of salt, but I'd expect it to involve some path that looks more like `/admin/config/memory_mb` (returning or accepting a number).  Or maybe just `/admin/config`(returning or accepting a JSON document).

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[cmd/dgraph/main.go, line 535 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1338#-Ks7rx04Eg5XO8fwGpD3:-Ks7rx04Eg5XO8fwGpD4:brecqs7) ([raw file](https://github.com/dgraph-io/dgraph/blob/fdba8cc6ace5d72d1215c8aea0e0606b651e53ad/cmd/dgraph/main.go#L535)):*
> ```Go
> 	var res x.QueryRes
> 	for k, v := range r.PostForm {
> 		if len(v) == 0 {
> ```

Why do we skip key/values with empty values?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1338#-:-Ks7uf3FAeSo2YYsP6TZ:bwysvpk)*

peterstace:
@manishrjain @janardhan1993 , I created another issues for the LRU cache change. #1355

peterstace:
Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[cmd/dgraph/main.go, line 535 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1338#-Ks7rx04Eg5XO8fwGpD3:-KsBGDsQbliyE3Sn5tlB:b-kbco9o) ([raw file](https://github.com/dgraph-io/dgraph/blob/fdba8cc6ace5d72d1215c8aea0e0606b651e53ad/cmd/dgraph/main.go#L535)):*

Why do we skip key/values with empty values?
</blockquote></details>

Here we're skipping over keys that have no values (each key can have zero or more values). I couldn't use curl to generate this sort of request, but a handcrafted request of this nature would be possible. Since a few lines down we access `v[0]`, it seems like a good idea to just skip any that have `len(v)==0`.

---

*[cmd/dgraph/main.go, line 552 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1338#-Ks7YiQPhG36mD6WuTsp:-KsBMSCQac1s-KaQMPgi:bu136s2) ([raw file](https://github.com/dgraph-io/dgraph/blob/fdba8cc6ace5d72d1215c8aea0e0606b651e53ad/cmd/dgraph/main.go#L552)):*

Hey @janardhan1993 -- can we instead do it so that the lru cache adjusts on a continuous basis -- it increases or decreases by a percentage of the current limit -- dependinging upon the memory usage.

This way, LRU cache would be more flexible and wouldn't need to be directly set from here.
</blockquote></details>

New issue created.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1338)*

peterstace:
@srh :  Yes, I agree with you (I also haven't designed an HTTP endpoint before). I designed it this way to keep things consistent with the other endpoints.

I did a bit more reading about this -- it looks like the idiomatic way to do this is as follows (basically your first suggestion):

- The endpoint name is `/admin/config/memory_mb`
- A `PUT` is used to change the value. The new value goes in the body as a plain ASCII encoded integer.
- A `GET` is used to get the value. The body of the result can just be a plain ASCII encoded integer.
- We don't necessarily need to handle `GET`, but could add it for completeness.

If everyone is happy with this, I'll make the change.

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1338#-:-KsBMo9jgR_RHS1sPvry:bq1sysl)*

srh:
I was thinking exactly the same thing.

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[cmd/dgraph/main.go, line 535 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1338#-Ks7rx04Eg5XO8fwGpD3:-KsBUqYIEDf8mKFVej_G:b-mtjahv) ([raw file](https://github.com/dgraph-io/dgraph/blob/fdba8cc6ace5d72d1215c8aea0e0606b651e53ad/cmd/dgraph/main.go#L535)):*

Here we're skipping over keys that have no values (each key can have zero or more values). I couldn't use curl to generate this sort of request, but a handcrafted request of this nature would be possible. Since a few lines down we access `v[0]`, it seems like a good idea to just skip any that have `len(v)==0`.
</blockquote></details>

Okay, I get what's going on.  I was wondering why we didn't error, or output an error for that specific key.  But also, I thought that v was a string -- it's actually a `[]string`. But what's happening is, there's a form.  When we parse the form, we (by calling `ParseForm()`) turn the url-encoded list of key1=value1&key2=value2&... pairs into a map of key => [values...].  There is no way to represent that a key has an empty list of values, aside from omitting the key altogether.  So len(v) will never be zero.  Does this sound right to you?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1338#-:-KsBWkkYNvTsu9a2J6b1:b-691qzq)*

peterstace:
Cool, I'll make that change.

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[cmd/dgraph/main.go, line 535 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1338#-Ks7rx04Eg5XO8fwGpD3:-KsBYQXQIPBKtIxBL5fv:b-uejy5d) ([raw file](https://github.com/dgraph-io/dgraph/blob/fdba8cc6ace5d72d1215c8aea0e0606b651e53ad/cmd/dgraph/main.go#L535)):*

Okay, I get what's going on.  I was wondering why we didn't error, or output an error for that specific key.  But also, I thought that v was a string -- it's actually a `[]string`. But what's happening is, there's a form.  When we parse the form, we (by calling `ParseForm()`) turn the url-encoded list of key1=value1&key2=value2&... pairs into a map of key => [values...].  There is no way to represent that a key has an empty list of values, aside from omitting the key altogether.  So len(v) will never be zero.  Does this sound right to you?
</blockquote></details>

Yes, that's how I understand it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1338#-:-KsBYNWQ4u6fyGMwkqCh:b-6qxaa1)*

peterstace:
There is now separate GET and PUT handlers, and those handlers use the traditional HTTP status codes (rather than e.g. always returning 200). The reason that the query (and other) handlers always return 200 even if something goes wrong is because that's the convention for GraphQL endpoints.

Some examples from my testing:

```
$ curl -v localhost:8080/admin/config/memory_mb
*   Trying ::1...
* TCP_NODELAY set
* connect to ::1 port 8080 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET /admin/config/memory_mb HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.54.1
> Accept: */*
>
< HTTP/1.1 200 OK
< Date: Wed, 23 Aug 2017 02:07:15 GMT
< Content-Length: 5
< Content-Type: text/plain; charset=utf-8
<
1024
* Connection #0 to host localhost left intact
```

```
$ curl -v -X PUT -d 2000 localhost:8080/admin/config/memory_mb
*   Trying ::1...
* TCP_NODELAY set
* connect to ::1 port 8080 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8080 (#0)
> PUT /admin/config/memory_mb HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.54.1
> Accept: */*
> Content-Length: 4
> Content-Type: application/x-www-form-urlencoded
>
* upload completely sent off: 4 out of 4 bytes
< HTTP/1.1 200 OK
< Date: Wed, 23 Aug 2017 02:08:47 GMT
< Content-Length: 0
< Content-Type: text/plain; charset=utf-8
<
* Connection #0 to host localhost left intact
```

```
$ curl -v -X PUT -d 500 localhost:8080/admin/config/memory_mb
*   Trying ::1...
* TCP_NODELAY set
* connect to ::1 port 8080 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8080 (#0)
> PUT /admin/config/memory_mb HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.54.1
> Accept: */*
> Content-Length: 3
> Content-Type: application/x-www-form-urlencoded
>
* upload completely sent off: 3 out of 3 bytes
< HTTP/1.1 400 Bad Request
< Date: Wed, 23 Aug 2017 02:09:07 GMT
< Content-Length: 32
< Content-Type: text/plain; charset=utf-8
<
memory_mb must be at least 1024
* Connection #0 to host localhost left intact
```

```
 $ curl -v localhost:8080/admin/config/memory_mb
*   Trying ::1...
* TCP_NODELAY set
* connect to ::1 port 8080 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET /admin/config/memory_mb HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.54.1
> Accept: */*
>
< HTTP/1.1 200 OK
< Date: Wed, 23 Aug 2017 02:09:22 GMT
< Content-Length: 5
< Content-Type: text/plain; charset=utf-8
<
2000
* Connection #0 to host localhost left intact
```

```
$ curl -v -X PUT -d aoeu localhost:8080/admin/config/memory_mb
*   Trying ::1...
* TCP_NODELAY set
* connect to ::1 port 8080 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8080 (#0)
> PUT /admin/config/memory_mb HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.54.1
> Accept: */*
> Content-Length: 4
> Content-Type: application/x-www-form-urlencoded
>
* upload completely sent off: 4 out of 4 bytes
< HTTP/1.1 400 Bad Request
< Content-Type: text/plain; charset=utf-8
< X-Content-Type-Options: nosniff
< Date: Wed, 23 Aug 2017 02:09:38 GMT
< Content-Length: 47
<
strconv.ParseFloat: parsing "": invalid syntax
* Connection #0 to host localhost left intact
```

```
$ curl -v -X GET localhost:8080/admin/config/memory_mb
Note: Unnecessary use of -X or --request, GET is already inferred.
*   Trying ::1...
* TCP_NODELAY set
* connect to ::1 port 8080 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET /admin/config/memory_mb HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.54.1
> Accept: */*
>
< HTTP/1.1 200 OK
< Date: Wed, 23 Aug 2017 02:09:50 GMT
< Content-Length: 5
< Content-Type: text/plain; charset=utf-8
<
2000
* Connection #0 to host localhost left intact
```

---

Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1338#-:-KsBpHidcQdvrzvL8JWm:b-np4wa2)*

srh:
OK.  The code looks acceptable to me, except for my comment I left, and also, the fact that the question of whether we should return a different HTTP status code upon error, or not, is something I don't have the experience to have an opinion about.

---

Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[cmd/dgraph/main.go, line 538 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1338#-KsC54PRS3d_LWkbRbzf:-KsC54PRS3d_LWkbRbzg:bk8xiz3) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c7d4578cf3c33cc75d30f887777beb3b0ec9580/cmd/dgraph/main.go#L538)):*
> ```Go
> func memoryLimitPutHandler(w http.ResponseWriter, r *http.Request) {
> 	var memoryMB float64
> 	if _, err := fmt.Fscanf(r.Body, "%f", &memoryMB); err != nil {
> ```

Can we ensure that this number we parse is the entirety of the request body?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1338#-:-KsC4cs6oWs3zVckOFuE:bn1utxr)*

srh:
Reviewed 3 of 5 files at r1, 1 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1338)*

peterstace:
Review status: 3 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[cmd/dgraph/main.go, line 538 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1338#-KsC54PRS3d_LWkbRbzf:-KsC9orTL4L1EUgxYBHm:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/3c7d4578cf3c33cc75d30f887777beb3b0ec9580/cmd/dgraph/main.go#L538)):*

Can we ensure that this number we parse is the entirety of the request body?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1338)*

srh:
:lgtm:

---

Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1338#-:-KsCIIQoLBD5ggB58CjD:bnfp4nl)*

PR #1347, opened 2017-08-21
correct "Blade Runner" uid.

fiatjaf:
Because it must be done.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1347)

### DIFF
--- a/wiki/content/query-language/index.md
+++ b/wiki/content/query-language/index.md
@@ -51,7 +51,7 @@ Query Example: "Blade Runner" movie data found by UID.
 
 {{< runnable >}}
 {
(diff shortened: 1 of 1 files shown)


manishrjain:
The UIDs change frequently because we delete and reload the entire data every time we make a release. There's no guarantee that the same entity would get the same UID. So, this change isn't useful. I'd recommend reverting the PR.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1347#-:-Ks67AYM3KGnH1XY4ILm:brty383)*

Issue #1345, opened 2017-08-21
Waiting for a WaterMark might return prematurely


srh:
You'll note in places like `waitForAppliedMark` or any other code that waits for a WaterMark to be satisfied, we have for loops that begin, `for n.applied.WaitingFor() {`.

What that function `WaitingFor()` does is read an atomic variable that doesn't get set to 1 until the first `Mark` makes it over the channel and gets processed.  Since we don't force these marks over the channel (WaterMark::Ch) to get processed, it's possible that the atomic variable has not yet been set to 1.  Thus it's possible (if we imagine the mark channel never getting consumed) all watermarks return immediately.

I don't understand the purpose of WaitingFor, so there might be something here I'm missing.

Edit: There's one place that waits on WaitingFor exclusively. Looking at the history it seems the variable was not introduced with a strong intent behind what it accomplishes, so I would feel comfortable reworking it. Snce I've already made waiting for watermarks not sleep-wait I'll figure this out too.

manishrjain:
Those are useful for testing purposes. @janardhan1993: Can you investigate this?

srh:
Is it related to IsTestRun/SetTestRun type stuff?  What is the point of turning off watermark behavior during test runs?

manishrjain:
Speaking roughly, mutations might not get proposed to Raft, but expect to be applied/synced to disk.

janardhan1993:
I had a look at the code. WaitingFor is used in mostly for reindexing,delete predicate mutation(We need previous entries to be synced to disk since we iterate over badger):  Since everything is serialised by raft, if you look at the code we force the pending marks over the channel before processing anything. 

Atleast for our present use cases it works. But as @srh pointed out it won't work if we want to wait for an index (for read linearazability) for which pending mark has not been pushed over the channel. 
I guess the intent of waitingFor was to check for all indices for which mark was pushed over the channel. That is the first thing we do in draft.go:Run method. 

```
 for _, entry := range rd.CommittedEntries {
   4         status := x.Mark{Index: entry.Index, Done: false}
```

srh:
How do we force pending marks over the channel?

janardhan1993:
My bad we just push it to the buffered channel and don't force it to be processed.

srh:
Fixed by #1377.

PR #1366, opened 2017-08-28
Make watermark not use timer

srh:
This avoids the question of fixing issue #1345 -- it keeps watermark behavior the same.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1366)

### DIFF
--- a/client/mutations.go
+++ b/client/mutations.go
@@ -372,7 +372,7 @@ RETRY:
 	// Mark watermarks as done.
 	if req.line != 0 && req.mark != nil {
 		atomic.AddUint64(&d.rdfs, uint64(req.size()))
-		req.mark.Ch <- x.Mark{Index: req.line, Done: true}
+		req.mark.Done(req.line)
 	}
 	return nil
 }
@@ -489,7 +489,7 @@ func (d *Dgraph) BatchSetWithMark(r *Req, file string, line uint64) error {
 	if sm.mark != nil && line != 0 {
 		r.mark = sm.mark
 		r.line = line
-		sm.mark.Ch <- x.Mark{Index: line}
+		sm.mark.Begin(line)
 	}
 
 L:
--- a/cmd/dgraph/main_test.go
+++ b/cmd/dgraph/main_test.go
@@ -1370,8 +1370,6 @@ func TestMain(m *testing.M) {
 
 	dir1, dir2, _ := prepare()
 
-	// we need watermarks for reindexing
-	x.AssertTrue(!x.IsTestRun())
 	// Parse GQL into internal query representation.
 	r := m.Run()
 	closeAll(dir1, dir2)
--- a/posting/list.go
+++ b/posting/list.go
@@ -481,7 +481,7 @@ func (l *List) addMutation(ctx context.Context, t *protos.DirectedEdge) (bool, e
 
 	if hasMutated {
 		if index != 0 {
-			l.water.Ch <- x.Mark{Index: index}
+			l.water.Begin(index)
 			l.pending = append(l.pending, index)
 		}
 		// if mutation doesn't come via raft
@@ -503,7 +503,7 @@ func (l *List) delete(ctx context.Context, attr string) error {
 
 	var gid uint32
 	if rv, ok := ctx.Value("raft").(x.RaftValue); ok {
-		l.water.Ch <- x.Mark{Index: rv.Index}
+		l.water.Begin(rv.Index)
 		l.pending = append(l.pending, rv.Index)
 		gid = rv.Group
 	}
@@ -640,7 +640,7 @@ func (l *List) syncIfDirty(delFromCache bool) (committed bool, err error) {
 	// deleteAll is used to differentiate when we don't have any updates, v/s
 	// when we have explicitly deleted everything.
 	if len(l.mlayer) == 0 && atomic.LoadInt32(&l.deleteAll) == 0 {
-		l.water.Ch <- x.Mark{Indices: l.pending, Done: true}
+		l.water.DoneMany(l.pending)
 		l.pending = make([]uint64, 0, 3)
 		return false, nil
 	}
@@ -721,7 +721,7 @@ func (l *List) syncIfDirty(delFromCache bool) (committed bool, err error) {
 		x.BytesWrite.Add(int64(len(data)))
 		x.PostingWrites.Add(1)
 		if l.water != nil {
-			l.water.Ch <- x.Mark{Indices: pending, Done: true}
+			l.water.DoneMany(pending)
 		}
 		if delFromCache {
 			x.AssertTrue(atomic.LoadInt32(&l.deleteMe) == 1)
--- a/posting/list_test.go
+++ b/posting/list_test.go
@@ -666,7 +666,6 @@ func TestAfterUIDCountWithCommit(t *testing.T) {
 var ps *badger.KV
 
 func TestMain(m *testing.M) {
-	x.SetTestRun()
 	x.Init()
 	Config.AllottedMemory = 1024.0
 	Config.CommitFraction = 0.10
--- a/query/query_test.go
+++ b/query/query_test.go
@@ -6777,7 +6777,6 @@ lossy                          : string @index(term) .
 `
 
 func TestMain(m *testing.M) {
-	x.SetTestRun()
 	x.Init()
 
 	dir, err := ioutil.TempDir("", "storetest_")
--- a/schema/parse_test.go
+++ b/schema/parse_test.go
@@ -265,7 +265,6 @@ func TestParse8_Error(t *testing.T) {
 var ps *badger.KV
 
 func TestMain(m *testing.M) {
-	x.SetTestRun()
 	x.Init()
 
 	dir, err := ioutil.TempDir("", "storetest_")
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -312,10 +312,10 @@ type SyncEntry struct {
 	Index  uint64
 }
 
-func addToEntriesMap(entriesMap map[chan x.Mark][]uint64, entries []SyncEntry) {
+func addToEntriesMap(entriesMap map[*x.WaterMark][]uint64, entries []SyncEntry) {
 	for _, entry := range entries {
 		if entry.Water != nil {
-			entriesMap[entry.Water.Ch] = append(entriesMap[entry.Water.Ch], entry.Index)
+			entriesMap[entry.Water] = append(entriesMap[entry.Water], entry.Index)
 		}
 	}
 }
@@ -350,10 +350,10 @@ func batchSync() {
 		pstore.BatchSet(wb)
 		wb = wb[:0]
 
-		entriesMap := make(map[chan x.Mark][]uint64)
+		entriesMap := make(map[*x.WaterMark][]uint64)
 		addToEntriesMap(entriesMap, entries)
-		for ch, indices := range entriesMap {
-			ch <- x.Mark{Indices: indices, Done: true}
+		for wm, indices := range entriesMap {
+			wm.Ch <- x.Mark{Indices: indices, Done: true}
 		}
 		entries = entries[:0]
 	}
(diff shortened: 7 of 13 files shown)


srh:
Review status: 0 of 13 files reviewed at latest revision, 4 unresolved discussions.

---

*[client/mutations.go, line 375 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1366#-KsbLyuIxaj0tSQ-cqWc:-KsbLyuIxaj0tSQ-cqWd:b-6ix1xd) ([raw file](https://github.com/dgraph-io/dgraph/blob/3f34f987383b63e9d96d9e2f5f9b9789cce3de15/client/mutations.go#L375)):*
> ```Go
> 	if req.line != 0 && req.mark != nil {
> 		atomic.AddUint64(&d.rdfs, uint64(req.size()))
> 		req.mark.Done(req.line)
> ```

Wound up encapsulating the channel.

---

*[cmd/dgraph/main_test.go, line 1374 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1366#-KsbM2HtJPIJtj5R_if_:-KsbM2HtJPIJtj5R_ifa:b-sso6y7) ([raw file](https://github.com/dgraph-io/dgraph/blob/d14bb29e1345f957440d4fab3a660af00f272860/cmd/dgraph/main_test.go#L1374)):*
> ```Go
> 
> 	// we need watermarks for reindexing
> 	x.AssertTrue(!x.IsTestRun())
> ```

I removed the SetTestRun/IsTestRun stuff.  I don't know what exactly its reason for existing was -- tests still work.

---

*[worker/groups.go, line 664 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1366#-KsbMKuO_1mlHQZn1oLz:-KsbMKuP5Twizm__5qQr:ba8h6sm) ([raw file](https://github.com/dgraph-io/dgraph/blob/3f34f987383b63e9d96d9e2f5f9b9789cce3de15/worker/groups.go#L664)):*
> ```Go
> 				return
> 			}
> 			n.syncAllMarks(ctx, lastIndex)
> ```

syncAllMarks never returned an error before.

---

*[worker/index.go, line 90 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1366#-KsbMRnmEmsks1_3GhdP:-KsbMRnmEmsks1_3GhdQ:b-e51goi) ([raw file](https://github.com/dgraph-io/dgraph/blob/3f34f987383b63e9d96d9e2f5f9b9789cce3de15/worker/index.go#L90)):*
> ```Go
> }
> 
> func (n *node) waitForAppliedMark(ctx context.Context, lastIndex uint64) {
> ```

The downside of centralizing WaitForMark is that we do lose the specific message.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1366)*

manishrjain:
Nice refactoring. Have some comments to simplify the waiter code.

Also, I think the reason for those test run cases was that query package was failing randomly. You might want to run the tests multiple times to ensure that the failures don't reoccur. We might have modified the tests in some way, which might have fixed the issue; but hard to say.

---

Reviewed 12 of 13 files at r1, 1 of 1 files at r2.

---

*[worker/index.go, line 91 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1366#-KsbjUgES6WMXpaQD7ue:-KsbjUgFzZ0P-SiOF984:b-1ny8u1) ([raw file](https://github.com/dgraph-io/dgraph/blob/52ec7f97c60b8fa7031d057c80457229a03ab2e3/worker/index.go#L91)):*
> ```Go
> 
> func (n *node) waitForAppliedMark(ctx context.Context, lastIndex uint64) {
> 	n.applied.WaitForMark(lastIndex)
> ```

Can remove this single line function.

---

*[x/watermark.go, line 179 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1366#-KsbkZqlYXERnmTGDpJx:-KsbkZqlYXERnmTGDpJy:b483g4e) ([raw file](https://github.com/dgraph-io/dgraph/blob/52ec7f97c60b8fa7031d057c80457229a03ab2e3/x/watermark.go#L179)):*
> ```Go
> 				// Some partly duplicated code with what's below.
> 				heap.Pop(&waiterIndices)
> 				toNotify := waiters[min]
> ```

This logic seems weird. We might have someone waiting for 101, but the heap only sees 10, 50 and 150. This map lookup would then not trigger; or so it seems.

A simpler solution might be to keep a heap of waiters and compare the waiter index against the min done index. Not sure if a heap allows duplication, but if it does, then we'd avoid having a map of list of chan; which is another benefit to readability.

P.S. Another thought was that you can attach a waiter to an index in the heap itself, thus making the wait done trigger even simpler. Just add the index as a done index when attaching waiter to the heap element.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1366#-:-KsbmlID-l3UVqNMwh1H:b-iobmb2)*

srh:
I'll run that test in a loop overnight.

---

Review status: 12 of 13 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[worker/index.go, line 91 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1366#-KsbjUgES6WMXpaQD7ue:-Ksbv2cLamAMUNfsZKr_:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/52ec7f97c60b8fa7031d057c80457229a03ab2e3/worker/index.go#L91)):*

Can remove this single line function.
</blockquote></details>

Done.

---

*[x/watermark.go, line 179 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1366#-KsbkZqlYXERnmTGDpJx:-Ksbvuk0ocyqQ17Q3ilr:bbfyzfw) ([raw file](https://github.com/dgraph-io/dgraph/blob/52ec7f97c60b8fa7031d057c80457229a03ab2e3/x/watermark.go#L179)):*

This logic seems weird. We might have someone waiting for 101, but the heap only sees 10, 50 and 150. This map lookup would then not trigger; or so it seems.

A simpler solution might be to keep a heap of waiters and compare the waiter index against the min done index. Not sure if a heap allows duplication, but if it does, then we'd avoid having a map of list of chan; which is another benefit to readability.

P.S. Another thought was that you can attach a waiter to an index in the heap itself, thus making the wait done trigger even simpler. Just add the index as a done index when attaching waiter to the heap element.
</blockquote></details>

If we have someone waiting for 101, then the heap will contain 101. `waiterIndices` is a different heap from `indices`.

We could keep a heap of waiters -- it can surely have duplicates -- the main reason I didn't do that was because I just mirrored the existing implementation with pending/indices.

If we put waiters on the pending/indices heap I think it would make the 10/50/150 while somebody waits on 101 case much more difficult.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1366#-:-KsbvN2ZgnYPzJzJnok2:b-tleaod)*

srh:
I have not seen test failures.

---

Review status: 12 of 13 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1366#-:-Ksg-ONeTRqodPkUioLL:byzac2b)*

manishrjain:
Review status: 12 of 13 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[x/watermark.go, line 179 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1366#-KsbkZqlYXERnmTGDpJx:-KsgO25IIP77ntMI5OaL:biyyg7g) ([raw file](https://github.com/dgraph-io/dgraph/blob/52ec7f97c60b8fa7031d057c80457229a03ab2e3/x/watermark.go#L179)):*
> If we put waiters on the pending/indices heap I think it would make the 10/50/150 while somebody waits on 101 case much more difficult.

Not really, in fact it should be simplified. Say, the heap has 10, 50 and 150. It's waiting on 10 right now. Then we ask to wait for 101. We add 101 in the heap (as done), and attach the waiting channel to it. Once 10 is done, and then 50 is done, heap moves to 101 -- sees that it's already done, and writes to the channel, and moves onto 150. This is just simpler logic to think through and hence code.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1366)*

srh:
Review status: 12 of 13 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[x/watermark.go, line 179 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1366#-KsbkZqlYXERnmTGDpJx:-KsgPi7h_R5JMefYAcUY:b-ieypzu) ([raw file](https://github.com/dgraph-io/dgraph/blob/52ec7f97c60b8fa7031d057c80457229a03ab2e3/x/watermark.go#L179)):*

> If we put waiters on the pending/indices heap I think it would make the 10/50/150 while somebody waits on 101 case much more difficult.

Not really, in fact it should be simplified. Say, the heap has 10, 50 and 150. It's waiting on 10 right now. Then we ask to wait for 101. We add 101 in the heap (as done), and attach the waiting channel to it. Once 10 is done, and then 50 is done, heap moves to 101 -- sees that it's already done, and writes to the channel, and moves onto 150. This is just simpler logic to think through and hence code.
</blockquote></details>

What if after we consider 101 to be done, 101 begins?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1366)*

manishrjain:
This particular change is a NOOP from that perspective. We already put
things in place so that the watermark doesn't miss marks. In fact,
watermark code has an asset failure for this case.

srh:
The watermark code has an assert that we don't create entries below doneUntil.  It does not assert that a waiter for level L can be added after a begin Mark for L has been made, but never before.

---

Review status: 12 of 13 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1366#-:-KsghSbYQ6a9YjCz1vWb:bh5zf2g)*

manishrjain:
Review status: 12 of 13 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[x/watermark.go, line 125 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1366#-KsgvLRPm0Btl9nMfYId:-KsgvLRPm0Btl9nMfYIe:b-gfxzsg) ([raw file](https://github.com/dgraph-io/dgraph/blob/ffb047f53146e0acb8c2b684a0eb52eba6b0035d/x/watermark.go#L125)):*
> ```Go
> 	// pending maps raft proposal index to the number of pending mutations for this proposal.
> 	pending := make(map[uint64]int)
> 	waiters := make(map[uint64][]chan struct{})
> ```

Just use one chan, and close it to represent done for all waiters.

---

*[x/watermark.go, line 208 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1366#-Ksgu-njK5kMZRedFKtH:-Ksgu-njK5kMZRedFKtI:b-wq1fyk) ([raw file](https://github.com/dgraph-io/dgraph/blob/ffb047f53146e0acb8c2b684a0eb52eba6b0035d/x/watermark.go#L208)):*
> ```Go
> 
> 	for mark := range w.markCh {
> 		if mark.waiter != nil {
> ```

Maybe better to do this:

```
if mark.index > 0 {
  ...
  continue
}
if len(mark.indices) > 0 {
  ...
  continue
}
if mark.waiter != nil {
   ...
  continue
}
```

Reads better. Also, mark.index is the most frequent code path, followed by indices, followed by waiter. This puts them in that order.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1366)*

manishrjain:
:lgtm:

---

Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1366#-:-Ksgwy5nQa2QAf4COavG:bnfp4nl)*

srh:
In master as of 5e1709e7369106be9434b1d0ef86a344914345ce.

PR #1378, opened 2017-08-30
Make syncMarks::create assert non-nil map

srh:
Its behavior didn't really make sense on a nil map, because it assigned to a local var.

But, my question is: Is the case of a nil map supposed to be handled somehow?  Is the assertion wrong?

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1378)

### DIFF
--- a/client/checkpoint.go
+++ b/client/checkpoint.go
@@ -93,9 +93,7 @@ func (d *Dgraph) writeCheckpoint() {
 }
 
 func (g syncMarks) create(file string) waterMark {
-	if g == nil {
-		g = make(map[string]waterMark)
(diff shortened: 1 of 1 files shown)


pawanrawal:
:lgtm: 
Assertion is good. I guess this worked since we never hit this case as `marks` is initialized in NewDgraphClient.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1378#-:-KslkBIKqR2155lHN33m:bsok12d)*

Issue #1344, opened 2017-08-20
Linearizable reads/writes


srh:
(We didn't have an issue for this.)  The feature we want is for read behavior to be linearizable, instead of happily reading outdated data.

The short of it is, we're going to use etcd/raft's ReadIndex functionality.  To avoid spamming Raft groups' leaders with MsgReadIndex messages, each follower node will only make one in-flight request at a time.  We'll do this from whatever node serves our Raft group.

That's the most basic version -- we could get more clever about how we do ReadIndex in the future to reduce latency.  We could allow multiple in-flight requests (and take care not to let them bunch up). 
 We could send our requests from other machines to avoid extra network hops. Some descriptions of alternatives are here: https://discuss.dgraph.io/t/linearizability/1683/3

We will also need to change the watermark implementation to not use sleep.

srh:
So, linearizability is implemented (albeit under-tested, and the simplest possible implementation as described).

One thing that isn't implemented, though is the question, what should the UI be?  How should the user specify a linearizable read vs. a non-linearizable one?

Option 1: an out-of-band flag.
Option 2: the flag gets parsed out of the query/schema read query somehow.
Option 3: it's not a mere flag -- the user could supply a {groupId => readIndex} bundle.

manishrjain:
All reads should be linearizable for now. If performance becomes an issue, we'll solve for it when it does.

srh:
I'm getting 14 second reads with a simple curl query while running dgraphloader.  Running the Raft ReadIndex operation is quick, but then it takes about 12 seconds between getting a linearizable read index and for `waitForAppliedMark` to return.  It looks like the time duration between a change reaching a majority of nodes and actually being "applied" can be large.

manishrjain:
> It looks like the time duration between a change reaching a majority of nodes and actually being "applied" can be large.

This is because we're using goroutines for each mutation which get scheduled whenever they do. @janardhan1993 is working on a smart scheduler which can prioritize mutations in a roughly FIFO order. That should help with the reads.

But, great to hear read linearizability is working! PR?

janardhan1993:
To minimise the latency, i guess we need to schedule all the mtuations which came after read request only when the applied watermark catches up.

srh:
Linearizability won't be fully working until #1345 because it does in fact trigger that scenario sometimes.

srh:
#1345 was fixed, so this can be closed.

PR #1383, opened 2017-08-30
Linearizable reads

srh:
Connected to #1344.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1383)


srh:
Review status: 0 of 13 files reviewed at latest revision, 1 unresolved discussion.

---

*[protos/facets.pb.go, line 1 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KspMrlNCewQnk1lQmm4:-KspMrlNCewQnk1lQmm5:bqp565p) ([raw file](https://github.com/dgraph-io/dgraph/blob/11a3e5641ce92e78305c11e7ac74719ae063bb4b/protos/facets.pb.go#L1)):*
> ```Go
> // Code generated by protoc-gen-gogo. DO NOT EDIT.
> ```

At some point I changed protos and changed them back -- and in regenerating, I undid some changes Pawan did with a recent regeneration.  I'm using version 3.3.0.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383)*

manishrjain:
Review status: 0 of 13 files reviewed at latest revision, 5 unresolved discussions, some commit checks broke.

---

*[worker/draft.go, line 757 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KsqpwuFofidr97t-d_A:-KsqpwuFofidr97t-d_B:bfbwhqt) ([raw file](https://github.com/dgraph-io/dgraph/blob/11a3e5641ce92e78305c11e7ac74719ae063bb4b/worker/draft.go#L757)):*
> ```Go
> 
> func (ls *linearizableState) feedRequest(req linearizableReadRequest) {
> 	ls.pendingRequests = append(ls.pendingRequests, req.readIndexCh)
> ```

Is `feedRequest` called serially compared to `slurpRequests`? If so, can you add a comment about why a lock isn't required here.

---

*[worker/draft.go, line 791 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KsrXpxoRjurBMXfrtIB:-KsrXpxoRjurBMXfrtIC:be8uy24) ([raw file](https://github.com/dgraph-io/dgraph/blob/11a3e5641ce92e78305c11e7ac74719ae063bb4b/worker/draft.go#L791)):*
> ```Go
> 	rctxCounter := ls.rctxCounter.Generate()
> 	ls.activeRequestRctx = rctxCounter[:]
> 	ls.activeRequestTimer.Reset(readIndexTimeout)
> ```

Why this timer? Can't you use a new `contextWithTimeout` instead?

---

*[worker/draft.go, line 799 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KsrY-ACORMlrVT3akDl:-KsrY-ACORMlrVT3akDm:b-ty0euu) ([raw file](https://github.com/dgraph-io/dgraph/blob/11a3e5641ce92e78305c11e7ac74719ae063bb4b/worker/draft.go#L799)):*
> ```Go
> }
> 
> const raftTickRate = 20 * time.Millisecond
> ```

If this const is being used in exactly one place, then it doesn't need to be defined elsewhere.

---

*[x/noncecounter.go, line 8 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KsrWs2A2TnqOmhiHkss:-KsrWs2A2TnqOmhiHkst:b-m77j4x) ([raw file](https://github.com/dgraph-io/dgraph/blob/11a3e5641ce92e78305c11e7ac74719ae063bb4b/x/noncecounter.go#L8)):*
> ```Go
> )
> 
> // NonceCounter can generate unique 128-bit numbers.  Only usable on a single thread.  (All it
> ```

NonceCounter seems unnecessarily more expensive than it's purpose. crypto libraries are expensive, also using 128-bit numbers are as well. Seems like a simple monotonically increasing counter should do the job.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383)*

srh:
Review status: 0 of 13 files reviewed at latest revision, 5 unresolved discussions.

---

*[worker/draft.go, line 757 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KsqpwuFofidr97t-d_A:-KsvB9RUyQFaeE1FYvbf:bnmvszq) ([raw file](https://github.com/dgraph-io/dgraph/blob/11a3e5641ce92e78305c11e7ac74719ae063bb4b/worker/draft.go#L757)):*

Is `feedRequest` called serially compared to `slurpRequests`? If so, can you add a comment about why a lock isn't required here.
</blockquote></details>

Yes.  Added a bunch of comments about that.

---

*[worker/draft.go, line 791 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KsrXpxoRjurBMXfrtIB:-Ksv2oquxi1eyJw3C7np:bjuj7zl) ([raw file](https://github.com/dgraph-io/dgraph/blob/11a3e5641ce92e78305c11e7ac74719ae063bb4b/worker/draft.go#L791)):*

Why this timer? Can't you use a new `contextWithTimeout` instead?
</blockquote></details>

If a context were used, it still would only be used the same way as this timer would.  It wouldn't be helpful to have it to pass to ReadIndex, for example.  ReadIndex's ctx parameter is only used while waiting for the node's run() thread to read it off `recvc`.  We have a timer because the raft library could discard the message if it doesn't have the leader configured, or because it could send off the message and never get a reply.

---

*[worker/draft.go, line 799 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KsrY-ACORMlrVT3akDl:-Ksv8XZ8oIxAiIeI_vBf:bpb386r) ([raw file](https://github.com/dgraph-io/dgraph/blob/11a3e5641ce92e78305c11e7ac74719ae063bb4b/worker/draft.go#L799)):*

If this const is being used in exactly one place, then it doesn't need to be defined elsewhere.
</blockquote></details>

OK, I can put it inside the function.  The reason I made it a const declaration is because it's useful to be able to search a codebase for "const" and see what arbitrary constants are there.  When put in a const there's more information.  You know with higher likelihood that nothing else depends on the number's value being what it is than if it's just quietly used in some expression.  But you can put a const inside a function.

---

*[x/noncecounter.go, line 8 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KsrWs2A2TnqOmhiHkss:-Ksv4b-lh0FQjUZt3mtW:btm7gl3) ([raw file](https://github.com/dgraph-io/dgraph/blob/11a3e5641ce92e78305c11e7ac74719ae063bb4b/x/noncecounter.go#L8)):*

NonceCounter seems unnecessarily more expensive than it's purpose. crypto libraries are expensive, also using 128-bit numbers are as well. Seems like a simple monotonically increasing counter should do the job.
</blockquote></details>

I believe if a node dies and reconnects it could get delivered old Raft messages.  For example, a node sends a MsgReadIndex, dies, restarts, reconnects, and the leader sends a MsgReadIndexResp back.  Assuming that's correct, we need some way to distinguish old values from new ones.  Reading 16 bytes off of /dev/urandom at startup once per group is cheap enough.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383)*

manishrjain:
Reviewed 12 of 13 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks broke.

---

*[worker/draft.go, line 799 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KsrY-ACORMlrVT3akDl:-KsvoI4bgagmINi0ckud:b-od0c09) ([raw file](https://github.com/dgraph-io/dgraph/blob/11a3e5641ce92e78305c11e7ac74719ae063bb4b/worker/draft.go#L799)):*

OK, I can put it inside the function.  The reason I made it a const declaration is because it's useful to be able to search a codebase for "const" and see what arbitrary constants are there.  When put in a const there's more information.  You know with higher likelihood that nothing else depends on the number's value being what it is than if it's just quietly used in some expression.  But you can put a const inside a function.
</blockquote></details>

Typically, you don't start by doing a search for consts. You read a logic, then see a const being used, then jump to it to read it's value. Having the const be used directly, avoid those extra jumps.

---

*[worker/draft.go, line 776 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KsvkoRW_ESvOCcsznL_:-KsvkoRW_ESvOCcsznLa:b-lxlr39) ([raw file](https://github.com/dgraph-io/dgraph/blob/806da55b856f081fc0c2edfeece7b8618985bdff/worker/draft.go#L776)):*
> ```Go
> // send another ReadIndex request for the next batch of requesters.  Called serially by the Run
> // loop.
> func cycleResponses(ls *linearizableState, n *node, indexOrNone uint64) error {
> ```

This whole logic could be simplified as follows:

```
	go func() {
		var requests []request
		for {
			select {
			case req := <-reqCh:
				requests = append(requests, req)
			default:
				// push to node for a new read index.
				// Get back a read index. Can be done with a select and timeout.
				for _, req := range requests {
					req.Ch <- struct{} // or the read index.
				}
				requests = requets[:0]
			}
		}
	}

```

---

*[worker/draft.go, line 824 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-Ksvo4YyJb3G-JuuR45b:-Ksvo4YyJb3G-JuuR45c:b-mom8zg) ([raw file](https://github.com/dgraph-io/dgraph/blob/806da55b856f081fc0c2edfeece7b8618985bdff/worker/draft.go#L824)):*
> ```Go
> 		case rd := <-n.Raft().Ready():
> 			for _, rs := range rd.ReadStates {
> 				if 0 != bytes.Compare(n.linState.activeRequestRctx, rs.RequestCtx) {
> ```

This can just push to a channel in linState, which the goroutines within it can handle to cycle responses etc. The timeout should be handled within linState and not exposed here.

---

*[worker/draft.go, line 898 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KsvnRtHK_QSwl6sccWA:-KsvnRtHK_QSwl6sccWB:bvsjz1g) ([raw file](https://github.com/dgraph-io/dgraph/blob/806da55b856f081fc0c2edfeece7b8618985bdff/worker/draft.go#L898)):*
> ```Go
> 			}
> 
> 		case req := <-n.linState.requests:
> ```

This doesn't really belong here. linState can have it's own goroutine.

---

*[worker/draft.go, line 1207 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KsvnDUpWmHKtGcnmLcl:-KsvnDUpWmHKtGcnmLcm:bizq4y0) ([raw file](https://github.com/dgraph-io/dgraph/blob/806da55b856f081fc0c2edfeece7b8618985bdff/worker/draft.go#L1207)):*
> ```Go
> func waitLinearizableRead(ctx context.Context, gid uint32) error {
> 	n := groups().Node(gid)
> 	replyCh := n.linState.readIndex()
> ```

No need to send back a channel, then get an index, then wait for that index. This can be done directly by calling `waitForRead()`, which can do all the 3.

---

*[x/noncecounter.go, line 8 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KsrWs2A2TnqOmhiHkss:-KsvmmUUiWVUS96CWIZk:bdvwyu0) ([raw file](https://github.com/dgraph-io/dgraph/blob/11a3e5641ce92e78305c11e7ac74719ae063bb4b/x/noncecounter.go#L8)):*

I believe if a node dies and reconnects it could get delivered old Raft messages.  For example, a node sends a MsgReadIndex, dies, restarts, reconnects, and the leader sends a MsgReadIndexResp back.  Assuming that's correct, we need some way to distinguish old values from new ones.  Reading 16 bytes off of /dev/urandom at startup once per group is cheap enough.
</blockquote></details>

I think the MSB can be a timestamp, and the LSBs can be a monotonically increasing counter. That would do the job. This class isn't really required for that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383)*

srh:
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks broke.

---

*[worker/draft.go, line 1207 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KsvnDUpWmHKtGcnmLcl:-KsvrkUStdH9KlB-NFLt:b-5vjb9o) ([raw file](https://github.com/dgraph-io/dgraph/blob/806da55b856f081fc0c2edfeece7b8618985bdff/worker/draft.go#L1207)):*

No need to send back a channel, then get an index, then wait for that index. This can be done directly by calling `waitForRead()`, which can do all the 3.
</blockquote></details>

What is waitForRead?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383)*

srh:
Alright, we've got it threaded now.

---

Review status: 12 of 13 files reviewed at latest revision, 9 unresolved discussions.

---

*[x/noncecounter.go, line 8 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KsrWs2A2TnqOmhiHkss:-KtARIzsofsEUTGTBWBy:b-9fq92z) ([raw file](https://github.com/dgraph-io/dgraph/blob/11a3e5641ce92e78305c11e7ac74719ae063bb4b/x/noncecounter.go#L8)):*

I think the MSB can be a timestamp, and the LSBs can be a monotonically increasing counter. That would do the job. This class isn't really required for that.
</blockquote></details>

A random number is more robust unique value than a timestamp and retrieving either from the operating system is cheap.  

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-:-KtARCTEGz-0HCNdPxeT:b-p6bm8q)*

manishrjain:
Review status: 12 of 13 files reviewed at latest revision, 13 unresolved discussions, some commit checks broke.

---

*[worker/draft.go, line 791 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KsrXpxoRjurBMXfrtIB:-KtES9hYoTNmqBrfS35f:b-186slr) ([raw file](https://github.com/dgraph-io/dgraph/blob/11a3e5641ce92e78305c11e7ac74719ae063bb4b/worker/draft.go#L791)):*

If a context were used, it still would only be used the same way as this timer would.  It wouldn't be helpful to have it to pass to ReadIndex, for example.  ReadIndex's ctx parameter is only used while waiting for the node's run() thread to read it off `recvc`.  We have a timer because the raft library could discard the message if it doesn't have the leader configured, or because it could send off the message and never get a reply.
</blockquote></details>

How about context.cancel if you decided to discard the read? Also, if you pass a ctx with timeout to ReadIndex, it can also use it to cancel a read which has timed out.

If you want to insist on timer, can you explain what particular benefit it brings over context? Why do you intend to break consistency across code base for this timer thing (which isn't used anywhere else)?

---

*[worker/draft.go, line 253 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KtENHC70IYpIBbng0jR:-KtENHC70IYpIBbng0jS:b-8wtb5n) ([raw file](https://github.com/dgraph-io/dgraph/blob/c32bf13a4239fe6709c1c71358a73c7c80481751/worker/draft.go#L253)):*
> ```Go
> 			Logger:          &raft.DefaultLogger{Logger: x.Logger},
> 			CheckQuorum:     true, // Lets us use lease-based linearizable ReadIndex
> 			ReadOnlyOption:  raft.ReadOnlyLeaseBased,
> ```

So, I suppose we're using this for performance? If so, can you please add an explanation for this choice -- because this goes against raft lib recommendation. Also, how is CheckQuorum related to this?

---

*[worker/draft.go, line 713 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KtENdY7Qzz8ylvjVluP:-KtENdY7Qzz8ylvjVluQ:b-c6pt1o) ([raw file](https://github.com/dgraph-io/dgraph/blob/c32bf13a4239fe6709c1c71358a73c7c80481751/worker/draft.go#L713)):*
> ```Go
> }
> 
> /* How does linearizability work?  It works as follows:  We maintain one linearizable read request
> ```

No need for the question, and the beginning of answer. Directly write:

// We maintain one line...

---

*[worker/draft.go, line 717 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KtENtSIQVLfYbF8ieE4:-KtENtSIQVLfYbF8ieE5:b-wqyzcl) ([raw file](https://github.com/dgraph-io/dgraph/blob/c32bf13a4239fe6709c1c71358a73c7c80481751/worker/draft.go#L717)):*
> ```Go
> */
> 
> type linearizableReadRequest struct {
> ```

Too long of a name, can be shortened to `linReadReq` or just read request.

---

*[worker/draft.go, line 719 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KtEO3SKPjLuv1aFoqm3:-KtEO3SKPjLuv1aFoqm4:b2d7r3o) ([raw file](https://github.com/dgraph-io/dgraph/blob/c32bf13a4239fe6709c1c71358a73c7c80481751/worker/draft.go#L719)):*
> ```Go
> type linearizableReadRequest struct {
> 	// A one-shot chan which we send a raft index upon
> 	readIndexCh chan<- uint64
> ```

Read prefix is already in the struct. Call it `indexCh`. So, the api reads like `readReq.indexCh`.

---

*[worker/draft.go, line 728 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KtEQkMCG9tkNzEqZlZm:-KtEQkMCG9tkNzEqZlZn:b-y4odcv) ([raw file](https://github.com/dgraph-io/dgraph/blob/c32bf13a4239fe6709c1c71358a73c7c80481751/worker/draft.go#L728)):*
> ```Go
> }
> 
> func runReadIndexLoop(
> ```

Can make this function member of node struct.

---

*[worker/draft.go, line 729 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KtENp00_PtFga1Df20R:-KtENp00_PtFga1Df20S:bntol4z) ([raw file](https://github.com/dgraph-io/dgraph/blob/c32bf13a4239fe6709c1c71358a73c7c80481751/worker/draft.go#L729)):*
> ```Go
> 
> func runReadIndexLoop(
> 	n *node, stop <-chan struct{}, finished chan<- struct{}, requestCh <-chan linearizableReadRequest,
> ```

100 chars. Also, func + args can fit in 2 lines instead of 3.

---

*[worker/draft.go, line 789 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KtEPtP8haecE4aUtukQ:-KtEPtP8haecE4aUtukR:b-8wlw00) ([raw file](https://github.com/dgraph-io/dgraph/blob/c32bf13a4239fe6709c1c71358a73c7c80481751/worker/draft.go#L789)):*
> ```Go
> 	// See also our configuration of HeartbeatTick and ElectionTick.
> 	const raftTickRate = 20 * time.Millisecond
> 	ticker := time.NewTicker(raftTickRate)
> ```

As I mentioned before, please don't create const if you're going to use the variable exactly once.  If you disagree with this, consider this to be a rule, rather than a suggestion. This is not negotiable.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383)*

srh:
Review status: 12 of 13 files reviewed at latest revision, 13 unresolved discussions, some commit checks broke.

---

*[worker/draft.go, line 791 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KsrXpxoRjurBMXfrtIB:-KtEqNY0fXXvJXr3bbQN:b-3tos3t) ([raw file](https://github.com/dgraph-io/dgraph/blob/11a3e5641ce92e78305c11e7ac74719ae063bb4b/worker/draft.go#L791)):*

How about context.cancel if you decided to discard the read? Also, if you pass a ctx with timeout to ReadIndex, it can also use it to cancel a read which has timed out.

If you want to insist on timer, can you explain what particular benefit it brings over context? Why do you intend to break consistency across code base for this timer thing (which isn't used anywhere else)?
</blockquote></details>

We wouldn't be adding a timer to an existing context.  We want time expiration to be treated differently than some outer ctx being canceled.

---

*[worker/draft.go, line 253 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KtENHC70IYpIBbng0jR:-KtEn396HnhAQV9D71F-:b-bqw3j5) ([raw file](https://github.com/dgraph-io/dgraph/blob/c32bf13a4239fe6709c1c71358a73c7c80481751/worker/draft.go#L253)):*

So, I suppose we're using this for performance? If so, can you please add an explanation for this choice -- because this goes against raft lib recommendation. Also, how is CheckQuorum related to this?
</blockquote></details>

CheckQuorum makes it actually operate, otherwise, as far as I can tell, it silently returns some dummy value.

---

*[worker/draft.go, line 728 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KtEQkMCG9tkNzEqZlZm:-KtEoxd0WamEki0oBd6j:b-tv18fr) ([raw file](https://github.com/dgraph-io/dgraph/blob/c32bf13a4239fe6709c1c71358a73c7c80481751/worker/draft.go#L728)):*

Can make this function member of node struct.
</blockquote></details>

Why would we?

---

*[worker/draft.go, line 789 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KtEPtP8haecE4aUtukQ:-KtEndTfC6vT_tB-ux4x:bvemyr0) ([raw file](https://github.com/dgraph-io/dgraph/blob/c32bf13a4239fe6709c1c71358a73c7c80481751/worker/draft.go#L789)):*

As I mentioned before, please don't create const if you're going to use the variable exactly once.  If you disagree with this, consider this to be a rule, rather than a suggestion. This is not negotiable.
</blockquote></details>

Okay, I'll get rid of it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383)*

srh:
Review status: 12 of 13 files reviewed at latest revision, 13 unresolved discussions.

---

*[worker/draft.go, line 717 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KtENtSIQVLfYbF8ieE4:-KtEvnp5-wbC-zgIjUf3:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/c32bf13a4239fe6709c1c71358a73c7c80481751/worker/draft.go#L717)):*

Too long of a name, can be shortened to `linReadReq` or just read request.
</blockquote></details>

Done.

---

*[worker/draft.go, line 719 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KtEO3SKPjLuv1aFoqm3:-KtEvoILM1IQU-Hz6Kmi:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/c32bf13a4239fe6709c1c71358a73c7c80481751/worker/draft.go#L719)):*

Read prefix is already in the struct. Call it `indexCh`. So, the api reads like `readReq.indexCh`.
</blockquote></details>

Done.

---

*[worker/draft.go, line 729 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KtENp00_PtFga1Df20R:-KtEvp87p-U9_sYVblRU:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/c32bf13a4239fe6709c1c71358a73c7c80481751/worker/draft.go#L729)):*

100 chars. Also, func + args can fit in 2 lines instead of 3.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383)*

srh:
Review status: 12 of 13 files reviewed at latest revision, 13 unresolved discussions.

---

*[worker/draft.go, line 713 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KtENdY7Qzz8ylvjVluP:-KtFH0CciRYcdqhESG8c:bh99ypo) ([raw file](https://github.com/dgraph-io/dgraph/blob/c32bf13a4239fe6709c1c71358a73c7c80481751/worker/draft.go#L713)):*

No need for the question, and the beginning of answer. Directly write:

// We maintain one line...
</blockquote></details>

Moved comment below too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383)*

manishrjain:
:lgtm:

---

Reviewed 3 of 3 files at r5.

---

*[worker/draft.go, line 804 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KtFJczTnEs9Du0v_j9C:-KtFJczTnEs9Du0v_j9D:b-r06z2k) ([raw file](https://github.com/dgraph-io/dgraph/blob/847137d796ac092f710337b0e515c35756cb3677/worker/draft.go#L804)):*
> ```Go
> 		// That way we know sending to readStateCh will not deadlock.
> 		finished := make(chan struct{})
> 		stop := make(chan struct{})
> ```

This can be n.stop. Doesn't need to be a separate variable.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-:-KtJhe2_oQxRVIPt7CHT:bnfp4nl)*

srh:

---

*[worker/draft.go, line 804 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383#-KtFJczTnEs9Du0v_j9C:-KtJrR50ILrWNIqjHHmt:b-3egpr3) ([raw file](https://github.com/dgraph-io/dgraph/blob/847137d796ac092f710337b0e515c35756cb3677/worker/draft.go#L804)):*

This can be n.stop. Doesn't need to be a separate variable.
</blockquote></details>

No, it cannot.  runReadIndexLoop must consume from readStateCh as long as Run's loop continues.  See the comment above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1383)*

srh:
In master as of 6e55ab609a8e3d053c9df128afa7af4ddae90fc1

PR #1391, opened 2017-09-01
Geo/near fix

### LINKED ISSUE #1390: near query only returns points within the specified region and not the intersecting polygons.
```
{
  countries(func: near(loc, [79.96, 29.24], 10000)) {
    loc
    name
  }
}
```

Should return the multi-polygon India because the point lies in India but it doesn't. This is because near checks that the multi-polygon should be within the sphere with centre at the given point and radius. Instead, it should just check if the multi-polygon intersects this sphere and return it if it does.

pawanrawal:
Fixes #1390 

Near behaves similarly to intersects now. We draw a polygon centred at the point given with the radius and find all geometries it intersects with.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1391)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,11 @@
 matrix:
   include:
     - os: linux
-      go: 1.8
+      go: 1.9
       language: go
     - os: osx
       language: go
-      go: 1.8
+      go: 1.9
 
 install: contrib/install.sh
 git:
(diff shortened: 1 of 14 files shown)


manishrjain:
:lgtm:

---

Reviewed 10 of 12 files at r1, 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1391#-:-Kt9R6dyD5oNGuT8Za44:bnfp4nl)*

pawanrawal:
Merged

PR #1408, opened 2017-09-06
WIP: Wal/index (Don't review)

janardhan1993:
problem:
1. we do a schema mutation, we update the schema but if we crash before index is rebuilt then on restart we don't rebuild because on replay we will see that it's already indexed so we don't do anything.
2. Similarly for indexing on scalar values.

We store the current value of schema in wal on schema mutation, current pl(should be small for scalar values) for indexed predicates on mutation. On restart we can restore the schema and postings for scalar predicates to the value they should have been at snapshot using our wal.
So we can replay normally after restoring using our wal.

3. reverse edges - no issues if we add the reverse edge always(we don't rely on hasmutated).

4. count edges - WIP.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1408)

### DIFF
--- a/posting/index.go
+++ b/posting/index.go
@@ -282,6 +282,7 @@ func updateCount(ctx context.Context, params countParams) error {
 		Attr:    params.attr,
 		Op:      protos.DirectedEdge_DEL,
 	}
+	// Read from wal and remove this // TODO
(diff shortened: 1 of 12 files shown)


janardhan1993:
Not needed now, will raise another PR for schema mutations.

PR #1420, opened 2017-09-07
Logic for Dgraph zero

manishrjain:
This is a breaking change. Needs another PR before Dgraph works again.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1420)


janardhan1993:
I will have another look later.

---

Reviewed 20 of 20 files at r1.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks broke.

---

*[cmd/dgraphzero/main.go, line 50 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtQ8aeWnGRRNLY2Pp4a:-KtQ7x0h8mcHmmj8lvoZ:b-sryro4) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/cmd/dgraphzero/main.go#L50)):*
> ```Go
> 		"addr:port of this server, so other Dgraph servers can talk to this.")
> 	port        = flag.Int("port", 8888, "Port to run Dgraph zero on.")
> 	raftId      = flag.Uint64("id", 0, "Unique raft index.")
> ```

may be call it nodeid ?

---

*[cmd/dgraphzero/raft.go, line 145 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtQAPTr2gvrB-WeAkF4:-KtQAPTr2gvrB-WeAkF5:bjbcyqe) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/cmd/dgraphzero/raft.go#L145)):*
> ```Go
> 		state := n.server.state
> 		group := state.Groups[p.Member.GroupId]
> 		group.Members[p.Member.Id] = p.Member
> ```

We should probably throw error if member id is already present. Would be helpful if users start nodes with same id. we can use member's addr to find out whether this update was from the same or different server.

---

*[cmd/dgraphzero/zero.go, line 79 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtQBktZjimXsV6RUEcZ:-KtQBktZjimXsV6RUEc_:b-omz2s0) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/cmd/dgraphzero/zero.go#L79)):*
> ```Go
> 
> 	for _, group := range s.state.Groups {
> 		for key, tab := range group.Tablets {
> ```

will the number of tablets every go high, if yes better to use map.

---

*[cmd/dgraphzero/zero.go, line 88 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtQCwIs-odHddYyTzxs:-KtQCwIs-odHddYyTzxt:b5ecanr) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/cmd/dgraphzero/zero.go#L88)):*
> ```Go
> }
> 
> func (s *Server) createProposals(dst *protos.Group) ([]*protos.ZeroProposal, error) {
> ```

This is used only for updates from a member ?

---

*[cmd/dgraphzero/zero.go, line 157 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtQE-YbY27C2VD781JG:-KtQE-YbY27C2VD781JH:b-rkcbd2) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/cmd/dgraphzero/zero.go#L157)):*
> ```Go
> 	pl := conn.Get().Connect(m.Addr)
> 	defer conn.Get().Release(pl)
> ```

We can have more replicas than numReplicas if connect request goes to both leader and replica of dgraph zero at same time.  It's safer to do it on leader always.

---

*[worker/draft.go, line 418 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtQEwAHAn-HsKcMgp7b:-KtQEwAHAn-HsKcMgp7c:b-qag3xl) ([raw file](https://github.com/dgraph-io/dgraph/blob/de64ac267d2a2e1768a667bc5f1f5373152f4284/worker/draft.go#L418)):*
> ```Go
> 				} else if rd.RaftState == raft.StateLeader && !leader {
> 					leaseMgr().resetLease(n.gid)
> 					go groups().syncMemberships()
> ```

We could still tell graph zero immediately on stepping down as leader.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-:-KtQFME_77yrCEk1X8Ti:b-ko3olc)*

janardhan1993:
Review status: all files reviewed at latest revision, 13 unresolved discussions, some commit checks broke.

---

*[cmd/dgraphzero/raft.go, line 254 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtT_GlZrv_UG5S3XnOZ:-KtT_GlZrv_UG5S3XnO_:b-rkpsty) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/cmd/dgraphzero/raft.go#L254)):*
> ```Go
> 			}
> 
> 			// TODO: Should we move this to the top?
> ```

As per raft thesis a node can't be a leader unless it's applied all the pending entries. This should probably be at top.
I will try to confirm this today.

---

*[worker/groups.go, line 147 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtTRmV_g9eA6lvHHsop:-KtTRmV_g9eA6lvHHsoq:b-9bxyof) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/worker/groups.go#L147)):*
> ```Go
> 			for i := 0; i < 10; i++ { // 10 attempts.
> 				var err error
> 				state, err = zc.Connect(gr.ctx, m)
> ```

This node would be added in membership. It would be better if we maintain whether the node has initialised or not in membershipstate. That way no new request would come to this server until it's initialised. 

---

*[worker/groups.go, line 158 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtTRTieqASuFuYy59oq:-KtTRTifhQKI8tDdJOiP:b-22haw6) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/worker/groups.go#L158)):*
> ```Go
> 			gr.state = state
> 
> 			// TODO: Is this comment useful?
> ```

This comment is no longer needed, earlier we used to add the node to membership later so if multiple nodes start at same time they will fetch information from group zero and see that there is no other node and each node will become single node cluster.

---

*[worker/groups.go, line 192 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtTSN8wG2KagQL90Hn8:-KtTSN8wG2KagQL90Hn9:bwn1odq) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/worker/groups.go#L192)):*
> ```Go
> 		for _, m := range group.Members {
> 			if m.Id == Config.RaftId {
> 				g.gid = gid
> ```

updating the value with read lock

---

*[worker/groups.go, line 384 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtTSfgDXId6ug28sszC:-KtTSfgDXId6ug28sszD:bozlli0) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/worker/groups.go#L384)):*
> ```Go
> // - Once iteration is over without errors, it would return back all new updates.
> // - These updates are then applied to groups().all state via applyMembershipUpdate.
> // func (g *groupi) syncMemberships() {
> ```

we don't sync membership information whenever node becomes a leader or steps down as a leader.

---

*[worker/mutation.go, line 329 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtTQeR8KCKuo5fxG4YL:-KtTQeR8KCKuo5fxG4YM:b-pityfh) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/worker/mutation.go#L329)):*
> ```Go
> 
> 	conn := p.Get()
> 	c := protos.NewZeroClient(conn)
> ```

client should be thread safe, so we should try to reuse client instead of creating everytime.

---

*[worker/worker.go, line 57 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtTQ7wcbb1X8I4sl9qR:-KtTQ7wcbb1X8I4sl9qS:b-tdj1ur) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/worker/worker.go#L57)):*
> ```Go
> 	pstore = ps
> 	// needs to be initialized after group config
> 	leasePred = "_lease_"
> ```

This is not needed now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420)*

manishrjain:
Review status: all files reviewed at latest revision, 13 unresolved discussions, some commit checks broke.

---

*[cmd/dgraphzero/main.go, line 50 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtQ8aeWnGRRNLY2Pp4a:-KtVF-5vegVB5MMExvs8:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/cmd/dgraphzero/main.go#L50)):*

may be call it nodeid ?
</blockquote></details>

Done.

---

*[cmd/dgraphzero/raft.go, line 145 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtQAPTr2gvrB-WeAkF4:-KtVF-szDN9xeyONrCf8:b4b3fp1) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/cmd/dgraphzero/raft.go#L145)):*

We should probably throw error if member id is already present. Would be helpful if users start nodes with same id. we can use member's addr to find out whether this update was from the same or different server.
</blockquote></details>

A server can restart, or be moved to another instance with its directories intact -- in that case, we shouldn't throw an error. Address could change as well -- that should be alright, I think.

---

*[cmd/dgraphzero/zero.go, line 79 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtQBktZjimXsV6RUEcZ:-KtVCCMRe4jySW9WD55N:b-hkix3t) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/cmd/dgraphzero/zero.go#L79)):*

will the number of tablets every go high, if yes better to use map.
</blockquote></details>

Could be high, but we won't need to run this very often. Every dgraph server would keep track of the tablets that it, or others are already serving. So, they would only need to ask for this, if they see it for the first time.

---

*[cmd/dgraphzero/zero.go, line 88 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtQCwIs-odHddYyTzxs:-KtVCOl_AJE1Q1mzy29W:b-b66kjk) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/cmd/dgraphzero/zero.go#L88)):*

This is used only for updates from a member ?
</blockquote></details>

Proposal is used for updates from members, but also 

---

*[cmd/dgraphzero/zero.go, line 157 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtQE-YbY27C2VD781JG:-KtVGYOS4KlnHwsYp_Wg:b-bysvss) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/cmd/dgraphzero/zero.go#L157)):*

We can have more replicas than numReplicas if connect request goes to both leader and replica of dgraph zero at same time.  It's safer to do it on leader always.
</blockquote></details>

What I've done is to check this while applying the proposal. If the number of servers in that group exceed the numReplicas, then we don't apply that proposal, and it goes back to the caller as an error. The caller can then retry.

---

*[worker/draft.go, line 418 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtQEwAHAn-HsKcMgp7b:-KtVH5lELfn6yvTIxeB2:b-dv8kk) ([raw file](https://github.com/dgraph-io/dgraph/blob/de64ac267d2a2e1768a667bc5f1f5373152f4284/worker/draft.go#L418)):*

We could still tell graph zero immediately on stepping down as leader.
</blockquote></details>

With group zero out, we don't really need to know who the leader is. But, I've added a TODO to consider if we need to.

---

*[worker/groups.go, line 147 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtTRmV_g9eA6lvHHsop:-KtVHHNIWnSKOZVYpBzD:btkb8v6) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/worker/groups.go#L147)):*

This node would be added in membership. It would be better if we maintain whether the node has initialised or not in membershipstate. That way no new request would come to this server until it's initialised. 
</blockquote></details>

What's the downside here?

---

*[worker/groups.go, line 158 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtTRTieqASuFuYy59oq:-KtVHWq9M8-Bg4LAPi9k:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/worker/groups.go#L158)):*

This comment is no longer needed, earlier we used to add the node to membership later so if multiple nodes start at same time they will fetch information from group zero and see that there is no other node and each node will become single node cluster.
</blockquote></details>

Done.

---

*[worker/groups.go, line 192 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtTSN8wG2KagQL90Hn8:-KtVHimUwxba5av9bbOv:bczbeg6) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/worker/groups.go#L192)):*

updating the value with read lock
</blockquote></details>

Switched to atomic.

---

*[worker/groups.go, line 384 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtTSfgDXId6ug28sszC:-KtVIi7nxxWlWIka47Jt:b-7yd1q4) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/worker/groups.go#L384)):*

we don't sync membership information whenever node becomes a leader or steps down as a leader.
</blockquote></details>

Added a TODO.

---

*[worker/mutation.go, line 329 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtTQeR8KCKuo5fxG4YL:-KtVIkr1XR8_w4BwIIyD:btgswjg) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/worker/mutation.go#L329)):*

client should be thread safe, so we should try to reuse client instead of creating everytime.
</blockquote></details>

We keep the connection around, creating a new client is a cheap op.

---

*[worker/worker.go, line 57 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-KtTQ7wcbb1X8I4sl9qR:-KtVItmc_KUb0eOQBM6P:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/438a30ee08c9e2972b3b698b517a6d794ad8c8ee/worker/worker.go#L57)):*

This is not needed now.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420)*

janardhan1993:
:lgtm:

---

Review status: 15 of 20 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1420#-:-KtVSB8MEwew4daGLv3S:bnfp4nl)*

PR #1491, opened 2017-09-19
Bulkloader: system test fixture

peterstace:
Test fixture for running system level tests for dgraph.

I want to make sure everyone is ok with this testing approach before moving forward any further with it.

How it works:

- It's a bash script that loads a set of RDFs/Schema using bulkloader.
- It then starts up dgraph so we can run queries on it.
- Then we compare the results of the queries with what we expect.
- We can easily add extra `suiteXX` directories that contain new RDF/Schema/Query/Expected-Result sets.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1491)


manishrjain:
:lgtm:

---

Reviewed 9 of 9 files at r1.

---

*[cmd/bulkloader/systest/suite01/rdfs.rdf, line 1 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1491#-KuOEgPCqLPJgFtumbyO:-KuOEgPCqLPJgFtumbyP:bssjcrz) ([raw file](https://github.com/dgraph-io/dgraph/blob/dab621756f3a47081f5485d0b63e2ab6667a9999/cmd/bulkloader/systest/suite01/rdfs.rdf#L1)):*
> ```XML
> <peter> <name> "Peter" .
> ```

Might want to add more names here.

<peter> -> `_:peter`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1491#-:-KuOEwUtWKz_F4jA1oF9:bnfp4nl)*

peterstace:
As discussed with @manishrjain and @pawanrawal , we want to do something a bit different and more robust in the medium/long term. Especially for making sure we only have end to end tests in a single place. For the short term though, these will still be useful.

PR #1545, opened 2017-09-27
Fix race with using upsert

pawanrawal:
The issue here was that after the Subgraph was executed, we were trying to access it using

```
sg := req.Subgraphs[i]
```
and that is wrong because the response doesn't belong to the subgraph at `i` index. Now the subgraph is part of the response along with the error. This does make the code a bit verbose which I don't like but I couldn't think of an easier way.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1545)

### DIFF
--- a/cmd/dgraph/main_test.go
+++ b/cmd/dgraph/main_test.go
@@ -1683,7 +1683,7 @@ func TestUpsert2(t *testing.T) {
 
 	res, err := runQuery(query)
 	require.NoError(t, err)
-	require.JSONEq(t, `{"data": {"uids":{"b":"0x271a","a":"0x2719"}}}`, res)
+	require.JSONEq(t, `{"data": {"uids":{"a":"0x271a","b":"0x2719"}}}`, res)
 
 	m := make(map[string]interface{})
 	require.NoError(t, json.Unmarshal([]byte(res), &m))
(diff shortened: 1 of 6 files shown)


pawanrawal:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[query/query.go, line 2518 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1545#-Kv022P5y9ywmehr6HH6:-Kv022P6j_Qgz2IcX75U:bc6bm71) ([raw file](https://github.com/dgraph-io/dgraph/blob/37ae95aa401959d89dad9cf20970b2d7a6138893/query/query.go#L2518)):*
> ```Go
> 			// We didn't get back any uids. So we would have to assign the uid and perform the
> 			// mutation (i.e. the upsert operation).
> 			sg := req.Subgraphs[i]
> ```

This was the issue. We would be accessing some other subgraph and checking its DestUids while the corresponding ProcessGraph might still be going on.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1545)*

pawanrawal:
Found a easier way closing this.

PR #1558, opened 2017-10-01
Prevent nil pointer exception crash

noh4ck:
Prevent a parser crash when using `val` on a predicate while also trying to expand it:
```graphql
{
  brCharacters(func: eq(name@en, "Blade Runner")) {
    name@en
    initial_release_date
    starring {
      # This makes dgraph crash!!!
      val(performance.actor) {
        name@en  # actor name
      }
      performance.character {
        name@en  # character name
      }
    }
  }
}
```

PS: Sorry for crashing the doc server...

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1558)

### DIFF
--- a/gql/parser.go
+++ b/gql/parser.go
@@ -2639,6 +2639,9 @@ func godeep(it *lex.ItemIterator, gq *GraphQuery) error {
 				count = seenWithPred
 			}
 		case itemLeftCurl:
+			if curp == nil {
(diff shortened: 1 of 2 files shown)


pawanrawal:
Thanks for the change @noh4ck. Could you also add a corresponding test in `gql/parser_test.go` and make sure that the existing tests pass?

PR #1569, opened 2017-10-03
Remove password storage type

### LINKED ISSUE #1561: Get rid of pwd:password rdf type
This storage type doesn't seem to be of much use. We can just check the schema type to decide if values for a predicate can be retrieved.

pawanrawal:
The storage type is not an RDF standard and was added when we didn't have schema types. I get rid of it and check the schema of the predicate to decide whether it can be fetched or not.

Resolves #1561

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1569)


manishrjain:
:lgtm: for the remove password storage type. But, I think we could simplify the change from y'day by just not returning password edge silently without error. Then we don't need to do any housekeeping re: expansion or not.

---

Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1569#-:-KvVS3mXi7wvs97HtywK:b-uvcohi)*

pawanrawal:
Yeah, that's a good idea just skip password edge. I will update this PR to do the same.

---

Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion.

---

*[worker/task.go, line 316 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1569#-KvVRnTv9tlC8O7oT24L:-KvVUyKYbZexYc4-4Gzo:b-ez5vu3) ([raw file](https://github.com/dgraph-io/dgraph/blob/1ddff28ce36f6092439105e3a849234382241a6b/worker/task.go#L316)):*

You know. It looks like we're going to length to return an error when user asks for password edge. We could just silently ignore returning it in all cases, no, expanded or asked for? Wouldn't it simplify the code from y'day?
</blockquote></details>

Agree it would simplify all the handling. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1569#-:-KvVTs4QCJEL93aMc0TZ:bb0gwks)*

manishrjain:
:lgtm:

---

Review status: 0 of 16 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1569#-:-KvVney1a219MpVFFKwo:bnfp4nl)*

pawanrawal:
In master as of 3435dce67c1f6b21965f0c12f4952c335b0e1da6

PR #1570, opened 2017-10-03
Bulkloader memory improvements

peterstace:
Improvements:
 * Uses an LRU cached backed by badger for the XID->UID mapping. This was the main memory hog.
 * Reduce buffer size on some channels.
 * Store map entries in a flat bytes slice while accumulating them. This has a double effect of reducing memory usage, and reducing CPU usage due to less GC overhead (even though we're doing more marshal/unmarshal work).

This is the change that allows us to load the 2 billion graph overflow dataset on our new server in under 1.5 hours.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1570)


manishrjain:
Reviewed 9 of 9 files at r1.

---

*[cmd/dgraph-bulk-loader/mapper.go, line 64 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1570#-KvVisysysR47MZZy5SB:-KvVisysysR47MZZy5SC:b-iy0psi) ([raw file](https://github.com/dgraph-io/dgraph/blob/6660b80be9dd2adc1896b389675297a45007d8b0/cmd/dgraph-bulk-loader/mapper.go#L64)):*
> ```Go
> 
> func (m *mapper) writeMapEntriesToFile(entriesBuf []byte, shardIdx int) {
> 	var buf []byte = entriesBuf
> ```

buf := entriesBuf

---

*[cmd/dgraph-bulk-loader/mapper.go, line 70 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1570#-KvVk0yPJYsfu2_8wU57:-KvVk0yPJYsfu2_8wU58:bsgyp4y) ([raw file](https://github.com/dgraph-io/dgraph/blob/6660b80be9dd2adc1896b389675297a45007d8b0/cmd/dgraph-bulk-loader/mapper.go#L70)):*
> ```Go
> 		x.AssertTrue(n > 0)
> 		buf = buf[n:]
> 		me := new(protos.MapEntry)
> ```

This doesn't seem like it would decrease memory usage. Previously we were doing this slowly over time, now we're doing it in lumpsum. But, the memory usage is the same.

---

*[cmd/dgraph-bulk-loader/uid_map.go, line 74 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1570#-KvVlUE2EUmLF7Z9Twpe:-KvVlUE2EUmLF7Z9Twpf:brmst5d) ([raw file](https://github.com/dgraph-io/dgraph/blob/6660b80be9dd2adc1896b389675297a45007d8b0/cmd/dgraph-bulk-loader/uid_map.go#L74)):*
> ```Go
> 
> 	var item badger.KVItem
> 	x.Check(m.kv.Get([]byte(xid), &item))
> ```

Couldn't we share this whole logic between dgraph loader and bulk loader?

---

*[cmd/dgraph-bulk-loader/uid_map.go, line 140 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1570#-KvVlgnTYHQqgYJjFTms:-KvVlgnTYHQqgYJjFTmt:bxudivj) ([raw file](https://github.com/dgraph-io/dgraph/blob/6660b80be9dd2adc1896b389675297a45007d8b0/cmd/dgraph-bulk-loader/uid_map.go#L140)):*
> ```Go
> 			batch = append(batch, &badger.Entry{
> 				Key:   []byte(m.xid),
> 				Value: uidBuf[:binary.PutUvarint(uidBuf[:], m.uid)],
> ```

Avoid complex expressions like these.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1570)*

peterstace:

---

*[cmd/dgraph-bulk-loader/mapper.go, line 64 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1570#-KvVisysysR47MZZy5SB:-KvWBkNDfPCu6IAO8K6B:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/6660b80be9dd2adc1896b389675297a45007d8b0/cmd/dgraph-bulk-loader/mapper.go#L64)):*

buf := entriesBuf
</blockquote></details>

Done.

---

*[cmd/dgraph-bulk-loader/mapper.go, line 70 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1570#-KvVk0yPJYsfu2_8wU57:-KvWBn0N71d3LRlts1tX:bg0vfdp) ([raw file](https://github.com/dgraph-io/dgraph/blob/6660b80be9dd2adc1896b389675297a45007d8b0/cmd/dgraph-bulk-loader/mapper.go#L70)):*

This doesn't seem like it would decrease memory usage. Previously we were doing this slowly over time, now we're doing it in lumpsum. But, the memory usage is the same.
</blockquote></details>

The memory usage would be (exactly) the same if all mappers were to sort at the same time. However, sorting is a small portion of the runtime compared to filling up the map entries buffer. So it's unlikely that all mappers would sort at the same time. Once one mapper has finished sorting, the next mapper to sort will use the same memory (so long as the GC frees it in time).

In practice, I found this change to have a significant impact on memory usage - the difference between the 2 billion RDF set running OOM after 5m compared to being able to run all the way through (64GB RAM).

We could potentially enforce that only 1 (or a configurable number) of mappers may sort at a time, to avoid the case where they all try to obtain large amounts of memory.

---

*[cmd/dgraph-bulk-loader/uid_map.go, line 74 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1570#-KvVlUE2EUmLF7Z9Twpe:-KvWCwOwGHAdRiqDMOMU:b2ze3bm) ([raw file](https://github.com/dgraph-io/dgraph/blob/6660b80be9dd2adc1896b389675297a45007d8b0/cmd/dgraph-bulk-loader/uid_map.go#L74)):*

Couldn't we share this whole logic between dgraph loader and bulk loader?
</blockquote></details>

Yes, I think that's a good idea. The performance requirements for bulk loader are a bit more restrictive (e.g. in bulk loader the UID map must be sharded), but we can definitely use the logic here for both. I'll make sure that's done before this gets merged.

---

*[cmd/dgraph-bulk-loader/uid_map.go, line 140 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1570#-KvVlgnTYHQqgYJjFTms:-KvWExuVgSkP-IYIU_CE:bzhnnji) ([raw file](https://github.com/dgraph-io/dgraph/blob/6660b80be9dd2adc1896b389675297a45007d8b0/cmd/dgraph-bulk-loader/uid_map.go#L140)):*

Avoid complex expressions like these.
</blockquote></details>

Good call, this one got a bit over the top. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1570)*

manishrjain:
:lgtm:

---

Reviewed 2 of 2 files at r2.

---

*[cmd/dgraph-bulk-loader/uid_map.go, line 74 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1570#-KvVlUE2EUmLF7Z9Twpe:-KvWGprPeHOq3nkyy-6K:bgad2uj) ([raw file](https://github.com/dgraph-io/dgraph/blob/6660b80be9dd2adc1896b389675297a45007d8b0/cmd/dgraph-bulk-loader/uid_map.go#L74)):*

Yes, I think that's a good idea. The performance requirements for bulk loader are a bit more restrictive (e.g. in bulk loader the UID map must be sharded), but we can definitely use the logic here for both. I'll make sure that's done before this gets merged.
</blockquote></details>

The map could be sharded in dgraphloader as well. It's a performance enhancement loader would also benefit from.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1570#-:-KvWH0O_gBkON-N-T4Kk:bnfp4nl)*

manishrjain:
Btw, moving the map into shared space between loader and bulk loader can be done in a separate PR.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1570#-:-KvWH2u9zBv8npmAX0S-:b-vt6u7z)*

PR #1599, opened 2017-10-06
Store NaN as 0

### LINKED ISSUE #1594: Cant parse JSON response when facets have NaN value
Do the following mutation

```
mutation {
  set {
    <1> <name> "Alice" (a=NaN,b=12) .
  }
}
```

And then the query

```
{
  me(func: uid(1)) {
    name @facets
  }
}
```

The server returns something that can't be parsed as a JSON.

pawanrawal:
This fixes #1594 

I thought about returning an error but I think just storing as 0.0 value should be ok as the user doesn't have to find all NaN and fix them themselves.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1599)

### DIFF
--- a/query/query_facets_test.go
+++ b/query/query_facets_test.go
@@ -74,7 +74,7 @@ func populateGraphWithFacets(t *testing.T) {
 	// missing name for 101 -- no name edge and no facets.
 
 	addEdgeToLangValue(t, "name", 320, "Test facet", "en",
-		map[string]string{"type": `"Test facet with lang"`, "nankey": "NaN"})
+		map[string]string{"type": `"Test facet with lang"`})
 
 	time.Sleep(5 * time.Millisecond)
 }
@@ -1202,7 +1202,7 @@ func TestFacetWithLang(t *testing.T) {
 	`
 
 	js := processToFastJSON(t, query)
-	require.JSONEq(t, `{"data": {"me":[{"@facets":{"name@en":{"type":"Test facet with lang", "nankey": 0}},"name@en":"Test facet"}]}}`, js)
+	require.JSONEq(t, `{"data": {"me":[{"@facets":{"name@en":{"type":"Test facet with lang"}},"name@en":"Test facet"}]}}`, js)
 }
 
 func TestFilterUidFacetMismatch(t *testing.T) {
--- a/query/query_facets_test.go
+++ b/query/query_facets_test.go
@@ -73,7 +73,8 @@ func populateGraphWithFacets(t *testing.T) {
 	addEdgeToValue(t, "name", 33, "Michale", nil)
 	// missing name for 101 -- no name edge and no facets.
 
-	addEdgeToLangValue(t, "name", 320, "Test facet", "en", map[string]string{"type": `"Test facet with lang"`})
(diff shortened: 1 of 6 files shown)


janardhan1993:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1599#-:-KvjVnqZOYWnQbiRnmkZ:bnfp4nl)*

peterstace:
I've run into this exact problem a few times before, although the solution I've most often seen is to encode `NaN` as `null` rather than `0.0`. It's a bit better for the user that way, because then they can tell the difference between an actual 0 and a NaN (which depending on their use case, may be important -- maybe they weren't expecting NaN and will want to investigate why it's in their data, but if it's zero they won't notice it as easily).

Does this occur for regular terminal edges values as well? (rather than just facets).

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1599#-:-KvjZFHLpU5OOBl1Nhav:b-f8zhqy)*

pawanrawal:
Yeah, it did occur for terminal edges too. Good call to check. I return an error if it occurs for terminal edges.

I am not sure encoding the facet as `null` string is a very good idea because most probably the user intended it to be an `int`/`float`. Also now that it's stored as a string it would be ignored from queries that aggregate (sum etc.) facets as we can't convert it to float. So we should either throw an error or just use 0 value.

---

Review status: 2 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1599#-:-Kvjgb_EJpQjGhr-6Jj_:b-uax79y)*

pawanrawal:
Actually lets just be consistent and return error. Making the change

---

Review status: 2 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1599#-:-KvjjMWCHKmL_vyQvoHH:bxlmiw4)*

peterstace:
Agreed, the more I think about this, I think it's best to just give an error. Could be surprising to users if we just changed their data for them (they might not notice). If there's a limitation to the system (i.e. cannot insert and retrieve NaN), better to tell the user explicitly. :lgtm: 

---

Review status: 2 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1599#-:-Kvjjq8X8RGwafW7qBbx:brc8onn)*

pawanrawal:
Sounds good. No to changing their data.

---

Review status: 2 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1599#-:-KvjkOgDC9eArArSjNIA:b-64p27d)*

pawanrawal:
In master as of ce068be02986b103c261c2d077db45a4a719bf58

PR #1622, opened 2017-10-11
Share XID map between bulk loader and dgraph client

peterstace:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1622)


manishrjain:
I'll let @janardhan1993 and @pawanrawal review this.

janardhan1993:
:lgtm:

---

Reviewed 7 of 7 files at r1.

---

*[cmd/dgraph-bulk-loader/loader.go, line 237 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1622#-Kw8fNwaDCJZ20QkG03C:-Kw8fNwaDCJZ20QkG03D:b8sjtdt) ([raw file](https://github.com/dgraph-io/dgraph/blob/cfbe5ebe338ca53988b1e44837f6df7af295ffdd/cmd/dgraph-bulk-loader/loader.go#L237)):*
> ```Go
> 	// Obtain a fresh uid range - since uids are allocated in increasing order,
> 	// the start of the new range can be used as the lease.
> 	lease, _, _ := ld.up.ReserveUidRange()
> ```

log error instead of ignoring

---

*[cmd/dgraph-bulk-loader/mapper.go, line 201 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1622#-Kw8fUDr0C5FZMFiSdAB:-Kw8d7dZnvfbEKWp6cVq:b14tk7i) ([raw file](https://github.com/dgraph-io/dgraph/blob/cfbe5ebe338ca53988b1e44837f6df7af295ffdd/cmd/dgraph-bulk-loader/mapper.go#L201)):*
> ```Go
> 
> func (m *mapper) lookupUid(xid string) uint64 {
> 	uid, isNew, _ := m.um.AssignUid(xid)
> ```

Don't ignore the error.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1622#-:-Kw8gjSSYM-3834Qt-qj:bnfp4nl)*

peterstace:
Review status: 5 of 7 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[cmd/dgraph-bulk-loader/loader.go, line 237 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1622#-Kw8fNwaDCJZ20QkG03C:-Kw8kWJeXsC8eZ7CNdtA:b-t0hkvs) ([raw file](https://github.com/dgraph-io/dgraph/blob/cfbe5ebe338ca53988b1e44837f6df7af295ffdd/cmd/dgraph-bulk-loader/loader.go#L237)):*

log error instead of ignoring
</blockquote></details>

In this particular case (and the other one in the bulk loader), there can't be an error here because the uid producer cannot give an error -- it's just atomically incrementing an integer. I had to add an error to the interface so that it would work ok for the live loader, where there legitimately could be an error.

I was ignoring these errors in a similar way to how you generally ignore errors from `bytes.Buffer` `buf.Write(...)` etc.

I've changed this to instead do an assertion that the error is nil,  and added a comment to explain why.

---

*[cmd/dgraph-bulk-loader/mapper.go, line 201 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1622#-Kw8fUDr0C5FZMFiSdAB:-Kw8oGpmaY4cCHdhTtR9:b-1xh38x) ([raw file](https://github.com/dgraph-io/dgraph/blob/cfbe5ebe338ca53988b1e44837f6df7af295ffdd/cmd/dgraph-bulk-loader/mapper.go#L201)):*

Don't ignore the error.
</blockquote></details>

Same as other error.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1622)*

pawanrawal:
:lgtm:
Didn't look at xidmap.go very carefully as I suppose it was ported over?

---

Review status: 5 of 7 files reviewed at latest revision, 6 unresolved discussions.

---

*[client/checkpoint.go, line 82 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1622#-Kw9-1HDPboK2Mp_2as7:-Kw9-1HDPboK2Mp_2as8:b-ijiwoi) ([raw file](https://github.com/dgraph-io/dgraph/blob/cfbe5ebe338ca53988b1e44837f6df7af295ffdd/client/checkpoint.go#L82)):*
> ```Go
> 		return 0, err
> 	}
> 	var uid uint64
> ```

Its a line no of the file and not a uid.

---

*[client/mutations.go, line 81 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1622#-Kw9-pMp4TYtDDvWhQnU:-Kw9-pMp4TYtDDvWhQnV:b-86uil3) ([raw file](https://github.com/dgraph-io/dgraph/blob/cfbe5ebe338ca53988b1e44837f6df7af295ffdd/client/mutations.go#L81)):*
> ```Go
> 	for {
> 		assignedIds, err := p.dc.AssignUids(context.Background(), &protos.Num{Val: 1000})
> 		if err != nil {
> ```

What do you think about handling err == nil first and returning?

Then this block doesn't need to be indented. Just an idea. Feel free to ignore.

---

*[xidmap/xidmap.go, line 32 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1622#-Kw92jd1IEIJod7TGpLn:-Kw92jd1IEIJod7TGpLo:b-vyeink) ([raw file](https://github.com/dgraph-io/dgraph/blob/cfbe5ebe338ca53988b1e44837f6df7af295ffdd/xidmap/xidmap.go#L32)):*
> ```Go
> 	up     UidProvider
> 	opt    Options
> 	noMap  block
> ```

A comment about this would be useful.

---

*[xidmap/xidmap.go, line 135 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1622#-Kw92PS1ZSPhiIhzQx5N:-Kw92PS1ZSPhiIhzQx5O:b-x7ib2n) ([raw file](https://github.com/dgraph-io/dgraph/blob/cfbe5ebe338ca53988b1e44837f6df7af295ffdd/xidmap/xidmap.go#L135)):*
> ```Go
> 
> // ReserveUid gives a single uid without creating an xid to uid mapping.
> func (m *XidMap) ReserveUid() (uint64, error) {
> ```

AllocateUid?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1622#-:-Kw92yovkoRuJeyeS1fJ:b-d9bmjs)*

PR #1651, opened 2017-10-18
Maintain uid/value/facet matrices in the same dimension

peterstace:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1651)


peterstace:
Single line change, so submitting without review.

PR #1653, opened 2017-10-23
Use linear search on non-sorted UID lists

### LINKED ISSUE #1642: Sorting on facet in recursive query doesn't return predicates?
The problem I have is that sorting my facet (weight) on an edge results in the string values of other predicates not being returned (but the nodes are returned.)
```
# THIS WORKS
{
  recurse( func: eq(xid, "Algeria"), depth: 5 ) {
    _uid_
    xid
    related @facets(weight)
  }
}

```
And returns [good.txt](https://github.com/dgraph-io/dgraph/files/1388027/good.txt)

```
# THIS GENERATES THE PROBLEM (FILTERING ADDED)
{
  recurse( func: eq(xid, "Algeria"), depth: 5 ) {
    _uid_
    xid
    related @facets(orderdesc: weight)
  }
}
```
And returns [bad.txt](https://github.com/dgraph-io/dgraph/files/1388026/bad.txt)

Additionally, it seems like the very same query is returning a different number of nodes (total at the bottom of each file) - and the query is the same, just one has sorting applied.

I run 0.8.3 on macOS.

peterstace:
Fixes #1642

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1653)

### DIFF
--- a/algo/uidlist.go
+++ b/algo/uidlist.go
@@ -390,12 +390,23 @@ func MergeSorted(lists []*protos.List) *protos.List {
 	return &protos.List{Uids: output}
 }
 
-// IndexOf performs a binary search on the uids slice and returns the index at
-// which it finds the uid, else returns -1
-func IndexOf(u *protos.List, uid uint64) int {
+// BinarySearchIndexOf performs a binary search on the uids slice and returns
+// the index at which it finds the uid, else returns -1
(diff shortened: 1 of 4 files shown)


peterstace:
Closing this, we should be able to make a solution that doesn't rely on linear search.

PR #1759, opened 2017-11-10
Fix issue where lang only predicates not expanded in expand(_all_)

peterstace:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1759)


peterstace:
Single line change, but feels pretty hacky so want to get reviewed anyway.

pawanrawal:
We have had a discussion about this before and I think the decision was to either return predicate in all languages or in none. This would randomly select a language which I am not sure is a good idea.

What do you think @mrjn?

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-:-KyZfWO3Bzw2dk_sNaz2:b-iy1w9o)*

peterstace:
That's right, it will look for an untagged version, and if one isn't found, return an arbitrary tagged version.

If we were to return the predicate in all languages, one drawback would be that we wouldn't be able to display the tag, since we only store the hash (as the UID) rather than the tag itself. We could do some hackery to get around this (e.g. instead of storing the hash of the UID, store the tag as a uint64 - we could store up to 8 chars).

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-:-KyZgtCmH7Nf2qnAq-v-:b-1fgvhi)*

pawanrawal:
That's a good point. I feel lang tags would anyway be small and could be stored directly. That is the only meta we store. Or we return just the list in an array without the language information.

manishrjain:
In expand(_all_), we should aim to return all the languages. The number of langs are fixed, and so is their fingerprint. So, we can handcode a fingerprint -> lang map, which can be used to decipher the language. No need to store it separately.

peterstace:
The number of languages that can be used for FTS is fixed, but in general, users could have any language tag, e.g. `@kl` for Klingon.

As you point out though, the *mapping* is fixed, so it should still be possible to recover the language tag from the fingerprint as long as we track it on the way in.

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-:-KynVcVp68YqbbfqooYp:b-5utb1j)*

manishrjain:
If Klingon becomes a problem, we can always fix it then. A hardcoded list of fingerprint -> lang is sufficient for most people though (we could even add Klingon to that list). We don't need to do anything better than this simple map, I think.

peterstace:
Agreed, I've been overthinking this. There are only a small number of possible language tags, so we can just add all of them in a `func init() { ... }`.

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-:-Kynmo9D85173AFK5_Hv:bj34x4e)*

peterstace:
I'm going to close and reopen when this is ready.

peterstace:
@pawanrawal and @manishrjain , this is ready to take a look at again.

manishrjain:
Reviewed 4 of 6 files at r3, 3 of 3 files at r4.

---

*[posting/list.go, line 880 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMEU5prsfVbOR-wAov:-KzMEU5prsfVbOR-wAow:b-vwgzaf) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/posting/list.go#L880)):*
> ```Go
> 	tagMap = make(map[uint64]string)
> 	const a = byte('a')
> 	const z = byte('z')
> ```

Emm.. this might not work. Sometimes langs have subtags. https://www.w3.org/International/articles/language-tags/

So, we should hard-code a list of languages; instead of generating all permutations. You can pick them up online.

---

*[posting/list.go, line 884 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMErbexm7bDoHyuGrT:-KzMErbexm7bDoHyuGrU:b6vqnsi) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/posting/list.go#L884)):*
> ```Go
> 	for tag[0] = a; tag[0] <= z; tag[0]++ {
> 		for tag[1] = a; tag[1] <= z; tag[1]++ {
> 			tagMap[farm.Fingerprint64(tag)] = string(tag)
> ```

Also, confirm that there's no collision.

---

*[protos/task.proto, line 61 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMEx9KizIg6cyQkcNH:-KzMEx9KizIg6cyQkcNI:b-qwokiw) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/protos/task.proto#L61)):*
> ```protobuf
> 	FilterTree facets_filter = 9; // filtering on facets : has Op (and/or/not) tree
> 
> 	bool ExpandAll = 10;
> ```

bool expand_all = 10;

// Don't we already have the logic to expand all?

---

*[systest/bulk_live_fixture_test.go, line 21 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMFGlceBGC2rTSw1dh:-KzMFGlceBGC2rTSw1di:bqlvepy) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/systest/bulk_live_fixture_test.go#L21)):*
> ```Go
> 
> func init() {
> 	cmd := exec.Command("go", "install", "github.com/dgraph-io/dgraph/dgraph")
> ```

Is this change accidental?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759)*

pawanrawal:

---

*[posting/list.go, line 897 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMIKTPHF8hqCWCVpQu:-KzMIKTPHF8hqCWCVpQv:b-vu3b0a) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/posting/list.go#L897)):*
> ```Go
> 	err := l.iterate(readTs, 0, func(p *protos.Posting) bool {
> 		var tag string
> 		if uid := p.GetUid(); uid != math.MaxUint64 {
> ```

This would also get postings corresponding to different values for list type. We set a meta for language. How about checking that?

---

*[protos/task.proto, line 61 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMEx9KizIg6cyQkcNH:-KzMM0HLneURfrENIyTj:b-n321ab) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/protos/task.proto#L61)):*

bool expand_all = 10;

// Don't we already have the logic to expand all?
</blockquote></details>

Add a comment about what this means.

---

*[query/query.go, line 498 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMKAscH54seTp7dXNg:-KzMKAscH54seTp7dXNh:bv4gvnw) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/query/query.go#L498)):*
> ```Go
> 				}
> 				if len(pc.LangTags) != 0 {
> 					x.AssertTrue(i < len(pc.LangTags)) // All langs are either present or absent.
> ```

Aah, return error instead of asserting.  

---

*[query/query.go, line 1750 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMJzuYKphUTjs5AhNO:-KzMJzuYKphUTjs5AhNP:b-qpfe7j) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/query/query.go#L1750)):*
> ```Go
> 				Attr:    pred,
> 			}
> 			temp.Params.Expand = child.Params.Expand
> ```

So currently `Expand` can be non-empty when the Subgraph is either of the two below.
```
expand(_all_)
expand(val(x))
```

This is a different case where we set Expand on an expanded child. This should probably be a different field called `Expanded` which can be a bool.

---

*[systest/mutations_test.go, line 48 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMM_dYw-gxLQZI4l-Z:-KzMM_dYw-gxLQZI4l-_:b-5qioks) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/systest/mutations_test.go#L48)):*
> ```Go
> 		CommitNow: true,
> 		SetNquads: []byte(`
>         <0x1> <name> "abc" .
> ```

Can you also add a case where `name` is of list type and see how it goes?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759)*

peterstace:

---

*[posting/list.go, line 880 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMEU5prsfVbOR-wAov:-KzM_xr_aCGgqi6PMfKH:bu67nbq) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/posting/list.go#L880)):*

Emm.. this might not work. Sometimes langs have subtags. https://www.w3.org/International/articles/language-tags/

So, we should hard-code a list of languages; instead of generating all permutations. You can pick them up online.
</blockquote></details>

Done. Added a hard coded list of languages that we can add to as needed.

---

*[posting/list.go, line 884 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMErbexm7bDoHyuGrT:-KzMa8vnt7wCsYdHQ6gl:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/posting/list.go#L884)):*

Also, confirm that there's no collision.
</blockquote></details>

Done.

---

*[protos/task.proto, line 61 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMEx9KizIg6cyQkcNH:-KzMb-p_FdCEgEFXzfko:b-x9mbhp) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/protos/task.proto#L61)):*

Add a comment about what this means.
</blockquote></details>

Added a comment.  @manishrjain , the existing expand all logic is at the level above - it gets all predicates and then expands each. That wouldn't work for languages, since we still need to expand all variants under a single predicate.

---

*[query/query.go, line 498 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMKAscH54seTp7dXNg:-KzMcf50cG5Mp6-v-HIX:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/query/query.go#L498)):*

Aah, return error instead of asserting.  
</blockquote></details>

Done.

---

*[query/query.go, line 1750 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMJzuYKphUTjs5AhNO:-KzMgCmELsataBirTBAj:b7jgk10) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/query/query.go#L1750)):*

So currently `Expand` can be non-empty when the Subgraph is either of the two below.
```
expand(_all_)
expand(val(x))
```

This is a different case where we set Expand on an expanded child. This should probably be a different field called `Expanded` which can be a bool.
</blockquote></details>

Done. I've now got `expandAll` as a field in the `params` type.

---

*[systest/bulk_live_fixture_test.go, line 21 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMFGlceBGC2rTSw1dh:-KzMdGFrag45jSSHemzU:b28rs3c) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/systest/bulk_live_fixture_test.go#L21)):*

Is this change accidental?
</blockquote></details>

That was intentional. Some of the tests in this package run with short=true, and we need to compile dgraph for them.

---

*[systest/mutations_test.go, line 48 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMM_dYw-gxLQZI4l-Z:-KzMd7xX6HkEfUYWtTc0:bnblobz) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/systest/mutations_test.go#L48)):*

Can you also add a case where `name` is of list type and see how it goes?
</blockquote></details>

I've already got an example using a predicate named `list` below. Is that what you mean?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759)*

peterstace:
Review status: 3 of 8 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[posting/list.go, line 897 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMIKTPHF8hqCWCVpQu:-KzMiucS7No-XVinTPIv:by612oc) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/posting/list.go#L897)):*

This would also get postings corresponding to different values for list type. We set a meta for language. How about checking that?
</blockquote></details>

Actually, turns out we store the whole language tag inside the posting inside the `Metadata` field. So a lot of the tag map etc won't be needed after all.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759)*

pawanrawal:
Review status: 3 of 8 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[posting/list.go, line 897 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMIKTPHF8hqCWCVpQu:-KzMkfG5CYvHrr1oP7EC:b-5fkpqt) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/posting/list.go#L897)):*

Actually, turns out we store the whole language tag inside the posting inside the `Metadata` field. So a lot of the tag map etc won't be needed after all.
</blockquote></details>

Oh, then we can just use that.

---

*[systest/mutations_test.go, line 48 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMM_dYw-gxLQZI4l-Z:-KzMkOmPDMJqCAvtux3Y:b-6e82mk) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/systest/mutations_test.go#L48)):*

I've already got an example using a predicate named `list` below. Is that what you mean?
</blockquote></details>

Sorry, I meant that an example where the predicate has value in different languages and is of list type. You can probably add some values to `list` in different languages.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759)*

peterstace:
Review status: 3 of 8 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[systest/mutations_test.go, line 48 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMM_dYw-gxLQZI4l-Z:-KzMq_oqu48_-8SSv5k3:bamas4e) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/systest/mutations_test.go#L48)):*

Sorry, I meant that an example where the predicate has value in different languages and is of list type. You can probably add some values to `list` in different languages.
</blockquote></details>

It went... weirdly.
```
<0x1> <list> "first@en" .
<0x1> <list> "second" .
```
```
"list": [
  "first@en",
  "second"
],
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759)*

peterstace:
Review status: 3 of 8 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[systest/mutations_test.go, line 48 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMM_dYw-gxLQZI4l-Z:-KzMquB-Va9p5gXjTzMg:b-9nhq3x) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/systest/mutations_test.go#L48)):*

It went... weirdly.
```
<0x1> <list> "first@en" .
<0x1> <list> "second" .
```
```
"list": [
  "first@en",
  "second"
],
```
</blockquote></details>

Woops^^ wrong syntax. What I showed above could be expected.

With:
```
<0x1> <list> "first"@it .
<0x1> <list> "first"@en .
<0x1> <list> "second" .
```
Result:
```
"list": [
  "first",
  "first",
  "second"
],
```
Similar when one has default lang: With:
```
<0x1> <list> "first" .
<0x1> <list> "first"@en .
<0x1> <list> "second" .
```
```
"list": [
  "first",
  "first",
  "second"
],
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759)*

peterstace:
Review status: 1 of 8 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[systest/mutations_test.go, line 48 at r4](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-KzMM_dYw-gxLQZI4l-Z:-KzMz9z-LQ2rH5KZ3-i-:b-fbbjyd) ([raw file](https://github.com/dgraph-io/dgraph/blob/8b59cefbe119e5894e4dbbb5f61f533f7271c813/systest/mutations_test.go#L48)):*

Woops^^ wrong syntax. What I showed above could be expected.

With:
```
<0x1> <list> "first"@it .
<0x1> <list> "first"@en .
<0x1> <list> "second" .
```
Result:
```
"list": [
  "first",
  "first",
  "second"
],
```
Similar when one has default lang: With:
```
<0x1> <list> "first" .
<0x1> <list> "first"@en .
<0x1> <list> "second" .
```
```
"list": [
  "first",
  "first",
  "second"
],
```

</blockquote></details>

I made some changes, interaction between lists and languages is more sensible now.

```
<0x1> <list> "first" .
<0x1> <list> "first_en"@en .
<0x1> <list> "first_it"@it .
<0x1> <list> "second" .
```
```
"list": [
  "second",
  "first"
],
"list@en": "first_en",
"list@it": "first_it"
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759)*

peterstace:
Sorry about all the back and forth - this is ready to look at again.

---

Review status: 1 of 8 files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-:-KzN1vbZjdguxZ9KZqaf:bb935hb)*

pawanrawal:
:lgtm: Glad that we got rid of the extra code for the map.

---

Reviewed 1 of 3 files at r4, 7 of 7 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1759#-:-KzN3XnPl1-0MA-scQ1S:bff5b8s)*

PR #1894, opened 2017-12-08
Live/bulk: use leader zero for ts and lease

peterstace:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1894)


janardhan1993:

---

*[xidmap/zero.go, line 26 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1894#-L01qdH4t2DT_NR1oRai:-L01qdH4t2DT_NR1oRaj:b5dkmha) ([raw file](https://github.com/dgraph-io/dgraph/blob/b3109f4471fd925143f374e30bb6c22d1c6f339b/xidmap/zero.go#L26)):*
> ```Go
> func (p *ZeroPool) Leader() (intern.ZeroClient, error) {
> 	for _, addr := range p.knownAddrs {
> 		conn, err := p.dial(addr)
> ```

We can avoid dialing same address everytime. You can use conn/pool.go package to reuse the pool code.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1894)*

peterstace:

---

*[xidmap/zero.go, line 26 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1894#-L01qdH4t2DT_NR1oRai:-L02BkMVjeyWYySgM0tk:b-pgafvi) ([raw file](https://github.com/dgraph-io/dgraph/blob/b3109f4471fd925143f374e30bb6c22d1c6f339b/xidmap/zero.go#L26)):*

We can avoid dialing same address everytime. You can use conn/pool.go package to reuse the pool code.
</blockquote></details>

I tweaked the code a bit so we only dial once per address. It makes the code simpler to just change the lice into a map than get this to work with the code in `conn/pool.go`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1894)*

janardhan1993:
:lgtm:

---

Reviewed 3 of 5 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1894#-:-L02JkiG6jnl0xvTf_Mz:bnfp4nl)*

manishrjain:
Cancel the LGTM by @janardhan1993, and work out a simpler logic for connection to zero.

---

---

*[xidmap/zero.go, line 24 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1894#-L02RZiNIf2SJeLHW7lA:-L02RZiNIf2SJeLHW7lB:b-9y7az3) ([raw file](https://github.com/dgraph-io/dgraph/blob/bf735cfec8e91dbf204ee78264cb9f33e4e68360/xidmap/zero.go#L24)):*
> ```Go
> }
> 
> func (p *ZeroPool) Leader() (intern.ZeroClient, error) {
> ```

I find this code a bit strange. It's connecting to the system, to get the leader, while iterating over all the known zeros. Also, it would do this every time `Leader()` is called, which is multiple times based on the xidmap.

This is a confusing behavior. Try to create a clear flow and simplify the logic here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1894#-:-L02S0p9zMyJ7JHDvGR-:bqfmztb)*

peterstace:

---

*[xidmap/zero.go, line 24 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1894#-L02RZiNIf2SJeLHW7lA:-L03OLrHhXpHsGQCbjeF:b-ydz2ea) ([raw file](https://github.com/dgraph-io/dgraph/blob/bf735cfec8e91dbf204ee78264cb9f33e4e68360/xidmap/zero.go#L24)):*

I find this code a bit strange. It's connecting to the system, to get the leader, while iterating over all the known zeros. Also, it would do this every time `Leader()` is called, which is multiple times based on the xidmap.

This is a confusing behavior. Try to create a clear flow and simplify the logic here.
</blockquote></details>

It's true that `Leader()` is called multiple times by the xid map. In particular, it will call `Leader()` whenever a uid range request fails (which will most likely be because that zero died or there's some kind of communication problem that has caused a new election cycle).

It does iterate over all zeros here, although it will stop iterating at the first zero that can successfully provide a membership state (so it will usually only look at one zero). It will only look at other zeros if the first fails (which is what we want to do anyway).

If you've got any concrete suggestions about how to make the flow more clear, I'm all ears.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1894)*

manishrjain:
:lgtm: One comment.

---

Reviewed 3 of 5 files at r1, 2 of 2 files at r2.

---

*[xidmap/zero.go, line 42 at r2](https://reviewable.io:443/reviews/dgraph-io/dgraph/1894#-L04gCBly4lTzp8_kxVp:-L04gCBmKYVf6sQQvj89:b64kfca) ([raw file](https://github.com/dgraph-io/dgraph/blob/bf735cfec8e91dbf204ee78264cb9f33e4e68360/xidmap/zero.go#L42)):*
> ```Go
> 			// Adds address to map map of known connections, but doesn't
> 			// overwrite connection if it already exists.
> 			p.knownZeros[member.Addr] = p.knownZeros[member.Addr]
> ```

p.knownZeros[member.Addr] = member?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1894#-:-L04gQK-MMvswy3T_6UR:bpjwta0)*

manishrjain:
So, a way to simplify is this:

- If we assume that we have a list of Zeros to begin with, we could just try to send the reqs directly, and switch over to another Zero in case we get an error about that instance not being a leader.
- If we assume that we have only one zero, then we can retrieve a list, and connect to the leader. Note that this instance only needs to be active when the loader is started, since we're going to store this list anyway.

I think we don't need to assume that we have a list of multiple zeros to begin with, and also check who's the leader. Taking in a single zero, and retrieving the membership state from it, should give us everything we need.

peterstace:
@manishrjain , I've implemented what you suggested. It now works as follows:

- Single zero address is provided by the user.
- We use that to get membership info about other nodes.
- Whenever a request fails, we round-robin over the other zero nodes until one works successfully (and stay with that node until a request fails).

Note that with this approach, the loader could get stuck if group zero's structure is modified after the loader starts (e.g. new nodes are added).

manishrjain:
Review status: 1 of 6 files reviewed at latest revision, 3 unresolved discussions.

---

*[xidmap/zero.go, line 17 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1894#-L07VVtxv8_QivqpcIxW:-L07VVty1tpOjx5fKZmA:bvskwe5) ([raw file](https://github.com/dgraph-io/dgraph/blob/36f6181e46ea09a2cd9fbc933a106bcc7a77b947/xidmap/zero.go#L17)):*
> ```Go
> }
> 
> func NewZeroPool(dial func(string) (*grpc.ClientConn, error), addr string) (*ZeroPool, error) {
> ```

ZeroLeader(addr string) // Initially would be passed by user, but later would use the member state.

---

*[xidmap/zero.go, line 32 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1894#-L07VQ3xIJjJpmIryFhW:-L07VQ3xIJjJpmIryFhX:b-w3tyn6) ([raw file](https://github.com/dgraph-io/dgraph/blob/36f6181e46ea09a2cd9fbc933a106bcc7a77b947/xidmap/zero.go#L32)):*
> ```Go
> 		return nil, err
> 	}
> 	for _, member := range state.GetState().Zeros {
> ```

Around this, just store the state. Pick the leader address, dial to it, and return that conn.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1894)*

peterstace:
Review status: 1 of 6 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[xidmap/zero.go, line 17 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1894#-L07VVtxv8_QivqpcIxW:-L080EWVYF1makDGtEMH:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/36f6181e46ea09a2cd9fbc933a106bcc7a77b947/xidmap/zero.go#L17)):*

ZeroLeader(addr string) // Initially would be passed by user, but later would use the member state.
</blockquote></details>

Done.

---

*[xidmap/zero.go, line 32 at r3](https://reviewable.io:443/reviews/dgraph-io/dgraph/1894#-L07VQ3xIJjJpmIryFhW:-L080FPKEd11ILMJye_Z:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/36f6181e46ea09a2cd9fbc933a106bcc7a77b947/xidmap/zero.go#L32)):*

Around this, just store the state. Pick the leader address, dial to it, and return that conn.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1894)*

manishrjain:
:lgtm:

---

Reviewed 1 of 5 files at r3, 4 of 4 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1894#-:-L0HTFNHBlrmtFbCVQ2P:bnfp4nl)*

PR #1932, opened 2017-12-19
Make sure at least 1 field is set in alter

peterstace:
I ended wasting a fair bit of time because I thought I was dropping all but was missing the underscore in the field name (JSON object). This change causes an error to come back to the user if the decoded Operation object has no fields set (if I make the mistake, then chances are someone else will make it as well).

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/1932)

### DIFF
--- a/dgraph/cmd/server/http.go
+++ b/dgraph/cmd/server/http.go
@@ -392,6 +392,11 @@ func alterHandler(w http.ResponseWriter, r *http.Request) {
 		op.Schema = string(b)
 	}
 
+	if err := checkOpValid(op); err != nil {
+		x.SetStatus(w, x.ErrorInvalidRequest, err.Error())
+		return
+	}
+
 	_, err = (&edgraph.Server{}).Alter(context.Background(), op)
 	if err != nil {
 		x.SetStatus(w, x.Error, err.Error())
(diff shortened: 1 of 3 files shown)


pawanrawal:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---

*[edgraph/server.go, line 264 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1932#-L0hRfR7eN6fspvNv2Sl:-L0hRfR7eN6fspvNv2Sm:b-e3vnij) ([raw file](https://github.com/dgraph-io/dgraph/blob/3171043bf2893bd0073349a66c769daf25ec1c29/edgraph/server.go#L264)):*
> ```Go
> 		op.StartTs = State.getTimestamp()
> 	}
> 	m := &intern.Mutations{Schema: updates, StartTs: op.StartTs}
> ```

Actually, just checked the code and it seems like StartTs shouldn't even be part of `protos.Operation` as all schema mutations happen at timestamp `1`. See updateSchema in `worker/mutation.go`. 

So we can get rid of StartTs from the `Operation` message.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1932#-:-L0hTKAzqzxGk1nlebT_:bnfp4nl)*

peterstace:

---

*[edgraph/server.go, line 264 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/1932#-L0hRfR7eN6fspvNv2Sl:-L0hU2aLtvPvN5ciWqtU:b-496hy7) ([raw file](https://github.com/dgraph-io/dgraph/blob/3171043bf2893bd0073349a66c769daf25ec1c29/edgraph/server.go#L264)):*

Actually, just checked the code and it seems like StartTs shouldn't even be part of `protos.Operation` as all schema mutations happen at timestamp `1`. See updateSchema in `worker/mutation.go`. 

So we can get rid of StartTs from the `Operation` message.
</blockquote></details>

That's a good point. It seems like a fairly separate change, so I'll create a new ticket.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/1932)*

PR #2035, opened 2018-01-18
Fix for TestLoaderXidMap on OSX

sboorlagadda:
zcat on osx seems to add .Z when running in POSIX mode (refer man zcat).
   So changing the test to use `gunzip -c filename.gz`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2035)


pawanrawal:
Thanks for the fix @sboorlagadda!

PR #2065, opened 2018-01-30
Added version subcommand to the CLI

### LINKED ISSUE #1993: sgraph -version flag

I think 1.1

Yes

Linux / intel cpu /16GB / SSD

docker exec -it dgraph dgraph -version

Expected to get version

Got
>Error: unknown shorthand flag: 'v' in -version
>Usage:
>  dgraph [command]
>
>Available Commands:
>  bulk        Run Dgraph bulk loader
>  help        Help about any command
>  live        Run Dgraph live loader
>  server      Run Dgraph data server
>  zero        Run Dgraph zero server
>
>Flags:
>      --block_rate int        Block profiling rate. Must be used along with block profile_mode
>      --config string         Configuration file. Takes precedence over default values, but is overridden to >values set with environment variables and flags.
>  -h, --help                  help for dgraph
>      --profile_mode string   Enable profiling mode, one of [cpu, mem, mutex, block]
>
>Use "dgraph [command] --help" for more information about a command.
>
>unknown shorthand flag: 'v' in -version

Also tried `-v` & `--version`

kshitij10496:
A new subcommand, `version`, is added to the dgraph commandline which specifies the current version of dgraph being used. 
Go's `version` subcommand is used as the guideline.

Closes #1993 

**Implementation:**
I have created a new `version` package for this command so as to maintain consistency across all the subcommands defined (such as `bulk`, zero`). If we prefer to keep this short and declare this command directly as a variable in the `main.go` file itself.

Ping @pawanrawal


kshitij10496:
Thanks for the quick review @pawanrawal 

> Could you also add a test for this?

Sure.

I had not added a test suite so as to ensure that the implementation was appropriate.
If we are happy with this implementation, I will add a unit test ASAP.

Also, should I squash both the implementation and the test into a single commit and force push the changes or are separate commits feasible?

pawanrawal:
Don't worry about the commits. I will squash and merge your branch later.

kshitij10496:
@pawanrawal I tried to write a test for the new subcommand using [this](https://stackoverflow.com/a/35827855/4927751) as the reference. 
My takeaway from this answer was that:
In order to ensure that a CLI command (built using a framework) works, we can simply test its executor function (in our case the `Run` function value). This helps in mocking the entire process of CLI tool installation and execution of the commands. 

In our case, we are using the [`PrintVersionOnly`](https://github.com/dgraph-io/dgraph/blob/master/x/init.go#L77) function to do our work. How should we proceed to test this function since it involves printing to the `StdOut`?

pawanrawal:
Its probably too much work and not worth it. I will go ahead and merge this.

kshitij10496:
Thank you @pawanrawal for helping me get started and merging my patch.
Onto the next issue!

PR #2128, opened 2018-02-14
Get updated lease from Zero if uid > maxLeaseId

pawanrawal:
We have had users complain that they get error uid > lease even when uid was returned by us. This can happen when there is some latency between Zero and Server. Users have to retry their mutation in this case which is worse than use making another RPC call to verify our state.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2128)

### DIFF
--- a/dgraph/cmd/zero/tablet.go
+++ b/dgraph/cmd/zero/tablet.go
@@ -226,7 +226,7 @@ func (s *Server) moveTablet(ctx context.Context, predicate string, srcGroup uint
 		Force:     true,
 	}
 	if err := s.Node.proposeAndWait(context.Background(), p); err != nil {
-		x.Printf("Error while reverting group %d to RW", srcGroup)
+		x.Printf("Error while reverting group %d to RW: %+v\n", srcGroup, err)
 	}
 	return err
 }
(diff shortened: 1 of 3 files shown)


manishrjain:
:lgtm: Couple of comments.

---

Reviewed 2 of 2 files at r1.

---

*[worker/groups.go, line 217 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/2128#-L5Md7_-juNnvQ8XtwrZ:-L5Md7_-juNnvQ8Xtwr_:b4qwc1l) ([raw file](https://github.com/dgraph-io/dgraph/blob/67a642a45dda47dd0840d378b14f062d9e7d2705/worker/groups.go#L217)):*
> ```Go
> func UpdateMembershipState(ctx context.Context) error {
> 	g := groups()
> 	p := g.AnyServer(0)
> ```

This should be the leader. We shouldn't get this info from anywhere else.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/2128#-:-L5MduIBvRYn7dNOjfBQ:bojlezm)*

pawanrawal:
Merged to master with 8d2d670ecb3ac071e8a37ede3b69c942c494c1df.

Issue #2138, opened 2018-02-18
Segfault on startup


aphyr:
On 1.0.3 and the current nightly (2018/02/17), dgraph alpha can segfault shortly after cluster startup when nodes are started concurrently. Here are [complete cluster logs](https://github.com/dgraph-io/dgraph/files/1734317/20180217T142809.000-0600.zip); node n5 crashed.

```
2018-02-17 12:28:28 Jepsen starting dgraph :server :--memory_mb 1024 :--idx 5 :--my n5:7080 :--zero n5:5080
2018/02/17 12:28:28 groups.go:86: Current Raft Id: 5
2018/02/17 12:28:28 gRPC server started.  Listening on port 9080
2018/02/17 12:28:28 HTTP server started.  Listening on port 8080
2018/02/17 12:28:28 worker.go:99: Worker listening at address: [::]:7080
2018/02/17 12:28:28 pool.go:118: == CONNECT ==> Setting n5:5080
2018/02/17 12:28:28 groups.go:109: Connected to group zero. Connection state: member:<id:5 group_id:1 addr:"n5:7080" > state:<counter:13 groups:<key:1 value:<members:<key:5 value:<id:5 group_id:1 addr:"n5:7080" > > > > zeros:<key:1 value:<id:1 addr:"n1:5080" leader:true > > zeros:<key:5 value:<id:5 addr:"n5:5080" > > > 
2018/02/17 12:28:28 pool.go:118: == CONNECT ==> Setting n1:5080
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x109788a]

goroutine 243 [running]:
github.com/dgraph-io/dgraph/worker.(*node).AmLeader(0x0, 0x1)
	/home/travis/gopath/src/github.com/dgraph-io/dgraph/worker/draft.go:697 +0x3a
github.com/dgraph-io/dgraph/worker.(*grpcWorker).PredicateAndSchemaData(0xc4201c8180, 0xc4200be900, 0x1a1bf60, 0xc4202da330, 0x0, 0x0)
	/home/travis/gopath/src/github.com/dgraph-io/dgraph/worker/predicate.go:220 +0x737
github.com/dgraph-io/dgraph/protos/intern._Worker_PredicateAndSchemaData_Handler(0x12ac480, 0xc4201c8180, 0x1a1b240, 0xc4202cc320, 0x0, 0x0)
[...]

Issue #2137, opened 2018-02-18
Cluster fails to converge on startup


aphyr:
On the current (2018/02/17) nightlies, when setting up a cluster of five nodes, all running dgraph zero and alpha, some alpha nodes get stuck loading snapshots and return "server not ready" errors for requests indefinitely. In [20180217T143148.000-0600.zip](https://github.com/dgraph-io/dgraph/files/1734313/20180217T143148.000-0600.zip), for instance, node `n1` repeatedly tries to fetch a snapshot, but never succeeds:

```
2018/02/17 12:32:08 draft.go:139: Node ID: 1 with GroupID: 1
2018/02/17 12:32:08 node.go:258: Group 1 found 0 entries
2018/02/17 12:32:08 draft.go:673: New Node for group: 1
2018/02/17 12:32:08 draft.go:678: Retrieving snapshot.
2018/02/17 12:32:08 draft.go:649: Error while calling fn: Cannot retrieve snapshot from peer, error: rpc error: code = Unknown desc = Group 1 not served.
. Retrying...
2018/02/17 12:32:09 draft.go:649: Error while calling fn: Cannot retrieve snapshot from peer, error: rpc error: code = Unknown desc = Not leader of group: 1
. Retrying...
2018/02/17 12:32:10 draft.go:649: Error while calling fn: Cannot retrieve snapshot from peer, error: rpc error: code = Unknown desc = Not leader of group: 1
. Retrying...
2018/02/17 12:32:11 draft.go:649: Error while calling fn: Cannot retrieve snapshot from peer, error: rpc error: code = Unknown desc = Not leader of group: 1
. Retrying...
2018/02/17 12:32:12 draft.go:649: Error while calling fn: Cannot retrieve snapshot from peer, error: rpc error: code = Unknown desc = Not leader of group: 1
. Retrying...
2018/02/17 12:32:13 predicate.go:51: Getting SNAPSHOT: Time elapsed: 05s, bytes written: 0 B
2018/02/17 12:32:13 draft.go:649: Error while calling fn: Cannot retrieve snapshot from peer, error: rpc error: code = Unknown desc = Not leader of group: 1
. Retrying...
2018/02/17 12:32:14 predicate.go:51: Getting SNAPSHOT: Time elapsed: 05s, bytes written: 0 B
2018/02/17 12:32:14 draft.go:649: Error while calling fn: Cannot retrieve snapshot from peer, error: rpc error: code = Unknown desc = Not leader of group: 1
[...]

PR #2140, opened 2018-02-19
Fix for 2137 and 2138

pawanrawal:
#2137 was caused because retrieveSnapshot happens in a blocking fashion, and say when it is called the first time we don't have leader information, then because membership update doesn't happen we never get the leader.

#2138 happens because we might receive a request even before a node is initialized and even if we are not the leader. Now `Leader()` only returns leader for a group, so if we can't find one we will update state and retry.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2140)

### DIFF
--- a/worker/draft.go
+++ b/worker/draft.go
@@ -416,6 +416,12 @@ func (n *node) applyAllMarks(ctx context.Context) {
 func (n *node) retrieveSnapshot() error {
 	pool := groups().Leader(groups().groupId())
 	if pool == nil {
+		// retrieveSnapshot is blocking at initial start and leader election for a group might
+		// not have happened when it is called. If we can't find a leader, get latest state from
+		// Zero.
+		if err := UpdateMembershipState(context.Background()); err != nil {
+			return fmt.Errorf("Error while trying to update membership state: %+v", err)
+		}
 		return fmt.Errorf("Unable to reach leader in group %d", n.gid)
 	}
 
@@ -623,6 +629,9 @@ func (n *node) joinPeers() error {
 	// Get leader information for MY group.
 	pl := groups().Leader(n.gid)
 	if pl == nil {
+		if err := UpdateMembershipState(context.Background()); err != nil {
+			return fmt.Errorf("Error while trying to update membership state: %+v", err)
+		}
 		return x.Errorf("Unable to reach leader or any other server in group %d", n.gid)
 	}
 
--- a/worker/groups.go
+++ b/worker/groups.go
@@ -106,7 +106,7 @@ func StartRaftNodes(walStore *badger.ManagedDB, bindall bool) {
 	if connState.GetMember() == nil || connState.GetState() == nil {
 		x.Fatalf("Unable to join cluster via dgraphzero")
 	}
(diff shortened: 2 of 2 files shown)


manishrjain:
Reviewed 2 of 2 files at r1.

---

*[worker/draft.go, line 417 at r1](https://beta.reviewable.io/reviews/dgraph-io/dgraph/2140#-L5fe1019mIxjfkX2mss:-L5fe1019mIxjfkX2mst:bp220d2) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7a5a476b0c3816fd4f0b49d60d9e2ddfc4c8f31/worker/draft.go#L417)):*
> ```Go
> 
> func (n *node) retrieveSnapshot() error {
> 	pool := groups().Leader(groups().groupId())
> ```

Why not do this in an infinite for loop?

```
for {
  pool := leader
  if pool != nil { break }
  if err := UpdateMembershipState(..) not nil, return err
}
```

---

*[worker/draft.go, line 630 at r1](https://beta.reviewable.io/reviews/dgraph-io/dgraph/2140#-L5feKQC8P4EsSeYyJjo:-L5feKQC8P4EsSeYyJjp:bkdp926) ([raw file](https://github.com/dgraph-io/dgraph/blob/f7a5a476b0c3816fd4f0b49d60d9e2ddfc4c8f31/worker/draft.go#L630)):*
> ```Go
> func (n *node) joinPeers() error {
> 	// Get leader information for MY group.
> 	pl := groups().Leader(n.gid)
> ```

INF for loop here as well, as above.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/dgraph-io/dgraph/2140)*

manishrjain:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/2140#-:-L5fgOpWTBfivdqZIDfn:bnfp4nl)*

pawanrawal:
Merged to master with 5563bd2703003fc40609bae79d25296bc7757977.

aphyr:
Hey, uh, is there a chance this might still allow crashes to happen? You mentioned that if nodes get requests when they're not the leader, that can cause them to crash--is the change to Leader() here supposed to prevent *clients* from contacting nodes which aren't leaders yet? What happens if a node makes a request to a node it *thinks* is the leader, but that leader steps down, then receives the request?

pawanrawal:
> is the change to Leader() here supposed to prevent clients from contacting nodes which aren't leaders yet?

Yes, earlier it used to return a node which was not a leader if there was no leader. Now, it returns an error so it would retry after sleeping for a second.

> What happens if a node makes a request to a node it thinks is the leader, but that leader steps down, then receives the request?

The node receiving the request for snapshot returns an error if it is not a leader, see https://github.com/dgraph-io/dgraph/blob/f28f7fcbacd0e02f4c83cf635a68b099c4efcf08/worker/predicate.go#L220.

In `JoinCluster`, the node receiving the request doesn't check if it is still the leader but that should be ok, as a follower could also propose for a node to joined to the cluster. Did you observe this crash again?

PR #2153, opened 2018-02-21
Make a bunch of changes to abort pending transactions so that snapshot happens more frequently..

pawanrawal:
1. Try snapshotting every 100 entries.
2. If `lastApplied - txnWatermark > 10000`, try aborting older transactions. This gives us an estimate of the number of entries in WAL. Ideally, this should be done based on size of in-memory WAL but the interface doesn't provide any such function. This would still mean that if they are doing batch mutations then this condition would not trigger until very late.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2153)

### DIFF
--- a/edgraph/config.go
+++ b/edgraph/config.go
@@ -63,7 +63,7 @@ var DefaultConfig = Options{
 	Tracing:             0.0,
 	MyAddr:              "",
 	ZeroAddr:            fmt.Sprintf("localhost:%d", x.PortZeroGrpc),
-	MaxPendingCount:     1000,
+	MaxPendingCount:     100,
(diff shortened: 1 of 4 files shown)


janardhan1993:
:lgtm:

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/2153#-:-L6GqGA58BqB72kcL9xO:bnfp4nl)*

pawanrawal:
Merged to master with 87aa10bbcb58e266649f03fc8a4fcb842cbf02f6.

Issue #2174, opened 2018-03-01
add numedges limit as a flag


jzhu077:
Feature request:
It would be nice if numEdges limit can be set as a flag because we have to set it to math.MaxUint64 to fit our use case.
https://github.com/dgraph-io/dgraph/blob/master/query/shortest.go#L214
https://github.com/dgraph-io/dgraph/blob/master/query/recurse.go#L167

pawanrawal:
This was added to master.

PR #2176, opened 2018-03-01
add max_num_edge_limit flag

jzhu077:
https://github.com/dgraph-io/dgraph/issues/2174

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2176)

### DIFF
--- a/dgraph/cmd/server/run.go
+++ b/dgraph/cmd/server/run.go
@@ -42,6 +42,7 @@ import (
 	"github.com/dgraph-io/dgraph/tok"
 	"github.com/dgraph-io/dgraph/worker"
 	"github.com/dgraph-io/dgraph/x"
+	"github.com/spf13/cast"
 	"github.com/spf13/cobra"
 )
 
@@ -111,6 +112,9 @@ func init() {
 		"Use 0.0.0.0 instead of localhost to bind to all addresses on local machine.")
 	flag.Bool("expose_trace", false,
 		"Allow trace endpoint to be accessible from remote")
+
+	flag.Uint64("query_edge_limit", 10000000,
+		"Limit for the maximum number of edges that can be returned in a query.")
 
 	// TLS configurations
 	x.RegisterTLSFlags(flag)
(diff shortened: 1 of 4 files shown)


pawanrawal:
I think a more flexible solution would be for `max_num_edge_limit` to take the number of edges, so making it `Uint64` with a default value of `1e6`. Then you could override that value with `math.MaxUint64`.

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/2176#-:-L6YhZqe86Na6fE2V5u4:b-b1thy4)*

pawanrawal:
Also, CI fails because of import cycle error.

jzhu077:
Updated. I have built and run locally.

pawanrawal:
That is a bit strange. Here are the logs from CI https://teamcity.dgraph.io/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=546&_focus=1428.

package `edgraph` imports `query` and `query` imports `edgraph` hence the dependency issue.

jzhu077:
I see the problem. I was building it in a directory with a vendor folder that contains dgraph-io.

jzhu077:
There are two options here, one is to remove the global variable Config in https://github.com/dgraph-io/dgraph/blob/master/edgraph/config.go#L29 then pass it as an argument to the relevant functions or create a separate package for the config struct. What do you reckon is the best option here? Should I wait for it to be fixed before I edit this PR?

pawanrawal:
I think you can just define a variable in `x` package (we use it to circumvent the circluar dependency issue) and set it equal to the value of the flag on start. Then this can be referenced from query.

jzhu077:
That's a good solution.PR updated.

dgraph/cmd/server/run.go
```diff
@@ -112,6 +113,9 @@ func init() {
 	flag.Bool("expose_trace", false,
 		"Allow trace endpoint to be accessible from remote")
 
+	flag.Uint64("num_edge_limit", 10000000,
+		"Number of edges limit")
```
pawanrawal: Please change the flag name to `query_edge_limit`. Otherwise, I am sure some users would think this is a limit on the number of edges the database can store. Also, the description can be clarified.
```
Limit for the maximum number of edges that can be returned in a query. 
```

jzhu077: Agreed

PR #2201, opened 2018-03-07
Add upsert directive

### LINKED ISSUE #2200: Add conflict directive to schema
The schema should allow for a `@conflict` directive. If a predicate has this directive, then we use the index keys for it for conflict detection. Something like

```
email: string @index(exact) @conflict
```

For other predicates, conflict detection for the index keys would not be done by default. We still plan to keep `IgnoreIndexConflict` flag but it won't be used anymore.

pawanrawal:
Resolves #2200

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2201)


manishrjain:
Largely looks good. Few questions. Get @janardhan1993 to do the final LGTM. I'll read your responses tomorrow (could be after you merge).

---

Reviewed 17 of 17 files at r1.

---

*[posting/list.go, line 332 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/2201#-L6yzzIvquLbTWN-HiEu:-L6yzzIvquLbTWN-HiEv:b-lz82p6) ([raw file](https://github.com/dgraph-io/dgraph/blob/472042ba1e449f4c85563ed25026afd1e20bb022/posting/list.go#L332)):*
> ```Go
> 		doAbort = false
> 		ignoreConflict = true
> 	} else if !schema.State().HasUpsert(t.Attr) && !x.Parse(l.key).IsData() {
> ```

This logic could be clarified.

doAbort = false

if data key ||
    is upsert { doAbort = true }

---

*[worker/mutation.go, line 296 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/2201#-L6z-XrjxiosRxCoGzsQ:-L6z-XrjxiosRxCoGzsR:b8tb4im) ([raw file](https://github.com/dgraph-io/dgraph/blob/472042ba1e449f4c85563ed25026afd1e20bb022/worker/mutation.go#L296)):*
> ```Go
> 	// It should have index directive.
> 	if s.Upsert && !(typ == types.StringID || typ == types.IntID) {
> 		return x.Errorf("@upsert directive is only supported for int/string type."+
> ```

Why this limitation? Can't we use this for time or float?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/2201#-:-L6z-jTsEHdqfk_hyitf:bmwdgvj)*

janardhan1993:
:lgtm:

---

Reviewed 17 of 17 files at r1.

---

*[worker/mutation.go, line 295 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/2201#-L6z4htekbPkSEhgkhIy:-L6z4htekbPkSEhgkhIz:b-4g7ovj) ([raw file](https://github.com/dgraph-io/dgraph/blob/472042ba1e449f4c85563ed25026afd1e20bb022/worker/mutation.go#L295)):*
> ```Go
> 	// Type should be int/string.
> 	// It should have index directive.
> 	if s.Upsert && !(typ == types.StringID || typ == types.IntID) {
> ```

Should be fine for all types ?

---

*[worker/mutation.go, line 301 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/2201#-L6z4bsoQnf7EkXBJ3VX:-L6z4bsoQnf7EkXBJ3VY:b5wpf2w) ([raw file](https://github.com/dgraph-io/dgraph/blob/472042ba1e449f4c85563ed25026afd1e20bb022/worker/mutation.go#L301)):*
> ```Go
> 
> 	if s.Upsert && len(s.Tokenizer) == 0 {
> 		return x.Errorf("Tokenizer is mandatory for: [%s] when specifying @upsert directive",
> ```

Probably tell the user that it should be indexed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/2201#-:-L6z508UWpf1ByyMCYUR:bnfp4nl)*

pawanrawal:
Review status: 13 of 17 files reviewed at latest revision, 4 unresolved discussions.

---

*[posting/list.go, line 332 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/2201#-L6yzzIvquLbTWN-HiEu:-L6zGm6AgpV5EKq02xAr:b-1jzvth) ([raw file](https://github.com/dgraph-io/dgraph/blob/472042ba1e449f4c85563ed25026afd1e20bb022/posting/list.go#L332)):*

This logic could be clarified.

doAbort = false

if data key ||
    is upsert { doAbort = true }
</blockquote></details>

Updated it 

---

*[worker/mutation.go, line 296 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/2201#-L6z-XrjxiosRxCoGzsQ:-L6z2EK6OGuyS9_8a0tc:b41qhgp) ([raw file](https://github.com/dgraph-io/dgraph/blob/472042ba1e449f4c85563ed25026afd1e20bb022/worker/mutation.go#L296)):*

Why this limitation? Can't we use this for time or float?
</blockquote></details>

Considering that `upsert` operation involves an `eq` query we need to store the exact value which we only do for `string @index(exact)` and `int @index(int)`.

We can allow non-lossy tokenizers but that could lead to false conflicts. What do you think about it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/2201)*

manishrjain:
Review status: 13 of 17 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[posting/list.go, line 332 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/2201#-L6yzzIvquLbTWN-HiEu:-L70gHNFOj4SHM4flpEi:b-3qtn9p) ([raw file](https://github.com/dgraph-io/dgraph/blob/472042ba1e449f4c85563ed25026afd1e20bb022/posting/list.go#L332)):*

Updated it 
</blockquote></details>

Looks better. But, this is more expensive. Previously the if else and or conditions would only execute some statements (isdate and hasupsert), not they would always be executed, which isn't great. So, use them within the else if, instead of outside.

---

*[worker/mutation.go, line 296 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/2201#-L6z-XrjxiosRxCoGzsQ:-L70gdsrPrqWLojsmkLR:b-71aq59) ([raw file](https://github.com/dgraph-io/dgraph/blob/472042ba1e449f4c85563ed25026afd1e20bb022/worker/mutation.go#L296)):*

Considering that `upsert` operation involves an `eq` query we need to store the exact value which we only do for `string @index(exact)` and `int @index(int)`.

We can allow non-lossy tokenizers but that could lead to false conflicts. What do you think about it?

</blockquote></details>

I think the choice of int and string is a bit arbitrary. Making this non-lossy is a better approach. But, then again, why restrict to anything? If they decide to put a lossy index and want upsert on it, we should just do it.

---

*[worker/mutation.go, line 301 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/2201#-L70h001TB9E3FYyGCLc:-L70h001TB9E3FYyGCLd:b-momftc) ([raw file](https://github.com/dgraph-io/dgraph/blob/472042ba1e449f4c85563ed25026afd1e20bb022/worker/mutation.go#L301)):*
> ```Go
> 
> 	if s.Upsert && len(s.Tokenizer) == 0 {
> 		return x.Errorf("Tokenizer is mandatory for: [%s] when specifying @upsert directive",
> ```

Use the language: "Index tokenizer"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/2201)*

PR #2240, opened 2018-03-19
send maxpending in connections state

### LINKED ISSUE #2236: Leader not copying snapshots to replicas if replicas start almost immediately after first server
I have a fairly large dataset (700+ MB of a text RDF file). I bring up the first server (leader) with this exported data via the bulk loader (map 2, shard 1). Once I start the second server after the first is up, data will not copy over to it even though the first server was determined to be the leader.

However, if I do the exact same thing but wait a few minutes for the "leader" to presumably catch up, then bring up the second server, it copies the data over. It takes roughly 1.5 hours for the data to copy (2.4GB) to the replica but it appears to work.

What version of Dgraph are you using? 1.04
Have you tried reproducing the issue with latest release? Yes
What is the hardware spec (RAM, OS)? Ubuntu 16.04, Docker, 32GB Ram
Steps to reproduce the issue (command/config used to run Dgraph). Large dataset for leader (700+MB RDF)
Expected behaviour and actual result: Leader to immediately start sending data to followers

janardhan1993:
Fixes #2236

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2240)


pawanrawal:
:lgtm: I have a question, what happens when both Zero and Server are shutdown and restarted? Does Zero update its MaxPending before connecting with the Server?

---

Reviewed 6 of 6 files at r1.

---

*[dgraph/cmd/zero/zero.go, line 344 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/2240#-L8-gf7la2FShYgymW22:-L8-gf7la2FShYgymW23:b-hwd6zm) ([raw file](https://github.com/dgraph-io/dgraph/blob/c6c1f0a3ce4c6ab388eb483bcce1b932db7b40ed/dgraph/cmd/zero/zero.go#L344)):*
> ```Go
> 		ms, err := s.latestMembershipState(ctx)
> 		cs := &intern.ConnectionState{State: ms}
> 		cs.MaxPending = s.orc.MaxPending()
> ```

initialize this in the literal above maybe?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/2240#-:-L8-hEg_oBFWhfh9Qzga:b-e3bhwa)*

PR #2249, opened 2018-03-21
Add a lang directive in schema and use it to optimize certain filter queries.

### LINKED ISSUE #2239: Performance drop on filtering
I'm using dgraph 1.0.3, on a 8 core 16G RAM CentOS 7 instance, with a small data set of 10k nodes and 60k edges.

The query I use is like this:

```
{  
  v(func: eq(pred1, "value1")) {
    pred2 @filter(eq(pred3, "value2")) {
      pred4
    }
  }
}
```

This query is taking around 160ms when the server is idle. If I run this query with multiple clients concurrently, it takes awfully more time as concurrency increases. In my case, 8 clients make the query takes up to 1s.

If I remove the `@filter(eq(pred3, "value2"))` filter, it only takes 2ms, and the time will not increase with much more concurrent clients.

The schema on `pred3` is `string @index(hash) .`, which is said to have the best indexing performance in the docs.

I profiled dgraph server, and find that `filterStringFunction` and `handleCompareFunction` are taking up most of the CPU time. I'm not familiar with the codebase, and didn't have the time to dig further. Would you please investigate this performance issue?

[pprof.dgraph.samples.cpu.007.pb.gz](https://github.com/dgraph-io/dgraph/files/1823671/pprof.dgraph.samples.cpu.007.pb.gz)

pawanrawal:
This should resolve #2239.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2249)


pawanrawal:
Review status: 0 of 14 files reviewed at latest revision, 2 unresolved discussions.

---

*[schema/schema.go, line 104 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/2249#-L86ofiwkFb-TGlqwedK:-L86ofiwkFb-TGlqwedL:b-mw0917) ([raw file](https://github.com/dgraph-io/dgraph/blob/8bf41372a971173dab0c7c3f5002245ade5f57d5/schema/schema.go#L104)):*
> ```Go
> // Get gets the schema for given predicate
> func (s *state) Get(pred string) (intern.SchemaUpdate, bool) {
> 	s.RLock()
> ```

Why was there a write lock for Get?

---

*[worker/mutation.go, line 329 at r1](https://reviewable.io:443/reviews/dgraph-io/dgraph/2249#-L86pL73LhoZjZ62Wnl3:-L86pL73LhoZjZ62Wnl4:b56xz6p) ([raw file](https://github.com/dgraph-io/dgraph/blob/8bf41372a971173dab0c7c3f5002245ade5f57d5/worker/mutation.go#L329)):*
> ```Go
> // If storage type is specified, then check compatibility or convert to schema type
> // if no storage type is specified then convert to schema type.
> func ValidateAndConvert(edge *intern.DirectedEdge, su *intern.SchemaUpdate) error {
> ```

Had to change this because cannot access `schema.State()` using `dgraph bulk`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dgraph-io/dgraph/2249)*

janardhan1993:
:lgtm: Just wondering shouldn't we throw error if user does mutations with lang and doesn't specify @lang in schema and then does a query.
we would give wrong results , we should atleast throw error during query if user tries to search for a language without specifying @lang in schema.

---

Reviewed 14 of 14 files at r1.

---

*[posting/index.go, line 348 at r1](https://beta.reviewable.io/reviews/dgraph-io/dgraph/2249#-L87itJq2UAkf1rqWsCX:-L87itJq2UAkf1rqWsCY:b-cktued) ([raw file](https://github.com/dgraph-io/dgraph/blob/8bf41372a971173dab0c7c3f5002245ade5f57d5/posting/index.go#L348)):*
> ```Go
> 			val, found, err = l.findValue(txn.StartTs, farm.Fingerprint64([]byte(t.Lang)))
> 		} else if schema.State().IsList(t.Attr) {
> 			val, found, err = l.findValue(txn.StartTs, farm.Fingerprint64([]byte(t.Value)))
> ```

This logic is there in posting/list.go, check if we need it there

---

*[schema/schema.go, line 104 at r1](https://beta.reviewable.io/reviews/dgraph-io/dgraph/2249#-L86ofiwkFb-TGlqwedK:-L87iype97KqKKg4mJDT:b6zrrkf) ([raw file](https://github.com/dgraph-io/dgraph/blob/8bf41372a971173dab0c7c3f5002245ade5f57d5/schema/schema.go#L104)):*

Why was there a write lock for Get?
</blockquote></details>

Good catch,

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/dgraph-io/dgraph/2249#-:-L87jgRN9z3hIsf73PxL:b-hb5sdh)*

pawanrawal:
Merged to master with f455229d773e9168002ad3d5ccc8ac083c54d91a.

PR #2251, opened 2018-03-22
Remove timeout from JoinCluster.

pawanrawal:
I couldn't find any issues about RAFT dropping pending conf change entries and therefore I feel this timeout is unnecessary. 

What does happen is that all entries that were proposed need to be applied, even if with [id as `Raft.None`](https://github.com/coreos/etcd/blob/c8cfdb3b558f758367b8f0c4566fce88b56c9aec/etcdserver/server.go#L1499).  So that might have been happening earlier, we could have been returning an error from applyConfChange without applying the pending entry, hence further proposals would be dropped. We don't do that now, so we don't need the timeout.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2251)

### DIFF
--- a/worker/draft.go
+++ b/worker/draft.go
@@ -714,11 +714,7 @@ func (n *node) joinPeers() error {
 	gconn := pl.Get()
 	c := intern.NewRaftClient(gconn)
 	x.Printf("Calling JoinCluster")
-	ctx, cancel := context.WithTimeout(n.ctx, time.Second)
-	defer cancel()
-	// JoinCluster can block indefinitely, raft ignores conf change proposal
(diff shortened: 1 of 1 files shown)


janardhan1993:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/dgraph-io/dgraph/2251#-:-L8A_AKB7JyA5X7guV6X:bnfp4nl)*

pawanrawal:
Merged to master with 807976c407b880cfac565cc709b6bee3c139f7f7.

PR #2263, opened 2018-03-27
Check if node is a peer from leader of a group.

### LINKED ISSUE #2072: Node can incorrectly start as a new node even though it was part of the cluster.
https://discuss.dgraph.io/t/unable-to-run-dgraph-in-a-multi-node-kubernetes-cluster/2147/39

From the discussion above, if a node became part of the cluster but crashed before writing Hard State or Snapshot to WAL it would start as a new node which is not correct. This is a problem because the cluster cannot get into a healthy state.

pawanrawal:
If a node crashed after joining the cluster and before writing any snapshot, we considered it as a new node and tried joining the cluster again. Now we check from the leader of the group if the node is already part of the group and consider it a restart if that's the case.

This fixes #2072.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2263)


pawanrawal:
Merged to master with 33aa2a5

Issue #2143, opened 2018-02-19
Read skew leading to write corruption


aphyr:
In the nightly build for 2018/02/19, with Jepsen, I can reliably reproduce what appears to be a snapshot isolation bug in which the sum of accounts gradually drifts higher and lower over time. This history also appears to exhibit a sequential consistency violation, where clients fail to read their own writes on the same nodes.

This particular version of the bank workload uses a three-part schema:

```
key: int @index(int) .
type: string @index(exact) .
amount: int .
```

Every account has type "account", and is identified by a numeric key. We query for accounts with

```
{ q(func: eq(key, $key)) { uid key amount } }
```

and write back the key and amount, but not the type.

In [this five-node cluster](https://github.com/dgraph-io/dgraph/files/1738137/20180219T142000.000-0600.zip), with replication factor three, shortly after a partition begins...

```clj
{:type :info, :f :start, :process :nemesis, :time 30928687283, :value [:isolated {"n1" #{"n2" "n5" "n3"}, "n4" #{"n2" "n5" "n3"}, "n2" #{"n1" "n4"}, "n5" #{"n1" "n4"}, "n3" #{"n1" "n4"}}]}
```

```clj
{:type :invoke, :f :read, :process 39, :time 54175450678}
{:type :ok, :f :read, :process 39, :time 54177149097, :value {0 48, 1 11, 2 7, 3 1, 4 0, 5 3, 6 24, 7 6}}
```

Process 39 initiates and completes a transfer of $3 from account 5 to 0.

```clj
{:type :invoke, :f :transfer, :value {:from 5, :to 0, :amount 3}, :process 39, :time 54177292423}
{:type :ok, :f :transfer, :value {:from 5, :to 0, :amount 3}, :process 39, :time 54224717067}

Process 39 fails to read its own transfer; this is legal under SI, but might violate the sequential-consistency-within-a-group property that I think Dgraph intends to provide.

```clj
{:type :invoke, :f :read, :process 39, :time 54224966995}
{:type :ok, :f :read, :process 39, :time 54227006450, :value {0 48, 1 11, 2 7, 3 1, 4 0, 5 3, 6 24, 7 6}}
```

Process 39 fails to make two transfers due to insufficient funds.

```clj
{:type :invoke, :f :transfer, :value {:from 4, :to 1, :amount 1}, :process 39, :time 54227194310}
[...]

PR #2287, opened 2018-03-30
add linerizability

### LINKED ISSUE #2148: Delete leaves dangling entities, allows double-upserts
On the current Dgraph nightly (2018/02/20), in a healthy five-node cluster, without any faults, concurrently upserting, deleting, and reading records by a single, indexed integer attribute `key` can result in queries for `key` returning long-lasting records with only a UID and no key attached; as if the UID were still present in that key's index, even though the triple has been deleted. Moreover, reads may actually see one of these dangling deleted records *concurrently* with a new record for that key. Moreover, reads may observe two *complete* records, even though one has been deleted.

This violates snapshot isolation, since reads are clearly not seeing a consistent point-in-time representation of state. It breaks internal consistency, since you can query for a record with `eq(key, 4)` and get back a record which doesn't have a `key` at all. Since a single process can actually see deleted data, it's definitely not sequential.

You can reproduce this on Jepsen 282623ca5b733097b9a524f70a451ca40151310e by running

```
lein run test -w delete -f --package-url https://github.com/dgraph-io/dgraph/releases/download/nightly/dgraph-linux-amd64.tar.gz --time-limit 120 --concurrency 100`.
```

For future users: you'll need 1.0.3 or 1.0.4, since this is on the nightlies and I don't think those have stable URLs. Test code is [here](https://github.com/jepsen-io/jepsen/blob/282623ca5b733097b9a524f70a451ca40151310e/dgraph/src/jepsen/dgraph/delete.clj#L35-L56).

The schema for this test is `key: int @index(int) .`. We read keys by querying

```
{ q(func: eq(key, $key)) {
  uid
  key
}}
```

... delete by querying for just the UID and issuing a delete for every attribute with `{uid: 0x123}`, and upsert by reading for an existing record with that key, and if none exists, inserting a new one. All operations occur within a transaction.

In [this run](https://github.com/dgraph-io/dgraph/files/1742617/20180220T182514.000-0600.zip), for key 0, an entity 0x298 already exists for the key, but is deleted by process 2. Processes 0, 1, 6, and 8 confirm the deletion of that entity: they find no matches for that query. Process 2 upserts a new record, which is visible on reads by process 5 and 6: `{:uid "0x2ab", :key 0}`. Concurrently with that upsert, process 4 deletes the key. Concurrent reads return dangling records: `{:uid "0x2ab"}` *with no key*. Other processes continue trying to delete key 0; their queries observe uid 0x2ab and issue a delete for it, but apparently with no effect. This dangling record, 0x2ab, persists for the remainder of the test--just over two minutes.

screenshot from 2018-02-20 19-41-38

In this diagram, time flows down, each process is a vertical track, operations are colored bars. Blue means success, pink is a known failed operation; e.g. it did not occur.

Moreover, queries may return multiple dangling records:

```clj
[{:uid "0xcf"} {:uid "0x110"}]
```

... a dangling record and a complete record:

```clj
[{:uid "0xcf"} {:uid "0x177", :key 1}]
```

... or multiple complete records for the same key:

```clj
[{:uid "0x148b"} {:uid "0x14f6", :key 13}]
[{:uid "0x148b"} {:uid "0x150c", :key 13}]
[{:uid "0x148b", :key 13} {:uid "0x150c", :key 13}]
```

For instance, in [this run](https://github.com/dgraph-io/dgraph/files/1742707/20180220T203635.000-0600.zip), shortly after an upsert, process 33 manages to observe two complete, coexisting records for the same key: 0x25fb, and 0x26e0. Note that 0x25fb was already deleted, which means it went from dangling to non-dangling. Note also that in order for the upsert to take place, it must have *failed* to observe 0x25fb, or it wouldn't have inserted anything. Something really odd is going on here.

screenshot from 2018-02-20 20-49-11

Even weirder, in [20180220T205917.000-0600.zip](https://github.com/dgraph-io/dgraph/files/1742740/20180220T205917.000-0600.zip), 0x7e7 is a dangling record which process 36 can *see* and apparently successfully delete (although the deletion doesn't actually delete the dangling record). However, right after reading 0x7e7, process 36 attempts a delete and finds nothing to delete. Then 0x7e7 goes right back to being a dangling pointer and is visible to process 36 again--this time as a *full* record, in conjunction with the newly upserted 0x808. 0x808 turns out to be deletable. 0x707 goes back to being a full record, then a dangling record after that.

screenshot from 2018-02-20 21-06-11

janardhan1993:
Fixes #2148  #2143 

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2287)


manishrjain:
Review status: 0 of 5 files reviewed at latest revision, all discussions resolved.

---

*[x/noncecounter.go, line 1 at r1](https://beta.reviewable.io/reviews/dgraph-io/dgraph/2287#-L8sm-ZaEBQi7wq9COF8:-L8sm-ZaEBQi7wq9COF9:ba7tcxc) ([raw file](https://github.com/dgraph-io/dgraph/blob/cea9389404410377a04c39a55cea31f3d8a87824/x/noncecounter.go#L1)):*
> ```Go
> package x
> ```

We don't need a noncecounter class. Just use the standard lib random number generation, that we use for proposal contexts.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/dgraph-io/dgraph/2287)*

janardhan1993:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[x/noncecounter.go, line 1 at r1](https://beta.reviewable.io/reviews/dgraph-io/dgraph/2287#-L8sm-ZaEBQi7wq9COF8:-L9EANx8BOldkA6qxAxp:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/cea9389404410377a04c39a55cea31f3d8a87824/x/noncecounter.go#L1)):*

We don't need a noncecounter class. Just use the standard lib random number generation, that we use for proposal contexts.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/dgraph-io/dgraph/2287)*

manishrjain:
:lgtm:

---

Reviewed 2 of 2 files at r3.

---

*[worker/draft.go, line 560 at r3](https://beta.reviewable.io/reviews/dgraph-io/dgraph/2287#-L9E_xvl9U187dJjMLIp:-L9E_xvl9U187dJjMLIq:b91pui1) ([raw file](https://github.com/dgraph-io/dgraph/blob/4163deed09f17acee48f65feacdc209e788bf1fd/worker/draft.go#L560)):*
> ```Go
> 			}
> 			activeRctx := make([]byte, 8)
> 			n.rand.Read(activeRctx[:])
> ```

This returns error. Assert it.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/dgraph-io/dgraph/2287#-:-L9EaAq-DzNk5FKa4ihD:bnfp4nl)*

PR #2332, opened 2018-04-12
Change counter of Mutations to be TX's to be more explicit

deckarep:
It was a bit confusing that this tool said mutations when I was expecting it to read as transaction count.  Theoretically they could be equivalent but this is truly a count of transactions.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2332)


pawanrawal:
Thanks for the fix @deckarep!

PR #2339, opened 2018-04-17
WIP: Fix for 2338

janardhan1993:


janardhan1993:
Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---

*[worker/scheduler.go, line 155 at r1](https://beta.reviewable.io/reviews/dgraph-io/dgraph/2339#-LAHo4zoDRElmboCNi2z:-LAHo4zoDRElmboCNi3-:bv7sdad) ([raw file](https://github.com/dgraph-io/dgraph/blob/b8c090884d0e8ecc3eac7cb18aef631abdf0b662/worker/scheduler.go#L155)):*
> ```Go
> 		for _, supdate := range proposal.Mutations.Schema {
> 			// This is neceassry to ensure that there is no race between when we start reading
> 			// from badger and new mutation getting commited via raft and getting applied.
> ```

We shouldn't do any check here based on state, since we udpate state without proposals also. This might get triggered only on some replicas due to async nature of stream.

We still need to fix the race though.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/dgraph-io/dgraph/2339)*

janardhan1993:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[worker/scheduler.go, line 155 at r1](https://beta.reviewable.io/reviews/dgraph-io/dgraph/2339#-LAHo4zoDRElmboCNi2z:-LAHoGDA8YhIw8UmxCjT:b-xm5eql) ([raw file](https://github.com/dgraph-io/dgraph/blob/b8c090884d0e8ecc3eac7cb18aef631abdf0b662/worker/scheduler.go#L155)):*

We shouldn't do any check here based on state, since we udpate state without proposals also. This might get triggered only on some replicas due to async nature of stream.

We still need to fix the race though.
</blockquote></details>

we can do this check if we update the state via proposals.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/dgraph-io/dgraph/2339)*

pawanrawal:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[dgraph/cmd/zero/oracle.go, line 302 at r2](https://beta.reviewable.io/reviews/dgraph-io/dgraph/2339#-LAL3KNC8A6G0VK4YppN:-LAL3KNC8A6G0VK4YppO:bnqmuvb) ([raw file](https://github.com/dgraph-io/dgraph/blob/95f33824f12d44756d0015829b837ad4df39007c/dgraph/cmd/zero/oracle.go#L302)):*
> ```Go
> 	// There might be race between this proposal trying to commit and predicate
> 	// move aborting it.
> 	if s.orc.aborted(src.StartTs) {
> ```

Nice fix, though I see that `proposeTxn` would never return any error. This might have been better handled by the functioning returning an error. Can you maybe add a TODO for this?

---

*[worker/groups.go, line 501 at r2](https://beta.reviewable.io/reviews/dgraph-io/dgraph/2339#-LAL7tIx-kYAL2F2P6np:-LAL7tIx-kYAL2F2P6nq:bxo2023) ([raw file](https://github.com/dgraph-io/dgraph/blob/95f33824f12d44756d0015829b837ad4df39007c/worker/groups.go#L501)):*
> ```Go
> 			}
> 			// In case there is no leader, we might miss this state update. So as soon as
> 			// a node becomes leader it asks for fresh state.
> ```

Is it asking or proposing the latest state? I don't follow the problem that we are trying to solve here.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/dgraph-io/dgraph/2339)*

Issue #2482, opened 2018-07-10
Adding array of geo gives "Input for predicate example.Loc of type scalar is uid"


commesan:
- What version of Dgraph are you using?
1.06

When using:

```

     type loc struct {
		Type   string    `json:"type,omitempty"`
		Coords []float64 `json:"coordinates,omitempty"`
	}

	type Tracker struct {
		Uid  string `json:"uid,omitempty"`
		Name string `json:"example.Name,omitempty"`
		Loc  []loc  `json:"example.Loc,omitempty"`
	}

	t := Tracker{
		Name: "Tom Baker",
		Loc: []loc{
			{
				Type:   "Point",
				Coords: []float64{1.1, 2},
			},
			{
				Type:   "Point",
				Coords: []float64{1.9, 5},
			},
		},
	}

```

And mapping:

```
        example.Name: string @index(exact) .
	example.Loc: [geo] @index(geo) .
```

I get Input for predicate example.Loc of type scalar is uid

When using geo as a list the SetJson parser seems to view the geojson types as nodes  #instead of scalars.

PR #2485, opened 2018-07-12
Correctly handle a list of type geo in json (#2482)

commesan:
When adding a list of geo types in a json string, the geojson string was being parsed as a list of connecting entities. I've added a check for geojson in the handling of the  []map[string]interface{} case.

Updated the list test to include a list of geojson.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2485)

### DIFF
--- a/edgraph/nquads_from_json.go
+++ b/edgraph/nquads_from_json.go
@@ -166,6 +166,26 @@ func checkForDeletion(mr *mapResponse, m map[string]interface{}, op int) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


commesan:
Sure, can add the function myself to parse the geo or add the TODO. Ran all tests, nothing broke.

commesan:
Updated branch with a handleGeoType function that is being called for the single geo object and list of geo objects.

manishrjain:
Thanks for your contribution!

PR #2498, opened 2018-07-25
Export mutation errors.

BlakeMScurr:
Exporting errors so I could check for them in my application like:
```
if errString != ErrPredicateMoving.Error() {
  // don't cancel the ingest, it's safe to do a retry loop/backoff
}
```

rather than relying on the fact that the error message won't change.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2498)


srfrog:
instead of simply exporting errors, i think we need error types. ie.,
```go
type Error struct {
   Severity string
   Code string
   System string
   Message string
}

return &Error{"NOTICE", "0A00", "worker", "Predicate is being moved, please retry later"}
// which further allows:
type WorkerError Error
// easier to inspect.
```

manishrjain:
We have an issue tracking this error organization. So, closing this PR.

Issue #2320, opened 2018-04-09
Timeout / Max Hops for shortest() Query?


billylindeman:
Sometimes when running a shortest() query between nodes that aren't closely connected dgraph will take a very long time and even run out of memory. Is there any plans to add a timeout or max hops param to have the query exit safely?

billylindeman:
After doing a bit of digging it looks like this is supported? https://github.com/dgraph-io/dgraph/blob/ddd7414f9b3ec7382a4aa521e9ba2d6e5cd81af7/query/query_test.go#L2296

Updated the ticket to reflect that it is undocumented in dgraph.io

BlankRain:
@billylindeman @pawanrawal  only `depth` is not enough. I also need a version of K-shortest function with  `timeout` param for once there is no path between two nodes, it still cost a lot . 
I need  fail as soon as possible instead of  waiting a long time .

BlankRain:
@billylindeman  @pawanrawal  
hi , I have add a `timeout` param to the `shortest` function.
more details you can find here https://github.com/dgraph-io/dgraph/pull/2561
However, I don't know how to make my PR pass .
@pawanrawal  could you help me to make my PR pass?

BlankRain:
@pawanrawal  yes you are right. I tested it.

PR #2561, opened 2018-09-04
Fix issue-2320: Add  "timeout" param to "KShortestPath"  function

BlankRain:
Related issue 
https://github.com/dgraph-io/dgraph/issues/2320
query may like this:

```
 query x{
 path as shortest(from: 0x5, to: 0x29e81, depth:2 ,numpaths: 2,timeout:1000) {
  actor.film
  }
 entity(func: uid(path)) {
      uid
  } 
}
```
If the time-cost longer than the value `timeout`, an error `: context deadline exceeded` returns.
So we can fail fast instead of blocking system.
Somethings the blocking looks like a `DDOS` attack. High use of CPU, slow or none response.

With the param `timeout`, we can stop this kind of `DDOS`.

By the way, the unit here is `Millisecond`,
If the value is less or equal to zero, means no timeout ,or use the `context`'s default timeout value .

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2561)

### DIFF
--- a/gql/parser.go
+++ b/gql/parser.go
@@ -2104,7 +2104,7 @@ func validKeyAtRoot(k string) bool {
 	switch k {
 	case "func", "orderasc", "orderdesc", "first", "offset", "after":
 		return true
-	case "from", "to", "numpaths":
+	case "from", "to", "numpaths", "timeout":
 		// Specific to shortest path
 		return true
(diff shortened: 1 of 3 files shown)


BlankRain:
setting grpc timeout param

PR #2611, opened 2018-09-25
An error was being returned when an empty response is expected.

### LINKED ISSUE #2610: Invalid error for semantically correct query
dgraph:latest

Yes

`my_id` is indexed on `exact` in schema. Performing this mutation:
```
{
  set {
    _:test <pred_one> "one" .
    _:test <test> "" .
    _:test <my_id> "123" .
  }
}
```

Query should return no results:
```
{
  t as var(func: eq(my_id, "123")) @filter(has(test)) @cascade {
    p as pred_test
  }
  query(func: eq(my_id, "123")) @filter(uid(t) AND eq(val(p), "my test")) {
    expand(_all_)
  }
}
```
Instead, this error appears:
`: Expected a valid value map. But got empty.`

Modifying the query to be of this format returns the expected no results found:
```
{
  t as var(func: eq(my_id, "123")) @filter(has(test)) @cascade {
    p as pred_test
  }
  query(func: uid(t)) @filter(eq(my_id, "123") AND eq(val(p), "my test")) {
    expand(_all_)
  }
}
```

srfrog:
When testing inequality value vars with non-matching values, the
response was sent as an error although it should return empty result
if the query has correct syntax.

Closes #2610

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2611)

### DIFF
--- a/query/query.go
+++ b/query/query.go
@@ -1634,7 +1634,8 @@ func (sg *SubGraph) replaceVarInFunc() error {
 
 func (sg *SubGraph) ApplyIneqFunc() error {
 	if sg.Params.uidToVal == nil {
-		x.Println("Expected a valid value map. But got empty.")
+		// Expected a valid value map. But got empty.
+		// Don't return error, return empty - issue #2610
 		return nil
 	}
 	var typ types.TypeID
--- a/query/query.go
+++ b/query/query.go
@@ -1634,7 +1634,8 @@ func (sg *SubGraph) replaceVarInFunc() error {
 
 func (sg *SubGraph) ApplyIneqFunc() error {
 	if sg.Params.uidToVal == nil {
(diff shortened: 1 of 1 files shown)


query/query.go
```diff
@@ -1634,7 +1634,8 @@ func (sg *SubGraph) replaceVarInFunc() error {
 
 func (sg *SubGraph) ApplyIneqFunc() error {
 	if sg.Params.uidToVal == nil {
-		return x.Errorf("Expected a valid value map. But got empty.")
+		x.Println("Expected a valid value map. But got empty.")
```
manishrjain: You can get rid of the println here. Just return nil

manishrjain:
LGTM. One comment.

PR #2613, opened 2018-09-25
Feature/simplify tls setup docs

srfrog:
Updated documentation for new TLS setup

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2613)

### DIFF
--- a/wiki/content/deploy/index.md
+++ b/wiki/content/deploy/index.md
@@ -1196,8 +1196,7 @@ The server option `--tls_client_auth` accepts different values that change the s
 | VERIFYIFGIVEN | Client certificate is verified if provided (default) |
 | REQUIREANDVERIFY | Always require a valid certificate (most secure) |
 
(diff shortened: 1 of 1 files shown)


danielmai:
Is there a reason why `dgraph cert` uses `--dir` and the other commands use `--tls_dir`? Can they use the same flag name? It seems like they serve the same purpose.

srfrog:
> Is there a reason why `dgraph cert` uses `--dir` and the other commands use `--tls_dir`? Can they use the same flag name? It seems like they serve the same purpose.

the command used to be called 'dgraph tls' but @manishrjain wanted to call it cert instead. i guess before it looked like 'dgraph tls --dir', etc.. maybe `--cert_dir` now would make sense ?

danielmai:
OK. I'll let you decide that with @manishrjain. I'm looking forward to the flags being consistent.

This new TLS section of the docs look great, @gus. Thanks!

PR #2678, opened 2018-10-19
Better control excessive traffic to Dgraph

manishrjain:
- Proposal retries generate a lot of traffic for an already jammed system. Now we consider each proposal retry to have an exponentially increasing weight (2^retry), which is limited against `pending_proposal` flag.
- For all the retries, the pending proposals rate limiter bleeds the count back slowly over time, to help with the congestion.
- Remove `btree` concept, which was the cause of deadlocks during LRU.gets. The reason it was put in place was because Dgraph used to not write commits to disk earlier (keeping them in memory). That has changed now, with each commit going to disk before a read is done. So, we don't need to maintain this data structure. This speeds up mutations considerably.

Note that the lack of btree also means that a txn won't be able to read back its own uncommitted writes to secondary indices. I think that's a rare use case and hence a fair tradeoff, given the complexity and cost of having to overlay this structure on the DB.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2678)

### DIFF
--- a/contrib/integration/testtxn/main_test.go
+++ b/contrib/integration/testtxn/main_test.go
@@ -456,7 +456,7 @@ func TestReadIndexKeySameTxn(t *testing.T) {
 	}
 
 	txn = s.dg.NewTxn()
-	defer x.Ignore(txn.Discard(context.Background()))
+	defer txn.Discard(context.Background())
 	q := `{ me(func: le(name, "Manish")) { uid }}`
 	resp, err := txn.Query(context.Background(), q)
 	if err != nil {
--- a/systest/mutations_test.go
+++ b/systest/mutations_test.go
@@ -574,7 +574,7 @@ func LangAndSortBugTest(t *testing.T, c *dgo.Dgraph) {
 	require.NoError(t, err)
 
 	txn = c.NewTxn()
-	defer x.Ignore(txn.Discard(ctx))
+	defer txn.Discard(ctx)
 	resp, err := txn.Query(ctx, `
 	{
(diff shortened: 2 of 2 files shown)


systest/mutations_test.go
```diff
@@ -570,6 +571,10 @@ func LangAndSortBugTest(t *testing.T, c *dgo.Dgraph) {
 			_:sang <name> "Sang Hyun"@en .
 		`),
 	})
+	require.NoError(t, err)
+
+	txn = c.NewTxn()
+	defer txn.Discard(ctx)
```
golangcibot: Error return value of `txn.Discard` is not checked

systest/mutations_test.go
```diff
@@ -1560,6 +1567,8 @@ func HasDeletedEdge(t *testing.T, c *dgo.Dgraph) {
 		ids = append(ids, uid)
 	}
 
+	txn = c.NewTxn()
+	defer txn.Discard(ctx)
```
golangcibot: Error return value of `txn.Discard` is not checked

contrib/integration/testtxn/main_test.go
```diff
@@ -453,6 +455,8 @@ func TestReadIndexKeySameTxn(t *testing.T) {
 		uid = u
 	}
 
+	txn = s.dg.NewTxn()
+	defer txn.Discard(context.Background())
```
golangcibot: Error return value of `txn.Discard` is not checked

Issue #2646, opened 2018-10-10
Errors refactor and organization


srfrog:
This feature is to organize all dgraph's errors into a new Go package that will allow us to describe and inspect errors better. Currently most errors are consumed by funcs in the `x` package. We need to extract `x/error.go` into its own package (errors) and build all the features we need to track distributed issues.

Design goals:
- New error type that is stackable and has contextual info.
- Context must include file, lines, and kind of operation.
- Reflection funcs to inspect context and/or kind.
- Unwrap stacktrace to describe chained events.
- Good performance and minimal allocs. 
- (Optional) User-friendly error codes (string)
- (Optional) Pluggable logging and tracing.

Related issues:
- Export mutation errors: #2498
- Improve error messages to point at specific line and character number #2609 
- Proper catching of errors like "readTs: 60534 less than minTs: 60535 for key" #2647 
- dgraph continues answering OK after OOM #3038

codexnull:
I would also like to have some distinction for errors based on audience. Errors meant to help developers diagnose a problem with dgraph itself should be different than those reporting a runtime error.  For example, users should not be faced with a stack trace because they mistyped a filename or command line option.

jgoodall:
In the meantime, is there a list of temporary errors. My understanding is that `Transaction has been aborted` and `readTs: <num> less than minTs: <num>` should both be retried. Is that true, and are there others?

campoy:
We are now using the package github.com/pkg/errors, therefore making this obsolete.

PR #2685, opened 2018-10-20
New max_retries config flag for alpha

### LINKED ISSUE #2680: Error while applying txn status to disk

v1.0.9-rc4

Yes.

4GB VM, Ubuntu

dgraph server --bindall --lru_mb=2048 --port_offset=0 --zero zero${bn}:5080 --my=dgraph${bn}:7080 --idx 1

Database starts normally, applies transactions or discards them.
Actual: Database won't start fully because it keeps retrying the same operation many times per second:
2018/10/19 03:07:38 draft.go:446: Error while applying txn status to disk (304251 -> 304252): Key with size 72716 exceeded 65528 limit. Key:
00000000  00 00 08 63 6d 70 5f 64  61 74 61 02 02 7b 22 66  |...cmp_data..{"f|
00000010  6c 6f 6f 72 70 6c 61 6e  22 3a 7b 22 63 6f 72 6e  |loorplan":{"corn|
00000020  65 72 73 22 3a 7b 22 66  39 30 64 61 35 65 33 2d  |ers":{"f90da5e3-|
00000030  39 65 30 65 2d 65 62 61  37 2d 31 37 33 64 2d 65  |9e0e-eba7-173d-e|
...

Can't see any options to discard pending transaction.  Trying to update schema to solve this, but database is never available.  Only apparent option is to delete all data and start over which is not acceptable, and now the database server is unusable.  Checked with badger tools: no obvious way to fix things at that level.

srfrog:
The `max_retries` sets a limit to the number of commit to disk retries. This is used to prevent persistent loops when we are stuck on a unsolvable failure state. The default limit is infinite, configurable at runtime via `--max_retries`.

Closes #2680

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2685)


dgraph/cmd/alpha/run.go
```diff
@@ -103,6 +102,9 @@ they form a Raft group and provide synchronous replication.
 	flag.BoolVar(&worker.Config.ExpandEdge, "expand_edge", true,
 		"Enables the expand() feature. This is very expensive for large data loads because it"+
 			" doubles the number of mutations going on in the system.")
+	flag.IntVar(&worker.Config.MaxRetries, "max_retries", 10,
```
manishrjain: Let's set this to infinite by default.

worker/draft.go
```diff
@@ -508,7 +507,13 @@ func (n *node) commitOrAbort(pkey string, delta *pb.OracleDelta) error {
 		if txn == nil {
 			return
 		}
+		retry := Config.MaxRetries
 		for err := txn.CommitToDisk(&writer, commit); err != nil; {
+			if retry <= 0 {
+				glog.Warningf("Max commit retry limit (%d) reached, ", Config.MaxRetries)
```
manishrjain: Add a message saying unable to persist a commit to disk. Giving up now would result in a complete or partial loss of committed txn.

sdwlig:
This will be good enough to manually recover.  Perhaps when certain errors are known to be permanent, that can be passed to the loop to cause it to end even if it is set to infinite.

srfrog:
> This will be good enough to manually recover. Perhaps when certain errors are known to be permanent, that can be passed to the loop to cause it to end even if it is set to infinite.

We have an issue to improve error handling and inspection, that should help with these situations #2646

worker/draft.go
```diff
@@ -41,6 +38,10 @@ import (
 	"github.com/dgraph-io/dgraph/schema"
 	"github.com/dgraph-io/dgraph/types"
 	"github.com/dgraph-io/dgraph/x"
+
```
golangcibot: File is not `goimports`-ed

PR #2687, opened 2018-10-22
Bug fix: Send commit timestamps in order

manishrjain:
Zero had a race condition, where it could send commit timestamps out of order. This can cause invalid reads anyway. But, even more serious than that, Alpha was recently changed to ensure that it does not write over a newer version of the key (https://github.com/dgraph-io/dgraph/commit/a4bd06f2f). If commits come out of order, some of these commits can be dropped on the floor.

This change fixes that by ensuring that Zero would send all the commits in the right order.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2687)


dgraph/cmd/root.go
```diff
@@ -75,7 +75,12 @@ func init() {
 	RootCmd.PersistentFlags().Bool("expose_trace", false,
 		"Allow trace endpoint to be accessible from remote")
 	rootConf.BindPFlags(RootCmd.PersistentFlags())
+
 	flag.CommandLine.AddGoFlagSet(goflag.CommandLine)
+	// Always set stderrthreshold=0. Don't let users set it themselves.
+	x.Check(flag.Set("stderrthreshold", "0"))
+	flag.CommandLine.MarkDeprecated("stderrthreshold",
```
golangcibot: Error return value of `flag.CommandLine.MarkDeprecated` is not checked

PR #2688, opened 2018-10-23
Fix docker-compose on getting started

TheEnbyperor:
dgraph alpha does not exist, so replace with dgraph server

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2688)

### DIFF
--- a/wiki/content/get-started/index.md
+++ b/wiki/content/get-started/index.md
@@ -96,7 +96,7 @@ services:
       - 8080:8080
       - 9080:9080
     restart: on-failure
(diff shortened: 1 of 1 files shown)


srfrog:
Thanks, but we are renaming server to alpha. This change will happen on the next release, unfortunately the docs are based from master and it was merged a bit ahead.

danielmai:
@srfrog is right. For now, please use the config file from the current release (v1.0.9): https://github.com/dgraph-io/dgraph/blob/v1.0.9/contrib/config/docker/docker-compose.yml

PR #2692, opened 2018-10-24
Adding auth token support

gitlw:
Performed testing using the following steps:
1. The alpha server is started with the --auth_token option
2.a The live loader command is run _without_ any change of schema through the --schema option, then the --auth_token option is _not_ required, i.e. the rdf file can still be processed successfully without the --auth_token option
2.b The live loader command is run _with_ a schema change through the --schema option, then the --auth_token option _must_ be provided, otherwise the Alter gRPC call fails with an error complaining about missing the auth token

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2692)

### DIFF
--- a/dgraph/cmd/live/run.go
+++ b/dgraph/cmd/live/run.go
@@ -22,7 +22,6 @@ import (
 	"compress/gzip"
 	"context"
 	"fmt"
-	"google.golang.org/grpc/metadata"
 	"io"
 	"io/ioutil"
 	"math"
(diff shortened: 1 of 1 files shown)


dgraph/cmd/live/run.go
```diff
@@ -22,6 +22,7 @@ import (
 	"compress/gzip"
 	"context"
 	"fmt"
+	"google.golang.org/grpc/metadata"
```
golangcibot: File is not `goimports`-ed

PR #2719, opened 2018-11-02
Replace bp128 with a simple to understand codec

manishrjain:
We had been using an encoder, which was pulled in from an abandoned repository, with no maintenance and little testing. The code was complex and no one touched it since it was put in place.

This PR replaces that encoding scheme using the same concepts, via protocol buffers and Varint encoding - both of which are solid and reliable.

This should fix the panic we've been seeing sometimes when loading/reading data.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2719)


dgraph/cmd/bulk/reduce.go
```diff
@@ -64,15 +64,10 @@ func (r *reducer) reduce(job shuffleOutput) {
...
-		} else {
-			pl.Uids = bp128.DeltaPack(uids)
-			val, err := pl.Marshal()
-			x.Check(err)
-			txn.SetWithMeta(currentKey, val, meta)
-		}
+		pl.Pack = codec.Encode(uids, 256)
+		val, err := pl.Marshal()
+		x.Check(err)
+		txn.SetWithMeta(currentKey, val, meta)
```
golangcibot: Error return value of `txn.SetWithMeta` is not checked

PR #2730, opened 2018-11-07
Restore schema states on error, and fixing flaky test

### LINKED ISSUE #2654: Predicate can be changed from list to scalar
docker image dgraph/dgraph:latest

1. Run some local dgraph
2. Go to Ratel Console "Alter" tab and submit this string:
`someP: [int] .`
3. Go to the Schema tab and verify that an list of integers is created.
4. Go back to the "Alter" tab and submit this:
`someP: int .`
5. A server responds with an error:

> Type can't be changed from list to scalar for attr: [someP] without dropping it first.

6. Go to the Schema tab and observe that `someP` is not a list anymore.

Actual result: server says "Error. I can't do that for you" but then does it anyway.
Expected result: server owns its words.

Same bug can be reproduced via the edit dialog in the Schema tab.

This is **not** a Ratel bug because predicate's list attribute changes in the `schema {}` response. And also because I'm Paul, I maintain Ratel.

I have only tested with predicates that have no data at all.

### LINKED ISSUE #2717: The test TestSystem/delete_all_reverse_index is flaky

v1.0.10-rc1-25-g3897db23

Yes

8G, Fedora

1. Start the dgraph cluster by running `run.sh` inside $GOPATH/src/github.com/dgraph-io/dgraph/dgraph
2. Run the test multiple times with the following script and eventually the test could fail. If you don't see the failure, try to increment the number of runs and try it again.

> #!/usr/bin/env bash
> for i in $(seq 1 10); do
>    go test -v -run=TestSystem/delete_all
> done

The test should always succeed.

gitlw:
fixes #2654 fixes #2717 
Tested with test.sh

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2730)

### DIFF
--- a/worker/mutation.go
+++ b/worker/mutation.go
@@ -99,7 +99,9 @@ func runMutation(ctx context.Context, edge *pb.DirectedEdge, txn *posting.Txn) e
 func runSchemaMutation(ctx context.Context, update *pb.SchemaUpdate, startTs uint64) error {
 	if err := runSchemaMutationHelper(ctx, update, startTs); err != nil {
 		// on error, we restore the memory state to be the same as the disk
-		schema.Load(update.Predicate)
+		if loadErr := schema.Load(update.Predicate); loadErr != nil {
+			glog.Errorf("failed to load schema: %v", loadErr)
+		}
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


worker/mutation.go
```diff
@@ -97,6 +97,8 @@ func runMutation(ctx context.Context, edge *pb.DirectedEdge, txn *posting.Txn) e
 // and further mutations are blocked until this is done.
 func runSchemaMutation(ctx context.Context, update *pb.SchemaUpdate, startTs uint64) error {
 	if err := runSchemaMutationHelper(ctx, update, startTs); err != nil {
+		// on error, we restore the memory state to be the same as the disk
+		schema.Load(update.Predicate)
```
golangcibot: Error return value of `schema.Load` is not checked

PR #2736, opened 2018-11-08
Print the Dgraph version when running live or bulk loader.

danielmai:
Related to e2783d6ef1997a7af1be98b3145398b823759e54.

Would be useful to know when users are sharing their logs from the live or bulk loader.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2736)

### DIFF
--- a/dgraph/cmd/bulk/run.go
+++ b/dgraph/cmd/bulk/run.go
@@ -70,7 +70,7 @@ func init() {
 	flag.Int("shufflers", 1,
 		"Number of shufflers to run concurrently. Increasing this can improve performance, and "+
 			"must be less than or equal to the number of reduce shards.")
-	flag.Bool("version", false, "Prints the version of Dgraph.")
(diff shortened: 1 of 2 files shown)


dgraph/cmd/bulk/run.go
```diff
@@ -70,7 +70,7 @@ func init() {
 	flag.Int("shufflers", 1,
 		"Number of shufflers to run concurrently. Increasing this can improve performance, and "+
 			"must be less than or equal to the number of reduce shards.")
-	flag.Bool("version", false, "Prints the version of dgraph-bulk-loader.")
+	flag.Bool("version", false, "Prints the version of Dgraph.")
```
manishrjain: Dgraph Bulk Loader.

If we just say Dgraph, a user could make it mean Dgraph Alpha or Zero version, which might be different from the bulk loader.

danielmai: Isn't the version all the same across all the subcommands, though?

PR #2754, opened 2018-11-16
Mismatched tests on map values produce incorrect logic.

### LINKED ISSUE #2733: Invalid division by zero
1.0.10

It worked with 1.0.9 but bugs with 1.0.10

Linux Debian Testing, 16GoDDR

In a query, I do a division and protect potential 0 value for the divisor with a 'max' function.
The error only happens with value variables, it works with numeric values.
I also tried to protect the division with a 'cond' without success.

Here is my query:

```
{
var(func: uid("0xea61")) {
	ct_thinks as count(thinks)
	knows {
		ct_knows as math(1)
	}
	sum_knows as sum(val(ct_knows))
}
me(func: uid("0xea61")) {
	avg: math(ct_thinks / max(1, sum_knows))
}
}
```

If sum_knows equals to 0 then avg should be 0.
If I replace ct_thinks with a literal value, then no more bug.

srfrog:
Several logic tests for map variables are assuming that the value is nil. This is partially correct, except empty initialized maps (with `make()`) are never nil. This leads to erroneous logic at different states. The best way to test for unassigned maps is to use `len()`.

This PR attempts to find any and all cases which map values were incorrectly tested using `nil` comparison and change those to `len()`.

e.g.,
```go
var m1 map[int]bool
m2 := make(map[int]bool)
if m1 == nil { /* true */ }
if m2 == nil { /* false */ }
if len(m1) == 0 { /* true */ }
if len(m2) == 0 { /* true */ }
```

Closes #2733

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2754)


srfrog:
@manishrjain there's one thing that changed - https://github.com/dgraph-io/dgraph/pull/2754/files#diff-8a2cc8263889287b334d1e7b4364decf

That test checked for an unmatched uid, previously that would return empty result (as old test show), now it returns an error **as intended.**

PR #2800, opened 2018-11-30
Simplify design and make tablet moves robust

manishrjain:
This PR changes the design of the tablet moves to this:

- Zero leader blocks any commit on the predicate to be moved.
- Gets a new timestamp, which the source group leader must wait for.
- Source group leader streams data to the destination group leader.
- Once done, Zero would propose the predicate be served by destination group.
- Zero defer unblocks the commits on the predicate.

There's no need to propose read-only among Zeros, and propose states among Alphas, etc. Instead of trying to block mutations on Alphas, we just block Commits on the Zero leader. If the leader changes, the move would be aborted, and the source group would continue as nothing happened. Destination group would then clean up the extra data during its cleanup cycle.

Tested this with increment tool with tons of tablet moves, and we don't lose any commit during this. Will test with Jepsen.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2800)


worker/predicate_move.go
```diff
@@ -188,14 +127,27 @@ func batchAndProposeKeyValues(ctx context.Context, kvs chan *pb.KVS) error {
...
 // Returns count which can be used to verify whether we have moved all keys
 // for a predicate or not.
 func (w *grpcWorker) ReceivePredicate(stream pb.Worker_ReceivePredicateServer) error {
+	if !groups().Node.AmLeader() {
+		return x.Errorf("ReceivePredicate failed: Not the leader of group")
+	}
+	// No new deletion/background cleanup would start after we start streaming tablet,
+	// so all the proposals for a particular tablet would atmost wait for deletion of
+	// single tablet. Only leader needs to do this.
+	mu := groups().blockDeletes
```
golangcibot: assignment copies lock value to mu: sync.Mutex (from `govet`)

dgraph/cmd/zero/http.go
```diff
@@ -140,6 +140,12 @@ func (st *state) moveTablet(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if !st.node.AmLeader() {
+		w.WriteHeader(http.StatusBadRequest)
+		x.SetStatus(w, x.ErrorInvalidRequest, "This Zero server is not the leader. Re-run command on leader.")
```
golangcibot: line is 104 characters (from `lll`)

worker/groups.go
```diff
@@ -759,83 +754,68 @@ OUTER:
...
 
-func (g *groupi) waitForBackgroundDeletion() {
-	// Waits for background cleanup if any to finish.
-	// No new cleanup on any predicate would start until we finish moving
-	// the predicate because read only flag would be set by now. We start deletion
-	// only when no predicate is being moved.
-	g.delPred <- struct{}{}
-	<-g.delPred
-}
+func (g *groupi) cleanupTablets() {
```
golangcibot: func `(*groupi).cleanupTablets` is unused (from `unused`)

PR #2804, opened 2018-12-03
Add Dart client to README and clients page.

katutz:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2804)


danielmai:
Congrats on releasing this new Dart client for Dgraph, @katutz. Have you thought about giving the client library a different name other than plain "dgraph"?

Thanks for your contribution!

katutz:
> Congrats on releasing this new Dart client for Dgraph, @katutz. Have you thought about giving the client library a different name other than plain "dgraph"?
> 
> Thanks for your contribution!

Thanks Daniel Mai! I thought in using some other name like ddgraph or dgraph_dart. But I was in doubt, because I saw some clients in Dart with the same name as theirs databases and others with a different name. Then I initially used the same name to reserve the dgraph package name in the Dart Pub (Dart package repository). Would it be interesting to switch to one of these names I thought or some other?

PR #2822, opened 2018-12-14
Go modules support.

srfrog:
This PR adds **go.mod** and ~locks~ tracks dependencies with **go.sum**. Effectively removing the need for the vendor dir.

The first time a build is ran go will fetch and cache all dependencies in `$GOPATH/pkg/mod`, from then on, that cache is updated.

> Go modules requires version 1.11.x or better.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2822)


KarthikNayak:
@srfrog why was this dropped?

alexkreidler:
Hey, any progress on this? It's a real blocker for anyone wanting to build anything cool and modern with Dgraph.

Thanks!

PR #2850, opened 2018-12-28
fixed bulk recognize has assigned uid bug

relunctance:
rdf eg: 
```
<0x4c4eab0>     <md5>   "a11aa8ed3695e504517ee1c4ea6f22c4"      .
```

live load work well , but bulk  can't recognize that has been assigned uid "0x4c4eab0" .

### DIFF
--- a/dgraph/cmd/bulk/mapper.go
+++ b/dgraph/cmd/bulk/mapper.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


manishrjain:
@codexnull Can you please review this? Two things to check for:

1. Does Live load do this?
2. The UID that's been handed, we need to make sure that bulk loader gets an appropriate lease from Zero to include that UID.

relunctance:
Thanks .
Your analysis is very good . many problems can be avoided , It's safe to close it.
But It can use third parties to maintain assigned UID , In this way, The user knows what uid is available.
Maybe it only applies to new clusters, not existing ones to expand group.

PR #2851, opened 2018-12-28
Refactor rebuild index logic.

martinmr:
* Delete the index.go file in the working module. The callers of the
  deleted methods will call the functions in the posting module
  directly.
* The rebuild index functions delete the index first.
* The rebuild index functions figure out themselves if the index needs
  to be rebuilt after the deletion, instead of delegating that task to
  the callers.
* Simplify delete logic.

No fuctional changes.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2851)


manishrjain:
`!current.Count == !old.Count`
This is confusing. No different from: `current.Count == old.Count`. Can you fix?

PR #2883, opened 2019-01-09
Refactor reindexing code to only reindex specific tokenizers.

martinmr:
Currently during reindexing, the whole index is being deleted and
rebuilt, even if there are tokenizers that did not change between
schemas. This change allows reindexing to happen only for tokenizers
that need it.

### DIFF
--- a/posting/index.go
+++ b/posting/index.go
@@ -38,8 +38,10 @@ import (
 
 var emptyCountParams countParams
 
-// IndexTokens return tokens, without the predicate prefix and index rune.
-func indexTokens(attr, lang string, src types.Val) ([]string, error) {
+// indexTokensforTokenizers return tokens, without the predicate prefix and
+// index rune, for specific tokenizers.
+func indexTokensForTokenizers(attr, lang string, tokenizers []tok.Tokenizer,
+	src types.Val) ([]string, error) {
 	schemaType, err := schema.State().TypeOf(attr)
 	if err != nil || !schemaType.IsScalar() {
 		return nil, x.Errorf("Cannot index attribute %s of type object.", attr)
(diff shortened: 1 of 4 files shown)


posting/index.go
```diff
@@ -70,15 +72,28 @@ func indexTokens(attr, lang string, src types.Val) ([]string, error) {
...
 
-// addIndexMutations adds mutation(s) for a single term, to maintain index.
-// t represents the original uid -> value edge.
+// indexTokens return tokens, without the predicate prefix and index rune, for
+// all tokenizers. This function is only being used in the tests as the code now
+// calls indexTokensForTokenizers with the complete list of tokenizers in cases
+// where the whole index needs to be rebuilt. This function is being kept for
+// convenience as manually getting the tokenizers in every test that need them
+// would add a lot of extra code.
+func indexTokens(attr, lang string, src types.Val) ([]string, error) {
```
golangcibot: func `indexTokens` is unused (from `unused`)

gitlw: It seems this is only being inside the index_test.go file, can we move this function into that test file?

posting/index.go
```diff
@@ -100,6 +115,15 @@ func (txn *Txn) addIndexMutations(ctx context.Context, t *pb.DirectedEdge, p typ
 	return nil
 }
 
+// addIndexMutations adds mutation(s) for a single term, to maintain index,
+// for all tokenizers.
+func (txn *Txn) addIndexMutations(ctx context.Context, t *pb.DirectedEdge, p types.Val,
```
gitlw: To better distinguish this method from the following one, maybe it can be renamed to addIndexMutationsForAllTokenizers?

posting/index.go
```diff
@@ -408,12 +432,27 @@ func deleteAllEntries(prefix []byte) error {
 	})
 }
 
+// deleteIndex deletes the index for the given attribute. All tokenizers are
+// deleted by this function.
 func deleteIndex(attr string) error {
```
gitlw: rename to deleteIndexForAllTokenizers?

posting/index.go
```diff
@@ -595,33 +650,74 @@ func needsIndexRebuild(old *pb.SchemaUpdate, current *pb.SchemaUpdate) indexOp {
...
+		return indexRebuildInfo{
+			op: indexNoop,
+		}
+	}
+
+	// Else, the tokenizers from the old schema that do not exist in the schema
+	// need to be deleted and the tokenizers from the current schema that did
+	// not exist in the old one need to be built.
+	tokenizersToDelete := make([]string, 0)
+	for token := range prevTokens {
```
gitlw: If we change the prevTokens and currTokens from map[string]bool to map[string]struct{} (since they are essentially sets), then we can reuse the x.Diff method to get the differences.

PR #2885, opened 2019-01-09
Gitlw/fix json facet

### LINKED ISSUE #2867: anyofterms does not work for facet created by JSON mutation

tried latest "latest" @ dockerhub and v1.0.12-rc1 

Yes

16gb, linux

Create any facet with a string value.

Query filter it with anyofterms

Expected: i can filter with anyofterms, so e.g. if the value of the facet is "READ" and i filter for anyofterms(permission, "READ WRITE"), i will get a result

Actual:
well, anyofterms does not work. the value is not returned. i can only use eq.
**BUT! This is only the case if i performed the mutation for this facet with the JSON API (from dgo). If i do it with the RDF syntax in ratel, it works!**

gitlw:
fixes #2867 
screenshot from 2019-01-09 10-57-02

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2885)


rdf/parse.go
```diff
@@ -275,17 +275,17 @@ func isNewline(r rune) bool {
...
-	"xs:date":            types.DateTimeID,
-	"xs:dateTime":        types.DateTimeID,
-	"xs:int":             types.IntID,
-	"xs:positiveInteger": types.IntID,
-	"xs:boolean":         types.BoolID,
-	"xs:double":          types.FloatID,
-	"xs:float":           types.FloatID,
-	"xs:base64Binary":    types.BinaryID,
-	"geo:geojson":        types.GeoID,
+	"xs:password":                                      types.PasswordID,
```
golangcibot: File is not `goimports`-ed (from `goimports`)

edgraph/nquads_from_json.go
```diff
@@ -28,7 +28,7 @@ import (
 	"github.com/dgraph-io/dgraph/types"
 	"github.com/dgraph-io/dgraph/types/facets"
 	"github.com/dgraph-io/dgraph/x"
-	geom "github.com/twpayne/go-geom"
+	"github.com/twpayne/go-geom"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

PR #2892, opened 2019-01-14
Regenerate protobuf with latest protoc and protoc-gen-gofast.

danielmai:
```
$ protoc --version
libprotoc 3.6.1
```

* Updated release.sh script to regenerate protos and stop if they're different from what's checked in the release tag.
* Regenerate proto for v1.0.12-rc3.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2892)


danielmai:
Use HTTPS git repo links so they don't require an SSH key to pull. The Ratel repo is public now, so the HTTPS link does not require authentication to pull. This will enable the release.sh script to be run in CI without authentication or require less set up for anyone to generate release binaries.

contrib/release.sh
```diff
@@ -51,6 +51,9 @@ go get -d google.golang.org/grpc
 go get -u github.com/prometheus/client_golang/prometheus
 go get -u github.com/dgraph-io/dgo
 # go get github.com/stretchr/testify/require
+go get -u github.com/dgraph-io/badger
```
manishrjain: Technically this isn't needed because we vendor badger in.

PR #2893, opened 2019-01-14
Fix int/float conversion to bool

srfrog:
This PR changes the logic in `Convert` to set int/float zero values as `false` when converting to bool. All other values will be `true`. Before this change, an int/float value of one (1) would convert to `false` bool.

Old behavior:
```
int: convert 1 to bool -> false
int: convert != 1 to bool -> true

float: convert 1 to bool -> false
float: convert != 1 to bool -> true

string: convert "" to bool -> error
```

New behavior:
```
int: convert 0 to bool -> false
int: convert != 0 to bool -> true

float: convert 0 to bool -> false
float: convert != 0 to bool -> true

string: convert "" to bool -> false
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2893)


manishrjain:
This is a breaking change, which would affect everyone using bools.

Can you add tests for this change?

PR #2902, opened 2019-01-16
Testing shortest path

### LINKED ISSUE #2520: k-shortest-path with uid `0x0`
# k-shortest-path with uid `0x0`
The version I am using is `v1.0.6`.
This is the latest release on windows, so I could not try it out on `v1.0.7`

I am using the `k-shortest-path` algorithm and allow arbitrary `from` and `to` uids. 

## Expected behaviour and actual result.

I would expect DGraph to return an empty result if no path can be found, as it normally happens when there is no path between two nodes.
However, since `0x0` seems to be kind of special, I would rather expect an error or something similar.
Something like that:
```json
{
  "data": {
    "path": []
  },
  "extensions": {
    "server_latency": {
      "encoding_ns": 3989100
    },
    "txn": {
      "start_ts": 5245,
      "lin_read": {
        "ids": {
          "1": 10706
        }
      }
    }
  }
}
```

### Actual result

I actually receive a path containing two nodes: `0x0` and the goal uid, for example:

```json
{
  "data": {
    "path": [
      {
        "nodeReference": "0x0"
      },
      {
        "nodeReference": "0x1b0c"
      }
    ],
    "_path_": [
      {
        "uid": "0x0"
      }
    ]
  },
  "extensions": {
    "server_latency": {
      "encoding_ns": 999300
    },
    "txn": {
      "start_ts": 5244,
      "lin_read": {
        "ids": {
          "1": 10706
        }
      }
    }
  }
}
```

gitlw:
fixes #2520

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2902)

### DIFF
--- a/systest/queries_test.go
+++ b/systest/queries_test.go
@@ -406,4 +406,30 @@ func ShortestPathWithUidZero(t *testing.T, c *dgo.Dgraph) {
 	txn := c.NewTxn()
 	defer txn.Discard(ctx)
 
+	require.NoError(t, c.Alter(ctx, &api.Operation{
+		Schema: `
+      name: string @index(exact) .
+      friend: uid .
+    `,
+	}))
+	assigned, err := txn.Mutate(ctx, &api.Mutation{
+		SetNquads: []byte(`
+_:a <name> "alice" .
+_:b <name> "bob" .
+_:c <name> "charlie" .
+_:a <friend> _:b .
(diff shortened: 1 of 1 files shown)


gitlw:
Closing after investigation and realizing that 0 is not a meaningful uid, as all uids should start from 1.

PR #2910, opened 2019-01-17
Allow querying all lang values of a predicate.

### LINKED ISSUE #2141: Return all languages for a string
I believe this is not yet supported - could you add a feature so that I could request a string predicate to return all the `@lang` it has.

So far, we can query for particular languages using the `@en:.` or `@.` or else. But the idea here is to return all existing languages for this predicate.

An example of query syntax could be:
```
{
    query(func: eq(topic.label@en, "Graph database")) {
        topic.label@_all_
    }
}
```

The returned data could either be flattened like it is now, or with an array of paired `"lang": "value"` :
```
{
    "query": {
        "topic.label@en": "Graph database",
        "topic.label@fr": "Base de donnee graphique",
        "topic.label@pl": "Blah blah blah"
    }
}

{
    "query": {
        "topic.label": [
            "en": "Graph database",
            "fr": "Base de donnee graphique",
            "pl": "Blah blah blah"
        ]
    }
}
```

martinmr:
This PR introduces the syntax pred@* which returns all the values of a
predicate which contain a language tag.

Fixes #2141

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2910)

### DIFF
--- a/gql/parser.go
+++ b/gql/parser.go
@@ -1561,6 +1561,10 @@ L:
 				function.Attr = val
 				attrItemsAgo = 0
 			} else if expectLang {
+				if val == "*" {
+					return nil, x.Errorf(
+						"The * symbol cannot be used as a valid language inside functions")
+				}
 				function.Lang = val
 				expectLang = false
 			} else if function.Name != uid {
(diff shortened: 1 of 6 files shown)


martinmr:
Ready for another round of review.

PR #2912, opened 2019-01-17
Allow min and max functions to take any number of arguments

martinmr:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2912)


martinmr:
Something is wrong with the parser. Specifically, it seems like I am popping everything from the valueStack when I see min or max but that is not what should happen. Not sure if there's a way to fix this since it seems like the current code depends heavily on knowing how many arguments were passed to the function.

PR #2920, opened 2019-01-21
Documentation for <predicate>@* syntax.

### LINKED ISSUE #2141: Return all languages for a string

martinmr:
Fixes #2141 

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2920)

### DIFF
--- a/wiki/content/query-language/index.md
+++ b/wiki/content/query-language/index.md
@@ -185,7 +185,7 @@ For example:
 - `name@*` => Look for all the values of this predicate and return them along with their language. For example, if there are two values with languages en and hi, this query will return two keys named "name@en" and "name@hi".
 
 
-{{% notice "note" %}}In functions, language lists are not allowed. Single language, `.` notation and attribute name without language tag works as described above. The @* syntax is also not allowed. {{% /notice %}}
+{{% notice "note" %}}In functions, language lists (including the * notation) are not allowed. Single language, `.` notation and attribute name without language tag works as described above. {{% /notice %}}
 
 {{% notice "note" %}}In case of full text search functions (`alloftext`, `anyoftext`), when no language is specified, default (English) Full Text Search tokenizer is used.{{% /notice %}}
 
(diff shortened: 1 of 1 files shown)


danielmai:
___
*[wiki/content/query-language/index.md, line 188 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/2920#-LWn5e8-5p7UF9tpOQfT:-LWn5e8-5p7UF9tpOQfU:b-mpn3hx) ([raw file](https://github.com/dgraph-io/dgraph/blob/41283db048fc3431cc168a1adf8f1839df3c07db/wiki/content/query-language/index.md#L188)):*
> ```Markdown
> 
> 
> {{% notice "note" %}}In functions, language lists are not allowed. Single language, `.` notation and attribute name without language tag works as described above. The @* syntax is also not allowed. {{% /notice %}}
> ```

Rephrase to "In functions, language lists (including `*` notation) ..."

It should be clear that @* is not allowed since @* is categorized as a language list feature (above, it's documented with "where the language list is set to *"). If it isn't clear, then this suggested revision buckets what's not allowed into a single sentence instead of a note structured as "Not Allowed. Allowed. Not Allowed."

PR #2928, opened 2019-01-23
Zero server shutdown endpoint

### LINKED ISSUE #2285: Feature request: Add local accessible shutdown REST endpoint to zero nodes
On Windows it is not possible to use Signals like SIGINT to stop processes. Also the golang documentation features this in different documents regarding Signal handling.

A big benefit for starting up the zero instance as a background service would be to add a /admin/shutdown endpoint that is locally accessible like the one for the server nodes.

1.0.4 - Windows

srfrog:
This PR adds a new HTTP endpoint at **/shutdown** to handle graceful service shutdowns. It also changes the process signal handling to work together with this command.

Example:
```sh
$ curl localhost:6080/shutdown
Server is shutting down...
# Done.
```

Closes #2285 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2928)

### DIFF
--- a/dgraph/cmd/zero/http.go
+++ b/dgraph/cmd/zero/http.go
@@ -22,9 +22,9 @@ import (
 	"net"
 	"net/http"
 	"strconv"
-	"sync"
 	"time"
 
+	"github.com/dgraph-io/badger/y"
 	"github.com/dgraph-io/dgraph/protos/pb"
 	"github.com/dgraph-io/dgraph/x"
 	"github.com/gogo/protobuf/jsonpb"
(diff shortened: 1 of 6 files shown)


dgraph/cmd/zero/http.go
```diff
@@ -224,6 +224,21 @@ func (st *state) getState(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	if r.Method != http.MethodGet {
+		w.WriteHeader(http.StatusBadRequest)
+		x.SetStatus(w, x.ErrorInvalidMethod, "Invalid method")
+		return
+	}
+
+	st.zero.shutDownCh <- struct{}{}
+	w.Write([]byte("Server is shutting down...\n"))
```
golangcibot: Error return value of `w.Write` is not checked (from `errcheck`)

srfrog:
> There's no authentication, right? So even if the dgraph server was started by root, it can still be shut down by any user?

That's correct. We need to add some access control to Zero.

srfrog:
I've created an issue specifically for zero access control - https://github.com/dgraph-io/dgraph/issues/2930
It could potentially ask an alpha for access or check the JWT itself. Or maybe even a simple HTTP digest.

PR #2967, opened 2019-02-01
Prevent dropping or altering of reserved predicates.

martinmr:
There is already logic to prevent certain internal predicates (ACL and
type preds) from being dropped by a DropAll request. This change adds
similar logic to prevent DropAttr and Schema requests from modifying or
dropping reserved predicates.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2967)


edgraph/server.go
```diff
@@ -283,6 +283,25 @@ func (s *Server) Alter(ctx context.Context, op *api.Operation) (*api.Payload, er
 		return empty, err
 	}
 
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)

martinmr:
I didn't realized that the test was broken after I addressed the comments in the previous review. I have fixed that and this is ready for another round of review.

PR #2985, opened 2019-02-07
Remove hacky test fix in ChecksumsMatch.

martinmr:
All tests in the query package have been modified to use the cluster
instead of the old ad-hoc setup so this fix is no longer needed.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2985)


manishrjain:
LGTM.

PR #2987, opened 2019-02-07
Optimize mutation and delta application

manishrjain:
Mutation and Txn Commit/Abort application happen in a serial order (among other less frequent operations). So, the latency of each operation has a cascading effect on the overall latency of transaction.

This PR optimizes mutation application by dividing up the task of each mutation proposal among goroutines. For a batch of 2000 NQuads, this would be divided up among 4 goroutines.

For commits, this PR makes 2 changes. It removes relying upon a Badger lookup to determine if the commit is being done at a lower version than the latest version of the key, instead picking this information up from the posting list. Secondly, it switches how txns are run to create fewer larger transactions instead of one txn per key-value.

With this PR, mutate latency goes from original 13.5ms to 7.9ms (>40% drop). Commit latency goes from 39ms to 19.7ms (50% drop). Either of these changes shouldn't have any side-effects and should be NOOPs in terms of functionality.

To make these measurements, this PR also adds metrics around each of these functions.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2987)


worker/draft.go
```diff
@@ -397,6 +434,17 @@ func (n *node) processApplyCh() {
...
+				var tags []tag.Mutator
+				switch {
+				case proposal.Mutations != nil:
+					tags = append(tags, tag.Upsert(x.KeyMethod, "apply.Mutations"))
+				case proposal.Delta != nil:
+					tags = append(tags, tag.Upsert(x.KeyMethod, "apply.Delta"))
+				}
+				// tags = append(tags, tag.Upsert(x.KeyError, perr.Error()))
+				ms := x.SinceInMilliseconds(start)
+				ostats.RecordWithTags(context.Background(), tags, x.LatencyMs.M(ms))
```
golangcibot: Error return value of `ostats.RecordWithTags` is not checked (from `errcheck`)

manishrjain:
No need to check that error.

PR #2995, opened 2019-02-08
Default value should not be nil

srfrog:
`types.Marshal()` expects vars to have non-nil values, this change
set empty string as the default value for a referenced empty
var. This should prevent nil-reference panic in `types.Marshal()`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2995)


martinmr:
This is breaking one of the tess.

srfrog:
Yes, I'm still working on that. Thanks

srfrog:
I can make that change in `types.Marshal()`, but the bug is an edge-case of the default value vars. I don't think returning error will be a good thing there.

srfrog:
Much better solution now in place.

PR #3007, opened 2019-02-12
Fix flaky delete scalar test.

martinmr:
I noticed TestDeleteScalarValue was flaky (was failing twice out of 100
times). I added logic to verify value can be queried before performing
any deletion. Ran the test 200 times and it didn't fail once.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3007)

### DIFF
--- a/dgraph/cmd/alpha/run_test.go
+++ b/dgraph/cmd/alpha/run_test.go
@@ -37,6 +37,7 @@ import (
 	"github.com/dgraph-io/dgraph/query"
 	"github.com/dgraph-io/dgraph/schema"
 	"github.com/dgraph-io/dgraph/worker"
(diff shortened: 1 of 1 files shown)


gitlw:
Before putting this patch, can we have a better understanding about why the failure can happen?

martinmr:
Not a lot of clues given that the error doe not happen very often. The test was failing at the step at which the data is queried after the first deletion (the deletion should not have any effect because the value of the object in the triple does not match the existing value).

My guess is that there's some timing issue which causes the query to be performed against an earlier version of the data. Querying multiple times before performing the deletion ensures that the tests sees the correct data before performing operations on it.

PR #3008, opened 2019-02-12
Allow multiple values for binary operators

### LINKED ISSUE #1050: min, max: Support more than two arguments
Version: 0.7.7

---

When aggregating using `min` or `max`, as a user I would like to do the following:

    max(v1, v2, v3)

rather than

    max(max(v1, v2), v3)

Therefore it might be useful if `min` and `max` supported more than two arguments.

Use case: https://gist.github.com/sungwoncho/ddba3a287ad87288331c493c2de5a5cb#file-get_active_answers

-------

Dataset:

[dgraph-1-2017-07-10-18-51.rdf.gz](https://github.com/dgraph-io/dgraph/files/1134923/dgraph-1-2017-07-10-18-51.rdf.gz)
[dgraph-schema-1-2017-07-10-18-51.rdf.gz](https://github.com/dgraph-io/dgraph/files/1134924/dgraph-schema-1-2017-07-10-18-51.rdf.gz)

Query:

```
# You might need to change the value of id
{
  var (id: 0xd18499a5ac49c3a6) {
    Has.Answer {
      answerTs as Timestamp
      ~Post {
        historyTs as Timestamp
      }
      Comment {
        commentTs as Timestamp
      }
      commentTsMax as max(var(commentTs))
      historyTsMax as max(var(historyTs))
      lastActive as math(max(max(answerTs, commentTsMax), historyTsMax))
      # The following line should work, if replaced with the above line
      # lastActive as math(max(answerTs, commentTsMax, historyTsMax))
    }
  }

  question(id: 0xd18499a5ac49c3a6) {
    Has.Answer(orderdesc: var(lastActive)) {
      Body {
        Text
      }
    }
  }
}
```

srfrog:
This PR allows using multiple values in binary operations: `max()`, `min()`.
It works by pre-evaluating the stack when 2 values already exist for the same operator.
This usage does not extend to `pow()` or `logbase()`.

Example query:
```
{
  var(func:uid(0x1)) {
    x as math(max(max(max(3,5),20),1))
    y as math(max(1,max(20,max(3,5))))
    z as math(max(1,20,3,5))
  }
  q(func:uid(0x1)) {
    val(x)
    val(y)
    val(z)
  }
}
```
Result:
```json
{
  "data": {
    "q": [
      {
        "val(x)": 20.000000,
        "val(y)": 20.000000,
        "val(z)": 20.000000
      }
    ]
  }
}
```
~~No breaking changes expected.~~ The math eval logic has changed when min/max binary operators are used.

Closes #1050

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3008)

### DIFF
--- a/gql/math.go
+++ b/gql/math.go
@@ -53,6 +53,15 @@ func (s *mathTreeStack) peek() *MathTree {
 	return s.a[len(s.a)-1]
 }
 
+func (s *mathTreeStack) reverse() *mathTreeStack {
+	l := len(s.a)
+	for i := l/2 - 1; i >= 0; i-- {
+		opp := l - 1 - i
+		s.a[i], s.a[opp] = s.a[opp], s.a[i]
+	}
+	return s
+}
+
 type MathTree struct {
 	Fn    string
 	Var   string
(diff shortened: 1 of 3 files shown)


gitlw:
Thanks for addressing this issue for the min and max functions.
However I feel we should report an error if more than 2 values are given to the logbase or pow function, unless you can find justifications that such operations make sense (I've never seen that).

srfrog:
The `pow()` function is treated as the `**` operator. e.g., `2 ** 2 ** 2` is `pow(2,2,2)`, we don't have the `**` operator in Dgraph.

I agree `logbase()` was a bit of a stretch, but it was already done when `pow()` worked. But someone working with spectral analysis could benefit from this since everything is done in a logarithmic scale.

gql/math.go
```diff
@@ -70,6 +79,10 @@ func isBinaryMath(f string) bool {
 	return f == "*" || f == "+" || f == "-" || f == "/" || f == "%"
 }
 
+func isBinary(f string) bool {
```
golangcibot: U1000: func `isBinary` is unused (from `unused`)

manishrjain:
@martinmr @gitlw can you review this change?

PR #3018, opened 2019-02-14
Store and query types.

martinmr:
This change adds the ability to store the types and query them similarly
to how the schema is currently being parsed.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3018)


martinmr:
Did more testing. Snapshots work as expected (was able to stop replicas and the entire cluster and still get the types back)

Issue #3023, opened 2019-02-15
 Dgraph alpha crashes when setting the parameter "--my" to the IPV6 address. 


chuerxiao:
If you suspect this could be a bug, follow the template.

- What version of Dgraph are you using?
v1.0.11

- Have you tried reproducing the issue with latest release?
v1.0.11 is the last release

- What is the hardware spec (RAM, OS)?
    * Distributor ID:	Ubuntu
      Description:	Ubuntu 16.04.5 LTS
      Release:	16.04
      Codename:	xenial
      memory 256GB
      cores 80

- Steps to reproduce the issue (command/config used to run Dgraph).
./dgraph alpha --zero [2001:ddb::64]:1139 --lru_mb 2048 -o=200 --my=[2001:ddb::64]:1142

- Expected behaviour and actual result.
```
badger2019/02/15 17:32:17 INFO: Replaying file id: 0 at offset: 0

badger2019/02/15 17:32:17 INFO: Replay took: 17.162µs

2019/02/15 17:32:17 [2001:ddb::64]:1142 is not valid address

github.com/dgraph-io/dgraph/x.AssertTruef

        /home/chujunlin/dgraph/src/github.com/dgraph-io/dgraph/x/error.go:79

github.com/dgraph-io/dgraph/worker.StartRaftNodes

        /home/chujunlin/dgraph/src/github.com/dgraph-io/dgraph/worker/groups.go:75

runtime.goexit

        /usr/local/go/src/runtime/asm_amd64.s:2361
```

chuerxiao:
https://github.com/dgraph-io/dgraph/blob/master/x/x.go

change
```
	if err := net.ParseIP(host); err == nil {
		return true
	}
```
to
```
	if ip := net.ParseIP(host); ip!= nil {
		return true
	}
```

jarifibrahim:
This should be fixed by https://github.com/dgraph-io/dgraph/pull/3027 (I couldn't assign the issue to myself)

PR #3027, opened 2019-02-17
ValidateAddress should return true if IPv6 is valid

jarifibrahim:
- Add tests for ValidateAddress function

Fixes https://github.com/dgraph-io/dgraph/issues/3023

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3027)

### DIFF
--- a/x/x.go
+++ b/x/x.go
@@ -279,7 +279,7 @@ func ValidateAddress(addr string) bool {
 	if p, err := strconv.Atoi(port); err != nil || p <= 0 || p >= 65536 {
 		return false
 	}
-	if err := net.ParseIP(host); err == nil {
+	if ip := net.ParseIP(host); ip != nil {
 		return true
 	}
 	// try to parse as hostname as per hostname RFC
--- a/x/x_test.go
+++ b/x/x_test.go
@@ -52,3 +52,45 @@ func TestDivideAndRule(t *testing.T) {
 
 	test(1755, 4, 439)
 }
+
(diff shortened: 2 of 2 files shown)


x/x.go
```diff
@@ -279,7 +279,7 @@ func ValidateAddress(addr string) bool {
 	if p, err := strconv.Atoi(port); err != nil || p <= 0 || p >= 65536 {
 		return false
 	}
-	if err := net.ParseIP(host); err == nil {
+	if ip := net.ParseIP(host); ip != nil {
```
jarifibrahim: ParseIP returns nil if the IP address is invalid. See https://golang.org/pkg/net/#ParseIP

x/x_test.go
```diff
@@ -52,3 +52,45 @@ func TestDivideAndRule(t *testing.T) {
...
+			name    string
+			address string
+			isValid bool
+		}{
+			{"Valid without port", "190.0.0.1", false},
+			{"Valid with port", "192.5.32.1:333", true},
+			{"Invalid without port", "12.0.0", false},
+			// the following test returns true because 12.0.0 is considered as valid
+			// hostname
+			{"Invalid with port", "12.0.0:3333", true},
```
jarifibrahim: I am not sure if this test should return true. Is this the intended behaviour?
A similar test on IPv6 returns false (see https://github.com/dgraph-io/dgraph/pull/3027/files#diff-12d9d50e4e8f838a59b4ad0e7ad05251R87)

Issue #3026, opened 2019-02-17
Smarter handling of out directory by bulk loader


codexnull:
I have more than once blown away data I did not mean to with the --out option of the bulk loader because it silently deletes and re-creates the directory. It's fortunate I have not yet tried to put the bulk output in their home directory by using --out=$HOME.

The bulk command should:

1) require the --out directory not exist or to be empty, and
2) require an additional option (e.g. --replace-out) to make it explicit the the directory is being replaced, not added to.

jarifibrahim:
Hey @codexnull :wave: , this seems like a simple change. Do you mind if I take this up?

codexnull:
Thanks for offering, but it was indeed a simple change and I had already started working on it. I actually ended up doing something even simpler: deleting just the out/<n> subdirectories (where <n> is the shard number) rather than the whole out/ directory.

codexnull:
Addressed in https://github.com/dgraph-io/dgraph/pull/3089.

PR #3034, opened 2019-02-18
Issue 3026: Don't blindly replace the bulk out directory.

codexnull:
Put the bulk shard directories in the out directory without destroying other data already there, if any.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3034)

### DIFF
--- a/dgraph/cmd/bulk/run.go
+++ b/dgraph/cmd/bulk/run.go
@@ -155,10 +155,11 @@ func run() {
 		log.Fatal(http.ListenAndServe(opt.HttpAddr, nil))
 	}()
 
-	// Delete and recreate the output dirs to ensure they are empty.
-	x.Check(os.RemoveAll(opt.DgraphsDir))
+	// Delete and recreate the shard output dirs to ensure they are empty without destroying
+	// existing non-shard data, if any.
 	for i := 0; i < opt.ReduceShards; i++ {
 		dir := filepath.Join(opt.DgraphsDir, strconv.Itoa(i), "p")
+		x.Check(os.RemoveAll(dir))
 		x.Check(os.MkdirAll(dir, 0700))
 		opt.shardOutputDirs = append(opt.shardOutputDirs, dir)
 	}


dgraph/cmd/bulk/run.go
```diff
@@ -155,10 +155,11 @@ func run() {
 		log.Fatal(http.ListenAndServe(opt.HttpAddr, nil))
 	}()
 
-	// Delete and recreate the output dirs to ensure they are empty.
-	x.Check(os.RemoveAll(opt.DgraphsDir))
+	// Delete and recreate the shard output dirs to ensure they are empty without destroying
+	// existing non-shard data, if any.
 	for i := 0; i < opt.ReduceShards; i++ {
 		dir := filepath.Join(opt.DgraphsDir, strconv.Itoa(i), "p")
+		x.Check(os.RemoveAll(dir))
```
jarifibrahim: @codexnull This would remove the directory if it already exists. I don't think this is the behaviour we want.

Running 
```
./dgraph bulk -f ./goldendata.rdf.gz -s goldendata.schema --map_shards=4 --reduce_shards=2 --http localhost:8000 --zero=localhost:5080 --out=myDir
```
twice would overwrite the output created by the first bulk command.

The description of  https://github.com/dgraph-io/dgraph/issues/3026 says we don't want to remove the directory if it exists and has some data.

jarifibrahim:
@codexnull do you think something like this would be a better approach?

```go
	// check if opt.DgraphsDir already exists
	if _, err := os.Stat(opt.DgraphsDir); !os.IsNotExist(err) {
		// OverwriteDgraphsDir should be set if directory is not empty
		if !isDirEmpty(opt.DgraphsDir) && !opt.OverwriteDgraphsDir {
			fmt.Fprintf(os.Stderr, "Directory %q already exists and it is not empty. Use --replace_out to overwrite it.\n",
				opt.DgraphsDir)
			os.Exit(1)
		}
		// Delete out directory
		x.Check(os.RemoveAll(opt.DgraphsDir))
	}
	// recreate the output dirs to ensure they are empty.
	for i := 0; i < opt.ReduceShards; i++ {
		dir := filepath.Join(opt.DgraphsDir, strconv.Itoa(i), "p")
		x.Check(os.MkdirAll(dir, 0700))
		opt.shardOutputDirs = append(opt.shardOutputDirs, dir)
	}

```

`OverwriteDgraphsDir` is a new flag I've added.

codexnull:
Re-submitting using a different approach.

PR #3051, opened 2019-02-20
Delete tablets which don't belong

manishrjain:
Tablet deletion:

- With recent changes, when a predicate is moved from one group to another, the tablet was left behind in the source group.
- This PR adds a way for Zero to instruct the source group to delete the predicate after successfully transferring ownership to the destination group.
- This PR also removes `cleanupTablets` method, so the source group does not make the decision of deleting tablets.
- Rollup operation now does the size calculation per tablet and sends it out to Zero.
- We no longer do a periodic `calculateTabletSize` anymore.
- When Zero receives a tablet size update, after applying the proposal, it determines which tablets need to be deleted, and sends an instruction to Alpha leader accordingly.

Potential issue: Considering the job of tablet deletion happens just after leader rollup, it is possible that the followers would still be doing their rollups when the tablet deletion happens. In that case, a race condition might happen where the deleted tablet could resurface because of roll-up operation (the key is read, then clean predicate deletes it, then the key is written back). In the future, we could cancel a roll-up operation if we determine it could cause a race condition like this.

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3051)


rstorr:
@codelingobot capture error messages shouldn't end in a period

rstorr:
@CodeLingoBot capture error messages shouldn't end in a period

PR #3057, opened 2019-02-22
Turn on ACL in the default docker-compose file

gitlw:
1. Without this change, client tests including dgraph4j needs to be separated into two groups,
one using the cluster with ACL feature turned off, and another one using the cluster with ACL feature turned on in order to test client side ACL logic.
This PR turns on the ACL feature in the default docker-compose file, and from there all regular tests and client side ACL tests can use the same cluster setup.

2. This PR also includes some minor server side changes to improve the ACL feature.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3057)


codexnull:
A lot of the changes are just to pass the `accessJwt` token around. How about just using a shared accessJwt? I know shared/global variables are usually discouraged but for testing I think it's OK and it would make the changeset a lot smaller.

ee/acl/acl.go
```diff
@@ -0,0 +1,550 @@
...
+		func(ctx context.Context, txn *dgo.Txn, groupId string) (AclEntity, error) {
+			group, err := queryGroup(ctx, txn, groupId)
+			return group, err
+		})
+}
+
+type AclEntity interface {
+	GetUid() string
+}
+type queryFunc func(context.Context, *dgo.Txn, string) (AclEntity, error)
```
golangcibot: U1000: type `queryFunc` is unused (from `unused`)

Issue #2854, opened 2018-12-29
alpha globals flag "--log_dir" in config.json is inoperative


relunctance:
If you suspect this could be a bug, follow the template.

- What version of Dgraph are you using?
```
Dgraph version   : v1.0.11-18-g4ca67ca
Commit SHA-1     : 4ca67ca
Commit timestamp : 2018-12-24 18:53:10 -0800
Branch           : master
Go version       : go1.11
```
- Have you tried reproducing the issue with latest release?
yes

- What is the hardware spec (RAM, OS)?
64G

- Steps to reproduce the issue (command/config used to run Dgraph).
-- cmd:
```
dgraph alpha --config /home/s/dgraph/conf/alpha.json
```
--alpha.json
```json
{
   ... 
   "log_dir": "/home/s/dgraph/logs"
   ... 
}
```

- Expected behaviour and actual result.
In this scene , expect all logs are writed "/home/s/dgraph/logs"  instead of  "/tmp" . 

```
log: exiting because of error: write /tmp/dgraph.dg.s.log.INFO.20181227-133013.23000: no space left on device
```

danielmai:
Although it’s not obvious, this is the expected behavior. The log flags like `log_dir` are handled by the glog library and are only configurable as command line flags and not via config files.

To prevent log files from being written and only writing to stderr, you can use the `--logtostderr` flag.

PR #3062, opened 2019-02-25
Set glog flags from configuration

jarifibrahim:
This PR fixes https://github.com/dgraph-io/dgraph/issues/2854

#### Why glog can't use values set in config file?
glog reads values from flags. When a user sets the value (of a glog flag) in a config file, glog doesn't recognize it because the flag value hasn't changed.

### How does this PR fix it?
With this PR, we set the value of glog flags from the values in config file.

**Note**: This is a hack. glog doesn't allow setting configuration programatically and thus we have to use this hack.

Signed-off-by: Ibrahim Jarif <jarifibrahim@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3062)

### DIFF
--- a/dgraph/cmd/root.go
+++ b/dgraph/cmd/root.go
@@ -126,7 +126,8 @@ func setGlogFlags(conf *viper.Viper) {
 	// https://github.com/golang/glog/blob/master/glog.go#L399
 	// and https://github.com/golang/glog/blob/master/glog_file.go#L41
 	glogFlags := [...]string{
-		"log_dir", "logtostderr", "alsologtostderr", "v", "stderrthreshold", "vmodule", "log_backtrace_at",
+		"log_dir", "logtostderr", "alsologtostderr", "v",
+		"stderrthreshold", "vmodule", "log_backtrace_at",
 	}
 	for _, gflag := range glogFlags {
 		// Set value of flag to the value in config
--- a/dgraph/cmd/root.go
+++ b/dgraph/cmd/root.go
@@ -17,7 +17,7 @@
 package cmd
 
 import (
-	goflag "flag"
+	"flag"
 	"strings"
 
 	"github.com/dgraph-io/dgraph/dgraph/cmd/alpha"
@@ -31,7 +31,6 @@ import (
 	"github.com/dgraph-io/dgraph/dgraph/cmd/zero"
 	"github.com/dgraph-io/dgraph/x"
 	"github.com/spf13/cobra"
-	flag "github.com/spf13/pflag"
 	"github.com/spf13/viper"
 )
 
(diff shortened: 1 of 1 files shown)


dgraph/cmd/root.go
```diff
@@ -112,6 +113,25 @@ func initCmds() {
...
+	// List of flags taken from
+	// https://github.com/golang/glog/blob/master/glog.go#L399
+	// and https://github.com/golang/glog/blob/master/glog_file.go#L41
+	glogFlags := [...]string{
+		"log_dir", "logtostderr", "alsologtostderr", "v", "stderrthreshold", "vmodule", "log_backtrace_at",
+	}
+	for _, gflag := range glogFlags {
+		// Set value of flag to the value in config
+		if stringValue, ok := conf.Get(gflag).(string); ok {
+			flag.Lookup(gflag).Value.Set(stringValue)
```
golangcibot: Error return value of `.Set` is not checked (from `errcheck`)

dgraph/cmd/root.go
```diff
@@ -112,6 +113,25 @@ func initCmds() {
...
+
+// setGlogFlags function sets the glog flags based on the configuration.
+// We need to manually set the flags from configuration because glog reads
+// values from flags, not configuration.
+func setGlogFlags(conf *viper.Viper) {
+	// List of flags taken from
+	// https://github.com/golang/glog/blob/master/glog.go#L399
+	// and https://github.com/golang/glog/blob/master/glog_file.go#L41
+	glogFlags := [...]string{
+		"log_dir", "logtostderr", "alsologtostderr", "v", "stderrthreshold", "vmodule", "log_backtrace_at",
```
golangcibot: line is 101 characters (from `lll`)

dgraph/cmd/root.go
```diff
@@ -31,7 +31,6 @@ import (
 	"github.com/dgraph-io/dgraph/dgraph/cmd/zero"
 	"github.com/dgraph-io/dgraph/x"
 	"github.com/spf13/cobra"
-	flag "github.com/spf13/pflag"
```
jarifibrahim: Aliasing `pflag` => `flag` and `flag` => `goflag` created unnecessary ambiguity. I've changed this to remove the ambiguity.

jarifibrahim:
We might also want to evaluate https://github.com/Sirupsen/logrus. It would be much cleaner (and easier) to configure the logging with logrus compared to glog.

Also, the logs from glog aren't parseable (you can parse them but searching for a specific key might be difficult). Logrus allows adding context information to logs https://github.com/sirupsen/logrus#default-fields which would be very useful when you have a lof of logs.

PR #3072, opened 2019-02-26
Rename metrics to match v1.0 metric names.

danielmai:
The namespace prefix is already set in the Namespace field when creating the
exporter, so the metric names do not need to include the namespace itself.

This change aligns the metric names to what they are in v1.0. They mostly follow the [Prometheus metric naming convention](https://prometheus.io/docs/practices/naming/). The one outlier is the latency buckets which are measured in milliseconds. Prometheus recommends using base units (seconds), but if we change the units to seconds then we should change the bucket distribution for a more even spread when most operations happen within milliseconds. I'm not sure what the distribution should be.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3072)

### DIFF
--- a/x/metrics.go
+++ b/x/metrics.go
@@ -41,11 +41,11 @@ var (
 		"Latency of the various methods", stats.UnitMilliseconds)
 
 	// value at particular point of time
-	PendingQueries = stats.Int64("pending_queries",
+	PendingQueries = stats.Int64("pending_queries_total",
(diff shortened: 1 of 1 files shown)


danielmai:
@srfrog Should we change the latency distribution units from milliseconds to seconds?

srfrog:
> @srfrog Should we change the latency distribution units from milliseconds to seconds?

I don't think we'll get enough resolution with seconds.

PR #3075, opened 2019-02-27
compose tool

codexnull:
Features supported right now:

* number of zeros in cluster
* number of alphas
* number of groups
* lru size
* enterprise features on or off

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3075)

### DIFF
--- a/contrib/compose/main.go
+++ b/contrib/compose/main.go
@@ -231,5 +231,6 @@ func main() {
 
 	out, err := yaml.Marshal(cfg)
 	x.Check(err)
+	fmt.Printf("# Auto-generated with: %v\n#\n", os.Args[:])
 	fmt.Printf("%s", out)
 }
--- a/contrib/compose/main.go
+++ b/contrib/compose/main.go
@@ -1,0 +1,83 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"path"
+
+	"github.com/dgraph-io/dgraph/x"
+	yaml "gopkg.in/yaml.v2"
+)
+
+type Volume struct {
+	Type     string
+	Source   string
+	Target   string
+	ReadOnly bool `yaml:"read_only"`
+}
+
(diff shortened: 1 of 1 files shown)


martinmr:
I think there's a comment by Manish about adding a flag that is still not addressed but after you are done with it, this PR LGTM

contrib/compose/main.go
```diff
@@ -0,0 +1,236 @@
...
+		"use a persistent data volume")
+	cmd.PersistentFlags().BoolVarP(&opts.EnterpriseMode, "enterprise", "e", false,
+		"enable enterprise features in alphas")
+	cmd.PersistentFlags().BoolVar(&opts.TestPortRange, "test_ports", true,
+		"use alpha ports expected by regression tests")
+
+	err := cmd.ParseFlags(os.Args)
+	if err != nil {
+		if err == pflag.ErrHelp {
+			cmd.Usage()
```
golangcibot: Error return value of `cmd.Usage` is not checked (from `errcheck`)

codexnull:
Dynamic docker-compose.yml generator. Features supported right now:

* number of zeros in cluster

* number of alphas

* number of groups

* lru size

* enterprise features on or off

* persistent data on or off

manishrjain:
Nice change!
I think there are two follow up items here:
1. The persistent volume is being done differently than how we do it now (by passing a path on the host machine).
2. Needs a way to turn on/off ACLs.

codexnull:
I didn't get the ACL support done yesterday but the tool is already usable so I wanted to get it out there. I'll add the ACL feature today.

The persistent volume is containerized for easier management and clean-up. I thought the way it was being done before, i.e. mapping a host path, had caused problems (unknown state on start, file ownership and permissions, etc).  But it should be possible to do that as well. I'll add that.

Issue #3076, opened 2019-02-28
Dgraph does not work with config file (unreleased master)


danielmai:
If you suspect this could be a bug, follow the template.

- What version of Dgraph are you using?
master

```
$ dgraph version

Dgraph version   : v1.0.12-rc3-155-g1bde8d552
Commit SHA-1     : 1bde8d552
Commit timestamp : 2019-02-27 12:16:12 -0800
Branch           : master
Go version       : go1.11.5
```

- Have you tried reproducing the issue with latest release?
Only happens in master.

- Steps to reproduce the issue (command/config used to run Dgraph).
Create a minimal config file for Zero or Alpha or any subcommand.

```
# Zero doesn't require any flags, so an empty config is valid
$ touch zero_config.yaml
```

```
$ dgraph zero --config=zero_config.yaml
2019/02/27 18:15:22 syntax error: expect file.go:234

github.com/dgraph-io/dgraph/x.Wrap
	/home/dmai/go/src/github.com/dgraph-io/dgraph/x/error.go:100
github.com/dgraph-io/dgraph/x.Check
	/home/dmai/go/src/github.com/dgraph-io/dgraph/x/error.go:42
github.com/dgraph-io/dgraph/dgraph/cmd.setGlogFlags
	/home/dmai/go/src/github.com/dgraph-io/dgraph/dgraph/cmd/root.go:135
github.com/dgraph-io/dgraph/dgraph/cmd.initCmds.func1
	/home/dmai/go/src/github.com/dgraph-io/dgraph/dgraph/cmd/root.go:116
github.com/dgraph-io/dgraph/vendor/github.com/spf13/cobra.(*Command).preRun
	/home/dmai/go/src/github.com/dgraph-io/dgraph/vendor/github.com/spf13/cobra/command.go:728
github.com/dgraph-io/dgraph/vendor/github.com/spf13/cobra.(*Command).execute
	/home/dmai/go/src/github.com/dgraph-io/dgraph/vendor/github.com/spf13/cobra/command.go:664
github.com/dgraph-io/dgraph/vendor/github.com/spf13/cobra.(*Command).ExecuteC
	/home/dmai/go/src/github.com/dgraph-io/dgraph/vendor/github.com/spf13/cobra/command.go:783
github.com/dgraph-io/dgraph/vendor/github.com/spf13/cobra.(*Command).Execute
[...]

PR #3077, opened 2019-02-28
Do not set nil value for log_backtrace_at flag

jarifibrahim:
https://github.com/dgraph-io/dgraph/pull/3062 added capabilities to set flag values from config file (we need this for glog).
This caused an issue with the log_backtrace_at flag.
The flag is defined as `flag.Var(&logging.traceLocation, "log_backtrace_at", "when logging hits line file:N, emit a stack trace")` where tracelocation is

```
// traceLocation represents the setting of the -log_backtrace_at flag.
type traceLocation struct {
	file string
	line int
}
```
If the value isn't specified the nil value for tracelocation type is set in the config, which is tracelocation{"", 0}.
But this value can't be set to the flag because of the check on line https://github.com/golang/glog/blob/master/glog.go#L374

This PR adds skips setting the flag to nil value if it nil in the config.

Fixes https://github.com/dgraph-io/dgraph/issues/3076

Signed-off-by: Ibrahim Jarif <jarifibrahim@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3077)

### DIFF
--- a/dgraph/cmd/root.go
+++ b/dgraph/cmd/root.go
@@ -132,6 +132,15 @@ func setGlogFlags(conf *viper.Viper) {
 	for _, gflag := range glogFlags {
 		// Set value of flag to the value in config
 		if stringValue, ok := conf.Get(gflag).(string); ok {
+			// Special handling for log_backtrace_at flag because the flag is of
+			// type tracelocation. The nil value for tracelocation type is
+			// ":0"(See https://github.com/golang/glog/blob/master/glog.go#L322).
+			// But we can't set nil value for the flag because of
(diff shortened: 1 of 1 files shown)


jarifibrahim:
@danielmai can you please review this PR?

danielmai:
It looks like this addresses it. Will do some testing.

jarifibrahim:
I see 5 checks on other PRs but this one has only 2 checks. I don't know why.

jarifibrahim:
> It looks like this addresses it. Will do some testing.

I wish we could have some tests.

PR #3085, opened 2019-03-02
Upgrade raft lib and fix group checksum

manishrjain:
Found 2 issues recently:
1. If a Zero leader becomes a follower, but Dgraph already has a proposal for it, raft lib would silently drop the proposal, causing Zero to retry indefinitely. Better error handling was added in the master version of raft lib, so in this PR, I vendor that in. That fixes the issue.

2. Group checksum was being calculated by looping over map of tablets directly. This caused the checksum to be different for Zeros, causing queries to block indefinitely. This PR sorts the tablets before calculating the checksum, fixing the issue. Figured out this issue via blockade test. Also fixed up tracing, so waiting for checksum is better indicated (took me a while to realize that it was the group checksum, not the timestamp wait which was blocking).

Longer version for issue 1:

The master version has better error handling and reporting, so when a proposal is dropped, either because there's no leader or because there's no forwarding is not allowed, it can return an error instead of being silent.

This is very useful in a bug we saw today where Dgraph Zero indefinitely tries to send a proposal, which got passed through `AmLeader()` checks, but then failed after `raft.Propose`. Because raft did not complain about the proposal, the proposal was getting retried indefinitely.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3085)


worker/mutation.go
```diff
@@ -39,8 +39,8 @@ import (
 )
 
 var (
-	ErrUnservedTabletMessage = "Tablet isn't being served by this instance"
-	errUnservedTablet        = x.Errorf(ErrUnservedTabletMessage)
+	errUnservedTablet  = x.Errorf("Tablet isn't being served by this instance.")
+	errPredicateMoving = x.Errorf("Predicate is being moved. Please retry later")
```
golangcibot: `errPredicateMoving` is unused (from `varcheck`)

PR #3096, opened 2019-03-05
Add best-effort query to http endpoint

### LINKED ISSUE #3093: Support best-effort read queries over HTTP
#3071 Introduced best-effort read queries over gRPC. They should also be supported over HTTP.

The way read-only queries are set in HTTP is by setting the query parameter `query?ro=true`. For best-effort reads, this could be done by setting `/query?be=true`, and it should automatically set the txn as read-only internally.

srfrog:
This change adds a new query parameter `be=` to support best effort queries using HTTP. Read-only is implied with best effort param, no need to set `ro=` also.

Usage:
```bash
$ curl 'localhost:8080/query?be=true' -d $'{ q(func:uid(0x1)) { uid }}'
```

Closes #3093 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3096)

### DIFF
--- a/dgraph/cmd/alpha/http.go
+++ b/dgraph/cmd/alpha/http.go
@@ -158,9 +158,17 @@ func queryHandler(w http.ResponseWriter, r *http.Request) {
 	d := r.URL.Query().Get("debug")
 	ctx := context.WithValue(context.Background(), query.DebugKey, d)
 	ctx = attachAccessJwt(ctx, r)
-	// If ro is set, run this as a readonly query.
-	if ro := r.URL.Query().Get("ro"); len(ro) > 0 && req.StartTs == 0 {
-		if ro == "true" || ro == "1" {
+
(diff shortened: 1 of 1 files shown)


danielmai:
Great, thanks @srfrog. These work as expected.

```
# Run a normal query
$ curl -XPOST localhost:8080/query -d '...'
```
```
# Run a read-only query
$ curl -XPOST localhost:8080/query?ro=true -d '...'
```
```
# Run a best-effort query
$ curl -XPOST localhost:8080/query?be=true -d '...'
$ curl -XPOST localhost:8080/query?ro=true&be=true -d '...' # ro is implied
$ curl -XPOST localhost:8080/query?be=true&ro=true -d '...' # order doesn't matter
$ curl -XPOST localhost:8080/query?ro=false&be=true -d '...' # be supercedes ro value
```

PR #3097, opened 2019-03-05
Improve the blockade script

gitlw:
- Fix leaked go routines
- Panic upon increment error
- Sleep for 10s for cluster to be ready to avoid very possible health check retry

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3097)


gitlw:
The increment tool was supposed to run forever until it succeeds, therefore there should never be an error emitted from the increment command. Hence the script was running fine, just in a different way than I had expected.

PR #3124, opened 2019-03-11
Improving ACL

gitlw:
- Since both the user ids and group ids are stored under the dgraph.xid predicate, after adding a user named "alice", querying for group using the string "alice" gave "alice" as a group, which was wrong. This PR adds a type to distinguish a user id from a group id.
- After modifying a user's group list, or changing a group's ACLs, we only showed the update was successful. This PR echos the user or group's info back the user by reading from Dgraph after the update.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3124)


ee/acl/acl.go
```diff
@@ -401,6 +409,8 @@ func chMod(conf *viper.Viper) error {
 		return fmt.Errorf("one of --pred or --pred_regex must be specified, but not both")
 	case len(predicate) == 0 && len(predRegex) == 0:
 		return fmt.Errorf("one of --pred or --pred_regex must be specified, but not both")
+	case perm > 7:
+		return fmt.Errorf("the perm value must be less than or equal to 7, the provided value is %d", perm)
```
golangcibot: line is 101 characters (from `lll`)

PR #3197, opened 2019-03-23
New txn operation

### LINKED ISSUE #3059: Support a new Upsert operation
This operation would allow a user to run a query, and use its results (via variables) to execute the mutation. This would then allow upserts to happen, without doing multiple network calls to the DB.

```
txn {
  query {
    me(func: eq(email, "someone@gmail.com")) {
      v as uid
    }
  }

  mutation @if(eq(len(v), 0)) {
    set {
      <uid(v)> <name> "Some One" .
      <uid(v)> <email> "someone@gmail.com" .
    }
  }
}
```

This txn would check for an account with `email=someone@gmail.com`, and only if it is present, would
it run the mutation.

This should return the result of query, normally. And also return the result of the mutation.

Additionally, we should be able to optionally do the mutation, iff the email is already present.
```
txn {
  query {
    me(func: eq(email, "someone@gmail.com")) {
      v as uid
    }
  }

  mutation @if(gt(len(v), 0)) {
    set {
      <uid(v)> <name> "Changed Name" .
    }
  }
}
```
This would introduce a new txn operator, and a new if directive applicable for mutations.

srfrog:
This PR adds a new type of mutation block called `txn{}` that executes a mutation(s) depending on the result of a conditional query in the same transaction. See original issue for initial problem description: #3059 

Some changes from the original issue:
1. Subject and object uid var values are retrieved using a keyword `uid` instead of predicate parsing (inside angle brackets).
2. Both subject and object can use uid vars.
3. More than one matching uid in query will cause error (see Update at bottom).

Test mutation:
```
// Schema:
name: string .
email: string @index(exact) .

// Mutation:
txn {
  query {
    me(func: eq(email, "someone@gmail.com"), first: 1) {
      v as uid
    }
  }

  mutation {
    set {
      uid(v) <name> "Some One" .
      uid(v) <email> "someone@gmail.com" .
    }
  }
}
```

## Breaking changes

- Changes to api.NQuad:
   - Added new field `SubjectVar` to hold name of variable name to replace subject Uid.
   - Added new field `ObjectVar` to hold name of variable name to replace object Uid.
- Changes to api.Mutation:
   - Added new field `CondQuery` to pass the conditional query from client.

See all commits: https://github.com/dgraph-io/dgo/pull/50/commits/ac2ab89e773b2b0bf68d6345ceea56ee70db0416

## Update

Requested changes from team's code-review:

- [x] Return an error if the variable is matched with more than one uid for now.
- [x] There should not be an error if the conditional query does not match any uid.
- [x] Add support to use the matched variable for object as well (besides subject).
- [ ] All filter logic is kept in the conditional query and the mutation is executed for each var request. A mutation with no vars requested is executed once.

Closes #3059

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3197)


MichelDiz:
I know it's in progress but I have questions Gus.

How this line `uid(v) <email> "someone@gmail.com" .`will look like?
It will have the angular brackets?`<0x1> <email> "someone@gmail.com .`

Can you add a test for multiple operation? like:

```
txn {
  query {
    me(func: eq(email, "someone@gmail.com")) {
      v as uid
    }
   me2(func: eq(email, "otherperson@gmail.com")) {
      F as uid
    }
  }

  mutation {
    set {
      uid(v) <friend> uid(F) .
    }
  }
}
```

srfrog:
> How this line uid(v) <email> "someone@gmail.com" .will look like?

The `uid(v)` is translated internally to `<0x1>`, for example. That is if **v=0x1**. Only the first value is used, eventually it could run multiple mutations for each of the uid's. But that was not specified for now.

> Can you add a test for multiple operation? 

The query is just like any regular query, so you can do whatever you want. The only requirement, for now, is that it yields a variable value. This might change as we add more features.

PR #3203, opened 2019-03-25
Fix lin reads timeouts and AssignUid recursion in Zero

manishrjain:
- The way we were doing linerizable reads was creating a new request context for every try. If the request times out, a new one was being used. When things got slower, this caused a traffic jam of new requests compounding because earlier ones timed out and retried.

- This PR fixes that by reusing the same `requestCtx`, so on a retry it would still be OK to receive response from the previous tries -- this allows wrapper requests to be considered done and things to move forward.

- AssignUids was causing a recursion where a Zero follower (A) forwarded the request to who it thought was the leader (B). But, if B was not (or no longer) the leader, it would try to forward it as well -- maybe to A or others. This caused a lot of recursive requests. This PR fixes that so a forwarded request is not forwarded any further, but outright rejected if received by a Zero follower.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3203)


conn/raft_server.go
```diff
@@ -217,6 +223,7 @@ func (w *RaftServer) RaftMessage(server pb.Raft_RaftMessageServer) error {
 		}
 		data := batch.Payload.Data
 
+		ctx, cancel := context.WithTimeout(ctx, time.Minute)
```
golangcibot: the cancel function is not used on all paths (possible context leak) (from `govet`)

PR #3205, opened 2019-03-25
Fix bug triggered by nested expand predicates.

### LINKED ISSUE #3198: Allow nested expand(_all_) for specified predicates
Dgraph returns an error when I make nested `expand(_all_)` like `expand(_all_) { external_node { expand(_all_)  } }`. I'm actually not sure whether this is a bug or by design. The [documentation](https://docs.dgraph.io/query-language/#expand-predicates) on `expand(_all_)` didn't mention this. Full repro follows.

Version:
```
Dgraph version   : v1.0.13
Commit SHA-1     : 691b3b35
Commit timestamp : 2019-03-09 19:33:59 -0800
Branch           : HEAD
Go version       : go1.11.5
```

Schema:
```
<name>: string .
<node>: uid .
```

Data:
```
{
  set
  {
    <_:x> <node> <_:x> .
    <_:x> <name> "test" .
  }
}
```

Problematic query:
```
{
    q(func: has(node)) {
        uid
    	expand(_all_) {
            uid
      	    node {
                uid
                expand(_all_)
            }
        }
    }
}
```
Error:
```
": Unhandled pb.node expand with parent node"
```

However, all of the following works:
```
{
    q(func: has(node)) {
        uid
    	expand(_all_) {
            uid
      	    node {
                uid
            }
        }
    }
}
```
```
{
    q(func: has(node)) {
        uid
        node {
            uid
            expand(_all_)
        }
    }
}
```
```
{
    q(func: has(node)) {
        uid
    	node {
            uid
      	    node {
                uid
                expand(_all_)
            }
        }
    }
}
```
```
{
    q(func: has(node)) {
        uid
    	expand(_all_) {
            uid
      	    expand(_all_) {
                uid
                expand(_all_)
            }
        }
    }
}
```
If this is not a bug, is it possible to support this kind of query?

martinmr:
Fixes #3198 

Certain queries with nested `expand(_all_)` predicates were being rejected
by the method valueVarAggregation. However, the query should have been
allowed and the results after ignoring the error are what was expected.
This adds an extra check to return nil from that method if the subgraph
is an expand subgraph.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3205)


martinmr:
I changed the fix. There's existing logic to clear the exact type of subgraphs that is causing the issue but it's executed before the expandSubgraph is run. Moving it at the end of ProcessGraph fixed the issue.

PR #3207, opened 2019-03-25
Gitlw/acl over tls

gitlw:
- Refactoring TLS code to allow a client to connect to the server over TLS without providing a certificate (one direction authentication). Now the logic for server side TLS set up and client side TLS set up are separated.
- Removed the logic that supports automatic reloading of server cert files (since it seems to complicate the code a lot, and the feature is probably not very important)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3207)


dgraph/cmd/alpha/run.go
```diff
@@ -64,7 +64,7 @@ const (
 
 var (
 	bindall bool
-	tlsConf x.TLSHelperConfig
+	tlsCfg  *tls.Config
```
golangcibot: U1000: var `tlsCfg` is unused (from `unused`)

PR #3217, opened 2019-03-27
UpsertSpan so successful ACL accesses are sent to Jaeger

gitlw:
spans need to be started inside Query, Mutate, and Alter (instead of doQuery, doMutate, doAlter), so the successful ACL accesses are captured by the span and sent to Jaeger

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3217)

### DIFF
--- a/edgraph/access_ee.go
+++ b/edgraph/access_ee.go
@@ -482,8 +482,7 @@ func authorizeAlter(ctx context.Context, op *api.Operation) error {
 	}
 
 	err := doAuthorizeAlter()
-	span := otrace.FromContext(ctx)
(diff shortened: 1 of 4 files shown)


srfrog:
I'm not so sure we should we be creating so many spans. It looks like we make 2 for query and mutate.

manishrjain:
Let's revert this PR. It causes 2 issues:

1. Might not create a span when we wanted to.
2. Might end a span which belongs to the caller.

The result is also that the spans no longer have a method name in our traces, when the caller passes a span.

PR #3222, opened 2019-03-28
Added crasher input and fix issues

gitlw:
More detailed explanation of the issue fixed by change in iri.go
--------------Here is a run of the program with non-crasher input ------------

```
AcceptRunRec
r := l.Next (r is >)
!c(r) returns true
break 
l.Backup()  pos now points to >

returns to IRIRef
r := l.Next (r is >)
l.Ignore (ignores the >)
```

------------Here is a run with crasher input (the first input added to the parser_test.go)--------

```
AcceptRunRec
r := l.Next (r is \\)

IsRIRChar(r, l) (r is \\)
r2 := l.Next() (r2 is 65533)
a<><\\ef bf bd
              |  pos is 8 here with (width being 3)
l.Backup()
a<><\\ef bf bd
      |
    pos is 5 here (width being 3)
return false to AcceptRunRec

AcceptRunRec break the for loop
l.Backup inside AcceptRunRec
a<><\\ef bf bd
  |
  pos is 2 here (width being 3)

return to IRIRef
r := l.Next (line 28) (r is > )
```

More detailed explanation of the issue fixed by change in parser.go
The last two inputs added to the parser_test.go causes a index out of range exception, since line 2440 in parser.go is trying to get an element from gq.NeedsVar when it's empty.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3222)


manishrjain:
lgtm

PR #3225, opened 2019-03-28
Gitlw/adding acl docs

gitlw:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3225)


Himan000:
Why ACL is an enterprise feature? It's just a common function for client to access database like than behavior when using mysql. If no access control,  then anyone who have the IP and port can access the data in dgraph.

gitlw:
@Himan000 You are absolutely right that without access control, anyone can access the data in dgraph.
However it's debatable whether this is a basic function, because ultimately any feature we make should be useful to at least some of our customers. In the same context, other DB vendors like Neo4j, CockroachDB, MongoDB all includes some feature around security in their enterprise offering.

martinmr:
Also, the set of people/devices with direct access to the database should be kept as small as possible.

Himan000:
Hi @gitlw @martinmr,
Yes, you'r right we have other way to keep accessing database narrow, e.g white list IP and port. But that's not a good way for database especial for security data. Never mind, it's your choise to keep off most non-enterprise user.

PR #3238, opened 2019-04-01
Add timeout option while running queries over HTTP

mangalaman93:
The endpoint /query now accepts a timeout key - /query?timeout=100ms.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3238)


danielmai:
Can you make the timeout accept a duration instead of just milliseconds? i.e., `?timeout=1s` or `?timeout=1m` for 1 second or 1 minute.

martinmr:
https://golang.org/pkg/time/#ParseDuration should work.

srfrog:
I find `context.WithTimeout` a bit easier to use since you only need to send the actual timeout. So it fits perfectly with `time.ParseDuration`

PR #3243, opened 2019-04-01
Always parse language tag.

martinmr:
Currently, the language tag is only being parsed if the value is a
string, which lead to issues in a different PR.
This change makes it so that the language tag is always parsed.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3243)

### DIFF
--- a/chunker/json/parse.go
+++ b/chunker/json/parse.go
@@ -127,10 +127,6 @@ func handleBasicType(k string, v interface{}, op int, nq *api.NQuad) error {
 		nq.ObjectValue = &api.Value{Val: &api.Value_IntVal{IntVal: i}}
 
 	case string:
-		// Here we split predicate and lang directive (ex: "name@en"), if needed. With JSON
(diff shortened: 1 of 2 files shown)


chunker/json/parse.go
```diff
@@ -261,11 +257,16 @@ func mapToNquads(m map[string]interface{}, idx *int, op int, parentPred string)
 		if op == DeleteNquads {
 			// This corresponds to edge deletion.
 			if v == nil {
-				mr.nquads = append(mr.nquads, &api.NQuad{
+				nq := &api.NQuad{
 					Subject:     mr.uid,
 					Predicate:   pred,
 					ObjectValue: &api.Value{Val: &api.Value_DefaultVal{DefaultVal: x.Star}},
-				})
+					}
```
golangcibot: File is not `goimports`-ed (from `goimports`)

PR #3257, opened 2019-04-05
Add TLS support to increment command

codexnull:
Change summary:

* Change --addr option to --dgraph to match the other commands.
* Support --tls_* options instead of ignoring them.
* Use x.SetupConnection() instead of grpc.Dial() to add TLS support.

### DIFF
--- a/dgraph/cmd/counter/increment.go
+++ b/dgraph/cmd/counter/increment.go
@@ -31,7 +31,6 @@ import (
 	"github.com/dgraph-io/dgraph/x"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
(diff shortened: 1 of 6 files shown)


codexnull:
You're right, it doesn't look good, but I was trying to be consistent with other commands. We already have `dgraph live --dgraph` and `dgraph acl --dgraph`. Maybe we should change all of these to `--alpha` instead? It would also help distinguish which of the two types of dgraph services you're connecting to.

manishrjain:
`--alpha` would make sense. Forgot that `dgraph live` is using `--dgraph` as a flag. This might be coming from the time when live loader used to be a separate binary. Yeah, should be renamed.

PR #3262, opened 2019-04-05
Remove _predicate_ from Dgraph.

martinmr:
From now on, expand queries will rely solely on the type of the node.
This change removes all mentions of the `_predicate_` internal pred that
was previously being used to serve this type of queries.

There is also another small change to allow the value of a list variable
to be used as a list of predicates in expand queries.

Also, some of the expand tests in the query package were simplified so that the
output of the test queries is not as long.

Most of the changes due to removing or replacing tests. These changes
are due to one of the following reasons.

1. Tests that were checking for the state of `_predicate_` and are no
longer valid.
2. Tests that were doing checks that should be done inside of the query
package. They have been removed if there was an existing query that was
doing the same or a new test has been added if no similar test existed.
3. Some of the expand tests inside the query package were removed
because there are tests verifying the expand functionality using types.
Since types are now the only way to make expand queries work, these
tests became obsolete.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3262)


dgraph/cmd/bulk/schema.go
```diff
@@ -37,12 +37,7 @@ type schemaStore struct {
 
 func newSchemaStore(initial []*pb.SchemaUpdate, opt options, state *state) *schemaStore {
 	s := &schemaStore{
-		m: map[string]*pb.SchemaUpdate{
-			"_predicate_": {
-				ValueType: pb.Posting_STRING,
-				List:      true,
-			},
-		},
+		m: map[string]*pb.SchemaUpdate{},
```
golangcibot: File is not `goimports`-ed (from `goimports`)

dgraph/cmd/zero/oracle.go
```diff
@@ -334,21 +333,7 @@ func (s *Server) commit(ctx context.Context, src *api.TxnContext) error {
...
-
-		// Ensure that we only consider checking _predicate_, if expand_edge flag is
-		// set to true, i.e. we are actually serving _predicate_. Otherwise, the
-		// code below, which checks if a tablet is present or is readonly, causes
-		// ALL txns to abort. See #2547.
-		if tablet := s.ServingTablet("_predicate_"); tablet != nil {
-			pkey := fmt.Sprintf("%d-_predicate_", tablet.GroupId)
-			preds[pkey] = struct{}{}
-		}
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)

martinmr:
Ran some benchmarks comparing the performance of the live and bulk loaders when `_predicate_` is enabled/disabled.

For the bulkloader (using the 21 million dataset), I observed an improvement but it was not drastic. The number of edges went down from 120 million to 99 million. We do not observe a 50% reduction in the number of edges because there are nodes that contain more than an edge for a given predicate.

During the mapping phase, it seems that the cost of adding a new `_predicate_` edge is exercised when the first edge is added and repeated edges are more of a no-op. The reduced number of edges speeds up the reduce phase but only by the same percent in the reduction of total edges.

In average, the version without `_predicate_` took around 1m45s to complete the bulk load while the current version took around 2m. The reduction in edges is slightly bigger than the reduction in total time.

For the live loader I used the smaller 1 million dataset. I could not observe any performance differences between the two versions. My hypothesis is that the process is throttled by transactions (which are not involved when the bulk loader is used) so the reduction in edges does not result in a reduction in the total time.

martinmr:
Re did the bulk loader benchmark, this time without any indices in the schema. This time I observe the 50% reduction although the time savings are not 50% but are nonetheless noticeable (1m5s vs 42s).

martinmr:
I ran the live loader benchmarks. Without `_predicate_`, the running time (with the 21 million dataset) is 3m40s. With `_predicate_` is 4m45s. I see a decrease in the running time but not by 50%, just like in the bulk loader.

PR #3273, opened 2019-04-09
Rename dgraph --dgraph option

codexnull:
Having a command line option `--dgraph` that's the same as the program name `dgraph` looks ugly. Rename it to `--alpha`. For consistency, do it across all  subcommands that use it (live, increment, and acl).

Change summary:
* Rename subcommand long options `--dgraph` and `--addr` to `--alpha` and short option `-d` to `-a`.
* Rename acl `--auth_token` short option from `-a` to `-t`.
* Update all tests to use the renamed options.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3273)

PR #3284, opened 2019-04-12
Fix a few systests using the removed --dgraph flag.

danielmai:
The flag name changed in #3273.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3284)


danielmai:
It's too bad we currently don't run these tests on every PR.

PR #3314, opened 2019-04-24
Use jsonpb library to unmarshal alter HTTP requests.

martinmr:
This library should be used instead of "encoding/json" to properly parse
json into a protobuf.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3314)


martinmr:
I checked the history and it doesn't look like the jsonpb library was used before at any point. It's likely that we didn't need it until now because api.Operation only contained strings and booleans so the default json library worked just fine.

PR #3325, opened 2019-04-26
Truncate Raft logs even when no txn commits are happening

manishrjain:
If there are no txn commits but many txns which allocate new read timestamps, the Raft log can still grow a lot because of MaxAssigned updates. This PR would truncate those logs as well to create new snapshot.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3325)


martinmr:
fyi, you still need to merge the changes from lucas PR to get teamcity working.

PR #3339, opened 2019-04-29
Add prometheus metrics for latest raft applied index and oracle max assigned ts.

danielmai:
This PR is like #3338 but for master branch which uses OpenCensus stats for metrics.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3339)


danielmai:
Doesn't seem like max assigned is being tracked properly over OpenCensus stats. It's always significantly behind the actual value of the latest timestamp.

PR #3355, opened 2019-05-02
Fix flakiness in snapshot_test.

martinmr:
- Retry mutations when transaction is aborted.
- Retrieve all the values and do the sum in the test instead of relying
on the sum function inside Dgraph.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3355)


martinmr:
@manishrjain The test is still flaky because of the potential streaming bug but these changes are safe to go in.

PR #3365, opened 2019-05-03
Remove custom HTTP Headers, cleanup API

### LINKED ISSUE #3142: Make HTTP endpoint compatible with GraphQL best practices
This is a Design Doc for Dgraph v1.1 - a breaking change to make HTTP API more compatible with GraphQL.

Further reading:
- [GraphQL Best Practices](https://graphql.org/learn/serving-over-http/)
- [Current Dgraph API](https://docs.dgraph.io/clients/#raw-http)
- [(Team DGraph Only) Internal Doc Comparing the above two](https://discuss.dgraph.io/t/proposal-http-headers-considered-harmful/4145)

In v1.1 Dgraph HTTP endpoints must have following properties: 

## /query

1. URL `/query` accepts HTTP POST with Content-Type `application/json` and POST body in JSON format:
```
{
  "query": "...",
  "variables": { "myVariable": "someValue", ... },
  "startTs": <number>,
  "debug": <true/false>,
}
```

2. JSON field `"query"` is required, all other fields are optional.

3. If the "application/graphql" Content-Type header is present, treat the HTTP POST body contents as the GraphQL query string and read all other fields from the query string e.g. `/query?variables=...&startTs=...&debug=true`

## /mutate

4. `/mutate` no longer accepts `X-Dgraph-...` custom headers. OPTIONS response no longer advertises the headers, if a POST request is received with any of these headers, their presence and values are ignored.

5. As of dgraph v1.0.13 `/mutate` accepts `X-Dgraph-CommitNow` and `X-Dgraph-MutationType`. Mutation type can be inferred from the Content Type header and thus no replacement is necessary.
CommitNow can be specified on the URL string: `/mutate/<startTs>?commitNow=true`

## /commit

`/commit` does not use any of the `X-Dgraph-...` so no changes to its behavior are necessary.

mangalaman93:
Fixes #3142
Why remove custom HTTP Headers?
  * Needs to be whitelisted for CORS
  * Parsing Unicode characters in headers is too hard
  * Uncommon to use, there are better approaches

We are still not completely GraphQL compatible. In fact,
GraphQL specs are not clear around this to begin with but
this commit cleans up HTTP API further.

Todo -
  * [x] Tests for application/json content type
  * [x] Remove headers from other endpoints
  * [x] Update documentation
  * [x] Raise issues with the clients that are using HTTP endpoint

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3365)


dgraph/cmd/alpha/http.go
```diff
@@ -325,29 +391,72 @@ func commitHandler(w http.ResponseWriter, r *http.Request) {
...
-	// Keys are sent as an array in the body.
-	reqText := readRequest(w, r)
-	if reqText == nil {
+	js, err := json.Marshal(response)
+	if err != nil {
+		x.SetStatusWithData(w, x.Error, err.Error())
 		return
 	}
 
+	writeResponse(w, r, js)
```
golangcibot: Error return value of `writeResponse` is not checked (from `errcheck`)

PR #3391, opened 2019-05-09
Bug Fix: Avoid Dgraph cluster getting stuck in infinite leader election

manishrjain:
Dgraph Alphas were calculating snapshots and checkpoints in the main Raft loop, which depending upon disk speed caused Ticks to not be done for seconds. This caused the followers to assume that the leader is unavailable, triggering an election. Because the checkpoint and snapshot calculation happens every 30s, the election was happening every 30s as well.

This PR moves both of those outside the main loop into their own goroutine (colocated with the code which shuts down Raft node). Tested successfully with a live cluster which was exhibiting these symptoms.

This PR also tracks how many heartbeats have come in and gone out from each node and prints them out under V(3). Useful for debugging.

The PR improves upon and uses x.Timer to track Raft.Ready components' latencies and report them in both Alphas and Zeros. This fixes the incorrect statement we were making about disk latency being the primary cause of Raft.Ready being slow.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3391)

### DIFF
--- a/conn/node.go
+++ b/conn/node.go
@@ -72,6 +72,8 @@ type Node struct {
 	// The stages are proposed -> committed (accepted by cluster) ->
 	// applied (to PL) -> synced (to BadgerDB).
 	Applied y.WaterMark
+
+	Heartbeats int64
 }
 
 type ToGlog struct {
@@ -155,6 +157,16 @@ func NewNode(rc *pb.RaftContext, store *raftwal.DiskStorage) *Node {
 	return n
 }
 
+func (n *Node) ReportRaftComms() {
+	ticker := time.NewTicker(10 * time.Second)
+	defer ticker.Stop()
+
+	for range ticker.C {
+		num := atomic.SwapInt64(&n.Heartbeats, 0)
+		glog.V(2).Infof("RaftComm: [%#x] Heartbeats exchanged since last report: %d", n.Id, num)
+	}
+}
+
 // SetRaft would set the provided raft.Node to this node.
 // It would check fail if the node is already set.
 func (n *Node) SetRaft(r raft.Node) {
@@ -233,6 +245,7 @@ func (n *Node) Send(msg raftpb.Message) {
 	if glog.V(2) {
 		switch msg.Type {
 		case raftpb.MsgHeartbeat, raftpb.MsgHeartbeatResp:
+			atomic.AddInt64(&n.Heartbeats, 1)
 		case raftpb.MsgReadIndex, raftpb.MsgReadIndexResp:
 		case raftpb.MsgApp, raftpb.MsgAppResp:
 		case raftpb.MsgProp:
--- a/conn/node.go
+++ b/conn/node.go
@@ -73,7 +73,8 @@ type Node struct {
 	// applied (to PL) -> synced (to BadgerDB).
 	Applied y.WaterMark
 
-	Heartbeats int64
+	heartbeatsOut int64
+	heartbeatsIn  int64
 }
 
 type ToGlog struct {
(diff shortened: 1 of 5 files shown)


danielmai:
___
*[conn/node.go, line 253 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/3391#-LePRXLL2UIkuUdehSi7:-LePRXLM8Cmp-7lCQegQ:bwl09j5) ([raw file](https://github.com/dgraph-io/dgraph/blob/f5f170e254f0d690f5e4154255305afdfb3b2f38/conn/node.go#L253)):*
> ```Go
> 		switch msg.Type {
> 		case raftpb.MsgHeartbeat, raftpb.MsgHeartbeatResp:
> 			atomic.AddInt64(&n.heartbeatsOut, 1)
> ```

This happens in glog.V(2) but the ticker in ReportRaftComms only zeroes out heartbeats in glog.V(3). Is that OK?

danielmai:
___
*[conn/raft_server.go, line 232 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/3391#-LePRp4MB9WvUkzrphsh:-LePRp4MB9WvUkzrphsi:b-fbl9fl) ([raw file](https://github.com/dgraph-io/dgraph/blob/f5f170e254f0d690f5e4154255305afdfb3b2f38/conn/raft_server.go#L232)):*
> ```Go
> 				switch msg.Type {
> 				case raftpb.MsgHeartbeat, raftpb.MsgHeartbeatResp:
> 					atomic.AddInt64(&n.heartbeatsIn, 1)
> ```

Same question here too about incrementing in V(2) and glogging/zero-ing in V(3).

conn/node.go
```diff
@@ -72,6 +72,9 @@ type Node struct {
 	// The stages are proposed -> committed (accepted by cluster) ->
 	// applied (to PL) -> synced (to BadgerDB).
 	Applied y.WaterMark
+
+	heartbeatsOut int64
```
mangalaman93: atomic package issues here? This is 64 bit. https://golang.org/pkg/sync/atomic/#pkg-note-BUG

conn/node.go
```diff
@@ -155,6 +158,20 @@ func NewNode(rc *pb.RaftContext, store *raftwal.DiskStorage) *Node {
 	return n
 }
 
+func (n *Node) ReportRaftComms() {
+	if !glog.V(3) {
```
mangalaman93: Shouldn't this work for level > 3 too?

worker/draft.go
```diff
@@ -964,14 +983,13 @@ func (n *node) Run() {
 				span.End()
 				ostats.RecordWithTags(context.Background(),
 					[]tag.Mutator{tag.Upsert(x.KeyMethod, "alpha.RunLoop")},
-					x.LatencyMs.M(x.SinceMs(start)))
+					x.LatencyMs.M(float64(timer.Total())/1e6))
```
mangalaman93: could divide by time.MilliSecond instead here

x/x.go
```diff
@@ -391,18 +395,24 @@ func (t *Timer) Start() {
...
 
 func (t *Timer) Total() time.Duration {
-	return time.Since(t.start)
+	return time.Since(t.start).Round(time.Millisecond)
 }
 
-func (t *Timer) All() []time.Duration {
-	return t.records
+func (t *Timer) String() string {
+	sort.Slice(t.records, func(i, j int) bool {
```
mangalaman93: Is sorting helpful in viewing the results?

PR #3392, opened 2019-05-09
Issue 3383: export \a character

codexnull:
An ASCII bell character added to the database escaped as \u0007 gets escaped to \a when exported. The \a escape sequence is not allowed by the lexer, so the exported file cannot be imported.

First I tried adding support for that sequence to the lexer for consistency with strconv.Quote, printf("%q"), and most quoting conventions. That worked for RDF but it turned out that JSON does not allow that escape sequence either. So the fix was instead to force it to \u0007 in the export.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3392)

### DIFF
--- a/chunker/rdf/parse_test.go
+++ b/chunker/rdf/parse_test.go
@@ -486,17 +486,11 @@ var testNQuads = []struct {
 	},
 	{
 		input:       `_:0a. <name> <bad> .`,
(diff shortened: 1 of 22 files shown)


martinmr:
Ok. I'll take this over. I'll close this PR for now.

PR #3393, opened 2019-05-10
Fix deadlock: Break circular lock acquisition

manishrjain:
There's a deadlock potential here due to circular lock acquisition. This PR breaks that circle.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3393)

### DIFF
--- a/posting/lists.go
+++ b/posting/lists.go
@@ -32,6 +32,7 @@ import (
 
 	"github.com/dgraph-io/badger"
 	"github.com/dgraph-io/badger/y"
+	"github.com/dgraph-io/dgo/protos/api"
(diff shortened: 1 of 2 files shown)


mehdiym:
Hello,
I'm using dGraph 1.0.15 with the go client, and I still have some rare deadlocks.

It doesn't happen with a database running for a long time.

It only happens just after a deep clear of the database (by removing p, w, zw directories) during the first tens/hundreds of queries/mutations.
I can only reproduce it with my test database as I regularly reset it this way.
If I just clear the database with a dropAll alteration, I doesn't happen anymore.

PR #3415, opened 2019-05-14
Call Sync explicitly to improve performance

manishrjain:
Instead of setting SyncWrites=true, we should call `db.Sync()` at the end of TxnWriter or when `raft.Ready().MustSync` is set to true. This is expected to be faster than when `SyncWrites` is blanket set to true without any change in data consistency behavior.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3415)


posting/writer.go
```diff
@@ -104,6 +104,7 @@ func (w *TxnWriter) SetAt(key, val []byte, meta byte, ts uint64) error {
 }
 
 func (w *TxnWriter) Flush() error {
+	defer w.db.Sync()
```
golangcibot: Error return value of `w.db.Sync` is not checked (from `errcheck`)

Issue #3411, opened 2019-05-14
Data Races in Alpha & Zero


mangalaman93:
* [ ] Data Race 1 in Alpha
 * [ ] Data Race 2 in Alpha
 * [ ] Data Race 3 in Alpha
 * [ ] Data Race 4 in Alpha
 * [x] Data Race 5 in Alpha #3432 
 * [x] Data Race 1 in Zero #3417 
 * [x] Data Race 2 in Zero #3417 
 * [x] Data Race 3 in Zero #3417 

Version: b4fb07777b6929ee873bfaa07d62cbb0f583acb7

## Data Race 1 in Alpha
```
alpha3    | ==================
alpha3    | WARNING: DATA RACE
alpha3    | Read at 0x0000028b9ec0 by goroutine 315:
alpha3    |   github.com/dgraph-io/dgraph/query.(*SubGraph).updateFacetMatrix()
alpha3    |       /home/aman/gocode/src/github.com/dgraph-io/dgraph/query/query.go:1395 +0x48f
alpha3    |   github.com/dgraph-io/dgraph/query.(*SubGraph).updateUidMatrix()
alpha3    |       /home/aman/gocode/src/github.com/dgraph-io/dgraph/query/query.go:1407 +0x3c
alpha3    |   github.com/dgraph-io/dgraph/query.(*SubGraph).populateVarMap()
alpha3    |       /home/aman/gocode/src/github.com/dgraph-io/dgraph/query/query.go:1448 +0x35c
alpha3    |   github.com/dgraph-io/dgraph/query.(*QueryRequest).ProcessQuery()
alpha3    |       /home/aman/gocode/src/github.com/dgraph-io/dgraph/query/query.go:2722 +0x15c8
alpha3    |   github.com/dgraph-io/dgraph/query.(*QueryRequest).Process()
alpha3    |       /home/aman/gocode/src/github.com/dgraph-io/dgraph/query/query.go:2771 +0x88
alpha3    |   github.com/dgraph-io/dgraph/edgraph.(*Server).doQuery()
alpha3    |       /home/aman/gocode/src/github.com/dgraph-io/dgraph/edgraph/server.go:653 +0xb23
alpha3    |   github.com/dgraph-io/dgraph/edgraph.(*Server).Query()
alpha3    |       /home/aman/gocode/src/github.com/dgraph-io/dgraph/edgraph/server.go:562 +0xbb
[...]

PR #3417, opened 2019-05-14
Fix race condition in oracle.go

martinmr:
Currently, we pass a pointer to a delta to a channel and then set the
group checksums. This results in a race condition because the pointer
is sent to multiple channels. This change changes the channel to send
a delta instead of a pointer. This should fix this particular race
condition.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3417)


mangalaman93:
Do not observe any race condition in zero anymore. Ran flock overnight.

mangalaman93:
Fixes zero data races in #3411

martinmr:
Thanks for checking @mangalaman93

PR #3427, opened 2019-05-15
Try another explaination of transactions

MichaelJCompton:
@martinmr @danielmai  I saw this https://github.com/dgraph-io/dgraph/pull/3423 and didn't think it explained what a transaction was ☹️   In particular:

- I didn't like "a unit of work"
- "A transaction can be aborted" ... really should mean 'will be'
- what does "same data" mean? doesn't help a user work out why their transactions are conflicting

So, I tried to rewrite it.  Feel free to let me know if I've made it worse.

**Also**: writing this made me realise that I don't really understand the conditions under which two transactions conflict, nor do I really understand exactly what `upsert` means in a Dgraph schema.  So is what I've written right?  Does it really work like that - e.g. if I've got a day index; does that mean every concurrent write on that index will conflict?  I had assumed before that `upsert` in the schema would control those kinds of conflicts and otherwise I'd only get conflicts if I tried to write same edges ... but then is that even a thing?

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3427)

### DIFF
--- a/wiki/content/clients/index.md
+++ b/wiki/content/clients/index.md
@@ -42,7 +42,7 @@ changes it makes - changes from concurrent transactions aren't visible.
 
 On commit, Dgraph will abort a transaction rather than commit when a
 conflicting, concurrently running transaction has already been committed.  Two
-transactions conflict when both: 
+transactions conflict when both transactions: 
 
 - write data to the same key in an index, or
 - write to the same edge of the same node.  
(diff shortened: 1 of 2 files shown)


MichaelJCompton:
@danielmai do you mind having a quick look when you get a chance.  I've rewritten again to make it detailed ... didn't write anything about not conflicting on different languages, but I think it should be ok.  Let me know if you think it's wrong or misleading.

Just wanted to make it quite clear about what a conflict is.

PR #3428, opened 2019-05-16
Adding datadog trace collector

gitlw:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3428)


manishrjain:
> This is done as suggested by the disclaimer section of
https://github.com/DataDog/opencensus-go-exporter-datadog
which says

These sort of things are worth adding as comments in the code, so future readers would know why you chose to sample everything. It does appear like an error.

PR #3432, opened 2019-05-16
Fix race condition in IsPeer.

martinmr:
IsPeer is unsafely accessing the _confstate field. This PR fixes this
race condition as well as making the node field in Raft unexported and
providing a safe method to update it.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3432)


mangalaman93:
@martinmr let me know when you want me to run flock on this change.
Related to #3411

mangalaman93:
Flock is running but I can't reliably reproduce this on master. Let me know if you have ideas around how to reproduce it.

martinmr:
No ideas but I think it's clear that the reads and writes on the confstate were not lock-protected so the race condition should be gone. I'll let Manish decide.

mangalaman93:
Makes sense, as long as PR logically looks good, should merge it. We will run more tests and keep an eye.

PR #3433, opened 2019-05-17
Remove goimports from list of golangci checks.

martinmr:
gofmt already covers this use case. Also, sometimes the output of gofmt
is incompatible with goimports.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3433)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -12,7 +12,6 @@ linters:
     - ineffassign
     - gas
     - gofmt
-    - goimports
     - golint
     - gosimple
     - govet


martinmr:
gofmt already formats the imports and sometimes its output is not compatible with goimports. Goimports also doesn't make much sense here since if there's missing imports here the code will not compile. It seems like goimports is more useful during development, not as an automated check.

danielmai:
It this removes false positives, then it sounds like a good idea.

PR #3438, opened 2019-05-17
Adding the use_raft_progress flag

gitlw:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3438)


gitlw:
After discussion with Daniel and Manish offline, it seems this is not a critical feature to add for now. Will reconsider it in the future if the need comes up.

danielmai:
This flag is no longer needed with #3444 merged that moved the Raft progress from p to w.

PR #3444, opened 2019-05-19
Move checkpoint key to WAL

manishrjain:
The checkpoint key was being stored in p directory. Can't think of a reason why I did that. Keeping the checkpoint key in WAL makes natural sense and allows someone to drop the w directory, without causing issues later. One of the issues is that if w is dropped, Raft index would start from 1 again, but p directory would have stored the progress to a much higher index. This would cause all new proposals to not be applied.

With this PR, w directory can be dropped and the checkpoint would also get dropped along with it.
This PR also adds two optimizations:
- Calculates checkpoints starting from the last checkpoint, instead of the first entry since snapshot.
- Does not calculate snapshot if the checkpoint - first < SnapshotAfter.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3444)


worker/draft.go
```diff
@@ -756,8 +728,10 @@ func (n *node) checkpointAndClose(done chan struct{}) {
 				// We use disk based storage for Raft. So, we're not too concerned about
 				// snapshotting.  We just need to do enough, so that we don't have a huge backlog of
 				// entries to process on a restart.
-				if err := n.proposeSnapshot(x.WorkerConfig.SnapshotAfter); err != nil {
-					x.Errorf("While calculating and proposing snapshot: %v", err)
+				if calculate {
+					if err := n.proposeSnapshot(x.WorkerConfig.SnapshotAfter); err != nil {
+						x.Errorf("While calculating and proposing snapshot: %v", err)
```
golangcibot: Error return value of `x.Errorf` is not checked (from `errcheck`)

PR #3509, opened 2019-06-04
Fix docs and add test for aborting a txn

mangalaman93:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3509)

### DIFF
--- a/dgraph/cmd/alpha/http.go
+++ b/dgraph/cmd/alpha/http.go
@@ -31,6 +31,7 @@ import (
 	"time"
 
 	"github.com/dgraph-io/dgo/protos/api"
+	"github.com/dgraph-io/dgo/y"
 	"github.com/dgraph-io/dgraph/edgraph"
 	"github.com/dgraph-io/dgraph/gql"
 	"github.com/dgraph-io/dgraph/query"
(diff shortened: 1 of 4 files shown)


danielmai:
___
*[dgraph/cmd/alpha/http.go, line 441 at r2](https://reviewable.io/reviews/dgraph-io/dgraph/3509#-LgYjF1zB3ix89mZTI5P:-LgYjF1zB3ix89mZTI5Q:b-viw1yy) ([raw file](https://github.com/dgraph-io/dgraph/blob/a2b16cfed5d3854abaa588231c07675a0c1857e0/dgraph/cmd/alpha/http.go#L441)):*
> ```Go
> 		}, nil
> 	case nil:
> 		return nil, errors.Errorf("transaction could not be aborted")
> ```

When does this case happen, actually? I haven't been able to see this error myself when requesting an abort.

gitlw:
The flow for requesting an abort is somewhat confusing.
1. client sets the TxnContext.Aborted to true, in order to ask for the abort operation
2. alpha forwards the request to zero, and zero will abort and return a nil error to alpha
3. In the CommitOverNetwork, alpha detects that the TxnContext.Aborted is set, and converts the nil error into a y.ErrAborted
4. Now this PR tries to handle the y.ErrAborted and convert it back into a nil error

It seems we can achieve the same result by removing step 3 and step 4.

PR #3530, opened 2019-06-06
Fixing the flaky test TestNodes

gitlw:
The TestNodes method fails once in a while with the error message:
=== RUN   TestNodes
--- FAIL: TestNodes (100.02s)
        Error Trace:    group_delete_test.go:91
        		group_delete_test.go:196
        Error:		Received unexpected error:
        	rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = "transport: Error while dialing dial tcp [::1]:9180: connect: connection refused"

Upon investigation, it turns out that the test is using the docker-compose file with 3 alpha nodes, and it assumes dg1 will be in group 1, dg2 will be in group2 and dg3 will be in group 3. During the test it would remove the nodes in group 2 and group 3, and then use the connection to *dg1* for verifying that a query still works.

However, it is possible that dg1 is actually in group 2 or group 3, instead of group 1. When that happens, dg1 will be shutdown by the remove node request, causing the final query that is still using connection to dg1 to fail. This is demonstrated by the fact that after the test fails, zero has the following state
```
{
    "cid": "0f8374a8-e16b-4500-b91a-7a07b76e361f",
    "counter": "2071",
    "groups": {
        "1": {
            "checksum": "9309113724815971540",
            "members": {
                "1": {
                    "addr": "dg3:7183",
                    "groupId": 1,
                    "id": "1",
                    "lastUpdate": "1559846886",
                    "leader": true
                }
            },
            "tablets": {
                "dgraph.type": {
                    "groupId": 1,
                    "predicate": "dgraph.type"
                },
                "directed_by": {
                    "groupId": 1,
                    "predicate": "directed_by"
                },
                "director.film": {
                    "groupId": 1,
                    "predicate": "director.film"
                },
                "http://www.w3.org/2000/01/rdf-schema#domain": {
                    "groupId": 1,
                    "predicate": "http://www.w3.org/2000/01/rdf-schema#domain"
                },
                "http://www.w3.org/2000/01/rdf-schema#range": {
                    "groupId": 1,
                    "predicate": "http://www.w3.org/2000/01/rdf-schema#range"
                },
                "http://www.w3.org/2002/07/owl#inverseOf": {
                    "groupId": 1,
                    "predicate": "http://www.w3.org/2002/07/owl#inverseOf"
                },
                "initial_release_date": {
                    "groupId": 1,
                    "predicate": "initial_release_date"
                },
                "name": {
                    "groupId": 1,
                    "predicate": "name"
                },
                "type.property.expected_type": {
                    "groupId": 1,
                    "predicate": "type.property.expected_type"
                },
                "type.property.reverse_property": {
                    "groupId": 1,
                    "predicate": "type.property.reverse_property"
                },
                "type.property.schema": {
                    "groupId": 1,
                    "predicate": "type.property.schema"
                }
            }
        }
    },
    "maxLeaseId": "270000",
    "maxRaftId": "3",
    "maxTxnTs": "10000",
    "removed": [
        {
            "addr": "dg1:7180",
            "groupId": 3,
            "id": "3",
            "lastUpdate": "1559846886",
            "leader": true
        },
        {
            "addr": "dg2:7182",
            "groupId": 2,
            "id": "2",
            "lastUpdate": "1559846886",
            "leader": true
        }
    ],
    "zeros": {
        "1": {
            "addr": "zero1:5180",
            "id": "1",
            "leader": true
        }
    }
}
```

So instead of using the connection to dg1, this PR changes the logic so that it asks zero which node is in group 1 (could be dg2 or dg3), and uses that node for queries. Hence it guarantees that the node in group 1 will not be affected by the remove node requests.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3530)

### DIFF
--- a/z/zero.go
+++ b/z/zero.go
@@ -30,8 +30,8 @@ type StateResponse struct {
 	Groups map[string]struct {
 		Members map[string]struct {
 			Addr       string `json:"addr"`
-			GroupId    int    `json:"groupId"`
-			Id         string `json:"id"`
+			GroupID    int    `json:"groupId"`
+			ID         string `json:"id"`
 			LastUpdate string `json:"lastUpdate"`
 			Leader     bool   `json:"leader"`
 		} `json:"members"`
--- a/z/zero.go
+++ b/z/zero.go
@@ -22,6 +22,13 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strconv"
+	"strings"
+
+	"github.com/dgraph-io/dgo"
+	"github.com/dgraph-io/dgo/protos/api"
+	"github.com/dgraph-io/dgraph/z"
+	"google.golang.org/grpc"
 )
 
 // StateResponse represents the structure of the JSON object returned by calling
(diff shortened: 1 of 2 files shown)


z/zero.go
```diff
@@ -28,7 +28,13 @@ import (
 // the /state endpoint in zero.
 type StateResponse struct {
 	Groups map[string]struct {
-		Members map[string]interface{} `json:"members"`
+		Members map[string]struct {
+			Addr       string `json:"addr"`
+			GroupId    int    `json:"groupId"`
```
golangcibot: struct field `GroupId` should be `GroupID` (from `golint`)

z/zero.go
```diff
@@ -28,7 +28,13 @@ import (
...
+			Id         string `json:"id"`
```
golangcibot: struct field `Id` should be `ID` (from `golint`)

z/zero.go
```diff
@@ -64,3 +77,39 @@ func GetState() (*StateResponse, error) {
 	}
 	return &st, nil
 }
+
+func GetClientToGroup(groupId string) (*dgo.Dgraph, error) {
```
golangcibot: func parameter `groupId` should be `groupID` (from `golint`)

PR #3412, opened 2019-05-14
New upsert block

mangalaman93:
Related to #3197 
Fixes #3059 

Example upsert operation -
```
   upsert {
      query {
        me(func: eq(email, "aman@dgraph.io")) {
          v as uid
        }
      }
      mutation {
          set {
            uid(v) <name> "Aman" .
            uid(v) <email> "aman@dgraph.io" .
          }
      }
    }
```

Probably in the next version -
 * [ ] Add support for multi valued variables
 * [ ] Support for val function
 * [ ] Add support for `@if`
 * [ ] Fuzz tests

Changes in other repos -
 * [x] Add tests in dgo repo https://github.com/dgraph-io/dgo/pull/68
 * [x] Python Client https://github.com/dgraph-io/pydgraph/pull/74
 * [x] Java Client https://github.com/dgraph-io/dgraph4j/pull/88
 * [x] JS Client https://github.com/dgraph-io/dgraph-js/commit/21563532863a01a639b4ddf4848e3de3de77b3e0

Todo 
 * [x] Add lots of tests for RDF and JSON
 * [x] Add documentation (docs.dgraph.io)
 * [x] ~~Add support for tracing~~
 * [x] JSON format for upsert block
 * [x] Add V(3) logging
 * [x] Support for uid function
 * [x] Fragments in upsert block
 * [x] ~~Two query blocks inside upsert block?~~
 * [x] no query block inside upsert block?
 * [x] no mutation block inside upsert block?
 * [x] no variable in query block?
 * [x] support for blank nodes in mutation block?
 * [x] Ensure upsert only generates 1 UID

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3412)

PR #3539, opened 2019-06-07
[Upsert #3] Add support for parsing functions in RDF (#3412)

mangalaman93:
Following are now possible (useful in upsert API)
  * uid(v) <predicate> "object" .
  * <0x01> <predicate> uid(foo) .

For now, we only support uid function. In future, we can
add support for more functions, starting with `val`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3539)


mangalaman93:
Closing this PR for #3412 , addressed all the comments.

Issue #3528, opened 2019-06-06
Count index missing a node after live import


mehdiym:
- What version of Dgraph are you using?
1.0.15

- What is the hardware spec (RAM, OS)?
16Go RAM, Linux Debian stable

- Steps to reproduce the issue (command/config used to run Dgraph).
  * set a schema including a @count index (and a @reverse index) on a predicate
  * live load a previous export
  * query with a count function at root : eq(count(predicate), 1)

- Expected behaviour and actual result.
Expected : No count different than 1
Actual : One node with a count greater than one

My request in ratel:
```
{
  me(func: eq(count(answer), 1)) {
    count(answer)
  }
}
```

The response:
```
{
  "extensions": {
    "server_latency": {
      "parsing_ns": 18314,
      "processing_ns": 2869208,
      "encoding_ns": 1473700
    },
    "txn": {
      "start_ts": 1163
    }
  },
  "data": {
    "me": [
      {
        "count(answer)": 2
      },
      {
        "count(answer)": 1
      },
      {
        "count(answer)": 1
      },
      {
        "count(answer)": 1
      }
    ]
  }
}
```

If I go to the schema tab, disable the @count index and re-enable it, then the query works as expected.

Note: My live import inserts many nodes but it is always that same node that gets wrong.

MichelDiz:
Hi, Can you share a RDF example that we can reproduce this?

danielmai:
Interesting that re-indexing the `@count` index makes the query work as expected.

I can reproduce the issue the steps you shared. Funny, when I run `dgraph live --conc 1` to turn off concurrent mutations (default concurrency is 10), the query works as expected. Even `--conc 2` shows the query issue. I suspect this issue is related with concurrent mutations to the same subject-predicate.

mehdiym:
I'm "glad" you can reproduce it.
Please note that I just reproduced the same bug but on nodes that were NOT imported by the live tool: on a fully generated database.
It happens during running my high concurrency integration tests.

danielmai:
This issue is fixed with #3545 that has been merged in master.

PR #3545, opened 2019-06-10
Handle concurrent update to count index

gitlw:
Before this change, two interleaving transactions can both update the count index regarding the same uid as shown in the test method TestCountIndexConcurrentTxns included in this PR. In the test, both txn1 and txn2 are trying to add an answer to the same uid <0x01>, and the sequnence of events are
1. txn1 starts at timestamp t0
2. txn2 starts at timestamp t1
3. txn1 commits at timestamp t2
4. txn2 commits at timestamp t3

each transaction would use a value of 0 for the count, and write an updated value of 1 to the count, leaving two entries in badger both having the key `answer 1`, and value `uid <0x01>`.
Later in the test, when we query for uids with `eq(count(answer), 1)`, the entries in badger would match the query, and thus the uid 0x01 is included in the query result.

The root cause of the problem is the concurrent write to the count index regarding the same uid. Hence this PR adds the key "(ctnIdx, attribute, uid)" as a conflict key to a transaction updating the count index, which cause subsequent transactions with concurrent writes to abort.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3545)


gitlw:
fixes #3528

PR #3552, opened 2019-06-11
External ids and the new upsert transaction.

MichelDiz:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3552)

### DIFF
--- a/wiki/content/mutations/index.md
+++ b/wiki/content/mutations/index.md
@@ -204,6 +204,28 @@ Now you can create new persons and attach its type using upsert transaction.
     }
 ```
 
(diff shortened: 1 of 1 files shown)


MichelDiz:
@MichaelJCompton okay, BTW do you think I should revamp the whole Extrernal IDs topic or let this PR as extra topic? I think the new txn is way better.

MichaelJCompton:
@MichelDiz I think the first part of xid is just intro, so I think this should be another section as you have it (feel free to suggest edits to the rest of the section as well though)

P.S. one small thing:

"Now you can create new persons" -> "Now you can create a new person"

PR #3560, opened 2019-06-14
Sending live loader error output to Stderr

gitlw:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3560)


danielmai:
___
*[dgraph/cmd/live/batch.go, line 125 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/3560#-LhMnhKo9oueBDOzYyiU:-LhMr9aYCH13885ndj74:bb4v3l1) ([raw file](https://github.com/dgraph-io/dgraph/blob/91ed74701ad2d49409a5511987142962ce746dc6/dgraph/cmd/live/batch.go#L125)):*
<details><summary><i>Previously, gitlw (Lucas Wang) wrote…</i></summary><blockquote>

I agree. So how about we add a -v flag to indicate whether these log lines should be shown? 
</blockquote></details>

Live loader already has `-v` as a global flag with glog around, and we make a deliberate choice to not use glog with tools like live loader. It makes it more confusing, but we could have a different flag to configure log verbosity for this.

manishrjain:
Is this being worked on? Or, can be closed?

PR #3583, opened 2019-06-20
Refactor Mutate function for processing mutation

mangalaman93:
Refactoring ensures the following -
  - The `doMutate` function was complex before doing things in sequence without any particular reasoning. Now we do the context check, health check, mutation allowed check first. Then we do authorization check, process query in case of upsert block and finally process mutation.
  - We are using `mu` (unparsed mutation) sometimes and `gmu` (parsed mutation) sometimes. Now it uses `gmu` once parsing is done.
  - We were passing and returning too many arguments before.
  - we are computing parsing and processing time more accurately now.

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3583)

### DIFF
--- a/edgraph/server.go
+++ b/edgraph/server.go
@@ -430,120 +430,84 @@ func (s *Server) Mutate(ctx context.Context, mu *api.Mutation) (*api.Assigned, e
 }
 
 func (s *Server) doMutate(ctx context.Context, mu *api.Mutation, authorize bool) (
-	resp *api.Assigned, rerr error) {
-
-	var l query.Latency
-	l.Start = time.Now()
-	gmu, err := parseMutationObject(mu)
-	if err != nil {
-		return resp, err
-	}
-	parseEnd := time.Now()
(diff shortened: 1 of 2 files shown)


edgraph/server.go
```diff
@@ -430,120 +430,84 @@ func (s *Server) Mutate(ctx context.Context, mu *api.Mutation) (*api.Assigned, e
...
-	if !isMutationAllowed(ctx) {
-		return nil, errors.Errorf("No mutations allowed.")
+	startParsingTime := time.Now()
+	gmu, err := parseMutationObject(mu)
+	if err != nil {
+		return resp, err
 	}
-	if mu.StartTs == 0 {
-		mu.StartTs = State.getTimestamp(false)
+	parsingTime += time.Now().Sub(startParsingTime)
```
golangcibot: S1012: should use `time.Since` instead of `time.Now().Sub` (from `gosimple`)

PR #3600, opened 2019-06-25
Migrating the GetDgraphClient logic into package x

gitlw:
The live loader, increment, and acl sub-commands all share similar options when creating the dgraph client, which include
// --alpha specifies a comma separated list of endpoints to connect to
// --tls_cacert, --tls_cert, --tls_key etc specify the TLS configuration of the connection
// --retries specifies how many times we should retry the connection to each endpoint upon failures
// --user and --password specify the credentials we should use to login with the server

This PR refactors the logic of checking those options into a method in package x, and uses that method in the different sub-commands.

A side effect of this PR is adding the user login feature in the live loader before processing data.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3600)

### DIFF
--- a/x/x.go
+++ b/x/x.go
@@ -565,6 +565,7 @@ func GetDgraphClient(conf *viper.Viper, login bool) (*dgo.Dgraph, CloseFunc) {
 	Checkf(err, "While loading TLS configuration")
 
 	ds := strings.Split(alphas, ",")
+	var conns []*grpc.ClientConn
 	var clients []api.DgraphClient
 
 	retries := 1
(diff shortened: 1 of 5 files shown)


x/x.go
```diff
@@ -539,3 +546,115 @@ func SpanTimer(span *trace.Span, name string) func() {
...
+func GetPassAndLogin(dg *dgo.Dgraph, opt *CredOpt) (CloseFunc, error) {
+	password := opt.Conf.GetString(opt.PasswordOpt)
+	if len(password) == 0 {
+		var err error
+		password, err = AskUserPassword(opt.Userid, "Current", 1)
+		if err != nil {
+			return nil, err
+		}
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
golangcibot: lostcancel: the cancel function is not used on all paths (possible context leak) (from `govet`)

x/x.go
```diff
@@ -539,3 +546,115 @@ func SpanTimer(span *trace.Span, name string) func() {
...
+	if err := dg.Login(ctx, opt.Userid, password); err != nil {
+		return nil, fmt.Errorf("unable to login to the %v account:%v", opt.Userid, err)
```
golangcibot: lostcancel: this return statement may be reached without using the cancel var defined on line 653 (from `govet`)

x/x.go
```diff
@@ -539,3 +546,123 @@ func SpanTimer(span *trace.Span, name string) func() {
...
+		for _, c := range conns {
+			c.Close()
+		}
+	}
+	return dg, closeFunc
+}
+
+type CredOpt struct {
+	Conf        *viper.Viper
+	UserId      string
```
golangcibot: struct field `UserId` should be `UserID` (from `golint`)

PR #3612, opened 2019-06-28
Add support for conditional mutation in Upsert Block

### LINKED ISSUE #3059: Support a new Upsert operation

mangalaman93:
Fixes #3059 
Related to #3412 

Todo -
 * [x] Lexing and Parsing
 * [x] Execution support
 * [x] Add support for multi valued variables
 * [x] Support for JSON API
 * [ ] Support for val function
❌ Support for multiple mutation in single upsert block
 * [ ] Fuzz tests
 * [ ] Update docs

Changes in other repos -
 * [ ] Add tests in dgo repo
 * [ ] Python Client
 * [ ] Java Client
 * [ ] JS Client

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3612)


edgraph/server.go
```diff
@@ -577,40 +579,79 @@ func doQueryInUpsert(ctx context.Context, mu *api.Mutation, gmu *gql.Mutation) (
 		return l, nil
 	}
 
+	upsertQuery := mu.Query
 	needVars := findVars(gmu)
+	isCondUpsert := strings.TrimSpace(mu.Cond) != ""
+	varName := fmt.Sprintf("__dgraph%d__", rand.Int())
```
golangcibot: G404: Use of weak random number generator (math/rand instead of crypto/rand) (from `gosec`)

mangalaman93: This is not for security!

Issue #3607, opened 2019-06-27
Query error after bulk load


danielmai:
If you suspect this could be a bug, follow the template.

- What version of Dgraph are you using?
master, starting with d336e41de.

- Have you tried reproducing the issue with latest release?
Yes

- Steps to reproduce the issue (command/config used to run Dgraph).

Run ./systest/21million/test-21million.sh.

The test fails after running this query:

Query:
```
{
  me(func: eq(count(genre), 13)) {
    name@en
    genre {
    	name@en
    }
  }
}
```

Response:
```
{
  "errors": [
    {
      "code": "ErrorInvalidRequest",
      "message": ": Unexpected meta: 0 for key: 00000000  00 00 05 67 65 6e 72 65  08 00 00 00 00 0d        |...genre......|\n"
    }
  ],
  "data": null
}
```

- Expected behaviour and actual result.
The test pass. This test passes in the previous commit before #3542.

PR #3628, opened 2019-07-03
Fix user meta of schema and type entries in bulk loader.

martinmr:
The user meta should not be BitCompletePosting since that indicates the
value is a posting list.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3628)


campoy:
Was there an issue for this? Please link it on the PR description with a `Fixes #xxx`

martinmr:
No issue for this. I was looking at another issue when I noticed this.

danielmai:
Would this fix #3607?

martinmr:
No. I found this issue while investigating that issue but it's not the fix.

PR #3648, opened 2019-07-10
Use manifest to only restore preds assigned to each group.

martinmr:
This PR changes the restore logic so that only the keys corresponding to
the predicates that were assigned to the group at the time of the last
backup are restored. This change is intended to properly deal with
predicate moves.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3648)


martinmr:
Marked as draft since the base is another branch currently on review but this change is ready for review.

PR #3662, opened 2019-07-12
Fix crash when trying to use shortest path with a password predicate.

### LINKED ISSUE #3657: Alpha crash with shortest path query with password predicate

docker latest and play.dgraph.io

Yes

Various N/A

Send a query with a predicate containing at least one predicate of type password.
For instance, you can use the internal `dgraph.password`, but any other will also cause the issue.

```
{
  path as shortest(from: 0x1, to: 0x2) {
	  <dgraph.password>
  }
  path(func: uid(path)) {
  	name
	}
}
```

I expected to get a result, instead I get a panic:

```
panic: runtime error: index out of range

goroutine 2248 [running]:
github.com/dgraph-io/dgraph/query.(*SubGraph).expandOut(0xc0015d2000, 0x4f997e0, 0xc012fc8f30, 0xc0148591a0, 0xc012fc22a0, 0xc016d28a80)
	/Users/francesc/src/github.com/dgraph-io/dgraph/query/shortest.go:186 +0xfca
created by github.com/dgraph-io/dgraph/query.shortestPath
	/Users/francesc/src/github.com/dgraph-io/dgraph/query/shortest.go:455 +0x226
```

martinmr:
The crash happens because querying this predicate does not populate the
uid matrix, resulting in an out-of-index error. This fix just checks the
size of the uid matrix before trying to access it.

Fixes #3657

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3662)

### DIFF
--- a/query/shortest.go
+++ b/query/shortest.go
@@ -185,7 +185,7 @@ func (sg *SubGraph) expandOut(ctx context.Context,
 				for mIdx, fromUID := range subgraph.SrcUIDs.Uids {
 					// This can happen when trying to go traverse a predicate of type password
 					// for example.
-					if len(subgraph.uidMatrix) < mIdx+1 {
+					if mIdx >= len(subgraph.uidMatrix) {
 						continue
 					}
 
--- a/query/shortest.go
+++ b/query/shortest.go
@@ -183,6 +183,12 @@ func (sg *SubGraph) expandOut(ctx context.Context,
 
 				// Send the destuids in res chan.
 				for mIdx, fromUID := range subgraph.SrcUIDs.Uids {
+					// This can happen when trying to go traverse a predicate of type password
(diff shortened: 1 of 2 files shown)


query/shortest.go
```diff
@@ -183,6 +183,12 @@ func (sg *SubGraph) expandOut(ctx context.Context,
 
 				// Send the destuids in res chan.
 				for mIdx, fromUID := range subgraph.SrcUIDs.Uids {
+					// This can happen when trying to go traverse a predicate of type password
+					// for example.
+					if len(subgraph.uidMatrix) < mIdx+1 {
```
campoy:
```suggestion
					if mIdx >= len(subgraph.uidMatrix) {
```

PR #3666, opened 2019-07-13
Adding the feature to allow an alpha go into lame duck mode

gitlw:
In the lame duck mode, the alpha server will reject all requests from the client.
This feature will be useful during a restore operation, or a point-in-time-recovery while the cluster is running.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3666)

### DIFF
--- a/dgraph/cmd/alpha/run.go
+++ b/dgraph/cmd/alpha/run.go
@@ -388,6 +388,7 @@ func setupServer() {
 	http.HandleFunc("/debug/store", storeStatsHandler)
 
 	http.HandleFunc("/admin/shutdown", shutDownHandler)
+	http.HandleFunc("/admin/lame", lameHandler)
 	http.HandleFunc("/admin/export", exportHandler)
 	http.HandleFunc("/admin/config/lru_mb", memoryLimitHandler)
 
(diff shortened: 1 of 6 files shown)


campoy:
I know this might seem overly politically correct, but I'd rather bring this up now before the code is committed and we can't change APIs easily.

"Lame" is an [ableist term](https://www.merriam-webster.com/dictionary/lame), would you consider using "draining" instead? "draining mode" not only avoids those connotations but also it's much easier to understand than "lame duck" which is a pretty technical term.

dgraph/cmd/alpha/run.go
```diff
@@ -279,13 +280,15 @@ func healthCheck(w http.ResponseWriter, r *http.Request) {
 	}
 
 	info := struct {
-		Version  string        `json:"version"`
-		Instance string        `json:"instance"`
-		Uptime   time.Duration `json:"uptime"`
+		Version      string        `json:"version"`
+		Instance     string        `json:"instance"`
+		Uptime       time.Duration `json:"uptime"`
+		DrainingMode bool          `json:drainingMode`
```
golangcibot: structtag: struct field tag `json:drainingMode` not compatible with reflect.StructTag.Get: bad syntax for struct tag value (from `govet`)

manishrjain:
This PR still needs work.

Issue #3642, opened 2019-07-09
Alpha crashes with fault address in groupby.go


jarifibrahim:
### What version of Dgraph are you using?
727b246b5147d5c652e4e752938d20d402722cc0

### Have you tried reproducing the issue with latest release?
Yes

### What is the hardware spec (RAM, OS)?
N/A

### Steps to reproduce the issue (command/config used to run Dgraph).
1. Run zero using 
```
dgraph zero -w /home/ibrahim/m2_drive/flock_backup/zero1/zw -o 0 --idx=1 --my=localhost:5080 --replicas=1 --logtostderr -v=2 --bindall
```
2. Run alpha using
```
dgraph alpha -p /home/ibrahim/m2_drive/flock_backup/alpha1/p -w /home/ibrahim/m2_drive/flock_backup/alpha1/w -o 100 --my=localhost:7180 --lru_mb=1024 --zero=localhost:5080 --logtostderr -v=2
```
4. Apply the following diff on flock agent's main.go
```diff
diff --git a/main.go b/main.go
index c2cf6ec..91e8da0 100644
--- a/main.go
+++ b/main.go
@@ -52,8 +52,8 @@ id_str: string @index(exact) @upsert .
 created_at: dateTime @index(hour) .
 hashtags: [string] @index(exact) .
 
-author: uid @count @reverse .
-mention: uid @reverse .
+author: [uid] @count @reverse .
+mention: [uid] @reverse .
 `
 
        cDgraphTweetQuery = `
```
3. Run flock agent
```
go run main.go -a ":9180"
```
4. Apply the following to flock client's main.go (without this change some of the queries in flock were failing)
```diff
diff --git a/client/main.go b/client/main.go
index c1eab19..5913475 100644
--- a/client/main.go
+++ b/client/main.go
@@ -547,7 +547,7 @@ type querySix struct {
 
 func (q *querySix) getParams(dgr *dgo.Dgraph) error {
        // we subtract 41 hours because that's the latest data we get from twitter
-       curTime := time.Now().Add(-41 * time.Hour)
+       curTime := time.Now().Add(-100 * time.Hour)
 
        query := fmt.Sprintf(`
 {
@@ -619,7 +619,7 @@ type querySeven struct {
 
 func (q *querySeven) getParams(dgr *dgo.Dgraph) error {
        // we subtract 41 hours because that's the latest data we get from twitter
-       curTime := time.Now().Add(-41 * time.Hour)
[...]

PR #3670, opened 2019-07-15
Workaround for issue #3642.

martinmr:
For some reason, there are nil values stored in the uidMatrix, which
results in a segmentation fault when trying to access the Uids field.
The root cause of this issue is still unclear but this change adds an
additional check to prevent the segmentation fault.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3670)

### DIFF
--- a/query/groupby.go
+++ b/query/groupby.go
@@ -221,7 +221,12 @@ func (sg *SubGraph) formResult(ul *pb.List) (*groupResults, error) {
 				if algo.IndexOf(ul, srcUid) < 0 {
 					continue
 				}
+
 				ul := child.uidMatrix[i]
+				// Workaround for issue #3642.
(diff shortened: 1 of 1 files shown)


query/groupby.go
```diff
@@ -221,7 +221,12 @@ func (sg *SubGraph) formResult(ul *pb.List) (*groupResults, error) {
 				if algo.IndexOf(ul, srcUid) < 0 {
 					continue
 				}
+
 				ul := child.uidMatrix[i]
+				// Workaround for issue #3642.
+				if ul == nil  {
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
				if ul == nil {
```

PR #3679, opened 2019-07-17
Increase Alpha trace logs.

danielmai:
This change bumps up the number of logs per span to capture all the events from Server.Query instead of getting truncated logs in the trace data of up to 64 of the latest annotations.

When I load the 21-million movie data set and run a query to get all the movies directed by Steven Spielberg, Jaeger shows 232 logs for the query. With this change, the logs would be truncated if there are more than 256 annotations in a single span, leaving us to the same problem of possibly losing valuable info for a query trace.

```
{
  director(func: allofterms(name@en, "steven spielberg")) {
    name@en
    director.film {
      name@en
      initial_release_date
      country {
        name@en
      }
      starring {
        performance.actor {
          name@en
        }
        performance.character {
          name@en
        }
      }
      genre {
        name@en
      }
    }
  }
}
```

image

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3679)

### DIFF
--- a/dgraph/cmd/alpha/run.go
+++ b/dgraph/cmd/alpha/run.go
@@ -517,7 +517,7 @@ func run() {
 	}
 	otrace.ApplyConfig(otrace.Config{
 		DefaultSampler:             otrace.ProbabilitySampler(x.WorkerConfig.Tracing),
-		MaxAnnotationEventsPerSpan: 64,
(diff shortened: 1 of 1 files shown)


danielmai:
Thanks @dgraph-bot :thinking: and @gitlw.

PR #3686, opened 2019-07-18
Use the stream framework to rebuild indices.

martinmr:
The stream framework can iterate over the index prefix keys concurrently
whereas currently there's only one thread iterating over the keys.
Changing to the stream framework should provide a speed up over the
current approach.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3686)

### DIFF
--- a/posting/index.go
+++ b/posting/index.go
@@ -530,7 +530,9 @@ func (r *rebuild) Run(ctx context.Context) error {
 		// so this function doesn't have any work to do.
 		return nil
 	}
-	stream.Orchestrate(ctx)
+	if err := stream.Orchestrate(ctx); err != nil {
+		return err
+	}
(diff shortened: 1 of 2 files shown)


posting/index.go
```diff
@@ -489,56 +490,51 @@ type rebuild struct {
...
 		}
+
+		return nil, nil
+	}
+	stream.Send = func(*bpb.KVList) error {
+		// The work of adding the index edges to the transaction is done by r.fn
+		// so this function doesn't have any work to do.
+		return nil
 	}
+	stream.Orchestrate(ctx)
```
golangcibot: Error return value of `stream.Orchestrate` is not checked (from `errcheck`)

martinmr:
The unit tests checking the rebuild process still pass. I manually checked that I could do full text queries after a rebuild with that index.

PR #3688, opened 2019-07-18
Rebuild reverse index before count reverse.

martinmr:
Not sure if this was causing issues before, but rebuilding the count
reverse index depends on the reverse index, so I am chaning the logic to
rebuild the reverse index first, followed by the count index.

Also adding a comment to when the count reverse index is rebuilt.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3688)


campoy:
Is there any chance you could have a test showing how this was previously broken and now behaves correctly?

PR #3692, opened 2019-07-19
Fix how enconding_ns is calculated.

martinmr:
Currently, the time taken to receive the query response is included as
part of the encoding time. This PR fixes that calculation.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3692)


gitlw:
Can you please define what the l.Transport means and the logic behind this equation?
l.Transport = time.Since(l.Start) - l.Parsing - l.Processing

martinmr:
I'll add a comment in the next PR to avoid having to run teamcity tests again, but transport is the time taken from the completion of the query to receiving the response (sometimes the query is processed by a different alpha). We are doing this because while debugging another issue, we noticed that the encoding time contains the time to receive the response.

PR #3694, opened 2019-07-19
Add transport_ns value to query responses.

### LINKED ISSUE #3690: encoding_ns might not be accurate
From #3668 the user reports that the value of `encoding_ns` is roughly 100ms from an AWS zone with a latency of 100ms. This probably means the transport time is being included as part of encoding_ns. If that's the case, Dgraph should be changed to report accurate stats. Possibly, that could involve adding another field called `transport_ns` to report the time taken to receive the response.

martinmr:
Fixes #3690

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3694)


martinmr:
Closing this as the changes are actually more involved. The delay cause by latency happens while waiting for a timestamp so I am going to add a stat for that stage. Also, the processing time should include the time needed  to get the schema or types.

PR #3707, opened 2019-07-23
adding the draining mode

gitlw:
The draining mode is used to disable all requests from the user, which can be useful during online restore or point-in-time-recovery.

Besides, this PR also refactored the implementation of the http /commit endpoint to call the Server.CommitOrAbort method, which already has the health check logic built in.

### DIFF
--- a/dgraph/cmd/alpha/http.go
+++ b/dgraph/cmd/alpha/http.go
@@ -440,13 +440,11 @@ func handleCommitOrAbort(startTs uint64, reqText []byte, abort bool) (map[string
 	}
 
 	var reqMap map[string][]string
-	if err := json.Unmarshal(reqText, &reqMap); err != nil && !useList {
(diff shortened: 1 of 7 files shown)


dgraph/cmd/alpha/http.go
```diff
@@ -432,29 +427,10 @@ func commitHandler(w http.ResponseWriter, r *http.Request) {
...
-		}, nil
-	case nil:
-		return nil, errors.Errorf("transaction could not be aborted")
-	default:
-		return nil, err
-	}
-}
-
-func handleCommit(startTs uint64, reqText []byte) (map[string]interface{}, error) {
+func handleCommitOrAbort(startTs uint64, reqText []byte, abort bool) (map[string]interface{}, error) {
```
golangcibot: line is 102 characters (from `lll`)

PR #3724, opened 2019-07-24
Reduce memory consumption in bulk loader

gitlw:
Sometimes bulk loader runs into OOM, and gets killed by the OS on my local machine, which has 32G of RAM.
I ran the pprof profiling, and it turns out most of the allocations happened inside the x.ReserveCap function, which allocates at least 2 times of an existing slice's current capacity.

While trying to avoid calling the x.ReserveCap function, I also find the map entries (pb.MapEntry) go through the following steps
1. they are marshalled to a shard's buffer (sh.entriesBuf) inside mapper.addMapEntry
2. in writeMapEntriesToFile, they are unmarshalled back into map entries (pb.MapEntry), sorted and then marshalled into the buf again

This PR tries to avoid the double marshaling by initially storing map entries in each shard, and only marshal them once after they are sorted inside the writeMapEntriesToFile.

Evaluation:
1. The bulk load test using the 21 million data set is still passing by running 
cd systest/21million
./test-21million.sh

2. After this change, the peak memory usage indicated by go_memstats_heap_inuse_bytes metric is reduced to 11.82GB vs 15.73GB before this change.
Screenshot from 2019-07-24 15-29-53
PS: the stats on the left side are based on running bulk loader after this change, and the stats on the right side are based on running bulk loader before this change.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3724)


dgraph/cmd/bulk/mapper.go
```diff
@@ -78,34 +79,24 @@ func less(lhs, rhs *pb.MapEntry) bool {
...
-		buf = buf[sz:]
-		entries = append(entries, me)
-	}
-
 	sort.Slice(entries, func(i, j int) bool {
 		return less(entries[i], entries[j])
 	})
 
-	buf = entriesBuf
+	buf := make([]byte, encodedSize, encodedSize)
```
golangcibot: S1019: should use make([]byte, encodedSize) instead (from `gosimple`)

gitlw:
By compressing the intermediate mapper out files into gzipped files, we are able to significantly reduce the disk space consumption. When bulk loader is run using the 21million data set, the disk space consumption is reduced from ~3GB to ~500MB
https://docs.google.com/spreadsheets/d/1S4wcgZQijl59Vqc_W-DzqHePdE3lzAYyPqKaIvD-QPs/edit?usp=sharing

PR #3728, opened 2019-07-29
Return GraphQL compliant errors

MichaelJCompton:
Add GraphQL spec compliant errors to x package.  That means alpha http endpoint should return GraphQL compliant errors.

Change from current structure is where `code` sits.  With this PR it moves to `extensions`.  Should we call this **breaking** and move into 1.1 release?

This changes means Alpha and GraphQL layer can use the same sorts of errors and responses and moves Dgraph closer to GraphQL compliant.

Maybe we should consider using GQL Locations structure for reporting locations of Dgraph query parsing errors as well?

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3728)

### DIFF
--- a/dgraph/cmd/alpha/http_test.go
+++ b/dgraph/cmd/alpha/http_test.go
@@ -38,15 +38,10 @@ import (
 	"github.com/dgraph-io/dgraph/z"
 )
 
-type respError struct {
-	Code    string `json:"code"`
-	Message string `json:"message"`
-}
-
 type res struct {
 	Data       json.RawMessage   `json:"data"`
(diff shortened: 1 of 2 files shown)


MichaelJCompton:
Doesn't semver say that breaking changes go into minor releases, so from 1.0.16->1.1.0 is the right spot?   i.e. just not in patch releases.

I think there's been some changes to Dgraph query parsing to report error locations.  Would make for nicer error messages to have that info.

MichaelJCompton:
P.S. GraphQL layer has the locations info and already reports that - e.g. on type errors or parse errors etc.  

It does it atm using a different struct, will be using this, once this merges.

PR #3745, opened 2019-08-02
adding back the go runtime metrics about CPU and memory

### LINKED ISSUE #3717: The default prometheus metrics are missing in current master branch

version 1.1 (not released yet)

Yes

32G RAM, Fedora

In the compose directory, do ./run.sh -a 1 -z 1 -m
Check the prometheus metrics at localhost:8180/debug/prometheus_metrics
and the default premetheus metrics such as go_goroutines, go_memstats_heap_alloc_bytes are not there (these stats are available in the release/v1.0 branch) 

The stats should still be there as they can be very important for troubleshooting.

gitlw:
fixes #3717 

Verified this change by checking the localhost:8180/debug/prometheus_metrics page, and confirmed that the following metrics are showing up
go_goroutines
go_threads
go_memstats_heap_alloc_bytes
dgraph_memory_inuse_bytes
go_memstats_heap_idle_bytes
...

Also removing our own NumGoRoutines metric, since it's already covered by prometheus.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3745)

### DIFF
--- a/x/metrics.go
+++ b/x/metrics.go
@@ -26,9 +26,10 @@ import (
 	"go.opencensus.io/trace"
 
 	"contrib.go.opencensus.io/exporter/jaeger"
-	"contrib.go.opencensus.io/exporter/prometheus"
+	oc_prom "contrib.go.opencensus.io/exporter/prometheus"
 	datadog "github.com/DataDog/opencensus-go-exporter-datadog"
 	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/spf13/viper"
 	"go.opencensus.io/stats"
 	"go.opencensus.io/stats/view"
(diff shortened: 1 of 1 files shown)


odeke-em:
Thank you for working on this @gitlw!

 LGTM, but it would be great to perhaps have a unit test to ensure that we don't regress.
We can perhaps make a test for this by:
1. A creating a handler that uses `x/metrics` and running it on a server using say https://golang.org/pkg/net/http/httptest/#NewServer
2. Scraping that endpoint after a couple of milliseconds
3. Verifying that those stats are present in the scrape

gitlw:
@odeke-em Great suggestion. I'll raise a separate PR for the test.

PR #3769, opened 2019-08-07
Bugfix: Use DestUids while doing the comparison for length function.

pawanrawal:
The current comparison was incorrect as we should be using DestUids because the uid variable is populated into them.

Related to #3756 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3769)


pawanrawal:
Doesn't look like I'll be able to merge this today because of CI tests which are failing because they are flaky.

PR #3771, opened 2019-08-07
Remove schema equality functions

animesh2049:
Remove all the equality check functions. Sort keys of schema.Types
in Stringify so that it generates the same schema everytime.
Also sort the Types while generating extra types. They are appending
query and mutation fields in arrays, so if they aren't added in same
order everytime, Stringify  will generate different schema strings.
We aren't sorting arrays in Stringify.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3771)


animesh2049:
Okay merging now.

PR #3778, opened 2019-08-08
Add a GraphQL request id

MichaelJCompton:
Adds middleware to add a request ID to all GraphQL requests.

That ID then appears in all logs and error messages.

Both logs and error messages are currently unsettled in GraphQL layer.  For logging, I expect we'll implement structured logging pretty early on.  I also think a PR soon will add returning some data (Dgraph query times and other metrics) in the GraphQL "extensions" field (as per Apollo metrics), so it's likely request ID should go in there and not in error messages.

For now, this is good enough that we can run a server, have some meaningful logs and correlate errors to log messages for debugging.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3778)


dgraph/cmd/graphql/api/context.go
```diff
@@ -0,0 +1,73 @@
...
+// throughout the GraphQL API.
+
+type apiKey string
+
+const (
+	requestID    = apiKey("requestID")
+	nilRequestID = "a1111111-b222-c333-d444-e55555555555"
+
+	gqlQuery    = apiKey("gqlQuery")
+	nilGQLQuery = "aGraphQLQuery {}"
```
golangcibot: `nilGQLQuery` is unused (from `varcheck`)

MichaelJCompton:
Merging without review from Pawan, because it's just enough to link logging and request IDs.  This all change as logging and responses get more sophisticated.

manishrjain:
Hey @MichaelJCompton , Look into open census. We use that in Dgraph. If you start using that, then we can track an individual query from the client, to the GraphQL server, to Dgraph DB internally. That'd be a better way to track queries.

PR #3786, opened 2019-08-10
Print commit SHA256 when invoking "make install"

martinmr:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3786)

### DIFF
--- a/dgraph/Makefile
+++ b/dgraph/Makefile
@@ -61,6 +61,7 @@ uninstall:
 	@go clean -i -x
 
 install:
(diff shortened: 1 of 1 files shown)


danielmai:
> Is this based on some request? I had a chat with Daniel about Dgraph binary printing out the SHA256 checksum of the binary itself.
> 

It'd be a bit weird (if not practically impossible) to embed the SHA256 sum into the Dgraph binary for `dgraph version`. You'd have to know upfront what the checksum is and to insert that into the binary such that `sha256sum` would be the same. It's a chicken-and-the-egg problem.

PR #3791, opened 2019-08-12
[cherry-pick] Add cluster-only and full-only flags to test.sh.

martinmr:
./test.sh -C or ./test.sh -F can be used to run only the custom
cluster tests or only the long running systests. On my machine
these take approx. 7 minutes and 11 minutes to finish
respectively. These flags can be used to run them as separate
jobs in parallel in CI so we can run all the tests for each PR.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3791)

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -159,19 +159,22 @@ function RunCustomClusterTests {
 # MAIN
 #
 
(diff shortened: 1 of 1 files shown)


martinmr:
The test suites for release/v1.0 are failing because the script doesn't know about this flags. This should fix it.

PR #3792, opened 2019-08-12
Set default type predicates as lists.

### LINKED ISSUE #3788: Dgraph isn't automatically setting values ​​as list in structured mutations as list.

1.1RC

Yes, latest RC

Add a list mutation in JSON before set the schema. 

Since Dgraph automatically recognizes Types (Like `[uid]` this is a "list") I think it should recognize a list through a structured mutation as a list. Avoiding the user lost some data from his list. In case he forgets to define the scheme before mutating.

I didn't tested it in RDF tho. Will do.

Check reference: https://github.com/dgraph-io/dgraph/issues/3787

martinmr:
If users add multiple entries to predicates of default type, the current
behavior is to overwrite existing values. This can lead to data loss in
cases where users didn't intend to do that or forgot to set a schema for
the predicate beforehand.

This change makes it so that predicates of default type are created as
lists.

Fixes #3788

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3792)

### DIFF
--- a/dgraph/cmd/bulk/schema.go
+++ b/dgraph/cmd/bulk/schema.go
@@ -102,7 +102,11 @@ func (s *schemaStore) validateType(de *pb.DirectedEdge, objectIsUID bool) {
 		sch, ok = s.schemaMap[de.Attr]
 		if !ok {
 			sch = &pb.SchemaUpdate{ValueType: de.ValueType}
-			if objectIsUID {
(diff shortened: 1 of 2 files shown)


martinmr:
I think I need this to load the arango benchmark data without having to figure out the schema (I don't think any of their benchmarks care about the schema).

martinmr:
Doesn't work as expected.

PR #3806, opened 2019-08-13
release/v1.0.13: Increase max trace logs per span in Alpha.

danielmai:
Bump up the number of logs per span to capture all the events
from Server.Query which can create a lot of logs.

Cherry-picking to release/v1.0.13 branch so we can cut a new v1.0.13-opencensus release for folks to easily switch upgrade their v1.0.13 binaries to capture more OpenCensus span logs.

This change also vendors in dependencies that happened before or around the time of the v1.0.13 release back in March.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3806)


danielmai:
Tests pass when running ./test.sh.

danielmai:
The tests won't pass since the test scripts have changed in master since v1.0.13. Fixing the test scripts so they work in TeamCity is not a priority for this PR.

PR #3824, opened 2019-08-16
Enterprise - Support reading a signed file with details of Enterprise license

pawanrawal:
This change gets rid of the `--enterprise_features` flag in Zero. Instead users who want to use the enterprise feature, would have to supply a file while starting a Zero node. This file would contain would be signed by our private key and would contain details about the maximum number of nodes they can run and the expiry of the license. This information would automatically be propagated to Alpha nodes along with other membership info.

TODO
---
- [x] Allow specifying the file using an HTTP endpoint in Zero as well.
- [x] Add test cases for `enterpriseDetails` function to check PGP verification is working as expected.
- [ ] Add integration tests to test the feature end to end.
- [x] Make sure ACL is only enabled when a valid file is supplied.
- [x] Generate a public key and store it in code.
- [x] Fix existing tests and make scripts that are breaking after this change.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3824)


pawanrawal:
PR is ready to merge. Waiting for @danielmai to share the signed file that we'll upload to Teamcity and read as an environment variable.

PR #3827, opened 2019-08-16
Adding e2e test for GraphQL server

hackintoshrao:
- Changes to default docker compose file in root/dgraph.
- Added containers to initialize and run the GrapQL server.
- Test schema to initialize the GraphqL server.
- Added one unit test which tests for the response of the mutation.
- Any changes to the request and response format of the add mutation
  should be caught by this test.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3827)


MichaelJCompton:
Looks cool.  My only comment is that I prefer to be able to see a test case and expected results close together in the test setup.  So much easier to grok what a test is doing if you can see the whole case on one screen.

It doesn't matter ATM with just a couple of cases, but as there's more cases, it gets hard to navigate and work out what connects with what.  So if we could set it up so we can see input and expected output together, that would be great.

hackintoshrao:
> 
> > ```go
> > // Converts a multiline string to single line.
> > // The GraphQL server expects the mutation/query to be a single line string.
> > ```
> 
> I didn't know that was the case. I wonder why? It's not anything we are adding, if it needs that, then it's the gqlparser.

I checked my composing a RAW request too. We need to convert that query into single line string.. Guess, that's what `insomnia` and `postman` does.

PR #3830, opened 2019-08-16
Replace errors.Errorf with fmt.Errorf in AssertTruefNoTrace.

martinmr:
errors.Errorf prints the trace, which is not the right thing to do in
this function (it's specifically called when the trace is not wanted).

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3830)

### DIFF
--- a/x/error.go
+++ b/x/error.go
@@ -29,6 +29,7 @@ package x
 // (3) You want to generate a new error with stack trace info. Use errors.Errorf.
 
 import (
+	"fmt"
 	"log"
 	"os"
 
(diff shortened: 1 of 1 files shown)


danielmai:
:lgtm:

This change shows the expected output (no traces) when running `dgraph alpha` without required flags:

```
$ dgraph alpha
[Decoder]: Using assembly version of decoder
2019/08/16 13:22:16 LRU memory (--lru_mb) must be specified. (At least 1024 MB)
```

PR #3834, opened 2019-08-18
Change to GraphQL "data: null" interpretation

MichaelJCompton:
GraphQL spec says: "If an error was encountered during the execution that prevented a valid response, the data entry in the response should be null."

I had originally interpreted that as meaning that if there's any sort of "error" (couldn't rewrite query, failed to execute Dgraph query, etc.) that we should be returning `{ "data": null }`, but I don't think that's right.

If there's an error down in the pipeline, it get's caught, we add to the GraphQL error list, and set that result to null, so you might thus end up with something like

```
data : {
  q1 : null,
  q2 { ... }
},
```

Which is what the spec says elsewhere should happen.  Sol, I'm interpreting "prevented a valid response" as meaning something really bad happened.  ATM that's just if we fail to actually form a "valid response" as json.

I might also wrap the HTTP pipeline with middleware that will catch any panics, log them and also return `{ "data": null }`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3834)

### DIFF
--- a/dgraph/cmd/graphql/resolve/resolver.go
+++ b/dgraph/cmd/graphql/resolve/resolver.go
@@ -116,7 +116,6 @@ func (r *RequestResolver) Resolve(ctx context.Context) *schema.Response {
 	}
 
 	if r.resp.Errors != nil {
-		r.resp.Data.Reset()
 		return r.resp
 	}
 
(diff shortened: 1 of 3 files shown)


MichaelJCompton:
yeah, there's a couple of tests that show something like that in the resolver tests, but there should be some more integration tests around errors and what gets reported back.

PR #3836, opened 2019-08-20
Add Apollo tracing to GraphQL layer

MichaelJCompton:
Adds the Apollo tracing spec to GraphQL layer results: https://github.com/apollographql/apollo-tracing

Apollo tracing can be used like here to return some tracing data in "extensions" of the response payload, and can also push tracing info to Apollo Engine for dev-ops stats etc.  Here we just do the first part.

I'm not really keen to wrap any testing around this or to lock it in at this stage.  For the moment, it's just a nice dev artefact to get some insight into what's going on inside the pipeline for every request.  I expect pretty soon we'll add a flag to turn it off, add some E2E benchmark testing, and probably add the ability to push these stats to Apollo Engine.  At that point, we should lock in its behaviour, so we know it's working in a way that those external tools expect.   (there's cards on the Asana board for all those points.)

With this PR, responses always come back like:

image

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3836)

### DIFF
--- a/dgraph/cmd/graphql/dgraph/dgo.go
+++ b/dgraph/cmd/graphql/dgraph/dgo.go
@@ -39,11 +39,26 @@ import (
 // GraphQuery to alpha, so we don't have to stringify -> reparse etc.  Also
 // allows exercising some of the particulars around GraphQL error processing
 // without needing a Dgraph instance the reproduces the exact error conditions.
+//
+// Implementations should use dgTimer to track the total Dgraph time.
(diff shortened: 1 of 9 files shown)


MichaelJCompton:
Dropping this for a few days.  

Once Query changes have gone into the API, will pick back up again.

pawanrawal:
Closing this PR. Will be incorporating relevant bits into #3942.

PR #3844, opened 2019-08-21
Apply GraphQL best practice to updates

MichaelJCompton:
Small redesign of arguments for updates.

There's no clear "best practice" for GraphQL, but there's lots of arguments for things like:

"Mutations should only ever have one input argument. That argument should be named input and should have a non-null unique input object type."  (https://blog.apollographql.com/designing-graphql-mutations-e09de826ed97)

Marc-André Giroux from GitHub also has some nice articles : https://medium.com/@__xuorig__

There's a couple of PRs coming that smooth out some not so nice parts in our initial cut of schema generation.

This one:
- Sets the return types for the update mutation to be nullable
- makes a single (nested) input to updates (the nested input is a precursor to allowing user to specify multiple behavioural updates on the same type: e.g. updatePostText, etc)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3844)


MichaelJCompton:
Waiting on : https://github.com/dgraph-io/dgraph/pull/3812 before merging

PR #3856, opened 2019-08-23
Add docs for upsert

mangalaman93:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3856)

### DIFF
--- a/wiki/content/howto/index.md
+++ b/wiki/content/howto/index.md
@@ -620,7 +620,13 @@ if len(login.Account) == 0 {
 
 ## Upserts
 
-Upsert-style operations are operations where:
+Consider the following example. Let's say we want to update age of a user
+in the database. To do this, we need to first query the UID of the user.
+We can now use the UID to update the age of the user. This whole operation
+requires us two network round trip to the database from the client, one
+for querying the UID and second for updating the age using the UID.
+
+In general, Upsert-style operations are operations where:
 
 1. A node is searched for, and then
 2. Depending on if it is found or not, either:
@@ -662,8 +668,8 @@ whichever way you wish.
 
 ### Upsert Block
 
-You can also use the `Upsert Block` to achieve the upsert procedure in a single mutation.
-The request will contain both the query and the mutation as explained
+You can also use the `Upsert Block` to achieve the upsert procedure in a single
+ mutation. The request will contain both the query and the mutation as explained
 [here]({{< relref "mutations/index.md#upsert-block" >}}).
 
 ### Conflicts
--- a/wiki/content/howto/index.md
+++ b/wiki/content/howto/index.md
@@ -620,13 +620,7 @@ if len(login.Account) == 0 {
 
 ## Upserts
 
-Consider the following example. Let's say we want to update age of a user
-in the database. To do this, we need to first query the UID of the user.
-We can now use the UID to update the age of the user. This whole operation
-requires us two network round trip to the database from the client, one
-for querying the UID and second for updating the age using the UID.
-
-In general, Upsert-style operations are operations where:
+Upsert-style operations are operations where:
 
 1. A node is searched for, and then
 2. Depending on if it is found or not, either:
--- a/wiki/content/mutations/index.md
+++ b/wiki/content/mutations/index.md
@@ -1004,27 +1004,27 @@ Result:
 
 ## Conditional Upsert
 
-The Upsert Block also allows specifying a conditional mutation block using a `@if`
+The Upsert Block also allows specifying a conditional mutation block using an `@if`
 directive. The mutation is executed only when the specified condition is true. If the
 condition is false, the mutation is silently ignored. The general structure of
-Conditional Upsert looks like as follows -
+Conditional Upsert looks like as follows:
 
 ```
 upsert {
   query <query block>
   [fragment <fragment block>]
-  mutation @if(...) <mutation block>
+  mutation @if(<condition>) <mutation block>
 }
 ```
 
-`@if` directive accepts conditions on variables defined in the query block and can be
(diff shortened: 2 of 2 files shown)


wiki/content/howto/index.md
```diff
@@ -620,6 +620,18 @@ if len(login.Account) == 0 {
 
 ## Upserts
 
+Consider the following example. Let's say we want to update age of a user
```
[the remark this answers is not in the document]
mangalaman93: I tried that, it didn't seem better to me. I have removed it for now. We can add more information here later.

wiki/content/mutations/index.md
```diff
@@ -851,10 +851,9 @@ cat data.json | jq '{set: .}'
 
 ## Upsert Block
 
-The Upsert block allows performing queries and mutations in a single request. The Upsert
-block contains one query block and one mutation block. Variables defined in the
-query block can be used in the mutation block using the `uid` function.
-
+The Upsert Block allows performing queries and mutations in a single request. The Upsert
+block contains one query block and one mutation block. Variables defined in the query
```
ashish-goswami:
```suggestion
Block contains one query block and one mutation block. Variables defined in the query
```

wiki/content/mutations/index.md
```diff
@@ -866,46 +865,44 @@ upsert {
...
 ```
 
-### Insert Use Case
-
-Now, let's say we want to create a new user with `email`, and `name` information.
-We also want to make sure that two users cannot have same email id in the database.
+Now, let's say we want to create a new user with `email` and `name` information.
+We also want to make sure that one email id has exactly one corresponding user in
+the database. To achieve this, we need to first query whether a user exists
+in the database with the given email id. If a user exists, we use the returned UID
```
campoy: If a user exists, we use its UID to update ...

wiki/content/mutations/index.md
```diff
@@ -923,40 +920,24 @@ Result:
       "uid(v)": "0x2"
```
campoy: This result is the one where the UID did not exist, make it clear.

campoy: Probably run it twice and show how the second time the result is different

wiki/content/mutations/index.md
```diff
@@ -977,22 +958,88 @@ Result:
...
+```
+
+Here, the query block queries for a user with `email` as `user@company1.io`. It stores
+the `uid` of the user in variable `v`. The mutation block then updates the `age` of the
+user by extracting the uid from the variable `v` using `uid` function.
+
+### Bulk Delete Example
+
+Let's say we want to delete all the users of `company1` from the database. This can be
+achieved in just one query using the Upsert Block -
```
ashish-goswami:
```suggestion
achieved in just one query using the Upsert Block:
```

wiki/content/mutations/index.md
```diff
@@ -977,22 +958,88 @@ Result:
...
+```
+
+The `@if` directive accepts a condition on variables defined in the query block and can be
+connected using `AND`, `OR` and `NOT`.
+
+### Example
+
+Let's say in our previous example, we know the `company1` has less than 100 employees.
+For safety, we want the mutation to execute only when the variable `v` stores less than
+100 but greater than 50 UIDs in it. This can be achieved as follows -
```
ashish-goswami:
```suggestion
100 but greater than 50 UIDs in it. This can be achieved as follows:
```

wiki/content/mutations/index.md
```diff
@@ -923,40 +920,24 @@ Result:
...
-The upsert first checks whether a user with the email `user@dgraph.io` exists. Because
-the database is currently empty, no such user exists. Hence, the variable `v` will be
-empty. In this case, the `uid` function returns a new UID for the variable `v` replacing
-with the new UID value wherever `uid(v)` is used.
-
-### Update Use Case
+The query part of the upsert block stores the UID of the user with the provided email
+id in the variable `v`. The mutation part then extracts the UID from variable `v` and
+stores the `name` and `email` information in the database. In case the user exists,
+the information is updated whereas if the user doesn't exist, `uid(v)` is treated
```
campoy: If the user exists, its information is updated. If the user doesn't exist, `uid(v)` is treated ...

wiki/content/mutations/index.md
```diff
@@ -923,40 +920,24 @@ Result:
...
 curl -H "Content-Type: application/rdf" -X POST localhost:8080/mutate?commitNow=true -d  $'
 upsert {
   query {
-    me(func: eq(email, "user@dgraph.io")) {
-      v as uid
-    }
+    v as var(func: eq(email, "user@company1.io"))
   }
 
   mutation {
```
campoy: This mutation would create a user without a name ... not sure this is a good example.

campoy: Maybe it's worth pointing to the conditional upserts here?

wiki/content/mutations/index.md
```diff
@@ -923,40 +920,24 @@ Result:
...
-### Update Use Case
+The query part of the upsert block stores the UID of the user with the provided email
+id in the variable `v`. The mutation part then extracts the UID from variable `v` and
+stores the `name` and `email` information in the database. In case the user exists,
+the information is updated whereas if the user doesn't exist, `uid(v)` is treated
+as a blank node and a new user is created as explained above.
 
-Now, we want to add the `age` information for the same user having email ID
-`user@dgraph.io`. We can use the Upsert block to do the same as follows:
+Now, we want to add the `age` information for the same user having the same email ID
```
ashish-goswami: same here

wiki/content/mutations/index.md
campoy: Replace the `-` with `:`

PR #3858, opened 2019-08-23
Option #2 for importing dgo v2 without go modules

campoy:
This moves the dgo code in `vendor` into a `v2` subdirectory.

Then all import paths to dgo from dgraph code need to be updated, but the vendored code itself remains the same.

Alternatively, we can go with option 1: #3859 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3858)


campoy:
Closed as we're going with #3859.

PR #3839, opened 2019-08-20
Add support for *API* required for Multiple Mutation

mangalaman93:
Related to #3817

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3839)

PR #3859, opened 2019-08-23
Option #1 for importing dgo v2 without go modules

campoy:
This option rewrites the import paths in the `dgo` code vendored in `dgraph`.

This is a smaller change but it does change the vendored code.

Alternatively, we can go with option #2: #3858

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3859)


campoy:
Closing this as it's a duplicate of #3839

PR #3860, opened 2019-08-23
vendor api.proto from dgo v2 and adopt Go modules

campoy:
As an alternative to #3858 and #3859

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3860)


campoy:
Closed as we're going with #3859, once badger v2 gets released we will move dgraph to Go Modules.

PR #3866, opened 2019-08-26
vendor gqlparser

MichaelJCompton:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3866)

PR #3867, opened 2019-08-26
More schema tests

MichaelJCompton:
Mostly this is about adding some more schema tests.  Also includes a change to how the schemas are output.  I found that the old version is a bit random and hard for a human to parse - if you input a schema and then look at what you get back, it's all over the place.

Seems easier to read if it's

- what I put in (at the top)
- what we always add
- what was generated

I had an idea to also move mutation and query to near the top ... but felt like the very bottom was just as easy to find for those.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3867)


MichaelJCompton:
CI issue is because it's waiting on vendor PR https://github.com/dgraph-io/dgraph/pull/3866

PR #3899, opened 2019-09-03
Create dgraph/standalone Docker image

campoy:
This will allow us to run dgraph in a single docker command:

```bash
$ docker run -tP dgraph/standalone
```

Please @danielmai, can you modify this so this image is built with each release and nightly too?
This will make our "getting started" much easier!

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3899)

### DIFF
--- a/contrib/standalone/Dockerfile
+++ b/contrib/standalone/Dockerfile
@@ -1,0 +1,11 @@
+FROM dgraph/dgraph:latest
+LABEL MAINTAINER="Dgraph Labs <contact@dgraph.io>"
+
+# Ratel port
+EXPOSE 8000
+# REST API port
+EXPOSE 8080
+
+ADD run.sh .
+RUN chmod +x run.sh
+CMD ["./run.sh"]
--- a/contrib/standalone/Dockerfile
+++ b/contrib/standalone/Dockerfile
@@ -5,6 +5,8 @@ LABEL MAINTAINER="Dgraph Labs <contact@dgraph.io>"
 EXPOSE 8000
 # REST API port
 EXPOSE 8080
+# gRPC API port
+EXPOSE 9080
 
 ADD run.sh .
 RUN chmod +x run.sh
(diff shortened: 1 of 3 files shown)


campoy:
Yeah, this is not the most robust way to run a cluster for sure, but it's maybe the easiest one.

I think that if you run the image with `-t` the `ctrl+c` would actually kill the container, wouldn't it?

danielmai:
> I think that if you run the image with `-t` the `ctrl+c` would actually kill the container, wouldn't it?

It does not. It continues running in the background. This can be verified with `docker ps`.

hackintoshrao:
Hey @danielmai , 

Looks like `docker run -it` solves the issue of the container running in the background.

danielmai:
> Looks like `docker run -it` solves the issue of the container running in the background.

OK. I added a TODO in run.sh. Feel free to merge.

campoy:
LGTM!

danielmai:
___
*[contrib/standalone/run.sh, line 6 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/3899#-LnqISr-8cJo5JYVHVla-r1-6:-LpEyDKsC6KbhUZUJaoD:bun2t6t) ([raw file](https://github.com/dgraph-io/dgraph/blob/bbd5ac22bafea1633fbce573520e9d319df48c18/contrib/standalone/run.sh#L6)):*
<details><summary><i>Previously, pullrequest[bot] wrote…</i></summary><blockquote>

I saw there was a recent PR making `--lru_mb` an optional parameter, is this just in place until that is available and then we can remove?

It seems like currently this will be taking the full total of memory on the system which might be a little bit aggressive if some of it is already in use. It could be worth an inline comment here just to explain what the goal is behind the `lru_mb` size being set
</blockquote></details>

Updated the setting to one-third of available memory, which is typically what we recommend.

PR #3910, opened 2019-09-04
Move api.proto file to Dgraph.

martinmr:
Note that this PR includes changes to the dgo code in the vendor
directory. When these changes are removed, Dgraph cannot be compiled
because it's vendored copy of dgo is using the api proto in dgo while
the rest of the code is using the copy in dgraph. I think including the
changes here is necessary to bootstrap the migration. The process would be
as follows.

1. Merge this PR into dgraph:master.
2. Update dgo to use the api proto from dgraph.
3. Vendor dgo into dgraph.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3910)


mangalaman93:
If `dgo` depends on any package in dgraph repo, it will require cloning the whole dgraph repo whenever a user is using the client. Not sure if that is a good idea.

PR #3920, opened 2019-09-04
Add ability to expand only on certain types.

martinmr:
Instead of doing expand(_all_), this change introduces expand(Type1, Type2) syntax to
expand only on certain types.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3920)


martinmr:
There are some issues with the query code that came up while making this change.

- Language tags are only expanded when using `expand(_all_)`. I am not sure if this is the desired behavior.
-  The following test is failing because the uid of the node is shown. This does not happen when using `expand(_all_)`. Removing the uid part from the query makes the test pass but I am not sure why there's a difference between doing `expand(_all_)` and `expand(<list of all the types>)`
```
func TestTypeExpandMultipleExplicitTypes(t *testing.T) {
	query := `{
		q(func: eq(make, "Toyota")) {
			expand(CarModel, Object) {
                           uid
                         }
		}
	}`
	js := processQueryNoErr(t, query)
	require.JSONEq(t, `{"data": {"q":[
		{"name": "Car", "make":"Toyota","model":"Prius", "year":2009}]}}`, js)
}
```

@pawanrawal could you take a look at the second item. I am not sure why the uid of the node is being returned along all the other predicates.

pawanrawal:
I had a look at it but couldn't find immediately why. I suppose what could be happening here is that the uid is being somehow added at the same level as other predicates instead of a child. It would require time for me to debug and I am busy with other things right now. If you can't figure it out, file it as a bug and then I or someone else could have a proper look later.

martinmr:
No worries. I was hoping it was something obvious I was missing. I'll take a closer look soon.

PR #3942, opened 2019-09-09
Add opencensus tracing to graphql layer.

pawanrawal:
I have attached a sample output from Jaeger.

I have also simplified the docker-compose file to just start one alpha and one zero. It also starts a Jaeger container so that we can see the traces.

Screenshot from 2019-09-09 10-36-53

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3942)


dgraph/cmd/graphql/resolve/query.go
```diff
@@ -52,20 +78,34 @@ func (qr *queryResolver) resolve(ctx context.Context) *resolved {
...
-	resp, err := qr.dgraph.Query(ctx, dgQuery)
+	sctx, span := otrace.StartSpan(ctx, "dgraph.Query")
+	resp, err := qr.dgraph.Query(sctx, dgQuery)
 	if err != nil {
 		glog.Infof("[%s] Dgraph query failed : %s", api.RequestID(ctx), err)
 		res.err = schema.GQLWrapf(err, "[%s] failed to resolve query", api.RequestID(ctx))
+		span.End()
 		return res
 	}
+	dgraphDuration.Duration = time.Now().Sub(tnow).Nanoseconds()
```
golangcibot: S1012: should use `time.Since` instead of `time.Now().Sub` (from `gosimple`)

dgraph/cmd/graphql/resolve/mutation.go
```diff
@@ -163,17 +171,59 @@ func (mr *mutationResolver) resolveMutation(ctx context.Context) *resolved {
...
 
+	dgraphQueryStart := time.Now().UTC()
+	dgraphDuration := &schema.LabeledOffsetDuration{Label: "query"}
+	dgraphDuration.StartOffset = dgraphQueryStart.Sub(mr.resolveStart).Nanoseconds()
+	trace.Dgraph = []*schema.LabeledOffsetDuration{dgraphDuration}
+
 	resp, err := mr.dgraph.Query(ctx, dgQuery)
 	if err != nil {
-		res.err = schema.GQLWrapf(err, "mutation %s created a node but query failed",
+		err = schema.GQLWrapf(err, "mutation %s created a node but query failed",
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pawanrawal:
Ended up not using opencensus for constructing the Apollo traces because of the following reasons.

1. Opencensus attributes only support primitive types like string, bool etc. We need a slice of `[]interface{}` for storing the `path` and are also using nested types for storing the Dgraph traces both of which we'll have to do a lot of book keeping for.
2. We'd have to write a converter which can convert opencensus attributes that don't have a type to a nested typed struct. In the current implementation we get the final struct directly.
3. We still only get top level traces from Opencensus. If we wanted to get predicate level tracing like Apollo gives, we'd have to build something that sits outside of the graphql layer and can receive data from both the layer and alpha nodes.
4. We have to write an exporter for getting the SpanData after finishing a span.
5. Apollo tracing and opencensus tracing should ideally be separate. Users should be able to turn them off/on independently. For e.g. setting `tracing` flag to 0 should still allow them to get Apollo tracing.

This does lock us in to returning tracing results in the Apollo tracing format. We might want to change the struct to be able to accommodate other formats in the future.

dgraph/cmd/graphql/resolve/mutation.go
```diff
@@ -59,6 +61,9 @@ type mutationResolver struct {
 	mutationRewriter dgraph.MutationRewriter
 	queryRewriter    dgraph.QueryRewriter
 	dgraph           dgraph.Client
+	// start time for the initial Resolve operation. Used to calculate offsets for other
+	// sub operations.
+	resolveStart time.Time
```
golangcibot: U1000: field `resolveStart` is unused (from `unused`)

pawanrawal:
Removed all Apollo specific stuff. We'll derive it from Opencensus Annotations if we need it later.

PR #3943, opened 2019-09-09
Generate GraphQL schema for search

MichaelJCompton:
Takes an input GraphQL schema with `@searchable(...)` and generates enough GraphQL schema to do enable search.

For example:  for something like

```
type Author {
	id: ID!
	name: String! @searchable(by: hash)
	dob: DateTime  
	posts: [Post]  
}

type Post {
	postID: ID!
	title: String! @searchable(by: term)
	text: String @searchable(by: fulltext)
	datePublished: DateTime  
}
```

It generates filters in `queryPost` and `queryAuthor` to search posts by title and text and authors by name.

It also adds arguments so that an Author's posts can searched.  E.g. Author becomes

```
type Author {
	...
	posts(filter: PostFilter, order: PostOrder, first: Int, offset: Int): [Post]
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3943)

### DIFF
--- a/dgraph/cmd/graphql/schema/gqlschema.go
+++ b/dgraph/cmd/graphql/schema/gqlschema.go
@@ -434,46 +434,48 @@ func addPaginationArguments(fld *ast.FieldDefinition) {
 //   ...
 // }
 func addFilterType(schema *ast.Schema, defn *ast.Definition) {
-	if hasSearchables(defn) {
-		filterName := defn.Name + "Filter"
-		filter := &ast.Definition{
-			Kind: ast.InputObject,
-			Name: filterName,
-		}
-
-		for _, fld := range defn.Fields {
-			if searchable := getSearchable(fld); searchable != "" {
-				filterTypeName := builtInFilters[searchable]
-				if schema.Types[fld.Type.Name()].Kind == ast.Enum {
-					// If the field is an enum type, we don't generate a filter type.
-					// Instead we allow to write `fieldName: enumValue` in the filter.
-					// So, for example : `filter: { postType: Answer }`
-					// rather than : `filter: { postType: { eq: Answer } }`
-					//
-					// Booleans are the same, allowing:
-					// `filter: { isPublished: true }
-					// but that case is aready handled by builtInFilters
-					filterTypeName = fld.Type.Name()
-				}
-
-				filter.Fields = append(filter.Fields,
-					&ast.FieldDefinition{
-						Name: fld.Name,
-						Type: &ast.Type{
-							NamedType: filterTypeName,
-						},
-					})
-			}
-		}
-
-		filter.Fields = append(filter.Fields,
-			&ast.FieldDefinition{Name: "and", Type: &ast.Type{NamedType: filterName}},
-			&ast.FieldDefinition{Name: "or", Type: &ast.Type{NamedType: filterName}},
-			&ast.FieldDefinition{Name: "not", Type: &ast.Type{NamedType: filterName}},
-		)
-
-		schema.Types[filterName] = filter
+	if !hasSearchables(defn) {
+		return
 	}
+
+	filterName := defn.Name + "Filter"
+	filter := &ast.Definition{
+		Kind: ast.InputObject,
+		Name: filterName,
+	}
+
+	for _, fld := range defn.Fields {
+		if searchable := getSearchable(fld); searchable != "" {
+			filterTypeName := builtInFilters[searchable]
+			if schema.Types[fld.Type.Name()].Kind == ast.Enum {
+				// If the field is an enum type, we don't generate a filter type.
+				// Instead we allow to write `fieldName: enumValue` in the filter.
+				// So, for example : `filter: { postType: Answer }`
+				// rather than : `filter: { postType: { eq: Answer } }`
+				//
+				// Booleans are the same, allowing:
+				// `filter: { isPublished: true }
+				// but that case is already handled by builtInFilters
+				filterTypeName = fld.Type.Name()
+			}
+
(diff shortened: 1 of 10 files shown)


MichaelJCompton:
'Bug' altert:

If you have a type like

```
type Post {
  ...
  tags: [String]
}
```

Then this will generate args into that like

```
type Post {
  ...
  tags(first: Int, offset: Int): [String]
}
```

I did that cause I saw that Dgraph accepts `Post.tagsHash (first: 2)` ... but closer inspection shows me that's it's just ignored.  So I'll remove that args generation on scalar lists.

dgraph/cmd/graphql/schema/testdata/schemagen/input/searchables
```diff
@@ -0,0 +1,25 @@
+type Post {
+	postID: ID!
+	title: String! @searchable(by: term)
+	text: String @searchable(by: fulltext)
+
+    tags: [String] @searchable(by: trigram)
```
pawanrawal: Indentation is a bit messed up here.

dgraph/cmd/graphql/schema/gqlschema.go
```diff
@@ -275,11 +363,196 @@ func addPatchType(schema *ast.Schema, defn *ast.Definition) {
...
+		// it's an enum - always has exact index
+		return "exact"
+	}
+	return search.Arguments.ForName(searchableArg).Value.Raw
+}
+
+// addTypeOrderable adds an input type that allows ordering in query.
+// Two things are added: an enum with the names of all the orderable fields,
+// that's called TOrderable; and an input type that allows saying order asc
+// or desc, that's called TOrder.  TOrder's fileds are TOrderable's.  So you
```
pawanrawal: `fields`

dgraph/cmd/graphql/schema/gqlschema.go
```diff
@@ -275,11 +363,196 @@ func addPatchType(schema *ast.Schema, defn *ast.Definition) {
...
+// might get:
+// enum PostOrderable { datePublished, numLikes, ... }, and
+// input PostOrder { asc : PostOrderable, desc: PostOrderable ...}
```
pawanrawal: Could you also mention `then` here and talk about how how it allows sorting by multiple predicates?

dgraph/cmd/graphql/schema/gqlschema.go
```diff
@@ -275,11 +363,196 @@ func addPatchType(schema *ast.Schema, defn *ast.Definition) {
...
+// Together they allow things like
+// order: { asc: datePublished }
+// and
+// order: { asc: datePublished, then: { desc: title } }
+func addTypeOrderable(schema *ast.Schema, defn *ast.Definition) {
+	if hasOrderables(defn) {
+		orderName := defn.Name + "Order"
+		orderableName := defn.Name + "Orderable"
+
+		schema.Types[orderName] =
```
pawanrawal: Looks a bit weird and I am a bit surprised that it compiles, `&ast.Defintion{` could go on this line.

dgraph/cmd/graphql/schema/gqlschema.go
pawanrawal: Please also mention `T` is the types name. Also should we have a validation that `defn` has the right type?

dgraph/cmd/graphql/schema/gqlschema.go
```diff
@@ -275,11 +363,196 @@ func addPatchType(schema *ast.Schema, defn *ast.Definition) {
...
+// or desc, that's called TOrder.  TOrder's fileds are TOrderable's.  So you
+// might get:
+// enum PostOrderable { datePublished, numLikes, ... }, and
+// input PostOrder { asc : PostOrderable, desc: PostOrderable ...}
+// Together they allow things like
+// order: { asc: datePublished }
+// and
+// order: { asc: datePublished, then: { desc: title } }
+func addTypeOrderable(schema *ast.Schema, defn *ast.Definition) {
+	if hasOrderables(defn) {
```
pawanrawal: Small suggestion
```
if !hasOrderables(defn) {
  return
}
...
```

dgraph/cmd/graphql/schema/gqlschema.go
```diff
@@ -275,11 +363,196 @@ func addPatchType(schema *ast.Schema, defn *ast.Definition) {
...
+// type R {
+//   f(filter: TFilter, ... ): T
+//   ...
+// }
 func addFilterType(schema *ast.Schema, defn *ast.Definition) {
-	schema.Types[defn.Name+"Filter"] = &ast.Definition{
-		Kind:   ast.InputObject,
-		Name:   defn.Name + "Filter",
-		Fields: getFilterField(),
+	if hasSearchables(defn) {
```
pawanrawal: Could return early here as well if no searchables are there.

pawanrawal: Looking at it again, we could have avoided this function because we iterate through all fields here. Do you think the current approach makes for cleaner code?

dgraph/cmd/graphql/schema/gqlschema.go
MichaelJCompton: done for type names.

Not sure what you mean about a validation?

dgraph/cmd/graphql/schema/gqlschema.go
MichaelJCompton: I think if we iterate without checking, then we still have to check this condition at the end before adding to the schema.  Seems better to have it up front.

PR #3949, opened 2019-09-09
Add support for expand on type

### LINKED ISSUE #3903: Support expand of types
Since v1.1.0 Dgraph support types and `expand(_all_)` is defined as the expansion of all the predicates found in all of the types linked to any of the nodes in the query.

So for instance, if we have two types Person and Aged:

```graphql
type Person {
   first_name: string
   last_name: string
}

type Aged {
   age: int
}
```

And we have the following dataset:

```
{
  set {
    _:a <dgraph.type> "Person" .
    _:a <dgraph.type> "Aged" .
    _:a <first_name> "John" .
    _:a <last_name> "Doe" .
    _:a <age> "42" .
     }
}
```

When we query all values of type `Person` and use `expand(_all_)`:

```graphql
{
  q(func: type(Person)) {
    dgraph.type
    expand(_all_)
  }
}
```

we end up receiving the field `age` even though it's not part of the type `Person` because `expand(_all_)` expands all of the predicates in all of the types linked to any of the values returned.

```graphql
{
  "data": {
    "q": [
      {
        "dgraph.type": [
          "Person",
          "Aged"
        ],
        "first_name": "John",
        "age": 42,
        "last_name": "Doe"
      }
    ]
  },
  "extensions": {...}
}
```

This is because our node `John Doe` is of type `Person` and `Aged`, as the result of the query shows.

I propose a way to expand only the fields related to a type by passing the type name as the parameter to `expand`.

So, we could send this query:

```graphql
{
  q(func: type(Person)) {
    dgraph.type
    expand(Person)
  }
}
```

This would return only the fields `first_name` and `last_name`. We could also pass multiple types to the expand query: `expand(Person, Aged)` and that would expand the union of the predicates.

balajijinnah:
close #3903 
This PR will add support to use type on expand function. 
# Changes made
- Removed restriction in expand function for using only the specified keyword like `_all_`. Now expand function can take any paramenter. But the given parameter should be assoicated with type. 
- subgraph is expanded from the type predicates. 

Now, user can query like this
```
{
  q(func: type(Person)) {
    dgraph.type
    expand(Person)
  }
}
```
Signed-off-by: பாலாஜி ஜின்னா <balaji@dgraph.io>

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3949)

### DIFF
--- a/query/query.go
+++ b/query/query.go
@@ -1703,6 +1703,18 @@ func getPredsFromVals(vl []*pb.ValueList) []string {
 	return preds
 }
 
+func getPredMapFromVals(vl []*pb.ValueList) map[string]struct{} {
+	preds := make(map[string]struct{}, len(vl))
+	for _, l := range vl {
+		for _, v := range l.Values {
+			if len(v.Val) > 0 {
+				preds[string(v.Val)] = struct{}{}
+			}
+		}
+	}
+	return preds
+}
+
 func uniquePreds(list []string) []string {
 	predMap := make(map[string]struct{})
 	for _, item := range list {
@@ -1788,9 +1800,21 @@ func expandSubgraph(ctx context.Context, sg *SubGraph) ([]*SubGraph, error) {
 			}
 			preds = append(preds, rpreds...)
 		default:
-			span.Annotate(nil, "expand default")
-			// We already have the predicates populated from the var.
-			preds = getPredsFromVals(child.ExpandPreds)
+			// If type exist in expand function, check whether that type exist for the node or not.
+			// If not then thorow error. If exist obtain all the predicate for the type and expand
+			// from there.
+			if len(child.Params.Expand) > 0 {
+				span.Annotate(nil, fmt.Sprintf("expand(%s)", child.Params.Expand))
+				_, ok := types[child.Params.Expand]
+				if !ok {
+					return out, errors.Errorf("Invalid expand type %d", child.Params.Expand)
+				}
+				preds = getPredicatesFromTypes(map[string]struct{}{child.Params.Expand: struct{}{}})
+			} else {
+				span.Annotate(nil, "expand default")
+				// We already have the predicates populated from the var.
+				preds = getPredsFromVals(child.ExpandPreds)
+			}
 		}
 		preds = uniquePreds(preds)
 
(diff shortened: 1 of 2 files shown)


query/query.go
```diff
@@ -1788,9 +1800,21 @@ func expandSubgraph(ctx context.Context, sg *SubGraph) ([]*SubGraph, error) {
...
-			// We already have the predicates populated from the var.
-			preds = getPredsFromVals(child.ExpandPreds)
+			// If type exist in expand function, check whether that type exist for the node or not.
+			// If not then thorow error. If exist obtain all the predicate for the type and expand
+			// from there.
+			if len(child.Params.Expand) > 0 {
+				span.Annotate(nil, fmt.Sprintf("expand(%s)", child.Params.Expand))
+				_, ok := types[child.Params.Expand]
+				if !ok {
+					return out, errors.Errorf("Invalid expand type %d", child.Params.Expand)
```
golangcibot: printf: Errorf format %d has arg child.Params.Expand of wrong type string (from `govet`)

query/query.go
```diff
@@ -1788,9 +1800,21 @@ func expandSubgraph(ctx context.Context, sg *SubGraph) ([]*SubGraph, error) {
...
+				}
+				preds = getPredicatesFromTypes(map[string]struct{}{child.Params.Expand: struct{}{}})
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
				preds = getPredicatesFromTypes(map[string]struct{}{child.Params.Expand: {}})
```

query/query.go
```diff
@@ -1788,9 +1800,21 @@ func expandSubgraph(ctx context.Context, sg *SubGraph) ([]*SubGraph, error) {
 			}
 			preds = append(preds, rpreds...)
 		default:
-			span.Annotate(nil, "expand default")
-			// We already have the predicates populated from the var.
-			preds = getPredsFromVals(child.ExpandPreds)
+			// If type exist in expand function, check whether that type exist for the node or not.
+			// If not then thorow error. If exist obtain all the predicate for the type and expand
```
mangalaman93: throw*

query/query.go
mangalaman93: why are we converting this into string?

balajijinnah:
I'm closing this.

PR #3958, opened 2019-09-10
Enable go modules and start using dgo/v2

gitlw:
dgo has enabled go modules, and the module path has been changed to "github.com/dgraph-io/dgo/v2". Thus we won't be able to continue the old practice of vendoring dgo to use its latest code, since the path "github.com/dgraph-io/dgo/v2" does not exist.
This PR enables go modules in dgraph and migrates the import path from ".../dgo" to ".../dgo/v2".

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3958)


manishrjain:
You can skip me if these are cosmetic changes.

Issue #3959, opened 2019-09-10
Bulk loader: Shard merge error during reduce phase


danielmai:
### What version of Dgraph are you using?
v1.1.0

### Have you tried reproducing the issue with the latest release?
Yes

### What is the hardware spec (RAM, OS)?
72 GB, Ubuntu 18.04

### Steps to reproduce the issue (command/config used to run Dgraph).

Run Dgraph Bulk Loader:

```sh
dgraph bulk -f /dgraph/g01.rdf.gz -s /dgraph/g01.schema.gz -z zero1:5180 --out /dgraph
/out --tmp /dgraph/tmp
```

### Expected behaviour and actual result.

Dgraph Bulk Loader should succeed. Once, the bulk loader failed with this error trace during the reduce phase:

```text
[22:08:35Z] MAP 04m54s nquad_count:153.9M err_count:0.000 nquad_speed:523.1k/sec edge_count:217.0M edge_speed:737.5k/sec
[22:08:36Z] MAP 04m55s nquad_count:153.9M err_count:0.000 nquad_speed:521.4k/sec edge_count:217.0M edge_speed:735.0k/sec
[22:08:37Z] MAP 04m56s nquad_count:153.9M err_count:0.000 nquad_speed:519.6k/sec edge_count:217.0M edge_speed:732.6k/sec
Shard /dgraph/tmp/shards/shard_0 -> Reduce /dgraph/tmp/shards/shard_0/shard_0
2019/09/10 22:08:38 rename /dgraph/tmp/shards/shard_0 /dgraph/tmp/shards/shard_0/shard_0: invalid argument

github.com/dgraph-io/dgraph/x.Check
	/tmp/go/src/github.com/dgraph-io/dgraph/x/error.go:42
github.com/dgraph-io/dgraph/dgraph/cmd/bulk.mergeMapShardsIntoReduceShards
	/tmp/go/src/github.com/dgraph-io/dgraph/dgraph/cmd/bulk/merge_shards.go:46
github.com/dgraph-io/dgraph/dgraph/cmd/bulk.run
	/tmp/go/src/github.com/dgraph-io/dgraph/dgraph/cmd/bulk/run.go:231
github.com/dgraph-io/dgraph/dgraph/cmd/bulk.init.0.func1
	/tmp/go/src/github.com/dgraph-io/dgraph/dgraph/cmd/bulk/run.go:50
[...]

PR #3960, opened 2019-09-11
Change the mapper output directory from $TMP/shards to $TMP/map_output

gitlw:
In #3959 , bulk loader crashes when trying to move a directory into itself with a new name
/dgraph/tmp/shards/shard_0
/dgraph/tmp/shards/shard_0/shard_0

The bulk loader logic is
1. the mapper produce output as 
.../tmp/shards/000
.../tmp/shards/001

2. read the list of shards under .../tmp/shards/
3. create the reducer shards as
.../tmp/shards/shard_0
.../tmp/shards/shard_1

4. move the list read in step 2 into the reducer shards created in step 3

Though I cannot reproduce the problem, but it seems creating of the reducer shard directory .../tmp/shards/shard_0 and listing all the mapper shards in step 2 are re-ordered. Something similar is mentioned in https://github.com/etcd-io/etcd/issues/6368

This PR avoids such possibilities by putting the mapper output into an independent directory
../tmp/map_output, so that the program works correctly even if the reordering happens.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3960)

### DIFF
--- a/dgraph/cmd/bulk/mapper.go
+++ b/dgraph/cmd/bulk/mapper.go
@@ -89,7 +89,7 @@ func (m *mapper) openOutputFile(shardIdx int) (*os.File, error) {
 	fileNum := atomic.AddUint32(&m.mapFileId, 1)
 	filename := filepath.Join(
 		m.opt.TmpDir,
-		"map_output",
+		mapShardDir,
 		fmt.Sprintf("%03d", shardIdx),
 		fmt.Sprintf("%06d.map.gz", fileNum),
 	)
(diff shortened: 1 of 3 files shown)


mangalaman93:
I have a question. Wouldn't it make more sense to run `fsync` after creating or moving directories/files? That will guarantee that operations are executed in order.

gitlw:
@mangalaman93 I think fsync would help with other types of problems.
For instance, if we are creating a subdirectory and then try to list all children under the parent, and we cannot find the newly created child, fsync could potentially help.
But in this case, fsync would not help because we are receiving a child directory which should not show up.

PR #3973, opened 2019-09-12
Enforce type required fields when @type directive is used.

martinmr:
Required (non-nullable in GraphQL lingo) fields mean that objects
that do not contain said fields should not be returned in the output.
This PR adds the initial implementation of required fields enforcement.

A couple things are missing:
- Dealing with non-nullable lists (e.g [string]!)
- @type directive does not work at the query root. This is a known
  issue that will be fixed in a separate PR.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3973)


martinmr:
Since we decided not to enforce types for now, I am closing this PR.

PR #3975, opened 2019-09-13
Allow @type directive in the query root.

### LINKED ISSUE #3974: @type directive not working at root level.
master

yes

Try to run a query with the type directive at the root level. The query is not parsed correctly.

Queries with @type directives at the root level should work the same as any ones with the directive at a non-root level.

martinmr:
Fixes #3974 

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3975)

### DIFF
--- a/gql/parser.go
+++ b/gql/parser.go
@@ -857,6 +857,10 @@ L:
 				if err := parseRecurseArgs(it, gq); err != nil {
 					return nil, err
 				}
+			case "type":
+				if err := parseType(it, gq); err != nil {
+					return nil, err
+				}
 			default:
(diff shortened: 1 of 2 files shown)


campoy:
Do we have a `@type` directive already that works somewhere else?

What's the difference in behavior between `@type(Person)` and `@filter(type(Person))`?

martinmr:
Right now, nothing. Eventually, more stuff than just filtering for the dgraph.type predicate will be part of type enforcement (for example: removing elements that are missing required fields, making sure the values match the value type described in the type definition, etc). The point of `@type` is to explicitly signal that all this type enforcement should be done on the query results.

PR #3962, opened 2019-09-11
Touchup types docs

MichaelJCompton:
Just some light touches around types docs to make language clearer.

Removes mention of `!`, which isn't supported yet.

Also removes mention of `@type(...)`, which we can add back later once https://github.com/dgraph-io/dgraph/issues/3961 is fixed

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3962)

PR #3993, opened 2019-09-13
Rewrite type system docs to explicitly mention type function and dire…

martinmr:
…ctive.

There's documentation for the type top-level function and the type
directive but it doesn't mention the @type directive expliclty, it
only shows an example query. This PR rewrites that section to make an
explicit mention so it's easier to find the documentation.

It also adds a note on what the difference between the type function,
directive, and their current equivalents is meant to be.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3993)

### DIFF
--- a/wiki/content/query-language/index.md
+++ b/wiki/content/query-language/index.md
@@ -2137,8 +2137,9 @@ types. Here's an example of how to set the types of a node:
 
 #### Using types during queries
 
-The type system can be used as a top level function in the query language. Here's an example:
+Right now, there are a couple ways in which the type system can be enforced during queries.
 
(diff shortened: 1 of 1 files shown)


martinmr:
so I forgot about https://github.com/dgraph-io/dgraph/pull/3962 which changes  the same sections. It removes mentions of @type.

But I am not sure if we should remove it (we could remove it from examples where it's not really needed) since then there would be an undocumented feature of the query language.

martinmr:
Closing this for now. I'll work on it again once the @type directive works, which was the original plan.

Issue #3916, opened 2019-09-04
Txn is too big to fit into one request at the very end of bulk import


eawer:
If you suspect this could be a bug, follow the template.

- What version of Dgraph are you using?
```
Dgraph version   : v1.0.17
Commit SHA-1     : 42deb299
Commit timestamp : 2019-08-30 12:46:31 -0700
Branch           : HEAD
Go version       : go1.12.7
```

- Have you tried reproducing the issue with latest release?
yes

- What is the hardware spec (RAM, OS)?
Ubuntu 16.04.6 LTS, 30GB

- Steps to reproduce the issue (command/config used to run Dgraph).

1. Run zero `docker run -it --rm -p 5080:5080 -p 6080:6080 -p 8080:8080 -p 9080:9080 -p 8000:8000 -v ~/freebase/dgraph:/dgraph --name dgraph dgraph/dgraph dgraph zero`
2. Start import
`docker exec -it dgraph dgraph bulk -r  freebase.rdf.gz -s freebase.schema -v 3 --zero localhost:5080`

- Expected behavior and actual result.
Expected - import without errors.
result - the process was running for about 8h, and failed on the 100% of reduce stage:
```
REDUCE 07h43m16s [100.00%] edge_count:6.227G edge_speed:524.0k/sec plist_count:1.202G plist_speed:101.2k/sec
REDUCE 07h43m17s [100.00%] edge_count:6.227G edge_speed:524.0k/sec plist_count:1.202G plist_speed:101.2k/sec
REDUCE 07h43m18s [100.00%] edge_count:6.227G edge_speed:524.0k/sec plist_count:1.202G plist_speed:101.2k/sec
REDUCE 07h43m19s [100.00%] edge_count:6.227G edge_speed:523.9k/sec plist_count:1.202G plist_speed:101.2k/sec
2019/09/04 13:02:29 Txn is too big to fit into one request
github.com/dgraph-io/dgraph/vendor/github.com/dgraph-io/badger.init.ializers
	/tmp/go/src/github.com/dgraph-io/dgraph/vendor/github.com/dgraph-io/badger/errors.go:44
runtime.main
	/usr/local/go/src/runtime/proc.go:188
runtime.goexit
	/usr/local/go/src/runtime/asm_amd64.s:1337

github.com/dgraph-io/dgraph/x.Wrap
	/tmp/go/src/github.com/dgraph-io/dgraph/x/error.go:91
github.com/dgraph-io/dgraph/x.Check
[...]

PR #3998, opened 2019-09-16
Fix Txn too big error in bulk loader

ashish-goswami:
Fixes https://github.com/dgraph-io/dgraph/issues/3916

I was able to reproduce this issue after running bulk loader with file generate using below code:

```go
package main

import (
	"fmt"
	"os"
)

func main() {
	f, err := os.Create("./many.rdf")
	if err != nil {
		fmt.Println("unable to open file ", err)
		return
	}
	for i := 0; i < 1000000; i++ {
		a := fmt.Sprintf("_:id%d <pred%d> \"val\" .", i, i)
		f.Write([]byte(a))
		f.Write([]byte("\n"))
	}
	f.Sync()
	f.Close()
}
```

I was thinking of using TxnWriter https://github.com/dgraph-io/dgraph/blob/d048d5c4b009cff63e25fb8807747cf13d4ce037/posting/writer.go#L29-L33, but found that, it creates a new transaction for every SetAt().

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3998)

### DIFF
--- a/dgraph/cmd/bulk/reduce.go
+++ b/dgraph/cmd/bulk/reduce.go
@@ -39,7 +39,6 @@ func (r *reducer) run() {
 		thr.Start()
 		NumReducers.Add(1)
 		NumQueuedReduceJobs.Add(-1)
-		r.writesThr.Start()
 		go func(job shuffleOutput) {
 			r.reduce(job)
 			thr.Done()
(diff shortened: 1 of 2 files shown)


ashish-goswami:
Closing this PR as fixed via https://github.com/dgraph-io/dgraph/pull/4296

PR #4017, opened 2019-09-18
Simplify type definitions.

martinmr:
Type definitons are now simpler as the type of the fields is not
required anymore. The new format is:

```
type Person {
  name
  address
  ssn
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4017)


campoy:
Hey, what's the issue related to this?

Changes to the schema or the query language like this are really really important and shouldn't be done without having a clear design in mind first.

What's the motivation behind this change?

martinmr:
This has already been agreed to. It's part of the conclusions at https://discuss.dgraph.io/t/should-we-remove-the-type-directive/5067/12?u=martinmr

PR #4061, opened 2019-09-25
Added variable substitution to FacetsFilter

### LINKED ISSUE #2481: QueryWithVars on facet filtering fails silently

  v1.0.6

   No

    Ubuntu, 128G

    nohup dgraph zero -w zw5 --replicas 1 > info1 &
    nohup dgraph server -p p5 -w w5 --lru_mb 20000 --zero localhost:5080 > info2 &
    nohup dgraph-ratel > info3 &

  i found a problem with query like this:

```
query withvar($a:string, $b:bool) {

query(func:uid($a)) {

candidate_school @facets(eq(on_edu,$b)) @groupby(name) {

  count(uid)
}
}
}

```
it doesn't have any result return, but when i use query like this:

```

{
query(func:uid(0x998a17)) {

candidate_school @facets(eq(on_edu,false)) @groupby(name) {

  count(uid)

}
}
}
```

it has result.

it seems that queryWithVars has a problem when query vars in facets.
@pawanrawal @sboorlagadda

refer ： https://discuss.dgraph.io/t/query-and-querywithvars/2796/9

harshil-goel:
Fixes #2481

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4061)


gql/parser_test.go
```diff
@@ -2032,6 +2032,23 @@ func TestParseStringVarInFilter(t *testing.T) {
 	require.Equal(t, "v0.7.3/beta", res.Query[0].Children[0].Filter.Func.Args[0].Value)
 }
 
+func TestParseVarInFacet(t *testing.T) {
+	query := `
+query works($since: string = "2018") {
```
balajijinnah: check the indent

harshil-goel: That's the indent for string queries.

Issue #4005, opened 2019-09-17
Values are not sorted properly


balajijinnah:
### What version of Dgraph are you using?
master
### Have you tried reproducing the issue with the latest release?
Yes

### Steps to reproduce the issue (command/config used to run Dgraph).
Just populate cluster with this data
̍https://github.com/dgraph-io/dgraph/blob/master/query/common_test.go#L269
and run the this query
```
{
		q(func:has(name), orderdesc: name, first:5) {
			 name
		 }
	 }
```

### Expected behaviour and actual result.
actual result 
```
{
  "data": {
    "q": [
      {
        "name": ""
      },
      {
        "name": ""
      },
      {
        "name": "Badger"
      },
      {
        "name": "name"
      },
      {
        "name": "expand"
      }
    ]
  },
  "extensions": {
    "server_latency": {
      "parsing_ns": 14640,
      "processing_ns": 375499571,
      "encoding_ns": 28760,
      "assign_timestamp_ns": 503321
    },
    "txn": {
      "start_ts": 98
    }
  }
}
```
Expected everything to be sorted

akamos:
Hi! 
 
I have a similar behaviour at Dgraph 1.1.0. I have data that contains values for name@de and name@ru predicates. 
If I try to order these name string predicates with a language tag (orderasc:name@ru) I get different results from each time (its not determined behaviour). Sometimes its sorted correctly, by @ru, and sometimes not! If its not correctly sorted, it's sorted by name@de. 

Check it out with following data and queries. Simply query, and the results differ from time to time.

Sample data: 
```
{
  set {
    _:1 <name@de> "Hello World" .
    _:1 <name@ru> "Привет!" . 
    _:1 <category_type> "Test" .

    _:2 <name@de> "ZZZ A little test" .
    _:2 <name@ru> "Авиякомпания" .
    _:2 <category_type> "Test" .

    _:3 <name@de> "Xylophon" .
    _:3 <name@ru> "Книга" .
    _:3 <category_type> "Test" .
	}
}
```

My query: 
```
{
  q(func:eq(category_type, "Test"),orderasc:name@ru)  {
    name@ru
    name@de
  }
}
```

akamos:
Are there any updates to this issue? It's for me a critical point in my production system and I would continue to use Dgraph.

arijitAD:
There are two PRs related to this issue. One is merged and other one is under review. I have linked both of them to this issue.

akamos:
[...]

PR #4062, opened 2019-09-25
Sort by value even it is indexed

balajijinnah:
Since we are not indexing language-wise. Uids are not in sorted order when we do reverse iteration.
This PR will p
Signed-off-by: பாலாஜி ஜின்னா <balaji@dgraph.io>

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4062)

### DIFF
--- a/worker/sort.go
+++ b/worker/sort.go
@@ -276,11 +276,9 @@ BUCKETS:
 	var multiSortOffsets []int32
 	for _, il := range out {
 		r.UidMatrix = append(r.UidMatrix, il.ulist)
-		if len(ts.Order) > 1 {
-			// TODO - For lossy tokenizer, no need to pick all values.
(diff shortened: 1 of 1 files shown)


balajijinnah:
Issue link https://github.com/dgraph-io/dgraph/issues/4005
example for what was happening before.
Let's consider the following mutation. 
```
	        <4097> <name> "Badger" .
		<4097> <name> "Барсук"@ru .
                <4098> <name> "SchoolBoy"
```
when we do a reverse iteration for the indexed attr name, 4097 and 4098 is picked as order because the russian name came first. Then we give this as a sort order. This is happening because, we are not doing language wise indexing. So, we pick some unsorted uid order. 

This PR, will do types.Sort after we get the required uid.

PR #4116, opened 2019-10-02
Fix how we handle errors in GraphQL mutations

MichaelJCompton:
In GraphQL when there are multiple mutations in a single request, the mutations are run serially.  If one of them fails, the remainder don't run.

This PR makes our handling of these errors a bit more subtle: there's really two classes of error 
1) the actual mutation failed
2) something happened in the query following the mutation that raised a GraphQL error.

(1) should stop future mutations, but (2) is inconsequential to the actual mutation, which must have succeeded by the time we run the query.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4116)


dgraph/cmd/graphql/resolve/mutation.go
```diff
@@ -64,11 +63,15 @@ type mutationResolver struct {
 }
 
 const (
-	createdNode = "newnode"
+	createdNode       = "newnode"
```
golangcibot: `createdNode` is unused (from `varcheck`)

PR #4119, opened 2019-10-02
Breaking change: Remove expand(_forward_) and expand(_reverse_).

martinmr:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4119)

### DIFF
--- a/gql/parser.go
+++ b/gql/parser.go
@@ -2802,10 +2802,6 @@ func godeep(it *lex.ItemIterator, gq *GraphQuery) error {
 					child.Expand = child.NeedsVar[len(child.NeedsVar)-1].Name
 				case "_all_":
 					child.Expand = "_all_"
-				case "_forward_":
-					child.Expand = "_forward_"
-				case "_reverse_":
-					child.Expand = "_reverse_"
 				default:
 					return item.Errorf("Invalid argument %v in expand()", item.Val)
 				}
--- a/gql/parser.go
+++ b/gql/parser.go
@@ -2802,6 +2802,10 @@ func godeep(it *lex.ItemIterator, gq *GraphQuery) error {
 					child.Expand = child.NeedsVar[len(child.NeedsVar)-1].Name
 				case "_all_":
 					child.Expand = "_all_"
+				case "_forward_":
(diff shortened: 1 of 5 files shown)


wiki/content/query-language/index.md
```diff
@@ -1800,10 +1800,6 @@ There are four ways to use the `expand` function.
 * If `_all_` is passed as an argument to `expand()`, all the predicates for each
```
campoy: Since we're at it, this statement is not accurate.

query/query.go
```diff
@@ -1824,7 +1824,7 @@ func expandSubgraph(ctx context.Context, sg *SubGraph) ([]*SubGraph, error) {
 		}
 
 		switch child.Params.Expand {
-		// It could be expand(_all_), expand(_forward_), expand(_reverse_) or expand(val(x)).
+		// It could be expand(_all_) or expand(val(x)).
```
campoy: What's val(x)? I have never any example like this

wiki/content/query-language/index.md
campoy: Since we're at it, this statement is not accurate.

query/query.go
MichelDiz: This was a variable that gets predicates values from the old `_predicate_` func. To expand a query with those preds. It was like "expand only the preds found in the `_predicate_` func".

martinmr: Pasted from my reply in Reviewable: "It takes a variable (a list of string values) and uses those values as the names of the predicates to expand. We should discuss whether to keep it but I left it untouched for now."

gql/parser.go
```diff
@@ -2802,10 +2802,6 @@ func godeep(it *lex.ItemIterator, gq *GraphQuery) error {
 					child.Expand = child.NeedsVar[len(child.NeedsVar)-1].Name
 				case "_all_":
 					child.Expand = "_all_"
-				case "_forward_":
-					child.Expand = "_forward_"
-				case "_reverse_":
-					child.Expand = "_reverse_"
```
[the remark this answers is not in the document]
martinmr: Added error message informing users about the deprecation of these features.

PR #4145, opened 2019-10-09
Return mutated uids which are part of an upsert operation in a map.

pawanrawal:
We now return a map of the variables that were part of an upsert mutation along with the list of uids that they map to when debug parameter is set to true.

The PR also refactors `http_test.go` to return a struct from `mutationWithTs`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4145)


edgraph/server.go
```diff
@@ -577,19 +606,22 @@ func (s *Server) doMutate(ctx context.Context, req *api.Request, authorize int)
...
 	}
 
 	mu := req.Mutations[0]
 	upsertQuery := req.Query
 	needVars := findVars(gmu)
 	isCondUpsert := strings.TrimSpace(mu.Cond) != ""
-	varName := fmt.Sprintf("__dgraph%d__", rand.Int())
+	// conditionalVar is a dummy var that we use to evaluate the result of
+	// conditional upsert.
+	conditionalVar := fmt.Sprintf("__dgraph%d__", rand.Int())
```
golangcibot: G404: Use of weak random number generator (math/rand instead of crypto/rand) (from `gosec`)

PR #4174, opened 2019-10-15
Add "set -e" to test script.

### LINKED ISSUE #4156: Test script continues regardless of command failures
If you run `test.sh` without a required tool present, the script continues despite failures.

For example

```
$ ./test.sh
INFO: Running only unit tests
INFO: Running tests using the default cluster
/dgraph/contrib/scripts/functions.sh: line 15: pushd: /src/github.com/dgraph-io/dgraph/dgraph: No such file or directory
Rebuilding dgraph ...
/dgraph/contrib/scripts/functions.sh: line 20: make: command not found
/dgraph/contrib/scripts/functions.sh: line 22: docker: command not found
/dgraph/contrib/scripts/functions.sh: line 23: docker-compose: command not found
```

I think that the shebang line should include the `-e` flag, or the script should have `set -e` early on, so that the script doesn't continue if a command fails.

martinmr:
This option should make the script fail when an individual command fails
instead of continuing.

Fixes #4156

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4174)

### DIFF
--- a/contrib/scripts/functions.sh
+++ b/contrib/scripts/functions.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -e
 # Containers MUST be labeled with "cluster:test" to be restarted and stopped
 # by these functions.
 
--- a/test.sh
+++ b/test.sh
@@ -23,6 +23,7 @@
 # Keep in mind that the test build will overwrite the "dgraph"
 # binary in your $GOPATH/bin with the Linux-ELF binary for Docker.
 
+set -e
 readonly ME=${0##*/}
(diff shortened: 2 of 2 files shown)


dsymonds:
FYI, using an explicit `set -e` inside the script is generally considered better practice than relying on the shebang, since the latter is only applied when the script is invoked directly (e.g. `./foo.sh`) but it is ignored when run by passing it to a shell as an argument (`bash foo.sh`). That can happen in a bunch of places, particularly in automation.

danielmai:
@dsymonds good point. @martinmr can you make another PR?

martinmr:
https://github.com/dgraph-io/dgraph/pull/4181

PR #4177, opened 2019-10-16
Added some schema rules to dgraph layer

harshil-goel:
- no [ID] lists of any kind
- changed my mind on lists.  Any combination of [T!]!, [T], [T!], [T]! is allowed, but no nested lists of any kind [[T]] (see also Input processing)
- there shouldn't be arguments on any fields: e.g. `type T { f(a: Int) }` isn't allowed

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4177)


MichaelJCompton:
one more note.

I'd prefer if the validation functions were were a single rule per function - atm the list validity check does two rules.  Can we change, so there's a single rule per function.

PR #4187, opened 2019-10-17
Change dgraph type from Type.FieldName to just FieldName.

pawanrawal:
We now also enforce the limitation that the type of a field across different type definitions must be the same. New tests have been added to verify this case.

Also, we aggregate all indexes for a field across all the types that it is defined in.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4187)


dgraph/cmd/graphql/resolve/mutation_rewriter.go
```diff
@@ -358,7 +358,7 @@ func rewriteObject(
 				if srcField != nil {
 					invType, invField := srcField.Inverse()
 					if invType != nil && invField != nil {
-						result[fmt.Sprintf("%s.%s", invType.Name(), invField.Name())] =
+						result[fmt.Sprintf("%s", invType.DgraphPredicate(invField.Name()))] =
```
golangcibot: S1025: the argument is already a string, there's no need to use fmt.Sprintf (from `gosimple`)

pawanrawal:
Closing this PR for now as we have decided that the default behavior would remain as it is right now, i.e. predicates in Dgraph are prefixed with the type name. We will later be adding a `@dgraph` directive which would allow overriding the predicate name.

PR #4222, opened 2019-10-27
Fix field look up for interfaces

MichaelJCompton:
When we look up the definition for a field, we should return the base definition from an interface (if it exists) not the one from the type

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4222)


MichaelJCompton:
Merging because needed for docs and GQL conf

PR #4225, opened 2019-10-29
Make release script work with go modules.

martinmr:
Force GO111MODULES to off except when building dgraph and badger.
This seems to work fine but at some point it might be worthy to
move to go modules entirely.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4225)

### DIFF
--- a/contrib/release.sh
+++ b/contrib/release.sh
@@ -13,19 +13,25 @@ RESET='\033[0m'
 # Don't use standard GOPATH. Create a new one.
 unset GOBIN
 GOPATH="/tmp/go"
+if [ -d $GOPATH ]; then
+   chmod -R 755 $GOPATH
+fi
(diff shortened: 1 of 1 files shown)


martinmr:
In order to get this to work I had to comment out the following.

- The golang version check. I am using 1.13. The script still works fine since I am manually setting the value of `GO111MODULE` instead of defaulting to `auto`.

- The ratel tests. They keep failing consistently. @paulftw Could you look into it? All you need to do to reproduce the error is to checkout this branch and run the release script (maybe comment out the golang version check in case you are using a different version).

martinmr:
@paulftw: I went back to the commit before go modules were introduced and the release script is failing when running the tests so I don't think the issue has anything to do with go modules.

PR #4226, opened 2019-10-29
Startup with introspection of empty schema

MichaelJCompton:
Adds introspection to the initial setup, so you can point `__schema` at it when it starts.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4226)

### DIFF
--- a/dgraph/cmd/graphql/admin/admin.go
+++ b/dgraph/cmd/graphql/admin/admin.go
@@ -178,8 +178,37 @@ type adminServer struct {
 	withIntrospection bool
 }
 
-// NewAdminResolver creates a GraphQL request resolver for the /admin endpoint.
-func NewAdminResolver(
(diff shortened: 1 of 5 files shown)


dgraph/cmd/graphql/admin/admin.go
```diff
@@ -178,8 +178,37 @@ type adminServer struct {
...
+
+	gqlSchema, err := schema.FromString("")
+	if err != nil {
+		panic(err)
+	}
+
+	rf := resolve.NewResolverFactory()
+	if withIntrospection {
+		rf.WithSchemaIntrospection()
+	}
```
MichaelJCompton: this is the only change.  The rest was just code move, because more is about to happen on startup and it belongs in the admin setup, not run.go

Issue #4227, opened 2019-10-30
Dgraph native from mjc/graphql doesnt support comments at the beggining of the file


aexol:
### What version of Dgraph are you using?
dgraph/standalone image
### Have you tried reproducing the issue with the latest release?
With the release from blog :)

### What is the hardware spec (RAM, OS)?
12gb ram i7 macOS

### Steps to reproduce the issue (command/config used to run Dgraph).
Add a comment at the beginning of GraphQL file

```graphql
# I am a comment making schema doesn't work
type Me{
"""
However, descriptions work
"""
  name:String!
}
```

### Expected behaviour and actual result.
works

MichaelJCompton:
This actually already works - it's the docs that's broken, not the server :-(

The line in the docs that loads the schema just pumps everything into one line, which means any `#` covers over all remaining lines. 

I'd originally written the command below.  I'll work on a friendlier way to load the schema.

```
jq -n --argjson schema "$(jq --raw-input --slurp < /Users/michael/go/src/github.com/dgraph-io/dgraph/dgraph/cmd/graphql/e2e_test_schema.graphql)" '{ query: "mutation addSchema($sch: String!) { addSchema(input: { schema: $sch }) { schema { schema } } }", variables: { sch: $schema }}' | curl -X POST -H "Content-Type: application/json" http://localhost:9000/admin -d @- | jq -r
```

MichaelJCompton:
Fixed in docs

PR #4234, opened 2019-11-01
Make comments in schemas work

MichaelJCompton:
Fixes https://github.com/dgraph-io/dgraph/issues/4227

Actually, the issue isn't a problem with the server - it was docs problem.

This PR does :
* Adds testing to ensure https://github.com/dgraph-io/dgraph/issues/4227 is ok
* Also makes `"""description"""` work as per GraphQL spec

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4234)


MichaelJCompton:
Looks like it's docs that's broken, not the server.

Will add more testing around `#` to check and then fixup a better way to load schema.

PR #4257, opened 2019-11-08
Added proper handling of int and float for math op

### LINKED ISSUE #4132: Math operations for int values fail for big values
v1.1.0-134-gd9767791

Yes

32gb ram, ubuntu 19.04

```
{
  set {
    _:user1 <money> "48038396025285290" .
  }
}

{
  q(func: has(money)) {
    f as money
    g: math(2+f)
  }
}
```

Expected Behaviour
```
"q": [
      {
        "money": 48038396025285290,
        "g": 48038396025285292
      }
    ]
```

Actual Result: 
```
"q": [
      {
        "money": 48038396025285290,
        "g": 48038396025285290
      }
    ]
```

This happens because all the math operations happen in float64 (https://github.com/dgraph-io/dgraph/blob/master/query/aggregator.go#L124), which only has a [precision of 53 bits](https://en.wikipedia.org/wiki/Double-precision_floating-point_format#IEEE_754_double-precision_binary_floating-point_format:_binary64). So int, being 64 bits can overflow when put into float64 and hence the error.

harshil-goel:
Fixes #4132

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4257)

### DIFF
--- a/query/aggregator.go
+++ b/query/aggregator.go
@@ -344,9 +344,8 @@ func ApplySince(a, res *types.Val) error {
 		a.Tid = types.FloatID
 		*res = *a
 		return nil
-	} else {
-		return errors.Errorf("Wrong type encountered for func since")
 	}
+	return errors.Errorf("Wrong type encountered for func since")
 }
 
 type MathUnaryFunction func(a, res *types.Val) error
--- a/query/aggregator.go
+++ b/query/aggregator.go
@@ -111,6 +111,314 @@ func compareValues(ag string, va, vb types.Val) (bool, error) {
 	return false, errors.Errorf("Invalid compare function %q", ag)
 }
 
+func ApplyIntAdd(a, b, c *types.Val) error {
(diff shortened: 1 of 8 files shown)


query/aggregator.go
```diff
@@ -111,73 +111,329 @@ func compareValues(ag string, va, vb types.Val) (bool, error) {
...
+	return errors.Errorf("Wrong type encountered for fun ceil")
+}
+
+func ApplySince(a, res *types.Val) error {
+	if a.Tid == types.DateTimeID {
+		a.Value = float64(time.Since(a.Value.(time.Time))) / 1000000000.0
+		a.Tid = types.FloatID
+		*res = *a
+		return nil
+	} else {
```
golangcibot: `if` block ends with a `return` statement, so drop this `else` and outdent its block (from `golint`)

PR #4274, opened 2019-11-14
Don't delete group if there is no member in the group

balajijinnah:
Manually tested by having two alpha with 1 replica count. removed one alpha and joined a new alpha. Alpha joined back to the existing group as expected
Signed-off-by: பாலாஜி <balaji@dgraph.io>

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4274)


balajijinnah:
@martinmr good catch about tablet. Lemme check and verify

PR #4279, opened 2019-11-15
Acquire read lock instead of exclusive lock for langBaseCache.

arijitAD:
Reduces lock contention in langBaseCache by using RWMutex instead of Mutex.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4279)


devinbost:
@arijitAD Could this PR have impacted upsert performance?

PR #4307, opened 2019-11-22
Infer schema as list if necessary.

martinmr:
If the same subject-predicate pair appears multiple times in a mutation
and the predicate needs to be created, create it as a list to avoid
losing data.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4307)


martinmr:
Closing in favor of the other PR.

PR #3202, opened 2019-03-25
Terraform script for running Dgraph on AWS

hackintoshrao:
- Terraform script for single node deployment.
- Added instructions in Readme.md.
- Tf.variable with the default configuration added.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/3202)

PR #4368, opened 2019-12-06
contrib: add terraform setup to run dgraph on AWS in standalone mode

fristonio:
* Add terraform setup for deploying dgraph on AWS in standalone mode(single EC2 instance in default VPC with default settings).
* Tested on AWS.
* Completes #3202 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4368)


fristonio:
___
*[contrib/config/terraform/aws/standalone/variables.tf, line 70 at r2](https://reviewable.io/reviews/dgraph-io/dgraph/4368#-LvRttlg7hFuYbbtEun8:-LvW-TTS5lBmPDzHokAn:b-q8hgx) ([raw file](https://github.com/dgraph-io/dgraph/blob/41469a95b697cd6965c4b5a5ddd7a98754ed108b/contrib/config/terraform/aws/standalone/variables.tf#L70)):*
<details><summary><i>Previously, danielmai (Daniel Mai) wrote…</i></summary><blockquote>

Make this false by default. Does that affect the output block?
</blockquote></details>

Ideally it shouldn't, the output block should be empty if there is no public IP. But I am thinking should we make this `false` by default, as the whole purpose of this standalone setup is to give user a deployed instance of dgraph to play with which he would ideally be looking to access outside the cluster. For the HA setup I think public ip false makes sense as we should give people the best security defaults for things that could go into production, but for the standalone setup I am not really sure. What's your take on this?

PR #4375, opened 2019-12-07
Fix how query responses are sent in upsert HTTP responses.

### LINKED ISSUE #4361: Predicate named "code" or "message" generates errors
master
yes

not relevant

Using dgraph:master and ratel ?dev:

1. mutate:
```
{
 set {
    _:x <name> "Bob" .
    _:x <code> "BobCode" .
    _:x <dgraph.type> "MyType" .
 }
}
```
2. create type MyType with both predicates
3. query:
```
{ res(func: type(MyType)) {
expand(_all_)
}}
```
4. query:
```
{ res(func: type(MyType)) {
code: code
name: name
}}
```
Step 4 generates error:
`Message: query alias [code] not allowed`

Moreover, you cannot edit schema using ratel (Schema->Types->MyType->Change Type generates blank page).

Both steps 3 and 4 should return the same result

I have found the comment in the source here: https://github.com/dgraph-io/dgraph/blob/master/edgraph/server.go#L1072 stating that "code", "message" and "uids" is used by the HTTP API.
But the problem for me is that some predicate names are disallowed (when using aliases) for no reason - is that documented anywhere?
Would it be possible if HTTP API used more private "_code" and "_message" names, as those are pretty standard as predicate names?  Or maybe the HTTP API implementation could be improved so that condition wouldn't be needed?

martinmr:
Right now the results of the upsert query are attached to the final
response at the same level as other field (e.g "message", "code") which
introduces the restriction that the names of these fields cannot be used
to name queries.

This PR gets rid of this restriction by attaching the response to the
queries in the "queries" field. This eliminates the need to check for
certain keywords in the query.

Fixes #4361

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4375)

### DIFF
--- a/dgraph/cmd/alpha/http.go
+++ b/dgraph/cmd/alpha/http.go
@@ -423,25 +423,8 @@ func mutationHandler(w http.ResponseWriter, r *http.Request) {
 	mp["code"] = x.Success
 	mp["message"] = "Done"
 	mp["uids"] = resp.Uids
-
-	// add query response if any, usual op if resp.Json == '{}' (i.e. l <= 2)
-	l := len(resp.Json)
-	if l > 2 && resp.Json[l-1] == '}' {
-		data, err := json.Marshal(mp)
-		if err != nil {
-			x.SetStatusWithData(w, x.Error, err.Error())
-			return
-		}
-
-		out := bytes.NewBuffer(resp.Json[:(l - 1)])
-		out.WriteRune(',')
-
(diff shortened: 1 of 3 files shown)


martinmr:
@mangalaman93 Is there any changes that have to be done for gRPC responses?

mangalaman93:
gRPC response is fine. Please get it approved by @manishrjain before we merge.

PR #4269, opened 2019-11-13
Remove duplicate code for query processing

mangalaman93:
The side effect of this is that now we return query response in an upsert.

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4269)

PR #4377, opened 2019-12-07
Add test for graphql vars in upsert

### LINKED ISSUE #4286: Upsert query variables fails
### Working Upsert

```
const query = `
    query {
        user as q(func: eq(user.id, "${request.userId}"))
    }
    `;

const mu = new dgraph.Mutation();
mu.setDelNquads(`uid(user) * * .`);

const req = new dgraph.Request();
req.setQuery(query);
req.setMutationsList([mu]);
req.setCommitNow(true);

const res = await dgraphClient.newTxn().doRequest(req);
```

When directly inserting the value of `request.userId` into the query string via string interpolation, the node returned by the query is successfully deleted.

### Failing upsert

```
const query = `
            query q($userId: string) {
                user as q(func: eq(user.id, $userId))
            }
        `;
    
const mu = new dgraph.Mutation();
mu.setDelNquads(`uid(user) * * .`);
    
const req = new dgraph.Request();
const vars = req.getVarsMap();
vars.set("$userId", `${request.userId}`);
req.setQuery(query);
req.setMutationsList([mu]);
req.setCommitNow(true);
    
const res = await dgraphClient.newTxn().doRequest(req);
```

When trying to use query variables and setting the corresponding entry in the variables map for `$userId` with `` `${request.userId}` ``, the node that should be returned by the query is not successfully deleted. 

Is this an unsupported feature for upserts or is something wrong with the code?

mangalaman93:
Fixes #4286 
Fixed by #4269, just adding a test for using graphql variables with an upsert.

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4377)

### DIFF
--- a/query/query1_test.go
+++ b/query/query1_test.go
@@ -24,6 +24,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/dgraph-io/dgo/v2/protos/api"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc/metadata"
 )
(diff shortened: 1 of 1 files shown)


mangalaman93:
When @prashant-shahi confirms, I will merge this PR.

PR #4404, opened 2019-12-12
allow @filter directive to be used with expand queries.

### LINKED ISSUE #3904: Support filter on expand
It would be very useful to filter what fields are included in an `expand` operation.

For instance, we could expand only the predicates pointing to values of type `Person` from a `Movie` as such:

```graphql
{
    q(func: type(Movie)) {
        expand(_all_) @filter(type(Person)) {
            first_name
        }
    }
}
```

This would expand the predicates of type `uid` or `[uid]` and show only the nodes associated to the type Person.

martinmr:
This PR introduces the new syntax `@fitler(type(<type_name>)`. It filters the results
of the query to only follow edges of the graph of the given type.
This implies that scalar fields will not be included in the response.

Fixes #3904

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4404)


query/query.go
```diff
@@ -1849,6 +1849,15 @@ func expandSubgraph(ctx context.Context, sg *SubGraph) ([]*SubGraph, error) {
...
 
+		// There's a types filter at this level so filter out any non-uid predicates
+		// since only uid nodes can have a type.
+		if len(child.Filters) > 0 {
+			preds, err = filterUidPredicates(ctx, preds)
+			if err != nil {
+				return out, err
+			}
+		}
+
```
golangcibot: func `getReversePredicates` is unused (from `unused`)

Issue #4383, opened 2019-12-09
[K8s / devops] Reviewing the liveness probe endpoint


hackintoshrao:
### What version of Dgraph are you using?
v1.1

The kubelet uses liveness probes to know when to restart a Container. 
For example, liveness probes could catch a deadlock, where an application is running, but unable to make progress. Restarting a Container in such a state can help to make the application more available despite bugs.

This is basically used to automate all the scenarios where an irresponsive database needs to restart.
But a wrong implementation would restart the pods when it's not necessary to do so.

It would be important to review it once and write a report on two things:

- Does the current liveness probe endpoint guarantee the detection of irresponsive database state?
- Review for any false positives.

fristonio:
Currently, we have the readiness probe and liveness probes in our helm [configuration](https://sourcegraph.com/github.com/dgraph-io/dgraph/-/blob/dgraph/cmd/alpha/run.go#L392) for dgraph alpha. There is a probable bug in the liveness probe in the sense that when a node tries to apply the raft snapshot, it sets the health status to unhealthy([here](https://github.com/dgraph-io/dgraph/blob/master/worker/draft.go#L836)), if probed at this time, Kubernetes will restart the container which should not be the case as the container is running.

Looking at a few implementations including [cockroachdb](https://sourcegraph.com/github.com/cockroachdb/cockroach@9d69708/-/blob/pkg/server/status.go#L623:24) I think that for the liveness probe we can have a bare HTTP handler returning status 200, which can detect cases like freeze, deadlocks, etc in the server. For readiness probe, the current implementation looks good which takes into account the state DB is operating in. Being more aggressive in terms of readiness probe would mean we will have to integrate health checks deep into our source code and probably brainstorm on what we will consider a not ready state for Dgraph.

PR #4405, opened 2019-12-12
contrib: update k8s liveness and readiness probes

fristonio:
* Fix liveness probe for dgraph alpha. Earlier there was a single endpoint for both liveness and readiness probes which can lead to some problems as described in #4383. This PR splits the single endpoint into two, one for readiness and one for liveness. The liveness probe does not take into consideration if the alpha is ready to take request, but ensure that the container is not in a frozen state and can serve some traffic.

The liveness probe can be triggered by providing a `live=1` GET parameter to the existing `/health` endpoint. The `/health` endpoint, however, corresponds to a readiness probe. This is similar to how cockroach DB handles these probes. 

* Add liveness probe to dgraph zero.
* Update k8s manifests to include both liveness and readiness probes.

This will require an update in the dgraph docker image `dgraph/dgraph`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4405)

### DIFF
--- a/dgraph/cmd/zero/http.go
+++ b/dgraph/cmd/zero/http.go
@@ -232,6 +232,13 @@ func (st *state) getState(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (st *state) pingResponse(w http.ResponseWriter, r *http.Request) {
+	x.AddCorsHeaders(w)
+
+	w.WriteHeader(http.StatusOK)
+	w.Write([]byte("OK"))
+}
+
 func (st *state) serveHTTP(l net.Listener) {
 	srv := &http.Server{
 		ReadTimeout:  10 * time.Second,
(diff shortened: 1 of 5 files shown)


dgraph/cmd/zero/http.go
```diff
@@ -232,6 +232,13 @@ func (st *state) getState(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (st *state) pingResponse(w http.ResponseWriter, r *http.Request) {
+	x.AddCorsHeaders(w)
+
+	w.WriteHeader(http.StatusOK)
+	w.Write([]byte("OK"))
```
golangcibot: Error return value of `w.Write` is not checked (from `errcheck`)

PR #4441, opened 2019-12-18
contrib: add prometheus monitoring manifest

fristonio:
* Add Prometheus Kubernetes manifest for monitoring.
* Update helm chart to respect the added monitoring manifest.
* Update `dgraph-ha.yaml` Kubernetes deployment manifest to respect the
  monitoring configuration.
* Update `dgraph-ha.yaml` Kubernetes manifest to use default `ClusterIP` service instead of `LoadBalancer`.
* Add Prometheus alert manager alerting rules.

A sample Prometheus target page is shown below:

screenshot-localhost_13000-2019 12 18-19_19_28

Alerting configuration page for Prometheus:

screenshot-localhost_9090-2019 12 19-18_13_18

Signed-off-by: fristonio <deepshpathak@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4441)


fristonio:
@danielmai I have updated the monitoring configuration with alerts rules for prometheus, can you take a look?
I have a few doubts related to the memory metrics alerting, as on what criteria should we consider as an alert situation. It would be great if you can elaborate a bit on that. Also, can you also pass a review on the default values for the configured alerts in the manifests?

danielmai:
___
*[contrib/config/kubernetes/dgraph-ha/dgraph-ha.yaml, line 19 at r2](https://reviewable.io/reviews/dgraph-io/dgraph/4441#-Lw_zwb4Cw3q6TNFepLo:-LxgHGTbA_cuXp2sBSQT:b-9mmbxm) ([raw file](https://github.com/dgraph-io/dgraph/blob/dd33e1b46befb007682dec20ab41c1326d22a927/contrib/config/kubernetes/dgraph-ha/dgraph-ha.yaml#L19)):*
<details><summary><i>Previously, fristonio (Deepesh Pathak) wrote…</i></summary><blockquote>

Ah, I changed this so I don't accidentally create multiple load balancers on our cloud provider given they are very expensive. Also, I think we should have them as `ClusterIP` itself if our main aim is to make them accessible locally outside of k8s, since for that purpose we can always use `kubectl port-forward` command. But if a customer creates a cluster using this configuration on his cloud provider it will cost him a lot of money as load balancers are pretty expensive and according to our configuration, it will create one for each service. Additionally it publically exposes the endpoints using the load balancers. Using ClusterIP it would be similar to what we are doing for Terraform and CloudFormation configuration for HA cluster where we don't expose anything publically.
</blockquote></details>

We should add documentation about how to use port-forward with ClusterIP.

PR #4454, opened 2019-12-20
Adds @noconflict directive to prevent conflict detection.

### LINKED ISSUE #4079: Concurrent write of nodes referencing the same node fails with Dgraph v1.1
v1.1

yes

16Gb, Linux 4.19.67-1-lts

run test 

```
package main

import (
	"context"
	"encoding/json"
	"strconv"
	"testing"

	"github.com/dgraph-io/dgo/v2"
	"github.com/dgraph-io/dgo/v2/protos/api"
	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"
	"google.golang.org/grpc"
)

func Test_it_writes_concurrently(t *testing.T) {
	conn, err := grpc.Dial(dgraphAddr(), grpc.WithInsecure())
	if err != nil {
		t.Fatalf("conntect: %v", err)
	}

	dg := dgo.NewDgraphClient(api.NewDgraphClient(conn))
	defer dg.Alter(context.Background(), &api.Operation{DropAll: true})

	schema := `
type node {
  name: string
  child: uid
}

name: string .
child: uid .
`
	err = dg.Alter(context.Background(), &api.Operation{Schema: schema})
	require.NoError(t, err)

	type node struct {
		ID   string `json:"uid"`
		Name string `json:"name"`

		Child *node `json:"child"`
	}

	child := node{ID: "_:blank-0", Name: "child"}
	js, err := json.Marshal(child)
	require.NoError(t, err)

	res, err := dg.NewTxn().Mutate(context.Background(), &api.Mutation{SetJson: js, CommitNow: true})
	require.NoError(t, err)

	in := []node{}
	for i := 0; i < 2; i++ {
		in = append(in, node{ID: "_:blank-0", Name: strconv.Itoa(i + 1), Child: &node{ID: res.GetUids()["blank-0"]}})
	}

	errChan := make(chan error)
	for i := range in {
		go func(n node) {
			js, err := json.Marshal(n)
			require.NoError(t, err)

			_, err = dg.NewTxn().Mutate(context.Background(), &api.Mutation{SetJson: js, CommitNow: true})
			errChan <- err
		}(in[i])
	}

	errs := []error{}
	for i := 0; i < len(in); i++ {
		errs = append(errs, <-errChan)
	}

	for _, e := range errs {
		assert.NoError(t, e)
	}
}
```

Expected behaviour: no error and two nodes are created
Actual behaviour: One mutate request returns  ```Transaction has been aborted. Please retry```

### Problem description

I can reproduce this to be working with Dgraph 1.0.16. If the nodes that are written concurrently don't reference the same child node everything works. Hence I guess the issue is with the child node reference.

arijitAD:
Fixes #4079
This is an experimental feature.
It adds @noconflict directive to prevent conflict detection at the predicate level. This can cause data loss, especially when using count indexes. This is not a recommended directive, but exists to help avoid conflicts for predicates which don't have high correctness requirements.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4454)


arijitAD:
@martinmr when using @noconflict directive, the user is aware that there could be data loss since we are allowing conflicting transactions to proceed. Hence I don't think there is any special handling need for @count directive. We could write about the consequence of using @noconflict in Dgraph docs.

PR #4496, opened 2020-01-03
Fix move tablet Jepsen test failure

manishrjain:
Jepsen bank tests with "move-tablet" nemesis were failing due to:

- The destination group was serving txns with start ts < move ts, before it had received the tablet.
- The source group was serving txns with starts ts > move ts, after it had deleted the tablet.

For transactions which are doing writes, this wasn't an issue because commits require consultation with Zero. During commit, Zero ensures that the group which served the data is also the current owner of that predicate. Otherwise, it aborts the transaction. So, this was only an issue for read-only transactions where Zero isn't consulted.

This PR fixes the issue by doing two things:

- It sets the move timestamp in Tablet and membership state, which the destination group uses to reject any transactions with start ts < move ts.
- After the move, the group checksum changes. When sending tablet deletion command to the source group, this checksum is included in the request. The source group uses this info to block until it has the latest membership information from Zero. That way, it knows not to serve the predicate before deleting it.

With these changes, Jepsen tests with "move-tablet" nemesis are passing on my machine.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4496)


aphyr:
I'm a bit confused here--@manish, you said that this problem only impacted read-only transactions, but in #4543, we saw permanent effects, which suggests a transaction encountered read skew and managed to promote it back into a write. Any idea why we saw persistent changes in account totals?

aphyr:
It also sounds like this might explain why we saw spurious null values, because the tablets aren't present at all, but... why would we see read skew where an actual *value* was present?

manishrjain:
> but in #4543, we saw permanent effects, which suggests a transaction encountered read skew and managed to promote it back into a write

My guess is this:
1. Race cond between read and tablet having just moved from src group to dst group.
2. Dgraph Alpha does not yet have the latest membership info, and services the wrong read (because the server thought it still was serving the predicate, after having deleted the predicate. This is the fix this PR makes).
3. Jepsen after doing the reads, does the writes in a separate network call, which goes to the dst group (membership info has been propagated by then), so can be written.
4. During commit, Zero gets info about conflict keys and the map of which group served which predicate for the writes, Zero checks that the write was done to the correct group (i.e. dst group) and lets it be committed. Dgraph does NOT keep track of reads -- graph systems can read a lot of data.

I think not servicing such reads would fix the problem and so far the testing has been supportive of that. This PR fixed majority of the failures with bank/move-tablet. I'm chasing some of the edge cases around this fix right now (ensuring followers do the same actions as the leader in terms of waiting for membership update before deleting their data).

manishrjain:
> It also sounds like this might explain why we saw spurious null values, because the tablets aren't present at all, but... why would we see read skew where an actual value was present?

That might be the different issue with posting list splits. In my branch, I've turned off the splits, and I haven't seen invalid values being served (only nil values, consistent with my above theory).

PR #4504, opened 2020-01-06
Multi-Tenancy support

balajijinnah:
This PR will add support for multi-tenancy.

Schema: Every Alter command takes namespace as a parameter, when schema node is parsed. The schema name is derived from prefixing namespace in the predicate.

When user query for schema, predicates for the current schema is filtered out and namespace in the predicate are trimmed.

Mutation: Predicate are prefixed with namespace while parsing RDF or JSON

Query: predicate is prefixed with namespace when creating taskQuery. So, no need to change in subgraph or outnode.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4504)


ChStark:
It would be great if this namespace could be a query parameter on the `/query` endpoint of the HTTP API

x/keys.go
```diff
@@ -162,16 +173,18 @@ func ReverseKey(attr string, uid uint64) []byte {
 // byte 0: key type prefix (set to DefaultPrefix)
 // byte 1-2: length of attr
 // next len(attr) bytes: value of attr
+// next two byte: length of namespace
+// next len(namespace) bytes: value of namespace
```
balajijinnah: use ascii delimitor 30. namespace should go first

worker/mutation.go
```diff
@@ -75,7 +75,7 @@ func runMutation(ctx context.Context, edge *pb.DirectedEdge, txn *posting.Txn) e
 		return err
 	}
 
-	key := x.DataKey(edge.Attr, edge.Entity)
+	key := x.DataKey(edge.Attr, namespace, edge.Entity)
```
balajijinnah: first arg should be namespace

PR #4548, opened 2020-01-10
Don't ignore errors when writing to buffers or string writers.

martinmr:
It's very unlikely that these calls will fail, so calling x.Check2
should be safe. This will get rid of the unhandled error warnings.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4548)


graphql/schema/gqlschema.go
```diff
@@ -1192,16 +1193,16 @@ func generateInputString(typ *ast.Definition) string {
 func generateEnumString(typ *ast.Definition) string {
 	var sch strings.Builder
 
-	sch.WriteString(fmt.Sprintf("%senum %s {\n", generateDescription(typ.Description), typ.Name))
+	x.Check2(sch.WriteString(fmt.Sprintf("%senum %s {\n", generateDescription(typ.Description), typ.Name)))
```
golangcibot: line is 104 characters (from `lll`)

PR #4531, opened 2020-01-09
Support filtering on nonindexed predicate

animesh2049:
Fixes https://github.com/dgraph-io/dgraph/issues/4305

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4531)

PR #4591, opened 2020-01-16
Add info about changes introduced by PR #4531

animesh2049:
Now we allow filtering based on non-idexed predicate for comparision functions.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4591)

### DIFF
--- a/wiki/content/query-language/index.md
+++ b/wiki/content/query-language/index.md
@@ -212,9 +212,9 @@ Query Example: Some of Bollywood director and actor Farhan Akhtar's movies have 
 
 ## Functions
 
-{{% notice "note" %}}Functions can only be applied to [indexed]({{< relref "#indexing">}}) predicates.{{% /notice %}}
-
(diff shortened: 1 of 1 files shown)


sleto-it:
Hi, one comment from me. In general, it would be extremely useful, when something changes or is a totally new feature, that we document the version when the change happened or when the feature was introduced

Examples for new features:

a) [...]. This option has been introduced in vX.Y
b) Starting from vX.Y, it is now possible to [...]

Examples for changes:

a) Beginning with vX.Y, [...]. Previously {quick description of old behavior)

**This will help a lot both users and support team**

Example / reference (from MySQL):

**1.** https://dev.mysql.com/doc/refman/8.0/en/with.html

Screen Shot 2020-01-16 at 2 31 49 PM

**2.** https://dev.mysql.com/doc/refman/8.0/en/comparison-operators.html

Screen Shot 2020-01-16 at 2 34 30 PM

Thanks,

PR #4660, opened 2020-01-23
Don't log GraphQL introspection queries

MichaelJCompton:
In v3, we are logging queries and what we generate from them.  That helps with debugging, but it pollutes the logs too much if we are also logging the introspection queries from GraphQL dev tools, because they come in way too often.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4660)

 

Docs Preview: [Dgraph Preview](https://dgraph-ee601b7fb6-42158.surge.sh)


graphql/resolve/resolver.go
```diff
@@ -324,12 +325,16 @@ func (r *RequestResolver) Resolve(ctx context.Context, gqlReq *schema.Request) *
 	}
 
 	if glog.V(3) {
-		b, err := json.Marshal(gqlReq.Variables)
-		if err != nil {
-			glog.Infof("Failed to marshal variables for logging : %s", err)
+		// don't log the introspection queries they are sent too frequently
+		// by GraphQL dev tools
+		if !op.IsQuery() || !strings.HasPrefix(op.Queries()[0].Name(), "__") {
```
pawanrawal: The second check in the OR block could also have compared the QueryType to make it a bit clearer.

PR #4664, opened 2020-01-24
Fix error check warnings in contrib/integration.

martinmr:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4664)


contrib/integration/bigdata/main.go
```diff
@@ -188,7 +188,9 @@ func mutate(c *dgo.Dgraph) error {
 			continue
 		}
 		payload := make([]byte, 16+rand.Intn(16))
-		rand.Read(payload)
+		if _, err := rand.Read(payload); err != nil {
```
golangcibot: G404: Use of weak random number generator (math/rand instead of crypto/rand) (from `gosec`)

PR #4668, opened 2020-01-24
Added liveness and readiness probes for k8s-HA

prashant-shahi:
In `v1.1.1` or present `latest` Dgraph docker image, some of the mentioned endpoints were not present. To make it work with that version, we can publish the patch with the endpoints in `v1.1.2` after merging changes from [prashant/adding-health-endpoint](https://github.com/dgraph-io/dgraph/tree/prashant/adding-health-endpoint) branch.

In `v1.2.0-rc1`, though the endpoints are present, the issue was `Alpha` pod not being able to communicate to `Zero` pod due to the `readinessProbe` in `Zero`, and the `Alpha` pod was waiting for the connection to happen to be ready.

Below configuration works for `v1.2.0-rc1` branch. I tested with JS-Flock.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4668)


hackintoshrao:
Hey @prashant-shahi ,

I don't think we need to make the readiness or liveness work in v1.1 containers.

Removing the readiness probe on zero is not the solution, we need to fix it, rather than removing it from the config.
So, let's identify the issue with the readiness probe for zero, and fix it from the source, that would be the right solution.

prashant-shahi:
@hackintoshrao Since, the connections between zero pods were to be made first for it *to be ready*, and they won't be able to ping each other unless they were marked ready. So, I had picked this solution to ignore readiness probe in zero. But I do get your concern.

I have pushed new changes to include the previously ignore probe along with other necessary changes to resolve this previously seen issue.

prashant-shahi:
I have tested this config file to work well with JS-Flock to load Tweet data.

danielmai:
___
*[contrib/config/kubernetes/dgraph-ha/dgraph-ha.yaml, line 99 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/4668#-M-LdOl80zdNzuCHIQkt:-M-LdOl80zdNzuCHIQku:b-esylud) ([raw file](https://github.com/dgraph-io/dgraph/blob/9bd78090315a68b90c685eee167df686aedf8f81/contrib/config/kubernetes/dgraph-ha/dgraph-ha.yaml#L99)):*
> ```YAML
>   # the sake of the other Dgraph Zero pods even before they're ready, since they
>   # have to be able to talk to each other in order to become ready.
>   publishNotReadyAddresses: true
> ```

Thanks for having the comments to explain why `publishNotReadyAddresses` is needed.

PR #4673, opened 2020-01-25
Escape MySQL column names

andres-lowrie:
This change wraps all the column names with backticks so that MySQL won't complain when a table has columns that are also reserved keywords in MySQL.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4673)

### DIFF
--- a/dgraph/cmd/migrate/dump.go
+++ b/dgraph/cmd/migrate/dump.go
@@ -91,7 +91,12 @@ func (m *dumpMeta) dumpTable(table string) error {
 	tableGuide := m.tableGuides[table]
 	tableInfo := m.tableInfos[table]
 
-	query := fmt.Sprintf(`select %s from %s`, strings.Join(tableInfo.columnNames, ","), table)
+	escapedColNames := make([]string, len(tableInfo.columnNames))
+	for _, c := range tableInfo.columnNames {
+		escapedColNames = append(escapedColNames, fmt.Sprintf("`"+"%s"+"`", c))
+	}
+
+	query := fmt.Sprintf(`select %s from %s`, strings.Join(escapedColNames, ","), table)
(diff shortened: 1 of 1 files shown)


andres-lowrie:
I was migrating a database that uses MySQL reserved keywords for column names thus the migrate tool  was showing the following MySQL error when trying to reference said column in a query:

```
while writing data file: while dumping table food: Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'groups,id,in_fndds,long_description,manufacturers,refuse_description,refuse_perc' at line 1
``` 

The problem was with the column name `groups`, this I figured surrounding the column name with backticks should allow users to easier migrate a larger variety of MySQL tables.

andres-lowrie:
> > Reviewed 1 of 1 files at r2.
> > _[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4673)_ status: all files reviewed, 1 unresolved discussion (waiting on @andres-lowrie and @manishrjain)
> 
> _[dgraph/cmd/migrate/dump.go, line 95 at r2](https://reviewable.io/reviews/dgraph-io/dgraph/4673#-LzhUr2gDzWZBazyu7pl:-LzhUr2gDzWZBazyu7pm:b-rlog18) ([raw file](https://github.com/dgraph-io/dgraph/blob/ffc4f52c8a18994d85056b6efc84e2488378b93c/dgraph/cmd/migrate/dump.go#L95)):_
> 
> Quoted 4 lines of code…
> using the same logic twice. This could be wrapped in a method.

I put in a function instead of method out of habit and figured that the wrapping logic was generic enough, idk

If this should be method I'm glad to change it

mangalaman93:
Feel free to re-open this PR against main, closing it for now. Thank you for the PR.

PR #4681, opened 2020-01-28
Don't write edge types into Dgraph types

MichaelJCompton:
When we started GraphQL, you had to write the edge types in with the Dgraph types.

like ...

```
type T {
  pred: String .
}
pred: String .
```

Since then, that's changed.  This updates the Dgraph schema we write so it removes those predicate types and thus we don't get warnings when we update the schema.

so we now write:

```
type T {
  pred
}
pred: String .
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4681)


MichaelJCompton:
The failing tests are failing on master - unrelated

PR #4685, opened 2020-01-28
support @deprecated

MichaelJCompton:
@deprecated is required by GraphQL spec ... there's nothing we have to do to support it other than allow it to pass through.  It's the client tools that should react to it.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4685)

  

Docs Preview: [Dgraph Preview](https://dgraph-2843bc9934-42173.surge.sh)

### DIFF
--- a/graphql/resolve/add_mutation_test.yaml
+++ b/graphql/resolve/add_mutation_test.yaml
@@ -25,6 +25,7 @@
           "Author.dob":"2000-01-01",
           "Author.posts":[]
         }
+
 -
   name: "Add multiple mutation with variables"
   gqlmutation: |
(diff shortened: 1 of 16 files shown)


MichaelJCompton:
The failing tests are failing on master - unrelated

MichaelJCompton:
@pawanrawal I've added the introspection tests.  Looks like there's a bug in gqlgen that means the result isn't right for enums.

Because fixing it isn't a code change at our side - it's just uncommenting that test - I'm suggesting that we merge this as is, and I'll add a fix for gqlgen and then enable the test separately.

PR #4693, opened 2020-01-29
Optimize types.ParseTime and add benchmark test for it

aakside:
benchmark                old ns/op     new ns/op     delta
BenchmarkParseTime-8     7103          4379          -38.35%

benchmark                old allocs     new allocs     delta
BenchmarkParseTime-8     39             16             -58.97%

benchmark                old bytes     new bytes     delta
BenchmarkParseTime-8     2144          896           -58.21%

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4693)

 

Docs Preview: [Dgraph Preview](https://dgraph-cf298c6467-43108.surge.sh)

### DIFF
--- a/types/scalar_types.go
+++ b/types/scalar_types.go
@@ -24,6 +24,11 @@ import (
 )
 
 const nanoSecondsInSec = 1000000000
+const dateFormatY = "2006" // time.longYear
+const dateFormatYM = "2006-01"
+const dateFormatYMD = "2006-01-02"
+const dateFormatYMDZone = "2006-01-02 15:04:05 -0700 MST"
+const dateTimeFormat = "2006-01-02T15:04:05"
 
 // Note: These ids are stored in the posting lists to indicate the type
 // of the data. The order *cannot* be changed without breaking existing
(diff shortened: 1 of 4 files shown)


mangalaman93:
@aakside Thank you for raising the PR. Will you check the CI build, seems like some of the tests fail. Also, will you provide the benchmark results with comparison using `benchcmp` tool?

aakside:
@mangalaman93 
> @aakside Thank you for raising the PR. Will you check the CI build, seems like some of the tests fail. Also, will you provide the benchmark results with comparison using `benchcmp` tool?

Done, and edited first comment to include benchmark comparison

aakside:
___
*[types/scalar_types.go, line 193 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/4693#-LzsQfRo0xsacu2wZGDA:-LzsZB0V9ZDF2wtwDsQ9:b-rxm92t) ([raw file](https://github.com/dgraph-io/dgraph/blob/61e4c59f9cec9877c43a8c38dcc1ebce82ffb7cb/types/scalar_types.go#L193)):*
>If && is correct, then I don't think the first condition is doing anything.

Counterexample: `val = "2006-04"`
* Without the first condition, we would incorrectly proceed to Parse val as dateFormatY
* With the first condition, we would not do that, and would instead Parse val as dateFormatYM

aakside:
___
*[types/scalar_types.go, line 197 at r2](https://reviewable.io/reviews/dgraph-io/dgraph/4693#-LzscUf696fTnP7DkiYq:-M-M_XZn5A5NXPg6dXh0:b98x36m) ([raw file](https://github.com/dgraph-io/dgraph/blob/c988329bd728bb5e0c036f4818fb766e0a8338de/types/scalar_types.go#L197)):*

Might check a dash here? Should optimize the rejection cases. In fact, the benchmark should be to check how fast can we reject strings which are not datetimes, but sort of look like them.

Could also look into building a larger regexp.
</blockquote></details>

Added a benchmark for rejections. I also tried checking for dashes and that made parse time slower so I left them out.

types/scalar_types.go
```diff
@@ -24,6 +24,12 @@ import (
 )
 
 const nanoSecondsInSec = 1000000000
+const dateFormatSY = "06"  // time.stdYear
```
golangcibot: const `dateFormatSY` is unused (from `unused`)

aakside:
___
*[types/scalar_types.go, line 207 at r7](https://reviewable.io/reviews/dgraph-io/dgraph/4693#-M-QFLSs-Akxg0K7Ty4K:-M-R5xUt8p1N_rEN7Apc:bpb3vwk) ([raw file](https://github.com/dgraph-io/dgraph/blob/a6b27f70c6c16db3ed83a53b655305ccd5459ddd/types/scalar_types.go#L207)):*

I'll double check these assumptions with the Go folks in Slack.

Or, check for a dash at 4th place, and if it's there, then go one by one for RFC, YMDZone and dateTimeFormat.
</blockquote></details>

I verified this slightly revised check is correct by referring to https://tools.ietf.org/html/rfc3339#section-5.6 and testing time.Parse() against several negative cases (e.g., without an offset, without offset colon and 'Z', only hours in UTC offsets, etc.) to ensure it fails as expected.

(I haven't yet gotten a response in the Gophers slack #general channel)

PR #4754, opened 2020-02-10
Remove HTTP admin endpoints

pawanrawal:
These endpoints are now served via the GraphQL Admin API and hence we are removing the HTTP endpoints. Another PR for updating the docs will follow.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4754)

 

Docs Preview: [Dgraph Preview](https://dgraph-e1f944314b-44322.surge.sh)


MichelDiz:
Just a point, wouldn't it be nice to leave that for compatibility? Some users might not like to be forced to use GraphQL for these operations. Sending curl commands would get bigger and more complex.

e.g:
do this
```
jq -n '
{ 
  query: "mutation {
  export(input: {
    format: "JSON"
  })
  {
    response {
      Response
    }
  }
}"' | 
curl -X POST -H "Content-Type: application/json" http://localhost:8080/admin -d @- | jq -r
```
Instead of just 

```
curl 'localhost:8080/admin/export?format=json'
```

PR #4769, opened 2020-02-13
graphql: Add test to query GQL Endpoint with TLS.

arijitAD:
Test that HTTPS/TLS options work with Graphql endpoint.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4769)


MichaelJCompton:
Should probably have a success and fail case - like the other tests in this file.

Issue #4799, opened 2020-02-18
[graphql] please ignore unsupported schema directive not throw error 


shynome:
<!-- If you suspect this could be a bug, follow the template. -->

### What version of Dgraph are you using?
v2.0.0-rc1

### Have you tried reproducing the issue with the latest release?
yes

### What is the hardware spec (RAM, OS)?

### Steps to reproduce the issue (command/config used to run Dgraph).

update schema with below content 
```graphql
enum Role {
  Admin
  User
}

directive @auth(r: Role = User) on FIELD_DEFINITION

type Product {
  id: ID!
  price: Float! @search @auth(r: Admin)
}
```
server throw error:
```log
I0218 09:17:04.812213       1 schema.go:68] Got updateGQLSchema request                                                                                                                                                                         
E0218 09:17:04.812743       1 panics.go:33] panic: runtime error: invalid memory address or nil pointer dereference.                                                                                                                            
 trace: goroutine 2738 [running]:                                                                                                                                                                                                               
runtime/debug.Stack(0xc023473000, 0x17a3820, 0x2371400)                                                                                                                                                                                         
   /usr/local/go/src/runtime/debug/stack.go:24 +0x9d                                                                                                                                                                                            
github.com/dgraph-io/dgraph/graphql/api.PanicHandler(0xc023473a28)                                                                                                                                                                              
   /tmp/go/src/github.com/dgraph-io/dgraph/graphql/api/panics.go:33 +0x57                                                                                                                                                                       
panic(0x17a3820, 0x2371400)                                                                                                                                                                                                                     
   /usr/local/go/src/runtime/panic.go:679 +0x1b2                                                                                                                                                                                                
github.com/dgraph-io/dgraph/graphql/schema.postGQLValidation(0xc0004a7f80, 0xc01ddb5b00, 0xa, 0x17, 0x0, 0x0, 0x0)                                                                                                                              
   /tmp/go/src/github.com/dgraph-io/dgraph/graphql/schema/gqlschema.go:340 +0x10f                                                                                                                                                               
github.com/dgraph-io/dgraph/graphql/schema.NewHandler(0xc01de31800, 0xbfb, 0xc023539530, 0x0, 0x0, 0x1)                                                                                                                                         
   /tmp/go/src/github.com/dgraph-io/dgraph/graphql/schema/schemagen.go:130 +0x47b                                                                                                                                                               
[...]

PR #4805, opened 2020-02-19
graphql: ignore unsupported schema directive

shynome:
#4799

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4805)


shynome:
Thanks your review
After add the test case I found the bug and now I have fixed that

PR #4820, opened 2020-02-20
graphql: Fix deleteGroup and deleteUser mutations.

pawanrawal:
deleteGroup should also delete the group from the user group lists that it belongs to. This was not happening before because we were not adding the delete mutation like we do for fields which have the `@hasInverse` directive. This happens now.

deleteUser should also delete the user from the groups that it belongs to. This was not happening because Dgraph didn't have the type info for `User` type and wasn't correctly deleting its edges. Now the type information is added to the list of initial types.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4820)

 

Docs Preview: [Dgraph Preview](https://dgraph-4bb7df26f6-45257.surge.sh)

### DIFF
--- a/x/x.go
+++ b/x/x.go
@@ -118,6 +118,12 @@ const (
 {"predicate":"dgraph.rule.predicate","type":"string","index":true,"tokenizer":["exact"],"upsert":true},
 {"predicate":"dgraph.rule.permission","type":"int"}
 `
+
+	InitialTypes = `
+	"types": [{"fields": [{"name": "dgraph.graphql.schema"}],"name": "dgraph.graphql"},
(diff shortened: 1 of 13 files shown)


x/x.go
```diff
@@ -118,6 +118,12 @@ const (
 {"predicate":"dgraph.rule.predicate","type":"string","index":true,"tokenizer":["exact"],"upsert":true},
 {"predicate":"dgraph.rule.permission","type":"int"}
 `
+
+	InitialTypes = `
+	"types": [{"fields": [{"name": "dgraph.graphql.schema"}],"name": "dgraph.graphql"},
+{"fields": [{"name": "dgraph.password"},{"name": "dgraph.xid"},{"name": "dgraph.user.group"}],"name": "User"},
```
golangcibot: line is 110 characters (from `lll`)

PR #4881, opened 2020-03-04
Disable @* language queries when the predicate does not support langs.

martinmr:
This changes makes "@." and "@*" behave the same way

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4881)

 

Docs Preview: [Dgraph Preview](https://dgraph-b943be8b6f-47260.surge.sh)


martinmr:
I checked. All valid combinations now return this error.

PR #4916, opened 2020-03-11
track operations and cancel when needed

mangalaman93:
we now track operations in the system such as rollup, indexing and snapshots. We cancel rollup whenever indexing or snapshot needs to be taken. We also do not allow multiple operations at the same time.

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4916)

### DIFF
--- a/worker/mutation.go
+++ b/worker/mutation.go
@@ -139,17 +139,23 @@ func runSchemaMutation(ctx context.Context, updates []*pb.SchemaUpdate, startTs 
 	}
 
 	// Ensure that rollup is not running.
-	if err := gr.Node.startTask(opBGIndex); err != nil {
+	wrtCtx := schema.GetWriteContext(context.Background())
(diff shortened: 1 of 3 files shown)


worker/mutation.go
```diff
@@ -139,15 +138,26 @@ func runSchemaMutation(ctx context.Context, updates []*pb.SchemaUpdate, startTs
 		time.Sleep(time.Second * 2)
 	}
 
+	// Ensure that rollup is not running.
+	wrtCtx := schema.GetWriteContext(context.Background())
+	if tmpCtx, err := gr.Node.startTask(wrtCtx, opIndexing); err != nil {
+		return err
+	} else {
```
golangcibot: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary) (from `golint`)

PR #4943, opened 2020-03-16
Add docs for background indexing

mangalaman93:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4943)

 
 
 

Docs Preview: [Dgraph Preview](https://dgraph-dafce04d46-48989.surge.sh)

### DIFF
--- a/wiki/content/query-language/index.md
+++ b/wiki/content/query-language/index.md
@@ -1840,7 +1840,7 @@ The `_all_` keyword requires that the nodes have types. Dgraph will look for all
 the types that have been assigned to a node, query the types to check which
 attributes they have, and use those to compute the list of predicates to expand.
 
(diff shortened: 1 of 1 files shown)


MichaelJCompton:
My other question is what is the error I get if I run a query when indexing is in progress?

I think that matters for the docs (I think it should be in the doc).   But I also thing it matters for usability.  

E.g. the examples in those clients libs require you to know a particular predicate and the expected indexes etc.  If my schema is a 200 line schema file, with multiple changes in it, then those functions in the clients lib don't really help me work out if indexing is still running - I'd have to parse the file myself, or know the changes and test for those specifically etc.

Whereas if I just try to run a query I'm interested in and it returns a helpful message that tells me that indexing is still running, then that's quite helpful.

PR #4998, opened 2020-03-21
Docker multistage builds

gja:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4998)


gja:
Closing this PR after discussions with @danielmai. It appears that this contrib dockerfile is not used directly for building as docker build ., but is instead set up with a CI server that builds ratel, badger and dgraph and puts all three into a single docker image.

PR #5000, opened 2020-03-23
Adding an Alpine Image

gja:
Please merge this in only once https://github.com/dgraph-io/dgraph/pull/4998 is merged in.

This adds make image-alpine to make an alpine image. Dgraph must be compiled on alpine since alpine uses musl-libc instead of glibc.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5000)


gja:
Closing this PR after discussions with @danielmai. It appears that this contrib dockerfile is not used directly for building as docker build ., but is instead set up with a CI server that builds ratel, badger and dgraph and puts all three into a single docker image.

PR #4951, opened 2020-03-17
Cherry-pick changes required for 1.2.2 release.

martinmr:
I will not squash this PR so the log is preserved.

My changes to the bulk loader are missing because the bulk loader has a lot of changes
and not all of them have been cherry-picked into this branch. Not sure what to do about
that since I didn't make all of those changes so I am not sure exactly what commits should
be cherry-picked.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/4951)

 

Docs Preview: [Dgraph Preview](https://dgraph-710d4db3af-48799.surge.sh)

PR #5009, opened 2020-03-23
Skip split keys during backup.

### LINKED ISSUE #5008: Split keys should be skipped during backup.
master

yes, also affects the beta of the march release

Try to backup a database with split keys. The process fails because those lists cannot be read directly
so they should be skipped.

Skip those lists.

martinmr:
Just like in the case of an export, the split keys should be skipped.

Fixes #5008

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5009)

### DIFF
--- a/worker/backup_processor.go
+++ b/worker/backup_processor.go
@@ -135,6 +135,12 @@ func (pr *BackupProcessor) WriteBackup(ctx context.Context) (*pb.Status, error) 
 			return false
 		}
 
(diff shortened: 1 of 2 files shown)


danielmai:
Does this affect other releases too aside from master / v20.03?

martinmr:
yeah, 1.2.2 should already be fixed by #4951 but I'll verify tomorrow just to be safe.

PR #5019, opened 2020-03-24
graphql: @dgraph(pred: "...") with @search

abhimanyusinghgaur:
We aren’t quite handling indexes correctly when the `@dgraph` predicate has a `@search`.

```graphql
type Country {
    ...
    name: String! @dgraph(pred: "name")
}

type Author {
    name: String! @dgraph(pred: "name") @search(by: [term])
    ...
}
```

This will write the predicate into the schema twice and only one will apply.  We should be smarter about this and do the following things:
* make sure we only write each predicate once.
* make sure all the predicate versions follow the below rules:
  1. all versions have the same dgraph type
  2. all versions either use @id, or none of them uses it
  3. all versions either use @secret, or none of them uses it
  4. a field from a type is not using the same predicate as another field in the interface that the type implements
  5. Two different interfaces, which are implemented by the same type, can't use same dgraph predicate for their fields
* set the indexes to the union of all the @search on the same pred name.

This PR takes care of all the above things.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5019)

### DIFF
--- a/graphql/schema/rules.go
+++ b/graphql/schema/rules.go
@@ -27,6 +27,7 @@ import (
 )
 
 func init() {
+	schemaValidations = append(schemaValidations, dgraphDirectivePredicateValidation)
 	defnValidations = append(defnValidations, dataTypeCheck, nameCheck)
 
 	typeValidations = append(typeValidations, idCountCheck, dgraphDirectiveTypeValidation,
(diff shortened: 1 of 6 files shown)


graphql/schema/schemagen.go
```diff
@@ -341,9 +398,19 @@ func genDgSchema(gqlSch *ast.Schema, definitions []string) string {
 				continue
 			}
 			fmt.Fprintf(&typeDef, "  %s\n", fld.name)
-			if !fld.inherited {
-				fmt.Fprintf(&preds, "%s: %s%s %s%s.\n", fld.name, f.typ, f.index, f.upsert,
+			if !fld.inherited && !predWritten[fld.name] {
+				indexStr := ""
+				if len(f.indexes) > 0 {
+					indexes := make([]string, 0)
+					for index, _ := range f.indexes {
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
					for index := range f.indexes {
```

graphql/schema/rules.go
```diff
@@ -39,6 +40,206 @@ func init() {
...
+	checkConflictingFieldsInImplementedInterfacesError := func(typ *ast.Definition) {
+		fieldsToReport := make(map[string][]string)
+		interfaces := typ.Interfaces
+
+		for i := 0; i < len(interfaces); i++ {
+			intr1 := interfaces[i]
+			interfacePreds1 := interfacePreds[intr1]
+			for j := i + 1; j < len(interfaces); j++ {
+				intr2 := interfaces[j]
+				for fname, _ := range interfacePreds[intr2] {
```
golangcibot, line 124: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
				for fname := range interfacePreds[intr2] {
```

PR #5048, opened 2020-03-26
Fix protobuf regeneration.

### LINKED ISSUE #5047: Cannot run "make regenerate"
master

yes

linux

run "make regenerate". it fails with the following:

```
➜  protos git:(master) make regenerate
Checking for semantic version 3.6.1 or newer
OK
Upgrading proto libraries before regenerating.
go: found github.com/golang/protobuf/protoc-gen-go in github.com/golang/protobuf v1.3.5
go: found github.com/gogo/protobuf/protoc-gen-gofast in github.com/gogo/protobuf v1.3.1
github.com/gogo/protobuf/gogoproto/gogo.proto: File not found.
pb.proto:29:1: Import "github.com/gogo/protobuf/gogoproto/gogo.proto" was not found or had errors.
make: *** [Makefile:44: regenerate] Error 1
```

It also changes the proto file:

```
diff --git a/protos/pb.proto b/protos/pb.proto
index 9d89b6de7..0574a1cc3 100644
--- a/protos/pb.proto
+++ b/protos/pb.proto
@@ -25,7 +25,7 @@ syntax = "proto3";
 package pb;

 import "api.proto";
-import "github.com/dgraph-io/badger/v2/pb/pb.proto";
+import "github.com/dgraph-io/badger/pb/pb.proto";
 import "github.com/gogo/protobuf/gogoproto/gogo.proto";

 /* import "gogoproto/gogo.proto"; */
```

I don't think this is caused by a change in the code because I went back to commits that I know for sure used to work and I am still seeing this error. Seems like the behavior of one of our libraries changed. But I haven't been able to narrow it down further.

make regenerate works

martinmr:
Commit c9b83927d3855d8542f1549befd140cc7b051738 breaks proto generation.
It seems like the protobuf is using the gogo library, which is not being
correctly sourced in the Makefile (because it's no longer in
$GOPATH/src).

Turning gomodules off when updating gogo and protoc fixes the issue for
now.

Fixes #5047

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5048)

### DIFF
--- a/protos/Makefile
+++ b/protos/Makefile
@@ -31,8 +31,8 @@ clean:
 check:
 	@./depcheck.sh && \
 		(echo "Upgrading proto libraries before regenerating." ; \
-		go get -u github.com/golang/protobuf/protoc-gen-go ; \
-		go get -u github.com/gogo/protobuf/protoc-gen-gofast)
+		GO111MODULE=off go get -u github.com/golang/protobuf/protoc-gen-go ; \
+		GO111MODULE=off go get -u github.com/gogo/protobuf/protoc-gen-gofast)
 
 .PHONY: regenerate
 regenerate: check clean


martinmr:
yeah, I'll let @abhimanyusinghgaur check if we can get by without that gogo library.

PR #5055, opened 2020-03-30
Get most tests working on OSX

gja:
### Description.

This PR gets most tests working on OSX.

* Apparently, if GOOS is set to linux (for cross compiling), then CGO_ENABLED is set to false as OSX's libc is not compatible with ubuntu's libc, which causes many things to fail. In this PR, OSX's compile for docker happens within a docker run command (which we should also use for linux eventually)
* Docker for Mac allocates 2GB for the linux subsystem, which is not enough for dgraph's tests (and things silently fail). test.sh now spits out a warning if you are running Docker for Mac with too little RAM (< 6GB)
* Some tests (like `load-json/load_test.go`), work by calling out to the dgraph binary. Thus we also need the dgraph binary compiled for the correct environment on the machine. This PR also introduces `testutil.DgraphBinaryPath()` which finds the dgraph binary using the default GOPATH if the GOPATH is unset

### Please indicate if this is a breaking change.

No

### Please indicate if this is an enterprise feature.

No

### Please indicate if documentation needs to be updated.

No

### Please indicate if end to end testing is needed.

No, it's updates to test.sh

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5055)

### DIFF
--- a/testutil/exec.go
+++ b/testutil/exec.go
@@ -18,6 +18,7 @@ package testutil
 
 import (
 	"fmt"
+	"go/build"
 	"io"
 	"os"
 	"os/exec"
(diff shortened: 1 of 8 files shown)


testutil/exec.go
```diff
@@ -21,6 +21,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"go/build"
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion

```

PR #5073, opened 2020-04-01
Include the total number of touched nodes in the query metrics.

martinmr:
Total is stored under a key called "_total".

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5073)


danielmai:
If there was a predicate named `_total`, then this total would conflict with that predicate's num uid count.

We should put the total as another key (called `num_uids_total`) under metrics:
```
{
  "metrics": {
    "num_uids": {
      "director.film": 7,
      "name": 46
    }
    "num_uids_total": 53
  }
}
```

PR #5147, opened 2020-04-09
Backup schema keys in incremental backups.

martinmr:
### Description.

The full schema should be backed up during incremental backups.
During restore, the schema should be cleared before processing each backup
to end up with only the final schema.

Also ensured that the Version of the schema keys is always one. AFAIK this is not
an issue but I think it's better to explicitly set the versions of these keys to 1 during
backups and restores.

### GitHub Issue or Jira number.

https://dgraph.atlassian.net/browse/DGRAPH-1224

### Affected releases.

20.03.1

### Changelog tags.

fixed

### Please indicate if this is a breaking change.

No

### Please indicate if this is an enterprise feature.

Yes

### Please indicate if documentation needs to be updated.

No

### Please indicate if end to end testing is needed.
 
I added unit tests to ensure the schema is correct but it'd be good to manually test this before.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5147)

### DIFF
--- a/worker/restore.go
+++ b/worker/restore.go
@@ -86,8 +86,12 @@ func loadFromBackup(db *badger.DB, r io.Reader, preds predicateSet) (uint64, err
 
 	// Delete schemas and types. Each backup file should have a complete copy of the
 	// schema.
-	db.DropPrefix([]byte{x.ByteSchema})
-	db.DropPrefix([]byte{x.ByteType})
+	if err := db.DropPrefix([]byte{x.ByteSchema}); err != nil {
+		return 0, err
+	}
+	if err := db.DropPrefix([]byte{x.ByteType}); err != nil {
+		return 0, err
+	}
 
 	loader := db.NewKVLoader(16)
 	var maxUid uint64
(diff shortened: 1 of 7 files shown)


worker/restore.go
```diff
@@ -84,6 +84,11 @@ func loadFromBackup(db *badger.DB, r io.Reader, preds predicateSet) (uint64, err
 	br := bufio.NewReaderSize(r, 16<<10)
 	unmarshalBuf := make([]byte, 1<<10)
 
+	// Delete schemas and types. Each backup file should have a complete copy of the
+	// schema.
+	db.DropPrefix([]byte{x.ByteSchema})
```
golangcibot, line 89: Error return value of `db.DropPrefix` is not checked (from `errcheck`)

worker/restore.go
```diff
@@ -84,6 +84,11 @@ func loadFromBackup(db *badger.DB, r io.Reader, preds predicateSet) (uint64, err
...
+	db.DropPrefix([]byte{x.ByteType})
```
golangcibot, line 90: Error return value of `db.DropPrefix` is not checked (from `errcheck`)

parasssh:
We should back port this 20.03 as well as 1.2

PR #5172, opened 2020-04-11
Script to start the test cluster

gja:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5172)

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -100,8 +100,10 @@ function FindCustomClusterTests {
 
 function FindDefaultClusterTests {
     touch $DEFAULT_CLUSTER_TESTS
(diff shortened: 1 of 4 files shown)


test.sh
```diff
@@ -100,8 +100,10 @@ function FindCustomClusterTests {
 
 function FindDefaultClusterTests {
     touch $DEFAULT_CLUSTER_TESTS
-    for PKG in $(grep -v -f $CUSTOM_CLUSTER_TESTS $MATCHING_TESTS); do
```
gja, line 103: I think grep -v -f works slightly differently on a mac, so I was always getting no tests matched whenever i passed in a matcher (like 'query').

joshua-goldstein:
`test.sh` is deprecated, we now use `t.go` to spin up our clusters @mangalaman93

PR #5206, opened 2020-04-14
Force users to explicitly tell restore command to run without zero.

martinmr:
The zero value passed to the command is optional, which can lead to
users forgetting about it and being surprised that the cluster does
not work (because the uid lease and the timestamp need to be updated
first).

This change makes the zero flag required unless the `--force_zero`
flag is explicitly passed. Not passing a zero value can still be useful
when only the p directories need to be generated.

Fixes DGRAPH-1251

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5206)

 
 

Docs Preview: [Dgraph Preview](https://dgraph-bf980ba833-54944.surge.sh)


martinmr:
Cherry-picked to the 1.2 and 20.03 branches.

PR #5278, opened 2020-04-22
Alpha: Expose compression_level option

jarifibrahim:
This PR exposes the `compression_level` flag in `alpha`. The flag can be set as 
```
dgraph alpha --badger.compression_level=xxx
```
A higher value of the compression level is more CPU intensive but offers better compression ratio. The default value is `3`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5278)


parasssh:
Understood. 
One follow up,
- Restore is a standalone tool today. Also, it does not insert via  alpha as you mentioned). So, it cannot know what alpha compression level is.

PR #5284, opened 2020-04-23
Restore: Handle incorrect encryption key

jarifibrahim:
Jira - DGRAPH-1281 and DGRAPH-1282

When the restore is started, we open a badger instance with the provided encryption key and then check if the backup file can be opened using the encryption key. If the backup cannot be opened using the encryption key, we return an error to the user but the badger DB stores the encryption key.
When the user tries to restore the same backup to the same badger instance without a key (because backup is unencrypted), badger complaints about the missing key. 

This PR fixes by opening badger after opening the backup file. This ensures we don't open an encrypted badger db for an unencrypted backup file.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5284)

### DIFF
--- a/worker/restore.go
+++ b/worker/restore.go
@@ -55,7 +55,7 @@ func RunRestore(pdir, location, backupId, keyfile string) LoadResult {
 			gzReader, err := gzip.NewReader(r)
 			if err != nil {
 				return 0, errors.Wrap(err,
-					"Unable to read backup. Ensure encryption key is correctly set")
+					"Unable to read the backup. Ensure encryption key is correct.")
 			}
 			// The badger DB should be opened only after creating the backup
 			// file reader. The following sequence of events can occur if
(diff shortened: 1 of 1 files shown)


worker/restore.go
```diff
@@ -47,6 +47,33 @@ func RunRestore(pdir, location, backupId, keyfile string) LoadResult {
...
 			dir := filepath.Join(pdir, fmt.Sprintf("p%d", groupId))
+			r, err := enc.GetReader(keyfile, r)
+			if err != nil {
+				return 0, err
+			}
+
+			gzReader, err := gzip.NewReader(r)
+			if err != nil {
+				return 0, errors.Wrap(err,
+					"Unable to read the backup. Ensure encryption key is correct.")
```
parasssh, line 58: keyfile is optional. Just return the err.

worker/restore.go
```diff
@@ -47,6 +47,33 @@ func RunRestore(pdir, location, backupId, keyfile string) LoadResult {
...
+			}
+			// The badger DB should be opened only after creating the backup
```
parasssh, line 60: Dont need this whole comment. Just the first sentence will suffice.

PR #5307, opened 2020-04-27
Add GraphQL admin endpoint to list backups.

martinmr:
An endpoint to read the backup manifests in the given location.

Fixes DGRAPH-1229

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5307)

 

Docs Preview: [Dgraph Preview](https://dgraph-f86e74c1b3-59162.surge.sh)


martinmr:
@pawanrawal For some reason I cannot get partial queries to work. Even if I query for only one field, I am getting the entire response back. I suspect it's because I am returning a list instead of a map but I am not sure. The schema seems to work fine and the graphql recognizes all the fields. Somehow, they are not being used and I end up with all the data.

Could you look at the code in graphql/admin and see if I am doing something wrong? Thanks.

pawanrawal:
There is a bug which was introduced recently in https://github.com/dgraph-io/dgraph/commit/17a9c79d166384bfa731fc6063e593dc2aae8d04. I have added a fix for it as part of https://github.com/dgraph-io/dgraph/pull/5317 which should be merged soon.

PR #5338, opened 2020-04-30
graphql: strip custom logic from exposed schema

abhimanyusinghgaur:
This PR removes custom directive from exposed schema. Particularly if the custom logic has any keys, private urls, etc embedded in it, it shouldn’t be public.

Fixes #GRAPHQL-347

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5338)


abhimanyusinghgaur:
No more required because of #5354.

PR #5359, opened 2020-05-04
Minor changes in fastJsonNode

ashish-goswami:
This PR further tries to simplify `fastJsonNode` processing. It has following changes:
* Return error response if encoded response is > 4GB in size.
* Replace `idMap` with `idSlice` in `encoder`.
* Refactor `AddListChild` and `AddMapChild` to remove `attr` from their arguments. Both functions
  accept parent, child and attr as arguments. This attr is equal to attr field of child, hence sending
  it explicitly can be avoided.
* Minimise calls to `idForAttr()`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5359)

### DIFF
--- a/query/outputnode.go
+++ b/query/outputnode.go
@@ -261,7 +261,7 @@ func (enc *encoder) AddListValue(fj fastJsonNode, attr uint16, v types.Val, list
 	return nil
 }
 
-func (enc *encoder) AddMapChild(fj fastJsonNode, val fastJsonNode) {
+func (enc *encoder) AddMapChild(fj, val fastJsonNode) {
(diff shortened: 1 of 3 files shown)


query/outputnode.go
```diff
@@ -859,6 +855,8 @@ type Extensions struct {
 	Metrics *api.Metrics    `json:"metrics,omitempty"`
 }
 
+const maxEncodedSize = int64(4 << 30) // 4GB
```
vvbalaji-dgraph, line 858: Would it make sense to have a query property file with such limits?

PR #5362, opened 2020-05-04
graphql: Do graphql query/mutation validation in the mock server.

pawanrawal:
@abhimanyusinghgaur  and I realized that GraphQL variables were not being defined in the query/mutation that were being sent to the remote server because of which those requests would have failed. I am modifying the mock server to validate the incoming request and populate the arguments from the variables so that the end to end tests are more real. We'll have to fix the code so that the generated queries/mutation define the variables that they actually use. 

TODO
---
- [x] Modify graphql mutations and graphql field batch operation endpoints to also do the validation.
- [x] Fix the code so that all tests pass.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5362)

 

Docs Preview: [Dgraph Preview](https://dgraph-4cd59732b0-61517.surge.sh)


pawanrawal:
Seems like some tests are failing because schema isn't getting updated fast enough. Will add a sleep for them as they pass locally.

PR #5379, opened 2020-05-06
Best effort queries do not wait for checksums to match.

martinmr:
Queries currently can get blocked when the group checksums don't match.
Best effort queries should not get blocked and should instead work with
the data that's available.

Fixes DGRAPH-1333

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5379)


martinmr:
@danielmai This fixes the issue when using the `---be` flag in the increment tool. I am still not sure if this is exactly what we want. Opening the PR to share my current progress.

PR #5468, opened 2020-05-19
Pstore and WALstore use different timeticker for value gc

JimWen:
Pstore and WALstore valueGC go routine in v1.2.x use the same timeticker  which would cause gc of p directory would't be apply sometimes. 

The related discuss thread is [here](https://discuss.dgraph.io/t/should-pstore-and-walstore-use-the-same-time-ticker-for-gc/6749/4).

Cherry-pic commit from master to fix this, and test on production env, it works fine and vlog size decreases significantly.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5468)

 

Docs Preview: [Dgraph Preview](https://dgraph-53d9c53006-64889.surge.sh)

### DIFF
--- a/dgraph/cmd/zero/run.go
+++ b/dgraph/cmd/zero/run.go
@@ -212,6 +212,10 @@ func run() {
 	x.Checkf(err, "Error while opening WAL store")
 	defer kv.Close()
 
+	gcCloser := y.NewCloser(1) // closer for vLogGC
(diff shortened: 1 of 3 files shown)


JimWen:
cla done

martinmr:
Thanks for finding the right commit.

PR #5488, opened 2020-05-20
Get lists of predicates and types before sending the snapshot.

martinmr:
There could be changes to the schema while the snapshot is being sent.
Therefore, the list of types and predicates that is sent at the end
should be computed before the snapshot is sent.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5488)

### DIFF
--- a/worker/snapshot.go
+++ b/worker/snapshot.go
@@ -220,16 +220,20 @@ func doStreamSnapshot(snap *pb.Snapshot, out pb.Worker_StreamSnapshotServer) err
 		return pk.IsSchema() || pk.IsType()
 	}
 
+	// Get the list of all the predicate and types at the time of the snapshot so that the receiver
+	// can delete predicates
+	predicates := schema.State().Predicates()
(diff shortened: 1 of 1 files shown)


worker/snapshot.go
```diff
@@ -220,16 +220,20 @@ func doStreamSnapshot(snap *pb.Snapshot, out pb.Worker_StreamSnapshotServer) err
...
 
-	// Indicate that sending is done. Send a list of all the predicate and types at the
-	// time of the snapshot so that the receiver can delete predicates
+	// Indicate that sending is done.
 	done := &pb.KVS{
 		Done:       true,
-		Predicates: schema.State().Predicates(),
-		Types:      schema.State().Types(),
+		Predicates: predicates,
+		Types:      types,
```
vvbalaji-dgraph, line 236: How is the snapshot used by an alpha? Would it have a problem if it interprets the data and encounters data for types that it doesn't know about?

PR #5503, opened 2020-05-22
Use rampMeter for Executor

ashish-goswami:
Currently we have `rampMeter` while applying proposals. This `rampMeter` puts a throttling on
memory used by underlying mutations present in proposals to be applied. However this works
correctly only in normal mode.
In ludicrous mode, we apply mutations for these proposals asynchronously. Hence above
rampMeter doesn't work as expected and this results in lot of memory usage in cases when there
are many proposals to apply.
This PR introduces a rampMeter for executor as well. Executor is responsible for applying mutations
in ludicrous mode.

**Testing:**
I tested this PR by running **live loader on 21M dataset** in ludicrous mode. On this PR live loader completes in **~5-6 minutes**. However when I changed `maxPendingEdgesSize` to 64KB(instead of 64MB currently) in this PR, live loader completes in **~10-11 minutes**.

**Benchmarking:**
I benchmarked using data generated by below script:
```go
package main

import (
  "bytes"
  "fmt"
  "os"
)

var (
  total int = 100000000
  pred      = 1024
)

func main() {
  f, err := os.OpenFile("test.rdf", os.O_CREATE|os.O_RDWR, 0755)
  if err != nil {
    panic(err)
  }
  defer f.Close()

  totalPerPred := total / pred

  buf := bytes.NewBuffer(nil)
  count := 1
  for i := 0; i < totalPerPred; i++ {
    for j := 0; j < pred; j++ {
      rec := fmt.Sprintf(`_:record_%d <pred_%d> "value_%d" .`, count, j, count)
      buf.WriteString(rec)
      buf.WriteString("\n")
      count++
      if count%100000 == 0 { 
        buf.WriteTo(f)
        buf.Reset()
      }   
    }   
  }

  buf.WriteTo(f)
  if err := f.Sync(); err != nil {
    panic(err)
  }
  fmt.Println("Done writing to file: ", count)
}
```
Above scripts generates ~100M records with 97K records/predicate and total of 1024 predicates.

**Master:**
Time to finish live loader: 8m44s
Alpha RAM(RES) usage: 10.9 GB

**This PR with maxPendingEdgesSize = 64KB:**
Time to finish live loader: 8m48s
Alpha RAM(RES) usage: 9.6 GB

**This PR with maxPendingEdgesSize = 64MB:**
Time to finish live loader: 8m32s
Alpha RAM(RES) usage: 10.5 GB

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5503)

 

Docs Preview: [Dgraph Preview](https://dgraph-222d09465b-67723.surge.sh)


vvbalaji-dgraph:
Thanks for the PR. Do we have data on memory usage in Ludicrous mode with and without this PR?

PR #5508, opened 2020-05-22
Use pre-allocated protobufs during backups.

martinmr:
Backups convert the values stored in disk into an intermediate format
(for backwards compatibility reasons). This creates a lot of
pb.PostingList and pb.BackupPostingList objects that are discarded
almost immediatly. Using preallocated protobufs (one per thread used by
the stream object) should avoid these allocations altogether.

Fixes DGRAPH-1312

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5508)

 

Docs Preview: [Dgraph Preview](https://dgraph-2396d9ce84-66219.surge.sh)

### DIFF
--- a/ee/backup/backup.go
+++ b/ee/backup/backup.go
@@ -34,12 +34,40 @@ import (
 	"github.com/dgraph-io/dgraph/x"
 )
 
+const (
+	// backupNumGo is the number of go routines used by the backup stream writer.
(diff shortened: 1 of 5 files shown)


martinmr:
I had to make some changes to cherry-pick this commit so I am letting the test suite run.

Issue #5403, opened 2020-05-08
Run mutations concurrently per predicate in Ludicrous mode


lgalatin:
### What version of Dgraph are you using?
20.03.1

### Have you tried reproducing the issue with the latest release?

yes
### Steps to reproduce the issue (command/config used to run Dgraph).
Please see https://discuss.dgraph.io/t/dgraph-cant-idle-without-being-oomkilled-after-large-data-ingestion/6543

minhaj-shakeel:
**Github issues have been deprecated.** 
This issue has been moved to [discuss](https://discuss.dgraph.io/t/8527). You can follow the conversation there and also subscribe to updates by changing your notification preferences.

 <img src="https://github.com/dgraph-io/benchmarks/blob/master/images/SetNotifications.png?raw=true" alt="drawing" width="40%"/>

PR #5535, opened 2020-05-28
Support for inserting single predicate concurrently in ludicrous_mode

ashish-goswami:
Fixes: https://github.com/dgraph-io/dgraph/issues/5403
Fixes: DGRAPH-1357

In ludicrous mode we apply mutations for single predicate in same order on all alpha nodes. This is
done by having one channel per predicate. Mutation edges are grouped based on predicates and
then pushed to respective predicate channel for further processing.
This works fine when we have edges distributed across predicates. In cases when we have less
predicates in edges/RDFs. This channel becomes bottleneck for predicate and affects throughput
of insertions.
This PR tries to address following things:
* Have fixed number of channels(workerChannels) instead of one channel per predicate. This can
  avoid creating channels/processing routine for predicates which have very few edges/records for
  them.
* Use (predicate, subject id) combination to distribute load across channels. This ensures equal load
  distribution across channels. This also takes care of the fact that mutation are applied in the same
  order on all alphas.

**Benchmarking:**
I did benchmarking on a system with ubuntu 18, 64GB RAM, 16 core processor.

**Master:**
Dataset    |                                                                           |Time to finish live loader| Alpha RAM
21M imdb dataset                                                                |5m27sec                        | 7.1 GB
100M RDFs with single predicate                                        |11m16s                          | 8 GB
~100M RDFs with 1024 predicates(~97K RDFs/predicate)|8m26s                            | 11 GB

**This PR:**
Dataset    |                                                                           |Time to finish live loader| Alpha RAM
21M imdb dataset                                                                |3m14sec                        | 6.6 GB
100M RDFs with single predicate                                        |6m35s                           | 7.3 GB
~100M RDFs with 1024 predicates(~97K RDFs/predicate)|8m45s                            | 9.5 GB

100M dataset was generated using below script:
```go
package main

import (
  "bytes"
  "fmt"
  "os"
)

var (
  total int = 100000000
  pred      = 1024
)

func main() {
  f, err := os.OpenFile("test.rdf", os.O_CREATE|os.O_RDWR, 0755)
  if err != nil {
    panic(err)
  }
  defer f.Close()

  totalPerPred := total / pred

  buf := bytes.NewBuffer(nil)
  count := 1
  for i := 0; i < totalPerPred; i++ {
    for j := 0; j < pred; j++ {
      rec := fmt.Sprintf(`_:record_%d <pred_%d> "value_%d" .`, count, j, count)
      buf.WriteString(rec)
      buf.WriteString("\n")
      count++
      if count%100000 == 0 { 
        buf.WriteTo(f)
        buf.Reset()
      }   
    }   
  }

  buf.WriteTo(f)
  if err := f.Sync(); err != nil {
    panic(err)
  }
  fmt.Println("Done writing to file: ", count)
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5535)

 

Docs Preview: [Dgraph Preview](https://dgraph-a847d5e832-67461.surge.sh)


worker/executor.go
```diff
@@ -109,7 +109,7 @@ func (e *executor) channelID(edge *pb.DirectedEdge) int {
 	b := make([]byte, len(attr)+8)
```
parasssh, line 109: Remove hard-coded "8".
`b := make([]byte, len(attr)+ unsafe.Sizeof(uid)`

worker/executor.go
```diff
@@ -109,7 +109,7 @@ func (e *executor) channelID(edge *pb.DirectedEdge) int {
...
 	x.AssertTrue(len(attr) == copy(b, attr))
 	binary.BigEndian.PutUint64(b[len(attr):len(attr)+8], uid)
```
parasssh, line 111: remove hard-coded "8".

worker/executor.go
ashish-goswami: No need to use `unsafe.Sizeof(uid)` as uint64 size is always 8.

worker/executor.go
ashish-goswami: same here.

parasssh: I mean not using hard-coded values.

PR #5549, opened 2020-05-29
Use -d instead of --detach when invoking docker-compose.

martinmr:
Some versions of docker-compose do not recognize the longer version of
this flag. Found this issue when trying to run the test suite in a
virtual machine.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5549)


martinmr:
Yeah, I didn't ran into issues when running the tests but jepsen was complaining about the docker-compose version.

danielmai:
Yup, exactly. Newer Docker Compose versions support different compose file versions. https://docs.docker.com/compose/compose-file/#compose-and-docker-compatibility-matrix

PR #5560, opened 2020-06-02
Replace shutdownCh and wait groups to a y.Closer for shutting down Alpha

pawanrawal:
Earlier we had `worker.ShutdownCh` which was used to pass information between `alpha/run.go` and `graphql/admin.go` which was causing confusion and was being used to stop the worker itself. We now have a `y.Closer` in `graphql/admin` which manages the WaitGroup and passing of signals between goroutines using channels. 

Fixes DGRAPH-1169

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5560)

### DIFF
--- a/dgraph/cmd/alpha/run.go
+++ b/dgraph/cmd/alpha/run.go
@@ -531,7 +531,7 @@ func setupServer(closer *y.Closer) {
 
 	go func() {
 		defer wg.Done()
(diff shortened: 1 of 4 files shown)


jarifibrahim:
Can't we use y.Closers instead of shutdown channel?

PR #5585, opened 2020-06-05
fix(dgraph): Fix snapshot calculation in ludicrous mode

ashish-goswami:
In `ludicrous mode` snapshot calculation is not happening correctly. This is resulting in no snapshot
being taken ever and hence `w` directory size is increasing with time.
For snapshot creation we need last applied raft index(such that all raft index entries before this index are also applied) and max timestamp out of all committed transactions. Committed
transactions are sent by Zero to all alpha leaders via subscription. In ludicrous mode we don't send
transaction to Zero while committing. Hence Zero doesn't have anything to send as part of Oracle
updates. Because of this, possible snapshot timestamp(maxCommitTs) in `calculateSnapshot()` is always 0. Hence no snapshot is ever taken.

This PR addresses two issues:
1. Correction while marking raft index done. Now any raft index will only be marked as done once
we have applied mutations for proposal(raft entry).
2. For ludicrous mode, use `StartTs` of any transaction for calculating `maxCommitTs` snapshot in
    `calculateSnapshot()` 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5585)

 
 

Docs Preview: [Dgraph Preview](https://dgraph-53dfc4a61e-77361.surge.sh)


worker/draft.go
```diff
@@ -1470,13 +1463,17 @@ func (n *node) calculateSnapshot(startIdx uint64, discardN int) (*pb.Snapshot, e
...
 			if proposal.Mutations != nil {
-				start := proposal.Mutations.StartTs
+				start = proposal.Mutations.StartTs
 				if start >= minPendingStart && snapshotIdx == 0 {
 					snapshotIdx = entry.Index - 1
 				}
 			}
-			if proposal.Delta != nil {
+			if x.WorkerConfig.LudicrousMode {
+				maxCommitTs = x.Max(maxCommitTs, start)
```
parasssh, line 1475: Add comment "StartTs and CommitTs is same for ludicrous mode"

PR #5599, opened 2020-06-08
Rollup: Reduce memory consumption of the map

jarifibrahim:
FIXES - DGRAPH-1613

The incremental roll-up uses a `map` to keep track of the keys that should be rolled up.
The current implementation never clears the map which stores the keys. The existing algorithm is
```
If key is present or key was rolled up 10 seconds ago, perform roll up
```
we never clean up the map in the existing algorithm and as a result the map consumed a lot of memory
```
3.77GB 32.51% 32.51%     3.93GB 33.95%  github.com/dgraph-io/dgraph/posting.(*incrRollupi).Process
```
```
(pprof) list Process
Total: 11.59GB
ROUTINE ======================== github.com/dgraph-io/dgraph/posting.(*incrRollupi).Process in src/github.com/dgraph-io/dgraph/posting/mvcc.go
    3.77GB     3.93GB (flat, cum) 33.95% of Total
         .          .    118:			for _, key := range *batch {
         .          .    119:				hash := z.MemHash(key)
         .          .    120:				if elem := m[hash]; currTs-elem >= 10 {
         .          .    121:					// Key not present or Key present but last roll up was more than 10 sec ago.
         .          .    122:					// Add/Update map and rollup.
    3.77GB     3.77GB    123:					m[hash] = currTs
         .   170.89MB    124:					if err := ir.rollUpKey(writer, key); err != nil {
         .          .    125:						glog.Warningf("Error %v rolling up key %v\n", err, key)
         .          .    126:					}
         .          .    127:				}
         .          .    128:			}
         .          .    129:			// clear the batch and put it back in Sync keysPool

```

The proposed change will clean up the keys every 2 minutes and remove the keys that have not changed in the last minute.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5599)

### DIFF
--- a/posting/mvcc.go
+++ b/posting/mvcc.go
@@ -110,7 +110,7 @@ func (ir *incrRollupi) Process(closer *y.Closer) {
 	m := make(map[uint64]int64) // map hash(key) to ts. hash(key) to limit the size of the map.
 	limiter := time.NewTicker(100 * time.Millisecond)
 	defer limiter.Stop()
-	cleanupTick := time.NewTicker(2 * time.Minute)
+	cleanupTick := time.NewTicker(5 * time.Minute)
 	defer cleanupTick.Stop()
 
 	for {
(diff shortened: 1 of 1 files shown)


posting/mvcc.go
```diff
@@ -109,10 +109,23 @@ func (ir *incrRollupi) Process(closer *y.Closer) {
...
 	for {
 		select {
 		case <-closer.HasBeenClosed():
 			return
+		case <-cleanupTick.C:
+			currTs := time.Now().UnixNano()
+			for hash, ts := range m {
+				// Remove entries from map which have been there for more
+				// than a two minutes.
+				if currTs-ts >= int64(2*time.Minute) {
```
jarifibrahim, line 125: This should be `10 seconds`.

PR #5607, opened 2020-06-09
feat(graphql+-): Allow parameterized @cascade directive.

parasssh:
Fixes DGRAPH-1279

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5607)

### DIFF
--- a/gql/parser.go
+++ b/gql/parser.go
@@ -2114,7 +2114,8 @@ func parseCascade(it *lex.ItemIterator, gq *GraphQuery) error {
 	// 4. @cascade\n someOtherPred
 	if items[0].Typ == itemLeftCurl || items[0].Typ == itemRightCurl || items[0].
 		Typ == itemAt || items[0].Typ == itemName {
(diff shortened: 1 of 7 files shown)


amaster507:
I am gonna dump this comment over here to:

I am for this as well, because we may want to require only some fields or sub-graphs but not all the way down.

FWIW, I think that the @cascade params should specifically include the sub-graphs to include as well.
```
{
  q(func: anyoftext(name, "Alice")) @cascade(name, friend) { # parameterized cascade
    name
    age
    children {   # not implied `@cascade`
      name
      age
      friend {   # not implied `@cascade`
        name
        age
      }
    }
    friend {     # `@cascade` by param
      name
      age
      friend  {  # implied `@cascade`
      	 name
      	 age 
      }
    }
  }
}
```
This would allow a query to require friends but not require children but return children as well if they exist without a second query.

---

This may already be applied somewhere and I just don't know it. Still getting my feet wet.

PR #5609, opened 2020-06-09
Added support for query and mutation

poonai:
NOTE : SKIPPING RESTORE TEST. WILL FIX IT WHILE DOING RESTORE AND BACKUP
This PR adds multi-tenancy support for query and mutation. And do the following to achieve that.
All predicates are prefixed by namespace to query and mutate their respective namespace to avoid conflicts. 

For querying - we prefix the attr when task query is constructed to avoid changes in parser and output json

For mutation - we prefix when rdf are converted into nquads

 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5609)

 
 
 

Docs Preview: [Dgraph Preview](https://dgraph-4c2948fc36-72241.surge.sh)


rstorr:
@codelingobot capture Intermediate vars declared then immediately assigned to a struct field can be collapased to a single line

rstorr:
@CodeLingoBot capture Intermediate vars declared then immediately assigned to a struct field can be collapased to a single line

PR #5185, opened 2020-04-13
[BREAKING] Namespace dgraph internal types/predicates with `dgraph.`

animesh2049:
Dgraph's internal types `User`, `Group` and `Rule` were too generic. Prefixed them with `dgraph.type` so that there is no chance of collision with user-defined types.

Also, reserved `dgraph.` as the namespace for dgraph's internal types/predicates so that no user could create any such type/predicate.

Fixes #4878.
Fixes #DGRAPH-1090.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5185)

 
 
 
 
 
 
 
 

Docs Preview: [Dgraph Preview](https://dgraph-d14923aaea-68875.surge.sh)

PR #5016, opened 2020-03-24
Add tool to upgrade ACLs

mangalaman93:
**TODO**
  * [x] Add docs
  * [x] Test the tool

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5016)

 
 
 

Docs Preview: [Dgraph Preview](https://dgraph-2d67a643bf-51338.surge.sh)

PR #5634, opened 2020-06-11
feat(Upgrade): Users can now upgrade to v20.07.0

abhimanyusinghgaur:
This PR adds the capability to upgrade to v20.07.0 from v20.03.0+
It has only considered the breaking changes introduced in #5185.

It also introduces a generic system to handle upgrades.

At present, upgrade tool expects to handle only breaking ACL changes, and does not expect to handle badger data format changes. The way it is supposed to work is by updating the Dgraph binary to new version, start Dgraph cluster with existing data directories (which could also have been restored by enterprise backup/restore), and then running the upgrade tool by specifying the version you want to upgrade from and the version you want to upgrade to.

Fixes #DGRAPH-1617.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5634)

 
 

Docs Preview: [Dgraph Preview](https://dgraph-ce37b2c34c-74143.surge.sh)


abhimanyusinghgaur:
linking #5016 for context.

Issue #5601, opened 2020-06-08
Alpha server start issue with ludicrous mode


ashish-goswami:
### What version of Dgraph are you using?
master(https://github.com/dgraph-io/dgraph/commit/222b35a1ec04bec31bf9eced7ff46cf4fe13587e)

### Steps to reproduce the issue (command/config used to run Dgraph).
1. Start zero server in ludicrous mode:
```
dgraph zero --ludicrous_mode
```
2. Start alpha server in ludicrous mode:
```
dgraph alpha --ludicrous_mode
```
Alpha node continuously prints below logs:
```
[Decoder]: Using assembly version of decoder
[Sentry] 2020/06/08 21:23:28 Integration installed: ContextifyFrames
[Sentry] 2020/06/08 21:23:28 Integration installed: Environment
[Sentry] 2020/06/08 21:23:28 Integration installed: Modules
[Sentry] 2020/06/08 21:23:28 Integration installed: IgnoreErrors
[Decoder]: Using assembly version of decoder
[Sentry] 2020/06/08 21:23:28 Integration installed: ContextifyFrames
[Sentry] 2020/06/08 21:23:28 Integration installed: Environment
[Sentry] 2020/06/08 21:23:28 Integration installed: Modules
[Sentry] 2020/06/08 21:23:28 Integration installed: IgnoreErrors
I0608 21:23:28.288587   20848 config.go:100] LRU memory (--lru_mb) set to 3935.75MB, 25% of the total RAM found (15743MB)
For more information on --lru_mb please read https://dgraph.io/docs/deploy/#config
I0608 21:23:28.288946   20848 util_ee.go:126] KeyReader instantiated of type <nil>
I0608 21:23:28.430722   20848 init.go:99] 

Dgraph version   : v2.0.0-rc1-392-g222b35a1e
Dgraph SHA-256   : 2da8881ffb4fc3ada6f840d7101a1437bad2511215a4b5e6462500dd442be0f7
Commit SHA-1     : 222b35a1e
Commit timestamp : 2020-06-05 17:39:49 -0700
Branch           : master
Go version       : go1.14.1

For Dgraph official documentation, visit https://dgraph.io/docs/.
For discussions about Dgraph     , visit https://discuss.dgraph.io.
To say hi to the community       , visit https://dgraph.slack.com.

Licensed variously under the Apache Public License 2.0 and Dgraph Community License.
Copyright 2015-2020 Dgraph Labs, Inc.


[...]

PR #5642, opened 2020-06-12
Fix alpha start in ludicrous mode

ashish-goswami:
Fixes: https://github.com/dgraph-io/dgraph/issues/5601
Fixes: DGRAPH-1669

Currently we are unable to start alpha server in ludicrous mode, because of reasons explained
below.

Things we do at alpha start:
1. At start we insert `dgraph.type`, `dgraph.graphql.schema` and `dgraph.graphql.xid` predicates as
    our initial schema.
2. We also insert `dgraph.graphql` type(having predicates `dgraph.graphql.schema` and
   `dgraph.graphql.xid`) as initial types.
3. We also upsert data related to graphql schema(which results in upserts for predicates 
   `dgraph.graphql.schema` and `dgraph.graphql.xid`).

While inserting types we verify that predicates used in the types should either already be present
in schema or should be part of same mutation where we are trying to insert this type.

In ludicrous mode we mark any proposal done without error as soon as it is retrieved as part of commited entries, as opposed to after applying proposal (this is done in normal mode).

Events which led to alpha getting stuck:
1. We propose schema mutation for `dgraph.type`, which is marked as done immediately. This
    proposal is still getting applied.
2. Since we marked above proposal done without error, we propose next schema mutation for
    `dgraph.graphql.schema`, which again is marked done without error. But applying of this proposal
    fails as indexing was already in progress for `dgraph.type`. But we never retry this schema
    mutation.
3. Next we propose `dgraph.graphql.xid`, which is applied successfully(indexing is done until now 
    for `dgraph.type`).
4. When we insert `dgraph.graphql` type, `verifyTypes()` fails, because it never finds
   `dgraphq.graphql.schema` predicate.

**Proposed solution:**
Assumption that proposal is never failed while applying, hence it should be marked as done without
error immediately, should hold only for data mutations and not schema mutations.

This PR checks if proposal is data mutation(having edges > 0), while marking it done
immediately. This leads to retrying of schema proposal if they fail once because indexing is already
in progress.

Note: We didn't observe this before commit(https://github.com/dgraph-io/dgraph/commit/aef7072bfcd6e47dccc4d8cc39d3e3d6166e8e08), because predicate `dgraph.graphql.schema` was getting inserted via graphql upserts(#3 of start steps). Now after
commit(https://github.com/dgraph-io/dgraph/commit/aef7072bfcd6e47dccc4d8cc39d3e3d6166e8e08) we cannot insert data for predicates starting with `dgraph.` prefix unless those are already present.

 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5642)

### DIFF
--- a/worker/draft.go
+++ b/worker/draft.go
@@ -1157,8 +1157,7 @@ func (n *node) Run() {
 						if span := otrace.FromContext(pctx.Ctx); span != nil {
 							span.Annotate(nil, "Proposal found in CommittedEntries")
 						}
-						if x.WorkerConfig.LudicrousMode &&
-							proposal.Mutations != nil && len(proposal.Mutations.Edges) > 0 {
+						if x.WorkerConfig.LudicrousMode && len(proposal.Mutations.GetEdges()) > 0 {
 							// Assuming that there will be no error while applying. But this
 							// assumption is only made for data mutations and not schema mutations.
 							n.Proposals.Done(proposal.Key, nil)
--- a/worker/draft.go
+++ b/worker/draft.go
@@ -1157,8 +1157,10 @@ func (n *node) Run() {
 						if span := otrace.FromContext(pctx.Ctx); span != nil {
 							span.Annotate(nil, "Proposal found in CommittedEntries")
 						}
-						if x.WorkerConfig.LudicrousMode {
-							// Assuming that there will be no error while proposing.
+						if x.WorkerConfig.LudicrousMode &&
+							proposal.Mutations != nil && len(proposal.Mutations.Edges) > 0 {
+							// Assuming that there will be no error while applying. But this
+							// assumption is only made for data mutations and not schema mutations.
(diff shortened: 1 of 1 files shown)


worker/draft.go
```diff
@@ -1157,8 +1157,9 @@ func (n *node) Run() {
 						if span := otrace.FromContext(pctx.Ctx); span != nil {
 							span.Annotate(nil, "Proposal found in CommittedEntries")
 						}
-						if x.WorkerConfig.LudicrousMode {
-							// Assuming that there will be no error while proposing.
+						if x.WorkerConfig.LudicrousMode && len(proposal.Mutations.GetEdges()) > 0 {
```
vvbalaji-dgraph, line 1160: Skipping ludicrous mode for schema mutations seems to be the right approach. If 'Getedges' is the right mechanism for identifying schema changes then it would be good to have a comment to describe our intent here.

Issue #5408, opened 2020-05-09
something wrong with tablet size calculation


Willem520:
### What version of Dgraph are you using?
Dgraph v20.03.1
### Have you tried reproducing the issue with the latest release?
yes
### What is the hardware spec (RAM, OS)?
centos(126G)
### Steps to reproduce the issue (command/config used to run Dgraph).
I noticed the tablet size is more than disk capacity
my machine disk capacity is
![1](https://user-images.githubusercontent.com/26085817/81467261-35503a80-920a-11ea-98f8-8d030823cccd.png)

the p directory size is
![2](https://user-images.githubusercontent.com/26085817/81467275-4b5dfb00-920a-11ea-9bd9-67dac8d85a33.png)

when I used /state endpoint, I got the result
![3](https://user-images.githubusercontent.com/26085817/81467292-62045200-920a-11ea-92ce-c58768f33810.png)

from the ratel, I got the result
![4](https://user-images.githubusercontent.com/26085817/81467299-67619c80-920a-11ea-9948-b2c1979ed353.png)

md5id and gid size is about 6.1TB,but my disk capacity is 2.9T
### Expected behaviour and actual result.
tablet size can calculate correctly.

Related to https://discuss.dgraph.io/t/ratel-predicate-capactiy

martinmr:
I haven't been able to reproduce this exact issue but it looks like something else is wrong. When calculating the sizes, the function skips all the tables with the following errors.

```
alpha1    | I0611 23:25:56.138946      14 draft.go:1245] Calculating tablet sizes. Found 4 tables
alpha1    | I0611 23:25:56.139087      14 draft.go:1254] Unable to parse key: Invalid size 25185 for key [33 98 97 100 103 101 114 33 104 101 97 100 255 255 255 255 255 255 255 254]
alpha1    | I0611 23:25:56.139145      14 draft.go:1254] Unable to parse key: Invalid size 25185 for key [33 98 97 100 103 101 114 33 104 101 97 100 255 255 255 255 255 255 255 254]
alpha1    | I0611 23:25:56.139560      14 draft.go:1254] Unable to parse key: Invalid size 25185 for key [33 98 97 100 103 101 114 33 104 101 97 100 255 255 255 255 255 255 255 254]
[...]

PR #5656, opened 2020-06-15
Fix: Change tablet size calculation to not depend on the right key.

martinmr:
In badger, the right key might be a badger specific key that Dgraph
cannot understand. To deal with these keys, a table is included in
the size calculation if the next table starts with the same key.

Related to DGRAPH-1358 and #5408.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5656)

### DIFF
--- a/worker/draft.go
+++ b/worker/draft.go
@@ -1240,8 +1240,27 @@ func (n *node) calculateTabletSizes() {
 	}
 	var total int64
 	tablets := make(map[string]*pb.Tablet)
+	updateSize := func(pred string, size int64) {
+		if pred == "" {
+			return
+		}
+
+		glog.Infof("Adding %v to table %v", size, pred)
+		if tablet, ok := tablets[pred]; ok {
+			tablet.Space += size
+		} else {
+			tablets[pred] = &pb.Tablet{
+				GroupId:   n.gid,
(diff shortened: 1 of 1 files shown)


jarifibrahim:
@parasssh @martinmr The `tableInfo.Left` has a higher chance of being an internal key. All the level 0 tables have the `head` as the leftmost key. So we would have to skip these tables anyway.

For tables other than level 0, they wouldn't usually have an internal key as the right-most key. So what do we gain by this change? What is the benefit of checking for the `left key of the next table` instead of the `right key of the current table`?

An important thing to remember is that all the versions of a single key will be stored in a single file. So if you have 100,000 versions of `foo`. They all will be stored in the same sst file.

parasssh:
> @parasssh @martinmr The `tableInfo.Left` has a higher chance of being an internal key then the `Right` key. All the level 0 tables have the `head` as the leftmost key. So we would have to skip these tables anyway.
> 

Hmm. I was under the impression that the badger internal keys are _only_ encountered towards the `Right` end of the table. 

So, I guess, in that case, `TableInfo`'s `Left` and `Right` should point to valid keys, not internal keys to make Table Size calculation work.

PR #5677, opened 2020-06-17
feat(GraphQl): Alpha now has flag for graphql extensions 

JatinDevDG:
This PR add "graphql extension" flag to alpha which will be used for turning extensions field on/off in graphql response.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5677)

### DIFF
--- a/graphql/schema/response.go
+++ b/graphql/schema/response.go
@@ -126,6 +126,7 @@ func (r *Response) WriteTo(w io.Writer) (int64, error) {
 
 // Output returns json interface of the response
 func (r *Response) Output() interface{} {
+
 	if r == nil {
 		return struct {
 			Errors json.RawMessage `json:"errors,omitempty"`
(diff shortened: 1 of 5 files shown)


graphql/schema/response.go
```diff
@@ -134,15 +134,24 @@ func (r *Response) Output() interface{} {
...
-	return struct {
-		Errors     []*x.GqlError   `json:"errors,omitempty"`
-		Data       json.RawMessage `json:"data,omitempty"`
-		Extensions *Extensions     `json:"extensions,omitempty"`
-	}{
-		Errors:     r.Errors,
-		Data:       r.Data.Bytes(),
-		Extensions: r.Extensions,
+	} else if x.Config.GraphqlExtension {
+		return struct {
```
parasssh, line 138: You can simplify this.

You can have the struct with all 3 fields. 

And if flag is turned on, set the Extensions. Otherwise, it will be nil and will be omitted (because of the "omitempty" against the field) anyways in the JSON response.

PR #5704, opened 2020-06-22
feat(Restore): Async restore operations.

martinmr:
Previously, restore operations blocked the HTTP response until they were completed.
This PR changes the flow so that the request returns immediately after the restore
is proposed to the cluster.

The request returns a unique ID for the restore operation that can be used to track the
status of the response.

Fixes DGRAPH-1697

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5704)

 
 
 
 

Docs Preview: [Dgraph Preview](https://dgraph-35ff497feb-74524.surge.sh)


parasssh:
Before we review, isn't backup also blocking which could also potentially take several minutes? Maybe we are ok with blocking behavior on restore as well?

PR #5724, opened 2020-06-24
test(restore): Fix flakiness in online restore tests.

martinmr:
The tests are flaky because the test waits for the leaders to complete the request. However,
the followers might be behind and the client can be connected to a follower. The fix is
to check that the client has exited draining mode before doing any queries.

Fixes DGRAPH-1740

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5724)

### DIFF
--- a/systest/online-restore/online_restore_test.go
+++ b/systest/online-restore/online_restore_test.go
@@ -26,6 +26,7 @@ import (
 	"runtime"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/dgraph-io/dgo/v200"
 	"github.com/dgraph-io/dgo/v200/protos/api"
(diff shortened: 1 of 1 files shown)


martinmr:
Tested this by running the test multiple times locally. I didn't see a single failure after the change.

ashish-goswami:
@martinmr `TestMoveTablets` is still failing in CI for same reason.

martinmr:
I moved the code around again and ran the test locally about 30 times in my local machine without any issues.

martinmr:
Ok. Third time is the charm. I think I got the right approach now and I saw a test pass on my local machine that would otherwise had failed without the check.

systest/online-restore/online_restore_test.go
```diff
@@ -59,6 +60,21 @@ func sendRestoreRequest(t *testing.T, backupId string) {
 	buf, err := ioutil.ReadAll(resp.Body)
 	require.NoError(t, err)
 	require.Contains(t, string(buf), "Restore completed.")
+
+	// Wait for the client to exit draining mode. This is needed because the client might
+	// be connected to a follower and might be behind the leader in applying the restore.
+	for i := 0; i < 10; i++ {
```
parasssh, line 66: Is 10 retries = 2.5 sec enough?

PR #5731, opened 2020-06-25
release/v20.07: Rollup: Reduce memory consumption of the map (#5599)

jarifibrahim:
FIXES - DGRAPH-1613

The incremental roll-up uses a `map` to keep track of the keys that should be
rolled up. The current implementation never clears the map which stores the
keys. The existing algorithm is
```
If key is present or key was rolled up 10 seconds ago, perform roll up
```
we never clean up the map in the existing algorithm and as a result the map
consumed a lot of memory 
```
3.77GB 32.51% 32.51%     3.93GB 33.95%  github.com/dgraph-io/dgraph/posting.(*incrRollupi).Process
```

The proposed change will clean up the keys every 2 minutes and remove the keys
that have not changed in the last 10 seconds.

(cherry picked from commit b77c23c665f6ae35aba5b1b54a0700d51c0cb8d0)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5731)

 

Docs Preview: [Dgraph Preview](https://dgraph-53de9bb7c6-72961.surge.sh)

### DIFF
--- a/posting/mvcc.go
+++ b/posting/mvcc.go
@@ -109,10 +109,22 @@ func (ir *incrRollupi) Process(closer *y.Closer) {
 
 	m := make(map[uint64]int64) // map hash(key) to ts. hash(key) to limit the size of the map.
 	limiter := time.NewTicker(100 * time.Millisecond)
+	defer limiter.Stop()
+	cleanupTick := time.NewTicker(5 * time.Minute)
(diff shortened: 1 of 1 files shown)


jarifibrahim:
Closing this PR since I cannot trigger the build. I'll send a new PR.

PR #5750, opened 2020-06-27
fix(GraphQL): Upgrade jwt-go to v4 to support multiple aud claims

### LINKED ISSUE #5740: GraphQL: cannot unmarshal array into Go struct field StandardClaims.aud of type string
JWT aud is an array, GraphQL endpoint is expecting a string it seems:

```
{
  "https://www.example.com/jwt/claims": {
    "USER": "github|XXXXX",
    "ROLE": "ADMIN"
  },
  "iss": "https://example.us.auth0.com/",
  "sub": "github| XXXXX",
  "aud": [
    "https://www.example.com",
    "https://example.us.auth0.com/userinfo"
  ],
  "iat": 1593154207,
  "exp": 1593240607,
  "azp": "X",
  "scope": "openid profile email",
  "permissions": []
}
```

dpeek:
Also removes forced exp check as claims.Valid performs this check, and exp is an optional claim according to spec.

If original check was intentional, could be updated to return error if `claims[exp] == nil`

Fixes #5740

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5750)

### DIFF
--- a/edgraph/access_ee.go
+++ b/edgraph/access_ee.go
@@ -29,7 +29,7 @@ import (
 	"github.com/dgraph-io/dgraph/schema"
 	"github.com/dgraph-io/dgraph/worker"
 	"github.com/dgraph-io/dgraph/x"
-	jwt "github.com/dgrijalva/jwt-go"
+	jwt "github.com/dgrijalva/jwt-go/v4"
 	"github.com/golang/glog"
 	otrace "go.opencensus.io/trace"
 	"google.golang.org/grpc/codes"
@@ -171,13 +171,6 @@ func validateToken(jwtStr string) ([]string, error) {
 	claims, ok := token.Claims.(jwt.MapClaims)
 	if !ok || !token.Valid {
 		return nil, errors.Errorf("claims in jwt token is not map claims")
-	}
-
-	// by default, the MapClaims.Valid will return true if the exp field is not set
-	// here we enforce the checking to make sure that the refresh token has not expired
-	now := time.Now().Unix()
-	if !claims.VerifyExpiresAt(now, true) {
-		return nil, errors.Errorf("Token is expired") // the same error msg that's used inside jwt-go
 	}
 
 	userId, ok := claims["userid"].(string)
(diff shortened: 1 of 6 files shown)


arijitAD:
@dpeek Can you run this test `TestAuthQueryRewriting` and check if it's passing?

pawanrawal:
Seems like there is a compilation error in `testutil/graphql.go` on line 160. Could you please fix it @dpeek?

```
cannot use time.Now().Add(time.Minute).Unix() (value of type int64) as *v4.Time value in struct literal
testutil/graphql.go

157		a.Namespace,
158		a.AuthVars,
159		jwt.StandardClaims{
160			ExpiresAt: time.Now().Add(time.Minute).Unix(),
161			Issuer:    "test",
162		},
163	}
```

dpeek:
So, I fixed the compilation error but I seem to have introduced a bunch of failures in other test suites – looks like tokens are also checked for ACL purposes? Couldn't find where the tokens for those tests are generated yet... will keep digging tonight unless you can offer some clues :)

arijitAD:
Yes, JWT is also used for ACL. After a user is authenticated in Dgraph, it sends a JWT. You can look at getAccessJwt() in `edgraph/access_ee.go`

pawanrawal:
Hey @dpeek, I had a deeper look at this and the failing tests. One of the failing test seems to https://github.com/dgraph-io/dgraph/blob/master/graphql/resolve/auth_test.go#L156. It fails because even though the specified JWT has an `aud`, we don't provide the right `Audience` while validating the JWT. According to the [spec](https://tools.ietf.org/html/rfc7519#section-4.1.3) if the JWT has an `aud`, then the principal processing the claim must identify itself against that. This means we need to have a way by which we can accept the `audience` from the user when the user updates the schema. Does that sound like what you are looking for?

arijitAD:
Thanks for your contribution. I have modified on top of your PR and merged it to master and 20.07 branch. #5927

PR #5757, opened 2020-06-29
docs(Docker): Reword docker compose to show the need to change volume mount directories.

amaster507:
Change wording to show the need to change the volume to something besides "/tmp/*"

Tested on AWS. If the mount directory of the volume is not changed away from tmp then the data will not persists even though the volume is persistent with EBS.

Figured this out by reading this topic on discourse: https://discuss.dgraph.io/t/dgraph-in-docker-unable-to-reload-data-on-restart/5927/4?u=amaster507

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5757)

### DIFF
--- a/contrib/config/docker/docker-compose.yml
+++ b/contrib/config/docker/docker-compose.yml
@@ -2,7 +2,7 @@
 # and Alpha in different Docker containers.
 
 # It mounts /tmp/data on the host machine to /dgraph within the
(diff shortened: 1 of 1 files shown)


joshua-goldstein (approved):
Looks good.  Thanks for making this more clear.

dshekhar95 (approved):
Looks good, good clarification

PR #5761, opened 2020-06-29
docs(Deploy): Add "poor man's ACL" wording in auth token docs.

danielmai:
Adding the "poor man's ACL" to the description of the `--auth_token` flag as per this Discuss post: https://discuss.dgraph.io/t/enterprise-features-dgraph-core-values-open-source-and-my-thoughts/7399/12.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5761)

 

Docs Preview: [Dgraph Preview](https://dgraph-748330f7b7-73848.surge.sh)


bucanero:
do we still want to merge this PR?

on a side note, I'm not a big fan of using the "poor man's" tag, as it gives the idea that our product is not for Enterprise-level applications. ( @aaroncarey )

aaroncarey:
> do we still want to merge this PR?
> 
> on a side note, I'm not a big fan of using the "poor man's" tag, as it gives the idea that our product is not for Enterprise-level applications. ( @aaroncarey )

I heartily agree! If we can reframe this as "simple auth (here and elsewhere)", that will:
1) Translate much better into other languages
2) Avoid creating the impression that Dgraph is software for hobbyists. Our competitors don't use this kind of term.
3) Avoid creating a negative impression from those who might come from a low-income background.

All of which is to say: we should change this term, here and elsewhere. We'd love to know who we need to convince to make this change a reality.

danielmai:
@bucanero @aaroncarey This PR is still valid. As per the original Discuss post, the goal is to make this feature easier to find via the search box. If it needs to be "Simple Auth" or "Simple ACL", then OK. The goal is for it to be discoverable. "Securing Alter Operations" has not been good enough.

bucanero:
> @bucanero @aaroncarey This PR is still valid. As per the original Discuss post, the goal is to make this feature easier to find via the search box. If it needs to be "Simple Auth" or "Simple ACL", then OK. The goal is for it to be discoverable. "Securing Alter Operations" has not been good enough.

Great, we're going to use `Simple ACL`. I'll commit the change and merge the PR.

PR #5788, opened 2020-07-01
fix(LiveLoader): correct the file name for syntactically wrong file in live loader logs

### LINKED ISSUE #5779: Dgraph Live reports the wrong input filename on parsing error
v20.03.3

Yes

Unrelated

Start a fresh temporary dgraph instance with docker:
```
docker run -it --rm dgraph/standalone:v20.03.3 /bin/bash
```

Inside, run the following commands:
```
dgraph zero > /dev/null 2>&1 < /dev/null &
dgraph alpha --lru_mb 1024 > /dev/null 2>&1 < /dev/null &
echo "<a> <p> <c> ." > a.rdf
echo "<b> <p> \"abc\"^^<http://www.w3.org/2001/XMLSchema#int>> ." > b.rdf
```
Wait a bit until dgraph alpha is up, then
```
dgraph live --files a.rdf,b.rdf
```

It fails correctly on the value of the second file with:
```
Found 2 data file(s) to process
Processing data file "a.rdf"
Processing data file "b.rdf"
Error while processing data file "a.rdf": During parsing chunk in processLoadFile: while parsing line "<b> <p> \"abc\"^^<http://www.w3.org/2001/XMLSchema#int>> .\n": while lexing <b> <p> "abc"^^<http://www.w3.org/2001/XMLSchema#int>> . at line 1 column 53: Invalid input: > at lexText
```

But it reports the wrong file as the source of the error:
```
Error while processing data file "a.rdf"
```
This make hunting the bad data extremely difficult if you are loading hundreds of files.

Report the correct file name, here `"b.rdf"`.

NamanJain8:
1. The error message for the unparsable input file for the live loader was printing incorrect filename corresponding to the error. Added a small `ferr` structure which contains the fields `error message` and `filename`.
2. Fixes #5779 
3. Fixes DGRAPH-5779

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5788)

 
 
 
 
 

Docs Preview: [Dgraph Preview](https://dgraph-95f85738b9-78510.surge.sh)


NamanJain8:
> It might be a json file also. Just check the whole error line. Find the line which contains the error, and then compare the whole line for the test. No need to split the file names.

Having a complete line would make it easy to check for errors in any type of file and it would be easy to understand if one later modifies these tests to just compare the whole error.
Comparing the whole line seems to be like adding more dependency on the format of the error message as these tests would have to be updated if the format of the error message changes. I tried to have minimal maintenance issues by including only some part of the line. What would be better? Adding regex for (ref | json) or complete line?

dgraph/cmd/live/run.go
```diff
@@ -479,11 +484,11 @@ func run() error {
 	fmt.Printf("Found %d data file(s) to process\n", totalFiles)
 
 	//	x.Check(dgraphClient.NewSyncMarks(filesList))
-	errCh := make(chan error, totalFiles)
+	errCh := make(chan fileError, totalFiles)
 	for _, file := range filesList {
 		file = strings.Trim(file, " \t")
 		go func(file string) {
-			errCh <- l.processFile(ctx, file, opt.key)
+			errCh <- fileError{file, l.processFile(ctx, file, opt.key)}
```
jarifibrahim, line 491:
```suggestion
			errCh <- errors.Wrapf(l.processFile(ctx, file, opt.key), "file: %s", file)
```

NamanJain8: Done.

NamanJain8:
> Add a message, not just the file name (see ibrahim's comment)

@martinmr, "Error while processing data file" is already there in error printing (see [here](Error while processing data file)). So I skipped that in Wrapf.

PR #5802, opened 2020-07-02
feat(GraphQL): GraphQL health now reported by /probe/graphql

abhimanyusinghgaur:
Fixes #GRAPHQL-507.

This PR adds `/probe/graphql` HTTP endpoint which returns HTTP status 200 OK, if the `/graphql` endpoint is up and running otherwise it returns HTTP status 503 Service Unavailable. It also returns a JSON body of the following form:
```
{"status":"init/up/updating schema"}
```
Status `init` is returned for 503, while rest of the two statuses return 200 code.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5802)

 

Docs Preview: [Dgraph Preview](https://dgraph-ee9f26feeb-75202.surge.sh)

### DIFF
--- a/graphql/admin/admin.go
+++ b/graphql/admin/admin.go
@@ -333,21 +333,22 @@ var (
 		"deleteGroup": {resolve.IpWhitelistingMW4Mutation},
 	}
 	// mainHealth tracks the health of the main GraphQL server.
-	// It is required for kubernetes probing.
-	mainHealth = &graphQLHealth{HttpStatusCode: http.StatusServiceUnavailable, StatusMsg: "init"}
+	mainHealth = &GraphQLHealth{HttpStatusCode: http.StatusServiceUnavailable, StatusMsg: "init"}
 )
 
-type graphQLHealth struct {
+// GraphQLHealth is used to report the health status of a GraphQL server.
+// It is required for kubernetes probing.
+type GraphQLHealth struct {
 	HttpStatusCode int
 	StatusMsg      string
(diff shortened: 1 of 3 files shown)


graphql/admin/admin.go
```diff
@@ -326,8 +325,40 @@ var (
...
 )
 
+// GraphQLHealth is used to report the health status of a GraphQL server.
+// It is required for kubernetes probing.
+type GraphQLHealth struct {
+	httpStatusCode int
+	statusMsg      string
+	// mux protects GraphQLHealth from simultaneous read/write, as an instance of this type could be
+	// used by multiple Go-routines
+	mux sync.RWMutex
```
gja, line 339: can't you replace all of this with sync.Value?

graphql/admin/admin.go
```diff
@@ -326,8 +325,40 @@ var (
...
+func (g *GraphQLHealth) Status() (int, string) {
+	g.mux.RLock()
+	defer g.mux.RUnlock()
+	return g.httpStatusCode, g.statusMsg
+}
+
+func (g *GraphQLHealth) up() {
+	g.mux.Lock()
+	defer g.mux.Unlock()
+	g.httpStatusCode = http.StatusOK
```
gja, line 351: I feel like this is the wrong layer for this. The HTTP status code belongs in the http layer. Probably just store an enum or whatever at this layer

PR #5804, opened 2020-07-02
test: Deal with incomplete tests in Jepsen tool

martinmr:
This PR does two things:

1. Add an option `--refresh-cluster` to down and up the test cluster before each test. I noticed
   that most errors that lead to incomplete tests go away if this is done.
1. Retry incomplete tests. Most incomplete tests fail early so it's not very expensive to retry
    the test. The tool now only returns an exit code of 1 if there's a test failure or a test was incomplete
    even after retries.

Related to DGRAPH-971

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5804)

 
 

Docs Preview: [Dgraph Preview](https://dgraph-c06910c0db-76986.surge.sh)

### DIFF
--- a/contrib/jepsen/main.go
+++ b/contrib/jepsen/main.go
@@ -153,6 +153,10 @@ var (
 		"Don't run any more tests after a failure.")
 )
 
+const (
+	maxRetries = 5
+)
+
 func command(cmd ...string) *exec.Cmd {
 	return commandContext(ctxb, cmd...)
 }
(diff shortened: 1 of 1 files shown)


contrib/jepsen/main.go
```diff
@@ -400,50 +408,80 @@ func main() {
...
 					if *exitOnFailure {
 						os.Exit(1)
 					}
 					defer os.Exit(1)
+					break retryLoop
+				case errTestIncomplete:
+					// Retry incomplete tests. Sometimes tests fail due to temporary errors.
+					tries++
+					if tries == maxRetries {
+						defer os.Exit(1)
```
parasssh, line 480: maybe a log here before exiting.

Issue #5687, opened 2020-06-18
ACL read-only user permissions reject queries using query functions and variables


danielmai:
### What version of Dgraph are you using?
v20.03.3

### Have you tried reproducing the issue with the latest release?
Yes

### What is the hardware spec (RAM, OS)?
Ubuntu Linux (Docker)

### Steps to reproduce the issue (command/config used to run Dgraph).

1. Run Dgraph cluster with ACLs enabled.
2. Log in as groot and set the following mutation:
   ```
   {"name": "Dgraph", "dgraph.type": "TypeName"}
   ```
3. Create a type called "TypeName` that has the `name` predicate.
2. Create an ACL group with read-only permissions to predicates.
3. Create a user and add it to the read-only group created in #2 (i.e., it's not a guardian user).

##### `type` function

Query: Run a query using the `type()` function:
```
{ 
  q(func: type(TypeName))
}
```
Alpha logs show access denied for predicates `""`:
```
I0618 22:30:58.799588      16 access_ee.go:727] ACL-LOG Authorizing user "rouser" with groups "rogroup" on predicates "" for "Read", allowed:false
```

##### `expand` function

Query: Run a query using `expand(_all_)`:
```
{ 
  q(func: has(name)) {
    expand(_all_)
  }
}
```
Alpha logs show access denied for predicates `",expand"`:
```
I0618 22:31:20.622734      16 access_ee.go:727] ACL-LOG Authorizing user "rouser" with groups "rogroup" on predicates ",expand" for "Read", allowed:false
```
##### `val` function and variables

Query: Run a query using `val(v)` and see an access denied log in Alpha:
```
{
  q1(func: has(name)) {
    v as name
  }
  
  q2(func: eq(val(v), "Dgraph")) {
    val(v)
  }
}
```
Expected response:
```
{
  "data": {
    "q1": [
      {
        "name": "Dgraph"
      }
    ],
    "q2": [
      {
        "val(n)": "Dgraph"
      }
    ]
  }
```
Actual response (q2 doesn't show up):
```
{
  "data": {
    "q1": [
      {
        "name": "Dgraph"
      }
    ]
  }
```

Alpha logs show access denied for predicates `"name,v,val"`:
```
I0618 22:43:16.186675      16 access_ee.go:727] ACL-LOG Authorizing user "rouser" with groups "rogroup" on predicates "name,v,val" for "Read", allowed:false
```
##### `uid` function

Query: Run a query using the `uid()` function:

```
{
[...]

PR #5945, opened 2020-07-12
fix(Query): Fix val queries when ACL is enabled

anurags92:
### Motivation
This PR partially fixes:

- DGRAPH-1711 -- `expand(_all_)` and `value variable` queries do not work with ACL turned on. This PR fixes the issues with `value variable` queries.

For a detailed discussion and repro examples refer [here.](https://github.com/dgraph-io/dgraph/issues/5687)

`value variables` queries when ACL is enabled. 

No.

### Testing 
See `TestValQueryWithACLPermissions` in `acl_test.go`.
Tests include various queries related to `val(variable)` under following circumstances:
1. User has access to no predicate
2. User has access to only one predicate 
3. User has access to all predicates.

### Fixes 
Partially
Fixes DGRAPH-1711

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5945)

### DIFF
--- a/edgraph/access_ee.go
+++ b/edgraph/access_ee.go
@@ -42,7 +42,7 @@ import (
 
 type predsAndvars struct {
 	preds []string
-	vars  map[string]struct{ predicate string }
+	vars  map[string]string
 }
 
 // Login handles login requests from clients.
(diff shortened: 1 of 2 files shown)


edgraph/access_ee.go
```diff
@@ -806,7 +833,18 @@ func authorizeQuery(ctx context.Context, parsedReq *gql.Result, graphql bool) er
...
+		for predicate := range blockedPreds {
+			if variable, found := predToVarsMap[predicate]; found {
+				// Add variables to blockedPreds to delete from Query
+				blockedPreds[variable] = struct{}{}
+				// Collect blocked Variables to remove from QueryVars
+				blockedVars[variable] = struct{}{}
+			}
+		}
 		parsedReq.Query = removePredsFromQuery(parsedReq.Query, blockedPreds)
+		parsedReq.QueryVars = removeVarsFromQueryVars(parsedReq.QueryVars, blockedVars)
```
vvbalaji-dgraph, line 847: After Vars are removed from QueryVars is it still called QueryVars?

anurags92: Yes, not all vars are removed. Only the blocked vars. QueryVars field should remain as is, because it is used internally for other purposes.

anurags92:
> Can you just check that what is `gq.Var` in the case of `val(x)`


Yes, as discussed offline,`gq.Var` would be empty And `gq.Attr = val`; `gq.NeedsVar = x`

PR #5666, opened 2020-06-17
perf(GraphQL): Nested auth queries no longer search through all possible records

arijitAD:
Fixes GRAPHQL-427
- [x] Bring all filter queries outside of the main query.
- [x] Link query hierarchy
- [x] Check if cascade directive is properly applied
- [x] Check how mutations will be affected
- [x] Add Unit tests.
- [x] Add E2E test checking touched_uids

GraphQL query
```
  gqlquery: |
    query {
      getProject(projID: "0x123") {
        projID
        columns {
          name
          colID
        }
      }
    }
```
Corresponding Dgraph query with auth filter
```
    query {
      getProject(func: uid(Project2)) @filter((uid(Project3) AND type(Project))) {
        projID : uid
        columns : Project.columns @filter(uid(Column1)) {
          name : Column.name
          colID : uid
        }
      }
      Project2 as var(func: uid(0x123))
      Project3 as var(func: uid(Project2)) @cascade {
        roles : Project.roles @filter(eq(Role.permission, "VIEW")) {
          assignedTo : Role.assignedTo @filter(eq(User.username, "user1"))
          dgraph.uid : uid
        }
        dgraph.uid : uid
      }
      Column1 as var(func: type(Column)) @cascade {
        inProject : Column.inProject {
          roles : Project.roles @filter(eq(Role.permission, "VIEW")) {
            assignedTo : Role.assignedTo @filter(eq(User.username, "user1"))
            dgraph.uid : uid
          }
          dgraph.uid : uid
        }
        dgraph.uid : uid
      }
    }
```
Nested auth query is unoptimized because it will fetch all the Columns since we query `Column1 as var(func: type(Column))` but we need to get only `Columns` that correspond to the `Project`.
```
    query {
      getProject(func: uid(Project2)) @filter((uid(Project3) AND type(Project))) {
        projID : uid
        columns : Project.columns @filter(uid(Column1)) {
          name : Column.name
          colID : uid
        }
      }
      Project2 as var(func: uid(0x123))
      Project3 as var(func: uid(Project2)) @cascade {
        roles : Project.roles @filter(eq(Role.permission, "VIEW")) {
          assignedTo : Role.assignedTo @filter(eq(User.username, "user1"))
          dgraph.uid : uid
        }
        dgraph.uid : uid
      }
      var(func: uid(Project2)) {
        Column2 as column {
           dgraph.uid: uid
         }
      }
      Column1 as var(func: uid(Column2)) @cascade {
        inProject : Column.inProject {
[...]

PR #5950, opened 2020-07-13
perf(GraphQL): Nested auth queries no longer search through all possible records (#5666)

arijitAD:
Fixes GRAPHQL-427(Release blocker)

Previously auth queries had type filter for all levels and hence it would fetch all nodes of a particular type. This PR only applies auth on the nodes that are part of the parent query. Hence, touching same/less number of nodes as it would have in the query without auth rules.

(cherry picked from commit 222f42a21fb4d85200101ae6421a0efb08c76746)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5950)

 

Docs Preview: [Dgraph Preview](https://dgraph-8ca11d144a-78367.surge.sh)

### DIFF
--- a/graphql/admin/update_group.go
+++ b/graphql/admin/update_group.go
@@ -50,7 +50,7 @@ func (urw *updateGroupRewriter) Rewrite(
 		}
 		for _, ruleI := range rules {
 			rule := ruleI.(map[string]interface{})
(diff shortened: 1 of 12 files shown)


amaster507:
This made a world of difference! Took a 9 second query down to a 869ms query. Excellent work! Thanks guys!

PR #5956, opened 2020-07-13
fix(dgraph): Dgraph version picked from git tags on master

### LINKED ISSUE #5721: Health endpoint doesn't show a version when built from master
Health endpoint doesn't show a version (at 8080/health) when built from master - This causes a block on Ratel UI. Also, any Dgraph version will get blocked by this. Not sure if this needs to be fixed on Dgraph itself or Ratel. Pinging @paulftw.

2 potential bugs (both true, not sure if both can be solved at Dgraph side)
1 - When using master, ratel thinks I'm using an old version.
2 - Got "Unsupported Content-Type".

Master

yes

Darwin 32GB

1. Build Dgraph from master or any PR.
2. Start Dgraph's cluster built from master and run Ratel.
3. Connect Ratel to http://localhost:8080 

Captura de Tela 2020-06-24 às 11 55 45

Captura de Tela 2020-06-24 às 11 58 24

4. Now open the console (F12)

You might see in some moment an error like:

```
Login Failed http://localhost:8080 Error: Pre v1.1 clients do not support Login methods
```

Also several errors like:

```
serverAclState error t: Unsupported Content-Type. Supported content types are application/json, application/graphql+-
```

This is a normal health endpoint

Captura de Tela 2020-06-24 às 11 45 27

And this is the one from master/PR builds.

Captura de Tela 2020-06-24 às 11 45 17

That it should work with master builds.

rahulgurnani:
Fixes DGRAPH-1734
Use next release version as the dgraph version number for dgraph build on master branch. Refer: https://discuss.dgraph.io/t/health-endpoint-doesnt-show-a-version-when-built-from-master/8566/ for discussion about this.

Earlier `git describe` was being used to get the recent tag which is yielding `v2.0.0-rc1-503-g144685a84` on master. `git describe` finds the last ancestor commit which is shared with any tag and yield that tag. It's unclear why it's yielding an old tag on master. May be something wrong with the commit history somewhere.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5956)

 

Docs Preview: [Dgraph Preview](https://dgraph-79bbaa9719-82639.surge.sh)

### DIFF
--- a/dgraph/Makefile
+++ b/dgraph/Makefile
@@ -19,8 +19,14 @@ BUILD          ?= $(shell git rev-parse --short HEAD)
 BUILD_CODENAME  = unnamed
 BUILD_DATE     ?= $(shell git log -1 --format=%ci)
 BUILD_BRANCH   ?= $(shell git rev-parse --abbrev-ref HEAD)
-# Set latest tag as BUILD_VERSION
-BUILD_VERSION ?= $(shell git for-each-ref refs/tags --sort=-taggerdate --format='%(refname:short)' --count=1)
+
+# use a large version number v99 on master branch
+ifeq ($(BUILD_BRANCH),master)
+BUILD_VERSION ?= 'v20.11.0-'$(BUILD)
+else
+BUILD_VERSION ?= $(shell git describe --always --tags)
+endif
+
 BUILD_TAGS    ?=
 GOPATH        ?= $(shell go env GOPATH)
 
--- a/dgraph/Makefile
+++ b/dgraph/Makefile
@@ -23,13 +23,15 @@ BUILD_BRANCH   ?= $(shell git rev-parse --abbrev-ref HEAD)
 # Use the next build version on master branch
 # NOTE: this needs to be updated with each new release
 ifeq ($(BUILD_BRANCH),master)
-BUILD_VERSION ?= 'v20.11.0-'$(BUILD)
+# Version on master is of the format <next-release-version>-g<commit-id>
+BUILD_VERSION ?= 'v20.11.0-g'$(BUILD)
 else
 BUILD_VERSION ?= $(shell git describe --always --tags)
 endif
 
 BUILD_TAGS    ?=
 GOPATH        ?= $(shell go env GOPATH)
+MODIFIED = $(shell git diff-index --quiet HEAD || echo "-mod")
 
 export GO111MODULE := on
 
--- a/dgraph/Makefile
+++ b/dgraph/Makefile
@@ -14,7 +14,6 @@
 # limitations under the License.
 #
 
-<<<<<<< HEAD
 BIN             = dgraph
 BUILD          ?= $(shell git rev-parse --short HEAD)
 BUILD_CODENAME  = unnamed
--- a/dgraph/Makefile
+++ b/dgraph/Makefile
@@ -20,7 +20,8 @@ BUILD_CODENAME  = unnamed
 BUILD_DATE     ?= $(shell git log -1 --format=%ci)
 BUILD_BRANCH   ?= $(shell git rev-parse --abbrev-ref HEAD)
 
-# use a large version number v99 on master branch
+# Use the next build version on master branch
+# NOTE: this needs to be updated with each new release
 ifeq ($(BUILD_BRANCH),master)
 BUILD_VERSION ?= 'v20.11.0-'$(BUILD)
 else
--- a/dgraph/Makefile
+++ b/dgraph/Makefile
@@ -14,16 +14,16 @@
 # limitations under the License.
 #
 
+<<<<<<< HEAD
 BIN             = dgraph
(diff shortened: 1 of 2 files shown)


rahulgurnani:
> Yes, very important to add the git version of the master HEAD in the version string. This is needed by Sentry to distinguish dev vs prod version while reporting panics.
> 
> > _[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5956#-:-MC8O-WY1cFLXtdmIrOf:b9ck0ov)_ status: all files reviewed, 1 unresolved discussion (waiting on @manishrjain, @pawanrawal, @rahulgurnani, and @vvbalaji-dgraph)
> 
> _[dgraph/Makefile, line 22 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/5956#-MC8OI9iDmgOQUD0OTm6:-MC8OI9iDmgOQUD0OTm7:b16incs) ([raw file](https://github.com/dgraph-io/dgraph/blob/b6e384510e84847e7a1886278c7680da24b849c9/dgraph/Makefile#L22)):_
> 
> > ```makefile
> > BUILD_BRANCH  ?= $(shell git rev-parse --abbrev-ref HEAD)
> > # Set latest tag as BUILD_VERSION
> > BUILD_VERSION ?= $(shell git for-each-ref refs/tags --sort=-taggerdate --format='%(refname:short)' --count=1)
> > ```
> 
> Wouldn't this keep flipping between 20.07.x, 20.03.x, 1.2.x etc depending on which was the last release. That is super confusing to see 20.07.x at one point and then 20.03.x later.

Yes, if a new tag is created, this would change and it's confusing. 
We can do one thing. If the branch is master, take version to be a large constant version, like v99 and append the commit id to it. 
Let me know how does this sound @parasssh @danielmai

rahulgurnani:
Updated based on discussion here: https://discuss.dgraph.io/t/health-endpoint-doesnt-show-a-version-when-built-from-master/8566/9?u=rahul

dgraph/Makefile
```diff
@@ -18,7 +18,14 @@ BIN            = dgraph
 BUILD         ?= $(shell git rev-parse --short HEAD)
 BUILD_DATE    ?= $(shell git log -1 --format=%ci)
 BUILD_BRANCH  ?= $(shell git rev-parse --abbrev-ref HEAD)
+
+# use a large version number v99 on master branch
+ifeq ($(BUILD_BRANCH),master)
+BUILD_VERSION ?= 'v99'
```
parasssh, line 24: I presume you will also have the latest git commit-id in the version string.

rahulgurnani: No. Why do we need it? Sentry already has a "commit" field corresponding to each log which keeps the git commit id.

danielmai:
___
*[dgraph/Makefile, line 22 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/5956#-MC8OI9iDmgOQUD0OTm6:-MDImbmO7LH_6dfQfHPH:b9vvdzd) ([raw file](https://github.com/dgraph-io/dgraph/blob/b6e384510e84847e7a1886278c7680da24b849c9/dgraph/Makefile#L22)):*
<details><summary><i>Previously, rahulgurnani (Rahul Gurnani) wrote…</i></summary><blockquote>

Yes, if a new tag is created, this would change and it's confusing.  We can do one thing. If the branch is master, take version to be a large constant version, like v99 and append the commit id to it.  Let me know how does this sound @parasssh @danielmai  
</blockquote></details>

Can we make master builds refer to v20.11.0-<sha> instead of v99? Whatever is today's master is going to eventually be v20.11. It would need to be updated every time we cut a release branch, but at least it's clear from a feature-standpoint that the master builds will contain features for v20.11. v99 is impossible for us to track.

dgraph/Makefile
```diff
@@ -18,7 +18,14 @@ BIN            = dgraph
 BUILD         ?= $(shell git rev-parse --short HEAD)
 BUILD_DATE    ?= $(shell git log -1 --format=%ci)
 BUILD_BRANCH  ?= $(shell git rev-parse --abbrev-ref HEAD)
+
+# use a large version number v99 on master branch
```
parasssh, line 22: Rectify comment.

dgraph/Makefile
```diff
@@ -18,7 +18,14 @@ BIN            = dgraph
...
+ifeq ($(BUILD_BRANCH),master)
+BUILD_VERSION ?= 'v20.11.0-'$(BUILD)
```
parasssh, line 24: Maybe add a note here that it needs to be bumped every few months.

rahulgurnani: Done

dgraph/Makefile
rahulgurnani: Updated

PR #5983, opened 2020-07-14
fix(graphQL): Validate subscriptions in Operation function (#5926)

pawanrawal:
We realised that as a result of b40c632, introspection queries started failing in Insomnia because subscription didn't exist in the types. Instead of adding more hacks there, we add validation for this inside Operation. We already do something similar for Queries/Mutations.

(cherry picked from commit e33850e1625a0c6e980224fdb862d7f585bcd441)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5983)

 

Docs Preview: [Dgraph Preview](https://dgraph-6bc865451d-79376.surge.sh)

### DIFF
--- a/graphql/e2e/directives/schema.graphql
+++ b/graphql/e2e/directives/schema.graphql
@@ -139,7 +139,7 @@ type Student implements People {
         taughtBy: [Teacher] @hasInverse(field: "teaches")
 }
 
-type Message {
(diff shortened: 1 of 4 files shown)


parasssh:
Please add the JIRA ticket in the description.

PR #5999, opened 2020-07-15
fix(dgraph): Panic because of nil map in groups.go

rahulgurnani:
Fixes DGRAPH-1935

delta.GroupChecksums is nil at this point in the code, so copying keys to this map causes the code to panic with message:
`assignment to entry in nil map`

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/5999)

### DIFF
--- a/worker/groups.go
+++ b/worker/groups.go
@@ -924,6 +924,7 @@ func (g *groupi) processOracleDeltaStream() {
 
 		for {
 			var delta *pb.OracleDelta
+			delta.GroupChecksums = make(map[uint32]uint64)
 			var batch int
 			select {
 			case delta = <-deltaCh:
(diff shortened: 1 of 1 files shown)


rahulgurnani:
@martinmr Addressed your review comment.

martinmr:
Can you run your tests locally. Lots of tests are failing so maybe the change is not as benign as we thought. But possible that this is just TeamCity acting up.

parasssh:
All CI tests pass locally.

parasssh:
> Can you run your tests locally. Lots of tests are failing so maybe the change is not as benign as we thought. But possible that this is just TeamCity acting up.

Actually, the code was faulty. Made the fix to do a make on the map just before using it in the for loop as Rahul had it earlier.

PR #6054, opened 2020-07-22
fix(GraphQL): fix bug in custom resolver, now body need not have all the fields.

JatinDevDG:
This PR fix bug in custom resolver, now body need not have all the fields(both nullable and non-nullable).

https://discuss.dgraph.io/t/evaluation-of-custom-field-failed-while-substituting-variables-into-body-for-remote-endpoint-with-an-error/8700

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6054)

 

Docs Preview: [Dgraph Preview](https://dgraph-ad97cfa78d-83180.surge.sh)


MichaelJCompton:
It does look like the test isn't quite right.  What I expect is that if the author's id is missing in the variables, then "author" shouldn't get serialized into the result body at all.

JatinDevDG:
> It does look like the test isn't quite right. What I expect is that if the author's id is missing in the variables, then "author" shouldn't get serialized into the result body at all.

fixed it.

JatinDevDG:
> Can you verify that optional arguments in custom query/mutation are either sent as null when supplied null or not sent if they are not given?
> 
> > Reviewed 4 of 4 files at r3.
> > _[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6054#-:-MDFNQrK6I4Ob5JxuAh8:b-qfcwab)_ status: all files reviewed, 5 unresolved discussions (waiting on @arijitAD, @JatinDevDG, and @MichaelJCompton)
> 
> _[graphql/e2e/custom_logic/custom_logic_test.go, line 1023 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/6054#-MDFN0aFDJnaW1usW022:-MDFN0aG2hyNisH0ApF4:bmvtbhu) ([raw file](https://github.com/dgraph-io/dgraph/blob/30d2da891d88f637f3e7950934f69b28c85e0afc/graphql/e2e/custom_logic/custom_logic_test.go#L1023)):_
> 
> > ```go
> > 			@custom(
> > 				http: {
> > 					url: "http://mock:8888/userName"
> > ```
> 
> Add a new endpoint to the mock server called `/userNameWithoutAddress`. In that verify that the body should not contain address.
added test.
> 
> _[graphql/schema/wrappers.go, line 1899 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/6054#-MDElwse1MZNKU0QKjKg:-MDElwse1MZNKU0QKjKh:b22quet) ([raw file](https://github.com/dgraph-io/dgraph/blob/30d2da891d88f637f3e7950934f69b28c85e0afc/graphql/schema/wrappers.go#L1899)):_
> 
> > ```go
> > 				return err
> > 			}
> > 			//if ok {
> > ```
> 
> Remove this

removed.

PR #6058, opened 2020-07-22
Pagination param "after" does not work when using func: uid(v)

### LINKED ISSUE #5242: Pagination param "after" does not work when using func: uid(v)

v1.2.0

Yes

4G, Linux

Go to 
https://dgraph.io/docs/query-language/#inequality
Try:
```
curl -H "Content-Type: application/graphql+-" localhost:8080/query -XPOST -d '
{
  var(func: has(name@en), first: 50) {
    test as uid
  }

  ss(func: uid(test), first: 2, after: 0x6) {
    uid
  }
}' | python -m json.tool | less
```

Response:
```
{
  "data": {
    "ss": [
      {
        "uid": "0x4"
      },
      {
        "uid": "0x6"
      }
    ]
  }
}
```

Expected:
Returns uid > 0x6

antblood:
---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6058)

1. Pagination param "after" does not work when using func: uid(v). For ex:
for query
```
{
  var(func: has(name@en), first: 50) {
    test as uid
  }

  ss(func: uid(test), first: 2, after: 0x6) {
    uid
  }
}
```
before the result was

```
{
  "data": {
    "ss": [
      {
        "uid": "0x4"
      },
      {
        "uid": "0x6"
      }
    ]
  }
}
```
now the result is
```
{
  "data": {
    "ss": [
      {
        "uid": "0x7"
      },
      {
        "uid": "0x8"
      }
    ]
  }
}
```
2. Fixes #5242.
3. Fixes [DGRAPH-1273](https://dgraph.atlassian.net/browse/DGRAPH-1273).
 

Docs Preview: [Dgraph Preview](https://dgraph-d9f6e30bc3-90479.surge.sh)


query/query1_test.go
```diff
@@ -669,6 +669,23 @@ func TestHasFuncAtRootWithAfter(t *testing.T) {
...
 }
 
+func TestHasFuncAtRootWithAfterOnUIDs(t *testing.T) {
+
+	query := `
+	{
+		var(func: has(name)) {
+			uids as uid
+		}
+		me(func: uid(uids), first: 2, after: 0x5) {
```
pawanrawal, line 679: This is looking good, one thing to verify is that if it also works at a nested level. So if the second query was something like

```
{
  me(func: uid(1, 2, 3, 4, 5) {
    friend @filter(uid(uids), after: ...) {
      name
    }
  }
}
```

pawanrawal (approved):

PR #6072, opened 2020-07-24
(release/v1.2) refactor: Simplify how list splits are tracked.

martinmr:
Currently the code keeps tracks of the split startUids and updates them
as the posting lists are split or empty parts are removed. Instead,
update the list of startUids at the end of each step based on the
contents of the map of startUids to posting list. This removes the need
of keeping track of the same piece of information in two different ways.

Also, fix a bug where the empty parts were removed from the list of
splits but not from the map. This caused empty posting lists to be
written to disk although it didn't affect normal execution because those
parts were not accessible from the main posting list.

(cherry picked from commit c51d007e7d53a0d313bab3e73e8536ff1fc455c0)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6072)

 

Docs Preview: [Dgraph Preview](https://dgraph-b8c97d2e40-82250.surge.sh)


martinmr:
Ran tests locally. Issues in TeamCity are due to the python binary missing.

PR #6083, opened 2020-07-28
fix(dgraph): Fix out of order issues with split keys in bulk loader.

martinmr:
Split keys for indexes can cause out-of-order issues due to the variable length of the term inside the key. 
To fix the issue, the split keys are written to a temporary DB first (using the writebatch to avoid the out
of order issues) and then copied to the main p directory.

Related to DGRAPH-1897

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6083)


dgraph/cmd/bulk/reduce.go
```diff
@@ -127,11 +136,23 @@ func (r *reducer) createBadger(i int) *badger.DB {
...
+func (r *reducer) createBadger(i int) *badger.DB {
+	db := r.createBadgerInternal(r.opt.shardOutputDirs[i])
 	r.dbs = append(r.dbs, db)
 	return db
 }
 
+func (r *reducer) createTmpBadger() *badger.DB {
+	tmpDir, err := ioutil.TempDir(r.opt.TmpDir, "split")
+	x.Check(err)
+	db := r.createBadgerInternal(tmpDir)
```
jarifibrahim, line 151: createbadgerInteral creates a badger instance with compression enabled. I don't think we need compression for temporary DB.

PR #6117, opened 2020-07-30
feat: Live Loader Now Supports Slash GraphQL

gja:
Live Loader can now support Slash GraphQL

Here is an example: 
```
./dgraph live --slash_grpc_endpoint=defiant-curve-2926.grpc.us-east-1.aws.thegaas.com:443 -f /path/to/file.json.gz -t api-token
```

Why introduce a new flag, `--slash_grpc_endpoint`?
* Connections to zero didn't support TLS, so we'd need a new flag to indicate this
* There was no way to start TLS, but use the system CA pool (--tls_ca_cert had to be a valid path). This is not needed for public auth
* Passing in the --alpha --zero and --tls_server_name with the same value felt clunky
* In Slash GraphQL, the GRPC authentication is passed in via `Authentication` metadata to all endpoints, as opposed to only passing in `X-Dgraph-AuthToken` to alter alone

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6117)

### DIFF
--- a/dgraph/cmd/live/run.go
+++ b/dgraph/cmd/live/run.go
@@ -20,6 +20,7 @@ import (
 	"bufio"
 	"compress/gzip"
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


rstorr:
@codelingobot capture errors should be checked immediately after they are declared

rstorr:
@CodeLingoBot capture errors should be checked immediately after they are declared

PR #6142, opened 2020-08-06
Stop forcing RAM mode for the w directory.

martinmr:


martinmr:
This fixes the issue but it would change the default. The flag badger.tables is used for both the p and w directories and the default is mmap. Not sure if we should leave the flag as is or add another flag for the w directory.

PR #6145, opened 2020-08-06
fix(Dgraph): Queue keys for rollup during mutation.

martinmr:
Right now, keys are only added for rollups during queries. Keys that
only receive queries will not be rolled up until they are queried, which
can cause issues and slowness.

Fixes DGRAPH-2202

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6145)


martinmr:
I verified with the live loader and I see rollups happening.

manishrjain:
When we do the mutations, we keep changes in memory and don't write to data to Badger. We write the data when we do the commit. The rollups should be triggered during the commit, not during the mutation. So, this change is wrong.

PR #6148, opened 2020-08-07
feat: add metrics cost to grpc and http header

balajijinnah:
---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6148)

### DIFF
--- a/dgraph/cmd/alpha/http_test.go
+++ b/dgraph/cmd/alpha/http_test.go
@@ -196,8 +196,8 @@ func queryWithTs(queryText, contentType, debug string, ts uint64) (string, uint6
 	return string(output), startTs, err
 }
 
-func queryWithTsForCost(queryText, contentType, debug string, ts uint64) (string,
-	uint64, string, error) {
+func queryWithTsForResp(queryText, contentType, debug string, ts uint64) (string,
+	uint64, *http.Response, error) {
 	params := make([]string, 0, 2)
 	if debug != "" {
 		params = append(params, "debug="+debug)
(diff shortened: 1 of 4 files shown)


x/x.go
```diff
@@ -146,6 +146,7 @@ const (
 	AccessControlAllowedHeaders = "X-Dgraph-AccessToken, " +
 		"Content-Type, Content-Length, Accept-Encoding, Cache-Control, " +
 		"X-CSRF-Token, X-Auth-Token, X-Requested-With"
+	DgraphCostHeader = "X-Dgraph-Cost"
```
gja, line 149: call in numUids please

dgraph/cmd/alpha/http_test.go
```diff
@@ -196,6 +196,37 @@ func queryWithTs(queryText, contentType, debug string, ts uint64) (string, uint6
 	return string(output), startTs, err
 }
 
+func queryWithTsForCost(queryText, contentType, debug string, ts uint64) (string,
```
gja, line 199: rather than making this a new function, you should probably pass options to queryWithTs(). Or if you don't want to do update all the callers.

```golang
type validationOpts struct {
  ResponseCallback func() 
}

func _queryWithTs(..., opts *validationOpts) {
...
context := map[string]interface{}
  if opt.ResponseCallback {
    err := opt.ResponseCallback(context, response)
  }

  return ..., context, error
}

func Test...() {
  ...
  queryWithTs(..., &callbackOptions{ResponseCallback: func() { // put cost into context }})
}
```

balajijinnah: I've made it to return resp instead of passing the validation function.

x/x.go
```diff
@@ -146,6 +146,7 @@ const (
 	AccessControlAllowedHeaders = "X-Dgraph-AccessToken, " +
 		"Content-Type, Content-Length, Accept-Encoding, Cache-Control, " +
 		"X-CSRF-Token, X-Auth-Token, X-Requested-With"
+	DgraphCostHeader = "numUids"
```
gja, line 149: Oh sorry, I should have been explicit 🙈.. Let's keep it similar to graphql.

Dgraph-TouchedUids

balajijinnah:
@darkn3rd I've added integration test. That should be enough.

PR #6156, opened 2020-08-11
fix: Don't allow idx flag to be set to 0 on dgraph zero

rahulgurnani:
**Fixes DGRAPH-2185**
The panic mentioned in the bug is caused when idx is set to 0. The raft library we are using [etcd/raft](https://github.com/etcd-io/etcd/tree/master/raft) doesn't allow ID in the raft config to be 0 as it is reserved [to denote None i.e. no leader](https://github.com/etcd-io/etcd/blob/master/raft/raft.go#L34).
This change just adds this check to gracefully exit if idx is set to 0.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6156)

 
 

Docs Preview: [Dgraph Preview](https://dgraph-37c3d9dcc2-85845.surge.sh)

### DIFF
--- a/dgraph/cmd/zero/run.go
+++ b/dgraph/cmd/zero/run.go
@@ -83,7 +83,7 @@ instances to achieve high-availability.
 		"addr:port of this server, so other Dgraph alphas can talk to this.")
 	flag.IntP("port_offset", "o", 0,
 		"Value added to all listening port numbers. [Grpc=5080, HTTP=6080]")
-	flag.Uint64("idx", 1, "Unique node index for this server.")
+	flag.Uint64("idx", 1, "Unique node index for this server. idx cannot be 0.")
 	flag.Int("replicas", 1, "How many replicas to run per data shard."+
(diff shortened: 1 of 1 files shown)


dgraph/cmd/zero/run.go
```diff
@@ -83,7 +83,7 @@ instances to achieve high-availability.
 		"addr:port of this server, so other Dgraph alphas can talk to this.")
 	flag.IntP("port_offset", "o", 0,
 		"Value added to all listening port numbers. [Grpc=5080, HTTP=6080]")
-	flag.Uint64("idx", 1, "Unique node index for this server.")
+	flag.Uint64("idx", 1, "Unique node index for this server. idx cannot be 0.")
```
rahulgurnani, line 86: @danielmai  Shall we update the docs here https://dgraph.io/docs/deploy/single-host-setup/#run-dgraph-zero ?

PR #6159, opened 2020-08-11
perf(GraphQL): Fix filter order with UID (#5928)

abhimanyusinghgaur:
There were cases that get rewritten like `updateStudent(func: type(Student)) @filter(uid(0x123)) { ... }`. This PR changes them to get rewritten like `updateStudent(func: uid(0x123)) @filter(type(Student)) { ... }` because, having `uid` as root func is more efficient than having `type` as root func.

(cherry picked from commit 64ce93bf6890eb20b09017453ae78fb63b3bf80e)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6159)

 

Docs Preview: [Dgraph Preview](https://dgraph-b202a7c9e9-85127.surge.sh)

### DIFF
--- a/graphql/resolve/auth_update_test.yaml
+++ b/graphql/resolve/auth_update_test.yaml
@@ -19,7 +19,7 @@
         uid
       }
       UserSecretRoot as var(func: uid(UserSecret1)) @filter(uid(UserSecretAuth2))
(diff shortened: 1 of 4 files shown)


pawanrawal:
Let's skip this one as it can introduce the bug that we already have in query into update as well. Once we fix that bug, we can look into this.

PR #6168, opened 2020-08-12
feat(GraphQL): Add HasFilter in the Schema

minhaj-shakeel:
---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6168)

 
 

Docs Preview: [Dgraph Preview](https://dgraph-5da1a43ec4-86307.surge.sh)


graphql/schema/gqlschema.go
```diff
@@ -694,6 +695,36 @@ func addFilterArgument(schema *ast.Schema, fld *ast.FieldDefinition) {
...
+		Name: filterName,
+	}
+
+	for _, fld := range defn.Fields {
+		if isID(fld) {
+			continue
+		}
+		filter.EnumValues = append(filter.EnumValues,
+			&ast.EnumValueDefinition{Name: fld.Name})
+
```
pawanrawal, line 713: this newline isn't required

graphql/schema/gqlschema.go
```diff
@@ -812,6 +839,13 @@ func addFilterType(schema *ast.Schema, defn *ast.Definition) {
 		}
 	}
 
+	// Has filter makes sense only if there is atleast one non ID field in the defn
+	if len(getNonIDFields(schema, defn)) > 0 {
```
pawanrawal, line 843: Actually sorry we should use `getFieldsWithoutIDType` here because that gives all fields except field having type `ID`. So it also gives fields with `@id` directive which is what we want because user can give the `Has` filter on a field with `@id` directive as well.

PR #6206, opened 2020-08-17
fix(Dgraph): Don't store start_ts in postings

martinmr:
Clear out the start_ts field in the postings of deltas before they are
marshalled to avoid storing that field in disk. This field is only meant
to be used during in-memory processing.

Related to https://discuss.dgraph.io/t/start-ts-not-being-cleared-before-postings-are-written-to-disk/9146

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6206)


martinmr:
The previous PR didn't work because I was using a read lock to modify the protobufs. I verified that the TestConcurrentQueryMutate test runs fine by running it 50 times.

martinmr:
Ok. I'll open the cherry-picks today so I don't forget but I won't push them until the releases are out.

PR #6211, opened 2020-08-17
CID never created if Zero stops early after first init (cherry pick to release/v1.2)

antblood:
Problem:

When we early(within 1s) stop a fresh dgraph zero. The CID is never assigned.

Changes:

Earlier when the dgraph zero wasn't started fresh, it was assumed that the CID is present in the entries.
Now I'm checking whether there's a CID is present in the entries on a restart. If it is present then that one is used, else a new CID is proposed.

Fixes DGRAPH-604
(cherry picked from commit 09ef988fa53e84c75455dec0dac43284ba89ed59)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6211)


antblood:
unwanted files included

PR #6267, opened 2020-08-24
chore(badger): Show cache metrics

jarifibrahim:
**This PR should not be cherry-picked to older versions. It uses the latest version of badger with the calloc changes.**

This PR adds support for monitoring cache metrics. We will print a warning message if we detect contention in the cache.

The cache is said to be in contention if
- The hit ratio is less than 0.4
- The mean time for an entry in cache is less than 5 seconds (item are being evicted very fast)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6267)


jarifibrahim:
Closing this PR in favour of https://github.com/dgraph-io/dgraph/pull/6426

PR #6291, opened 2020-08-27
Update badger to v2.2007.2

jarifibrahim:
This badger update brings the following changes from badger.
```
b41e77a Fix(cleanup): Avoid truncating in value.Open on error (#1465) (#1489)
a0d4903 fix(cleanup): Do not close cache before compaction (#1464) (#1488)
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6291)


jarifibrahim:
We don't need this PR anymore since master is already on latest badger.

PR #6314, opened 2020-08-28
fix(Dgraph): skip backing up nil lists.

martinmr:
Version 1.2 returns a nil list from ReadPostingList. The backup was
trying to call the method SingleListRollup. The fix is to ignore those
lists. They will be backed up when the main key is accessed.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6314)

 

Docs Preview: [Dgraph Preview](https://dgraph-2f08db6222-89552.surge.sh)


martinmr:
Tests pass locally.

PR #6319, opened 2020-08-29
release/v1.2 - fix(task): Return error on closed DB (#6075)

jarifibrahim:
Fixes DGRAPH-2181

The queries in dgraph are processed in separate goroutines. The pstore badger DB
could be closed while the query was being processed. This causes panics such as
```
panic: runtime error: invalid memory address or nil pointer dereference
	panic: Unclosed iterator at time of Txn.Discard.
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x11d2311]

goroutine 19298 [running]:
github.com/dgraph-io/badger/v2.(*Txn).Discard(0xc05586bc20)
	/go/pkg/mod/github.com/dgraph-io/badger/v2@v2.0.1-rc1.0.20200718033852-37ee16d8ad1c/txn.go:517 +0xc1
panic(0x19e26a0, 0x2988560)
	/usr/local/go/src/runtime/panic.go:969 +0x166
github.com/dgraph-io/badger/v2/skl.(*Skiplist).IncrRef(...)
```
This PR attempts to reduce the number of such crashes. This PR doesn't fix the actual
issue but it tries to reduce the probability of such crashes by checking if badger is not
closed before accessing it.

An ideal fix would be to stop all the goroutines started by dgraph while closing DB so
that we don't try to read from a closed badger DB.

(cherry picked from commit 3cea0fe26c27d62e9ebc5f9574cdc9ba2e31fa13)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6319)

 

Docs Preview: [Dgraph Preview](https://dgraph-38ac76a875-89711.surge.sh)

### DIFF
--- a/posting/mvcc.go
+++ b/posting/mvcc.go
@@ -225,6 +225,9 @@ func ReadPostingList(key []byte, it *badger.Iterator) (*List, error) {
 
 // TODO: We should only create a posting list with a specific readTs.
 func getNew(key []byte, pstore *badger.DB) (*List, error) {
+	if pstore.IsClosed() {
+		return nil, badger.ErrDBClosed
+	}
 	txn := pstore.NewTransactionAt(math.MaxUint64, false)
 	defer txn.Discard()
 


jarifibrahim:
I ran the tests locally and it works fine.

PR #6323, opened 2020-08-31
fix(raft): Only leader should check the quorum

jarifibrahim:
The Checkquorum() function is used by the leader to determine if it is
still a leader and not behind a network partition (split-head situation).

The current code would checkquorum even for followers which isn't necessary.
This PR removes the quorum check for followers.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6323)

### DIFF
--- a/dgraph/cmd/zero/raft.go
+++ b/dgraph/cmd/zero/raft.go
@@ -61,15 +61,16 @@ func (n *node) amLeader() bool {
 
 func (n *node) AmLeader() bool {
 	// Return false if the node is not the leader. Otherwise, check the lastQuorum as well.
-	return n.amLeader() && func() bool {
-		// This node must be the leader, but must also be an active member of
-		// the cluster, and not hidden behind a partition. Basically, if this
-		// node was the leader and goes behind a partition, it would still
-		// think that it is indeed the leader for the duration mentioned below.
-		n.mu.RLock()
-		defer n.mu.RUnlock()
-		return time.Since(n.lastQuorum) <= 5*time.Second
-	}()
+	if !n.amLeader() {
+		return false
+	}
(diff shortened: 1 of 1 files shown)


manishrjain (changes requested):
Run this with blockade.

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6323#-:-MGJIonhAkJZAGM_Gq_2:b-7p6ae1)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jarifibrahim, @manishrjain, and @vvbalaji-dgraph)

___
*[dgraph/cmd/zero/raft.go, line 64 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/6323#-MGJI5of4nszdgdVrWRg:-MGJI5of4nszdgdVrWRh:b-p9l6h1) ([raw file](https://github.com/dgraph-io/dgraph/blob/93fbecd88f42e62db9c2c7c48d7517a9aa3c96df/dgraph/cmd/zero/raft.go#L64)):*
> ```Go
> func (n *node) AmLeader() bool {
> 	// Return false if the node is not the leader. Otherwise, check the lastQuorum as well.
> 	return n.amLeader() && func() bool {
> ```

This is unusual way of writing an if statement.

manishrjain (commented):
If it works with Blockade, then good to go.

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6323#-:-MGJIrRT1ud3YRjOFE1r:blflozz)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jarifibrahim, @manishrjain, and @vvbalaji-dgraph)

PR #6338, opened 2020-08-31
fix(Dgraph): add flag to set up compression in zero.

martinmr:
---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6338)

 

Docs Preview: [Dgraph Preview](https://dgraph-7f60b0034a-90556.surge.sh)


martinmr:
@jarifibrahim: What should the default value for this flag be? Right now it's set to three which is the same value as the alpha but compression was not enabled in zero before so a value of zero might be better.

jarifibrahim (commented):
We should keep compression on in zero. We had it enabled but it got disabled when we changed badger default options.

This PR :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6338#-:-MG9Gn-O-kHK6HmgDfGV:b-hbarpc)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @manishrjain and @vvbalaji-dgraph)

PR #6359, opened 2020-09-01
fix(shutdown): Force exit if CTRL-C is caught before initialization

jarifibrahim:
Forcefully kill alpha if CTRL-C is caught before node initialization
has completed.

Fixes DGRAPH-2377

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6359)

### DIFF
--- a/dgraph/cmd/alpha/run.go
+++ b/dgraph/cmd/alpha/run.go
@@ -80,6 +80,8 @@ var (
 
 	// need this here to refer it in admin_backup.go
 	adminServer web.IServeGraphQL
+
+	initDone uint32
 )
 
 func init() {
@@ -569,6 +571,7 @@ func setupServer(closer *y.Closer) {
 	glog.Infoln("gRPC server started.  Listening on port", grpcPort())
 	glog.Infoln("HTTP server started.  Listening on port", httpPort())
 
+	atomic.AddUint32(&initDone, 1)
 	admin.ServerCloser.Wait()
 }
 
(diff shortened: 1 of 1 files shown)


dgraph/cmd/alpha/run.go
```diff
@@ -746,7 +749,13 @@ func run() {
 			}
 			numShutDownSig++
 			glog.Infoln("Caught Ctrl-C. Terminating now (this may take a few seconds)...")
-			if numShutDownSig == 3 {
+
+			switch {
+			case atomic.LoadUint32(&initDone) < 2:
```
parasssh, line 754: condition is < 2 because you have two checkpoints that need to be reached to consider it initialized?

dgraph/cmd/alpha/run.go
```diff
@@ -771,6 +780,7 @@ func run() {
 			x.UpdateCorsOrigins(origins)
 			return
 		}
+		atomic.AddUint32(&initDone, 1)
```
parasssh, line 783: Alpha does not seem to reach this line of code. As a result, initDone is never reaching the value 2.

dgraph/cmd/alpha/run.go
jarifibrahim: Yes. One checkpoint in serve(..) function and the other one is after raft nodes are initialized.

dgraph/cmd/alpha/run.go
jarifibrahim: Yes, there was an infinite loop above this line. It should be fixed now.

PR #6367, opened 2020-09-02
don't propose trail license if the licences is provided by the user.

balajijinnah:
because, raft ready already proposing enterprise license. So, avoiding
the race condition.

Signed-off-by: பாலாஜி <balaji@dgraph.io>

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6367)

### DIFF
--- a/dgraph/cmd/zero/raft.go
+++ b/dgraph/cmd/zero/raft.go
@@ -458,7 +458,7 @@ func (n *node) proposeNewCID() {
 	}
 
 	// Apply trial license only if not already licensed.
-	if n.server.license() == nil {
+	if n.server.license() == nil && Zero.Conf.GetString("enterprise_license") == "" {
 		if err := n.proposeTrialLicense(); err != nil {
 			glog.Errorf("while proposing trial license to cluster: %v", err)
 		}
--- a/dgraph/cmd/zero/raft.go
+++ b/dgraph/cmd/zero/raft.go
@@ -457,7 +457,7 @@ func (n *node) proposeNewCID() {
 		time.Sleep(3 * time.Second)
 	}
 
-	// Apply trial license only if not already licensed.
+	// Apply trial license only if not already licensed and no enterprise license provided.
 	if n.server.license() == nil && Zero.Conf.GetString("enterprise_license") == "" {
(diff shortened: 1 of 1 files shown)


dgraph/cmd/zero/raft.go
```diff
@@ -458,7 +458,7 @@ func (n *node) proposeNewCID() {
 	}
 
 	// Apply trial license only if not already licensed.
-	if n.server.license() == nil {
+	if n.server.license() == nil && Zero.Conf.GetString("enterprise_license") == "" {
```
jarifibrahim, line 461: The license will be applied for the second time in node.Run()
https://github.com/dgraph-io/dgraph/blob/1783e0edafb995b344eb5cfd22008f9ab0b29be8/dgraph/cmd/zero/raft.go#L795-L803
Is it okay to reapply the license? I think the other call would complain.

I think we apply the trial license here so that we can get a CID but I'm not 100% sure.

balajijinnah: The problem is trail license is applied after enterprise license in some cases. I thought the same as moving the trail license to Run method. But, this is little cleaner and simple approch.

jarifibrahim: How about moving the `apply trial` code from here to where we actually apply the license?
So we apply either the license user has provided or the trial license.

balajijinnah: That would lead to applying the trail license again and again. I believe they kept the trail code in CID generation because they want to apply the trail only when the CID is generated.

dgraph/cmd/zero/raft.go
```diff
@@ -458,7 +458,7 @@ func (n *node) proposeNewCID() {
 	}
 
 	// Apply trial license only if not already licensed.
```
jarifibrahim, line 460: Update this comment, please.

dgraph/cmd/zero/raft.go
jarifibrahim: Okay. This should be okay to do in that case.

jarifibrahim (approved):
LGTM! Please get @parasssh and @martinmr's review on this.

PR #6370, opened 2020-09-02
fix(Dgraph): Parse Content-Type in headers properly

ajeetdsouza:
Fixes DGRAPH-2227.

A request sent to a query or mutation handler with the header `Content-Type: application/json; charset=utf-8` fails even though it is perfectly valid. This PR fixes parsing of the `Content-Type` header.

Currently, I've added a check - `charset`, if defined, must be `utf-8`. We may or may not want to remove that check.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6370)

### DIFF
--- a/dgraph/cmd/alpha/http.go
+++ b/dgraph/cmd/alpha/http.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"mime"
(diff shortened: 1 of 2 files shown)


parasssh (approved):
Minor comment. Add UT. LGTM otherwise.

PR #6385, opened 2020-09-03
fix(GraphQL): fix introspection completion bug

abhimanyusinghgaur:
Fixes GRAPHQL-673.
This PR fixes the issue where introspection queries would break if there were two types implementing same interface and having fields with same name in those two types and that repeating field is also a field in introspection query.
This was introduced after #6228.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6385)


graphql/schema/wrappers.go
```diff
@@ -673,7 +673,15 @@ func (f *field) DgraphAlias() string {
 	// if this field is repeated, then it should be aliased using its dgraph predicate which will be
 	// unique across repeated fields
 	if f.op.inSchema.repeatedFieldNames[f.Name()] {
-		return f.DgraphPredicate()
+		dgraphAlias := f.DgraphPredicate()
```
pawanrawal, line 676: This could have been simplified by appending the if below to the if above in this block with an and condition.

pawanrawal (approved):

graphql/e2e/schema/introspection.graphql
```diff
@@ -0,0 +1,113 @@
+query {
```
pawanrawal, line 1: Could we have this file at one place and reference it from everywhere else? I thought we already had it somewhere else too.

PR #6392, opened 2020-09-04
fix(GraphQL): don't generate orderable enum value for list fields

abhimanyusinghgaur:
Fixes GRAPHQL-650.

For this user schema:
```
type Starship {
        id: ID!
        name: String! @search(by: [term])
        length: Int64
        tags: [String]
        createdAt: DateTime
}
```
we were generating complete GraphQL schema with StarshipOrderable as:
```
enum StarshipOrderable {
  name
  length
  tags
  createdAt
}
```
that would cause a DQL query to be formed which errors out, see this GraphQL query:
```
query {
  queryStarship(order: {asc: tags}) {
    id
    name
    length
    tags
  }
}
```
It gives back:
```
{
  "errors": [
    {
      "message": "Dgraph query failed because Dgraph execution failed because : Sorting not supported on attr: Starship.tags of type: [scalar]"
    }
  ],
  "data": {
    "queryStarship": []
  }
}
```
which means we should not allow even list of scalar along with object types in orderable. Only scalar field should be allowed in orderable. So, the correct orderable should be without tags field:
```
enum StarshipOrderable {
  name
  length
  createdAt
}
```

This PR fixes the above bug.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6392)

### DIFF
--- a/graphql/schema/gqlschema.go
+++ b/graphql/schema/gqlschema.go
@@ -1020,7 +1020,7 @@ func addTypeOrderable(schema *ast.Schema, defn *ast.Definition) {
 	}
 
 	for _, fld := range defn.Fields {
-		if orderable[fld.Type.Name()] {
+		if fld.Type.NamedType != "" && orderable[fld.Type.NamedType] {
 			order.EnumValues = append(order.EnumValues,
 				&ast.EnumValueDefinition{Name: fld.Name})
 		}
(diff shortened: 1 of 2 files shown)


pawanrawal (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6392#-:-MGOl7t54vNsSmhfqBKW:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @MichaelJCompton)

PR #6393, opened 2020-09-04
Add value variable tests

animesh2049:
Added some tests which involve use of value variables in mutations, when ACL is enabled.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6393)


anurags92:
Looks good to me.

PR #6463, opened 2020-09-16
build(jemalloc): Install Jemalloc if missing via Makefile.

jarifibrahim:
This PR adds a `jemalloc` target to makefile to install jemalloc. The `make install` command will install jemalloc if it's not already installed.

The `make jemalloc` target currently fails on docker with `command sudo` not found. I don't have a solution for it.

**The change doesn't work on macOS or Windows. Dgraph builds with jemalloc by default on Linux only.**

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6463)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,9 @@ BUILD_BRANCH   ?= $(shell git rev-parse --abbrev-ref HEAD)
 BUILD_VERSION  ?= $(shell git describe --always --tags)
 
 MODIFIED = $(shell git diff-index --quiet HEAD || echo "-mod")
+HAS_JEMALLOC = $(shell ldconfig -p | grep jemalloc | wc -l)
+JEMALLOC_URL = "https://github.com/jemalloc/jemalloc/releases/download/5.2.1/jemalloc-5.2.1.tar.bz2"
+
 
 SUBDIRS = dgraph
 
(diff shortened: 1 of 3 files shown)


danielmai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6463)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jarifibrahim, @manishrjain, and @vvbalaji-dgraph)

___
*[dgraph/Makefile, line 98 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/6463#-MHMFZR--9YDDEB9zq2-:-MHMFZR--9YDDEB9zq20:b72mnco) ([raw file](https://github.com/dgraph-io/dgraph/blob/14f6243f43d35fabe6566325fe96fb219d8a47c4/dgraph/Makefile#L98)):*
> ```
> ldconfig ; \
> ```
Does this need `sudo` too? I get this error locally, and it happens in CI too.

```
$ ldconfig                   
/sbin/ldconfig.real: Can't create temporary cache file /etc/ld.so.cache~: Permission denied
```

Do we need to run ldconfig? Isn't `sudo make install` sufficient?

danielmai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6463)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @jarifibrahim, @manishrjain, and @vvbalaji-dgraph)

___
*[dgraph/Makefile, line 66 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/6463#-MHMIIxR2FwaH86-u_am:-MHMIIxSEP3RZhGEayOV:b-h15hh3) ([raw file](https://github.com/dgraph-io/dgraph/blob/14f6243f43d35fabe6566325fe96fb219d8a47c4/dgraph/Makefile#L66)):*
> ```
> ldconfig -p
> ```
We're using the static library (`libjemalloc.a`), not the shared library (`libjemalloc.so`), so we shouldn't need to use ldconfig, as it doesn't tell us about the .a file:

```
$ ldconfig -p | grep jemalloc
	libjemalloc.so.2 (libc6,x86-64) => /usr/local/lib/libjemalloc.so.2
	libjemalloc.so (libc6,x86-64) => /usr/local/lib/libjemalloc.so
```

We should check if the path to the archive exists: `/usr/local/lib/libjemalloc.a`

PR #6474, opened 2020-09-17
fix (graphql): disallowing field names with as

aman-bansal:
Signed-off-by: aman-bansal <bansalaman2905@gmail.com>
This is related to GRAPHQL-564. `as` is Dgraph reserved keyword. It's being used by DQL to identify variables. This PR is to restrict the naming for fields with the name `as`.

Though interesting things have been found while working on this. There is a difference in how we check `as` keyword while query parsing. For example, in normal DQL queries, the parser uses this `strings.ToLower(peekIt[0].Val) == "as"` (dgraph/gql/parser.go:2552) this means `as`, `As`, `AS`, `aS` all are the same.

But while parsing facets, its been treated like this `!trySkipItemVal(it, "as")` (dgraph/gql/parser.go:1984).

Though having a different use case, keyword significance is the same. Would love to hear thoughts about these implementations.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6474)

 
 
 
 
 

Docs Preview: [Dgraph Preview](https://dgraph-5d5088eacc-96554.surge.sh)

### DIFF
--- a/graphql/schema/gqlschema_test.yml
+++ b/graphql/schema/gqlschema_test.yml
@@ -2398,6 +2398,47 @@ invalid_schemas:
     {"message": "Subscription is a reserved word, so you can't declare a type with this name. Pick a different name for the type.", "locations": [{"line":1, "column":6}]},
     ]
 
+  -
+    name: "as is reserved keyword - type Name"
+    input: |
+      type As {
+        id: ID!
+        name: String
+      }
+    errlist: [
+      { "message": "As is a reserved word, so you can't declare a type with this name. Pick a different name for the type.", "locations": [ { "line": 1, "column": 6 } ] },
+    ]
+
+  - name: "as is reserved keyword - field name"
+    input: |
+      type X {
+        as: ID!
+        name: String
+      }
+    errlist: [
+      { "message": "Type X; Field as: as is a reserved keyword and you cannot declare a field with this name.", "locations": [ { "line": 2, "column": 3 } ] },
+    ]
+
+  - name: "as is reserved keyword - type name using @dgraph directive"
+    input: |
+      type X @dgraph(type:"as") {
+        id: ID!
+        name: String
+      }
+    errlist: [
+      { "message": "Type X; type argument 'as' for @dgraph directive is a reserved keyword.", "locations": [ { "line": 1, "column": 9 } ] },
+    ]
+
+  - name: "as is reserved keyword - field name using @dgraph directive"
+    input: |
+      type X {
+        id: ID!
+        name: String @dgraph(pred:"as")
+      }
+    errlist: [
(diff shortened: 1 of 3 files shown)


abhimanyusinghgaur (changes requested):
we also have a `@dgraph(type: "", pred: "")` directive, using which one can override the GraphQL generated name for any field or type. I think that also needs to have this restriction.

@ashish-goswami can say more about its usage in facets.

> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6474#-:-MI37hyK2OHOnbTBpD2s:b-z5ioah)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @aman-bansal, @danielmai, @MichaelJCompton, and @pawanrawal)

___
*[wiki/content/graphql/schema/reserved.md, line 10 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/6474#-MI372XJ62MJmBVubbe8:-MI372XJ62MJmBVubbe9:b-h22gl9) ([raw file](https://github.com/dgraph-io/dgraph/blob/15b8a20ddd72f2b17d5173e411fcbdf8d129476f/wiki/content/graphql/schema/reserved.md#L10)):*
> ```
> Names `Int`, `Float`, `Boolean`, `String`, `DateTime` and `ID` are reserved and cannot be used to define any other identifiers.
> Similarly, Names like `uid`, `Subscription`, `as`, `Query` and `Mutation` are also reserved and cannot be used to define any other Identifiers.
> ```
Maybe just combine the two lines, or state it all like:
```
The following names are reserved and can't be used to define any other identifiers:
* Int
* Float
...
* as/As/aS/AS
* Query
...
```

abhimanyusinghgaur (changes requested):
Nice work!

Sorry, I should have pointed it the last time itself, but we also need to add tests for the same. Have a look in `graphql/schema/gqlschema_test.yml` and just add `invalid_schemas` using `as` at all the different places, i.e., as a type name, field name, and the same cases with `@dgraph`.

> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6474#-:-MII7Ydw0WmuzGRisXmn:bmd1aqs)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @danielmai, @MichaelJCompton, and @pawanrawal)

abhimanyusinghgaur (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6474#-:-MIJ9Rsd9Z3cJgdidF2e:bnfp4nl)* status: 2 of 3 files reviewed, all discussions resolved (waiting on @abhimanyusinghgaur, @danielmai, @MichaelJCompton, and @pawanrawal)

abhimanyusinghgaur:
Actually, the target branch should be `master`.

PR #6485, opened 2020-09-17
fix(trie): Do not unmap buffer in recursive calls

jarifibrahim:
We were unmapping buffers between recursive calls which would invalidate
the existing objects (and their pointers) and any attempt to access the objects
would cause a sigfault.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6485)


jarifibrahim:
Closing this PR. We don't need it anymore.

PR #6497, opened 2020-09-18
Do not allow enabling compression without cache

jarifibrahim:
Alpha/zero should not be used if compression/encryption is enabled. This
PR adds checks so that we don't start alpha/zero if compression or
encryption is enabled.

**This PR also sets cache size for alpha to 2GB and disables compression for Zero.**

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6497)


jarifibrahim:
This has been fixed in badger. We don't need this PR.

PR #6540, opened 2020-09-21
fix(dgraph): Update alpha http handler to handle params in content-type header

### LINKED ISSUE #5174: Graphql : Failing on wrong content-type message

v20.03.0

Yes

Mac for client and running dgraph/standalone in a container

This code is making a graphql call using Machine-Box and it is failing with the following error.

https://play.golang.org/p/T6YVYIs3AOD

```
Unsupported Content-Type. Supported content types are application/json, application/graphql+-
```

This is not accurate since Machine-Box is setting the `content-type` properly on line 117. Machine-Box sets the `content-type` in the http call like this.

https://github.com/machinebox/graphql/blob/v0.2.2/graphql.go#L117
```
113 r, err := http.NewRequest(http.MethodPost, c.endpoint, &requestBody)
114 if err != nil {
115     return err
116 }
117 r.Header.Set("Content-Type", "application/json; charset=utf-8")
118 r.Header.Set("Accept", "application/json; charset=utf-8")
119 for key, values := range req.Header {
120     for _, value := range values {
121         r.Header.Add(key, value)
122     }
123 }
124 c.logf(">> headers: %v", r.Header)
```

As you can see, the `content-type` is correct. Adding the charset should not cause Dgraph to fail.

The code provided should work.  The code inside of dgraph that is not parsing the `content-type` correctly can be found here.

https://github.com/dgraph-io/dgraph/blob/master/dgraph/cmd/alpha/http.go#L179

rahulgurnani:
This fixes #5174 

Earlier, if we send "Content-Type" as "application/json; charset=utf8" query failed. The content-type http header allows params and our code didn't take those into account.
Docs for Content-Type header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type

Added dependency to gddo/httputil/header package which has a utility to parse the headers. 
Also added a test for this change.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6540)


parasssh:
Branch has conflicts. Please merge and update your PR.
Also, should we be doing this across the entire code base? Seems it'd be good for any HTTP handler.

rahulgurnani:
#6370 already addresses this issue

PR #6556, opened 2020-09-22
feat(query): Support for between func with count at root

ashish-goswami:
---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6556)

 

Docs Preview: [Dgraph Preview](https://dgraph-9c6b5d07ee-96161.surge.sh)


worker/task.go
```diff
@@ -2166,27 +2176,33 @@ func preprocessFilter(tree *pb.FilterTree) (*facetsTree, error) {
...
-		illegal = count <= 0
+		illegal = countl <= 0
 	case "gt":
-		illegal = count < 0
+		illegal = countl < 0
 	case "ge":
-		illegal = count <= 0
+		illegal = countl <= 0
+	case "between":
+		illegal = countl <= 0 || counth <= 0
```
parasssh, line 2205: another check could be countl > counth

worker/task.go
```diff
@@ -2234,6 +2250,15 @@ func (qs *queryState) evaluate(cp countParams, out *pb.Result) error {
 	for itr.Seek(countKey); itr.Valid(); itr.Next() {
 		item := itr.Item()
 		var key []byte
+		key = item.KeyCopy(key)
+		k, err := x.Parse(key)
+		if err != nil {
+			return err
+		}
+		if cp.fn == between && int64(k.Count) > counth {
+			break
```
parasssh, line 2259: I thought we will continue here to the next item and not break, no?

worker/task.go
ashish-goswami: We are not returning any error in case of improper bounds. Returned response would be empty.

worker/task.go
ashish-goswami: We need to break here as soon as we cross the upper bound in case of between.

worker/task.go
parasssh: I think we should just check that case here as we are already checking other illegal cases.

worker/task.go
parasssh: Is it because all future keys count will be higher. (i.e keys are sorted?)

worker/task.go
ashish-goswami: Agreed. But the point is not to return error from here, to have between behaviour consistent.

worker/task.go
ashish-goswami: yes.

parasssh (approved):

PR #6560, opened 2020-09-23
Initialize posting list in moveToNextPart

animesh2049:
If `deleteBelowTs` is not zero `it.uidPosting` won't be intialized in `pitr.init()`. This could cause panic when we are doing `pitr.posting()`. Our assumption was that `piter.valid()` would return false in such cases. This assumption might not always be true. If `it.uids` is empty, we try to move to next valid part of the list. In the function `moveToNextPart` we are intitializing many components of pIterator but not `uidPosting`. If `it.uids` is filled here, `piter.valid()` would become true and we
would end up into a panic at `pitr.posting()`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6560)

### DIFF
--- a/posting/list.go
+++ b/posting/list.go
@@ -184,6 +184,7 @@ func (it *pIterator) moveToNextPart() error {
 	}
 	it.plist = plist
 
+	it.uidPosting = &pb.Posting{}
 	it.dec = &codec.Decoder{Pack: it.plist.Pack}
 	// codec.SeekCurrent makes sure we skip returning afterUid during seek.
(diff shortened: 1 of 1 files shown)


martinmr (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6560#-:-MHvsLI21WvIEM6qT8fR:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @manishrjain and @martinmr)

parasssh (approved):

PR #6596, opened 2020-09-29
fix(GraphQL Query): Remove extra fields when querying interfaces

anurags92:
### Motivation
For queries like 
```
 queryCharacter {
	name
	... on Human {
		name
		totalCredits
	}
	... on Droid {
		name
		primaryFunction
	}
}
```
Dgraph response includes the field `name` multiple times. This change fixes that by keeping a `map` of the fields already seen and removing them from the response.

This PR fixes: GRAPHQL-638

Query response for query on fragments in GraphQL

Yes

### Testing

Added following tests in:

In `fragments.go` tests are added to check that repeated fields are not present in the response and order is preserved.

### Fixes 
Fixes GRAPHQL-638
https://github.com/99designs/gqlgen/issues/1300

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6596)

### DIFF
--- a/graphql/e2e/common/fragment.go
+++ b/graphql/e2e/common/fragment.go
@@ -125,51 +125,6 @@ func fragmentInQuery(t *testing.T) {
 	cleanupStarwars(t, result.QueryStarship[0].ID, "", "")
 }
 
-/*
-qcRep1: queryCharacter {
-				name
-				... on Human {
-					name
-					totalCredits
-				}
-				... on Droid {
-					name
-					primaryFunction
-				}
-			}
-			qcRep2: queryCharacter {
-				... on Human {
-					totalCredits
-				}
-				name
-				... on Droid {
-					primaryFunction
-				}
-			}
-*/
-
-/*
-"qcRep1": [
-            {
-				"name": "Han"
-                "totalCredits": 10,
-            },
-            {
-                "name": "R2-D2",
-                "primaryFunction": "Robot"
-            }
-		],
-		"qcRep2": [
-            {
-                "totalCredits": 10,
-                "name": "Han"
-            },
-            {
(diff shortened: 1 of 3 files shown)


abhimanyusinghgaur (approved):
:lgtm: 

I don't think it breaks backward compatibility. It was a bug, so shouldn't be considered as functionality for being backward compatible.

> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6596#-:-MISnrPa0rcZFE2wUlTs:b-pi67de)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @anurags92, @MichaelJCompton, and @pawanrawal)

___
*[graphql/resolve/resolver.go, line 1272 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/6596#-MISoaQP-6OHCwHZ8cDa:-MISoaQP-6OHCwHZ8cDb:b-g2zk5m) ([raw file](https://github.com/dgraph-io/dgraph/blob/6ecf657e457b92d42f4efe0bf114a7992398ebdf/graphql/resolve/resolver.go#L1272)):*
> ```
> Below map keeps track of fields which have been seen as part of
> 	// interface to avoid double entry in the resulting response
> ```
we should put a more elaborative comment for this edge case:
```
seenField is used to keep track of fields that have been included in the response.
It helps in avoiding double entry in the response for fields that are queried twice at the same level, once under an interface and second time under a fragment on some object type implementing that interface. For example, consider `name` in the following query:
query {
  queryCharacter {
    ... on Human {
      totalCredits
    }
    name
    ... on Droid {
      name
      primaryFunction
    }
  }
}
```

PR #6634, opened 2020-10-02
Pagination param "after" does not work when using func: uid(v) (#6365)

antblood:
Pagination param "after" does not work when using func: uid(v)

for query
```
{
  var(func: has(name@en), first: 50) {
    test as uid
  }

  ss(func: uid(test), first: 2, after: 0x6) {
    uid
  }
}
```

the response was

```
{
  "data": {
    "ss": [
      {
        "uid": "0x4"
      },
      {
        "uid": "0x6"
      }
    ]
  }
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6631)

 

Docs Preview: [Dgraph Preview](https://dgraph-70fd76f0af-98614.surge.sh)

### DIFF
--- a/query/query.go
+++ b/query/query.go
@@ -1982,6 +1982,11 @@ func ProcessGraph(ctx context.Context, sg, parent *SubGraph, rch chan error) {
 				return sg.DestUIDs.Uids[i] < sg.DestUIDs.Uids[j]
 			})
 		}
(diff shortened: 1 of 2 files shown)


jarifibrahim (approved):
I don't understand the code but it's a cherry-pick so LGTM

PR #6652, opened 2020-10-05
fix(TestPlugins) Increase sleep timer to prevent connection retries on sleeping cluster

anurags92:
### Motivation

TestPlugins is flaky in because it sometimes tries to connect to the cluster while the cluster is still not up. 

This PR fixes: DGRAPH-1637

TestPlugins, Systest

No

### Testing
N/A

### Fixes 

Fixes DGRAPH-1637

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6652)

### DIFF
--- a/systest/cluster_setup_test.go
+++ b/systest/cluster_setup_test.go
@@ -82,7 +82,7 @@ func (d *DgraphCluster) StartZeroOnly() error {
 	}
 
 	// Wait for dgraphzero to start listening and become the leader.
-	time.Sleep(time.Second * 4)
+	time.Sleep(time.Second * 6)
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


parasssh (commented):
Are we sure it is only a timing issue and that increasing the timeout is enough to fix this?

anurags92:
> Are we sure it is only a timing issue and that increasing the timeout is enough to fix this?

If you see the test failures across few months in past, all the errors are: `bulk_live_fixture_test.go:240: rpc error: code = Unknown desc = Please retry again, server is not ready to accept requests`. It is difficult to reproduce this since doesn't happen locally. Updating the sleep-time, will revert and check again if the problem persists.

parasssh (approved):

PR #6691, opened 2020-10-09
feat(dgraph): enabling TLS config in http zero

aman-bansal:
Enabling TLS based encryption in Dgraph zero. 
Fixes https://dgraph.atlassian.net/browse/DGRAPH-1366. Based on discussion here https://discuss.dgraph.io/t/enabling-tls-configuration-for-internal-ports-making-dgraph-more-secure/10860/2
The same variables are used which needs to be defined for alpha tls configuration

Things to consider:
1. With TLS enabled, tls_enabled_route has to be defined.
2. If defined all routes will be accessible with TLS. But ones mentioned in --tls_enabled_route will not be accessible via HTTP.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6691)

 
 

Docs Preview: [Dgraph Preview](https://dgraph-93b2fd387f-103012.surge.sh)

### DIFF
--- a/dgraph/cmd/zero/http.go
+++ b/dgraph/cmd/zero/http.go
@@ -246,9 +246,18 @@ func (st *state) serveHTTP(l net.Listener) {
 		IdleTimeout:  2 * time.Minute,
 	}
 
+	tlsCfg, err := x.LoadServerTLSConfig(Zero.Conf, "node.crt", "node.key")
+	x.Check(err)
+
 	go func() {
 		defer st.zero.closer.Done()
-		err := srv.Serve(l)
+		switch {
+		case tlsCfg != nil:
+			srv.TLSConfig = tlsCfg
+			err = srv.ServeTLS(l, "", "")
+		default:
+			err = srv.Serve(l)
+		}
 		glog.Errorf("Stopped taking more http(s) requests. Err: %v", err)
 		ctx, cancel := context.WithTimeout(context.Background(), 630*time.Second)
 		defer cancel()
--- a/dgraph/cmd/zero/http.go
+++ b/dgraph/cmd/zero/http.go
@@ -243,7 +243,7 @@ func (st *state) pingResponse(w http.ResponseWriter, r *http.Request) {
 	_, _ = w.Write([]byte("OK"))
 }
 
-func (st *state) serveHTTP(l net.Listener) {
+func (st *state) startListenHttpAndHttps(l net.Listener) {
 	m := cmux.New(l)
 	startServers(m)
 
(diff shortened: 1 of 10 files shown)


pawanrawal (changes requested):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6691)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @aman-bansal, @manishrjain, and @vvbalaji-dgraph)

___
*[dgraph/cmd/zero/http.go, line 246 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/6691#-MJlyqR539X7s2VSlEOy:-MJlyqR539X7s2VSlEOz:b1z6xhf) ([raw file](https://github.com/dgraph-io/dgraph/blob/83d2cf0225cd38f1fe3552a3ac2236208fd1b218/dgraph/cmd/zero/http.go#L246)):*
> ```Go
> }
> 
> func (st *state) serveHTTP(l net.Listener) {
> ```

should be renamed as it serves both HTTP and HTTPS
___
*[dgraph/cmd/zero/http.go, line 278 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/6691#-MJm1ucP7ZCsgQ6ezRBr:-MJm1ucP7ZCsgQ6ezRBs:bbr43gz) ([raw file](https://github.com/dgraph-io/dgraph/blob/83d2cf0225cd38f1fe3552a3ac2236208fd1b218/dgraph/cmd/zero/http.go#L278)):*
> ```Go
> 
> func startServers(m cmux.CMux) {
> 	// if tls enabled http rule will check that the requested route should not be part of tls_enabled_routes
> ```

The naming needs to be fixed here.

parasssh:
-all tests  for the combination of tls config  (on/off) and enabled routes  (0 , all, some).
- bulk loader system test with and without TLS. 
- System test with alpha, zero and using Ratel. Full TLS enabled.

parasssh (approved):

aman-bansal:
Both bulk live or other clients connect with GRPC port of zero. This change doesn't affect any of those. We already have tests running to check those features.

aman-bansal:
I changed the enable flag to disable because there are many endpoints like /debug which isn't initiated directly. We can introduce some side effects with negation logic. Also this align more with manish comment

```
Another thing to do would be to have a flag, which can tell the HTTP endpoints which can be plain-text. For e.g., “/health, /debug”, etc. That way, a very secure usage can keep them empty. But, selectively open some to easier `debugging.
````

pawanrawal (approved):
> Reviewed 4 of 10 files at r1, 4 of 6 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6691)* status: all files reviewed, 4 unresolved discussions (waiting on @aman-bansal, @manishrjain, and @vvbalaji-dgraph)

___
*[dgraph/cmd/zero/run.go, line 97 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/6691#-MKAgialB8NFMbm_3i-m:-MKAgialB8NFMbm_3i-n:b-ryvl5j) ([raw file](https://github.com/dgraph-io/dgraph/blob/a228342bb579fd3fa3cae4ec5680104a836c90b3/dgraph/cmd/zero/run.go#L97)):*
> ```Go
> 	flag.Bool("tls_use_system_ca", true, "Include System CA into CA Certs.")
> 	flag.String("tls_client_auth", "VERIFYIFGIVEN", "Enable TLS client authentication")
> 	flag.String("tls_disabled_route", "", "comma separated zero endpoint which will disabled from TLS encryption."+
> ```

which will be .... 
___
*[tlstest/zero_https/no_tls/no_tls_test.go, line 58 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/6691#-MKAhIVs7Z59c1l8wm47:-MKAhIVs7Z59c1l8wm48:b-rnxd9f) ([raw file](https://github.com/dgraph-io/dgraph/blob/a228342bb579fd3fa3cae4ec5680104a836c90b3/tlstest/zero_https/no_tls/no_tls_test.go#L58)):*
> ```Go
> 	defer func() { _ = do.Body.Close() }()
> 	body, err := ioutil.ReadAll(do.Body)
> 	if err != nil {
> ```

require.NoError(t, err)

PR #6694, opened 2020-10-09
fix(config): Set glog -v flag correctly from config files.

danielmai:
---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6694)

 
 

Docs Preview: [Dgraph Preview](https://dgraph-0529b2d193-107199.surge.sh)

### DIFF
--- a/dgraph/cmd/root.go
+++ b/dgraph/cmd/root.go
@@ -157,18 +157,17 @@ func setGlogFlags(conf *viper.Viper) {
 	}
 	for _, gflag := range glogFlags {
 		// Set value of flag to the value in config
-		if stringValue, ok := conf.Get(gflag).(string); ok {
-			// Special handling for log_backtrace_at flag because the flag is of
-			// type tracelocation. The nil value for tracelocation type is
-			// ":0"(See https://github.com/golang/glog/blob/master/glog.go#L322).
-			// But we can't set nil value for the flag because of
-			// https://github.com/golang/glog/blob/master/glog.go#L374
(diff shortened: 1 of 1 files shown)


dgraph/cmd/root.go
```diff
@@ -157,18 +157,17 @@ func setGlogFlags(conf *viper.Viper) {
...
-			// type tracelocation. The nil value for tracelocation type is
-			// ":0"(See https://github.com/golang/glog/blob/master/glog.go#L322).
-			// But we can't set nil value for the flag because of
-			// https://github.com/golang/glog/blob/master/glog.go#L374
-			// Skip setting value if log_backstrace_at is nil in config.
-			if gflag == "log_backtrace_at" && stringValue == ":0" {
-				continue
-			}
-			x.Check(flag.Lookup(gflag).Value.Set(stringValue))
+		stringValue := conf.GetString(gflag)
```
jarifibrahim, line 160: @danielmai why is this change needed? it looks like we aren't changing anything.

danielmai: This change reads all glog config values as strings and then sets them accordingly. Before, we were only setting values that could correctly be cast as a string (`conf.Get(gflag).(string)`). Any non-string values were ignored. So, even though `{"v": 2}` (an int) is a valid setting, it wouldn't be set. It'd have to be `{"v": "2"}` (a string) to work. This change makes both configs (int or string) work.

Calling `GetString` interprets the value as a string.

PR #6704, opened 2020-10-12
[Stale]Make ACL login error message more generic

animesh2049:
In case of invalid login request we don't reveal info about user. So make error message more generic
Fixes DGRAPH-2468

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6704)

### DIFF
--- a/edgraph/access_ee.go
+++ b/edgraph/access_ee.go
@@ -149,13 +149,16 @@ func (s *Server) authenticateLogin(ctx context.Context, request *api.LoginReques
 	// authorize the user using password
 	user, err := authorizeUser(ctx, request.Userid, request.Password)
 	if err != nil {
+		glog.Error(err)
 		return nil, invalidLogin
 	}
 
 	if user == nil {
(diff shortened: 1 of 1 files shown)


edgraph/access_ee.go
```diff
@@ -145,19 +145,17 @@ func (s *Server) authenticateLogin(ctx context.Context, request *api.LoginReques
 	}
 
 	// authorize the user using password
-	var err error
-	user, err = authorizeUser(ctx, request.Userid, request.Password)
+	invalidLogin := errors.New("Invalid username or password")
+	user, err := authorizeUser(ctx, request.Userid, request.Password)
 	if err != nil {
-		return nil, errors.Wrapf(err, "while querying user with id %v",
```
jarifibrahim, line 151: Actually, I would return the same error here and log it as glog.Error() and then return the invalidLogin error. This was the alpha logs would have information about why login failed but the user wouldn't know about it.

PR #6717, opened 2020-10-13
chore(compose): Set --idx for Alphas

jarifibrahim:
If the compose tool doesn't set the idx for alphas, they will be assigned IDs automatically by raft and that leads to confusion. This PR fixes it by explicitly assigning id to all the alphas.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6717)


jarifibrahim:
We don't need this PR anymore since there is an issue with how zero processes the idx flag.

PR #6788, opened 2020-10-22
feat(Query): Enable persistent queries in dgraph

anurags92:
### Motivation
Enable persistent queries as discussed in [RFC on discuss.](https://discuss.dgraph.io/t/rfc-persistent-queries-and-cached-results/10636)

This PR fixes: GRAPHQL-721

This feature works out-of-the-box with Apollo client for GraphQL

Query package

No

### Testing

Added following tests in:

-  `persistedQuery` in `common.go`. It is an end-to-end test which checks following cases:

     1. If incorrect sha is given throw with `provided sha doesn't match the query`
     2. If only sha is given and it is not present in dgraph throw with `PersistedQueryNotFound`
     3. If both query and sha are given and they match, then execute query and persis the query-sha pair. Check for no error.
     4. Use sha provided in `step iii` to execute the query and check that it throws no error.
     5. Do `step iv` via a `GET` request

### Fixes 
Fixes GRAPHQL-721

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6788)

### DIFF
--- a/edgraph/graphql.go
+++ b/edgraph/graphql.go
@@ -175,16 +175,26 @@ func ProcessPersistedQuery(ctx context.Context, gqlReq *schema.Request) error {
 		return nil
 	}
 
-	queryForSHA := fmt.Sprintf(`query{
-			me(func: eq(dgraph.graphql.p_sha256hash, %s)){
-				dgraph.graphql.p_query
-			}
-		}`, sha256Hash)
+	queryForSHA := `query Me($sha: string){
+						me(func: eq(dgraph.graphql.p_sha256hash, $sha)){
+							dgraph.graphql.p_query
+						}
+					}`
+	variables := map[string]string{
+		"$sha": sha256Hash,
+	}
 	req := &api.Request{
 		Query:    queryForSHA,
+		Vars:     variables,
 		ReadOnly: true,
 	}
-	storedQuery, _ := (&Server{}).doQuery(context.WithValue(ctx, IsGraphql, true), req, NoAuthorize)
+
+	storedQuery, err := (&Server{}).doQuery(ctx, req, NoAuthorize)
+
+	if err != nil {
+		glog.Errorf("Error while querying sha %s", sha256Hash)
+		return err
+	}
 
 	type shaQueryResponse struct {
 		Me []struct {
@@ -193,8 +203,10 @@ func ProcessPersistedQuery(ctx context.Context, gqlReq *schema.Request) error {
 	}
 
 	shaQueryRes := &shaQueryResponse{}
-	if err := json.Unmarshal(storedQuery.Json, shaQueryRes); err != nil {
-		return err
+	if len(storedQuery.Json) > 0 {
+		if err := json.Unmarshal(storedQuery.Json, shaQueryRes); err != nil {
+			return err
+		}
 	}
 
 	if len(shaQueryRes.Me) == 0 {
--- a/edgraph/graphql.go
+++ b/edgraph/graphql.go
@@ -171,7 +171,7 @@ func ProcessPersistedQuery(ctx context.Context, gqlReq *schema.Request) error {
 	query := gqlReq.Query
 	sha256Hash := gqlReq.Extensions.PersistedQuery.Sha256Hash
 
-	if len(sha256Hash) == 0 {
+	if sha256Hash == "" {
 		return nil
 	}
 
@@ -210,10 +210,12 @@ func ProcessPersistedQuery(ctx context.Context, gqlReq *schema.Request) error {
 	}
 
 	if len(shaQueryRes.Me) == 0 {
-		if len(query) == 0 {
+		if query == "" {
 			return errors.New("PersistedQueryNotFound")
 		}
-		if !hashMatches(query, sha256Hash) {
+		if match, err := hashMatches(query, sha256Hash); err != nil {
+			return err
+		} else if !match {
 			return errors.New("provided sha does not match query")
 		}
 
@@ -261,9 +263,12 @@ func ProcessPersistedQuery(ctx context.Context, gqlReq *schema.Request) error {
 
 }
 
-func hashMatches(query, sha256Hash string) bool {
+func hashMatches(query, sha256Hash string) (bool, error) {
 	hasher := sha256.New()
-	hasher.Write([]byte(query))
+	_, err := hasher.Write([]byte(query))
+	if err != nil {
+		return false, err
+	}
 	hashGenerated := hex.EncodeToString(hasher.Sum(nil))
-	return hashGenerated == sha256Hash
+	return hashGenerated == sha256Hash, nil
 }
--- a/edgraph/graphql.go
+++ b/edgraph/graphql.go
@@ -19,13 +19,17 @@ package edgraph
 import (
 	"bytes"
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"time"
 
 	"github.com/dgraph-io/dgo/v200/protos/api"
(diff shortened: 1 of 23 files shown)


x/keys.go
```diff
@@ -556,11 +556,12 @@ var internalPredicateMap = map[string]struct{}{
...
-	"dgraph.type.User":       {},
-	"dgraph.type.Group":      {},
-	"dgraph.type.Rule":       {},
-	"dgraph.graphql.history": {},
+	"dgraph.graphql":                 {},
+	"dgraph.type.User":               {},
+	"dgraph.type.Group":              {},
+	"dgraph.type.Rule":               {},
+	"dgraph.graphql.history":         {},
+	"dgraph.graphql.persisted_query": {},
```
gja, line 564: this may be too much, check with pawan

schema/schema.go
```diff
@@ -678,6 +689,16 @@ func initialSchemaInternal(all bool) []*pb.SchemaUpdate {
 		}, &pb.SchemaUpdate{
 			Predicate: "dgraph.graphql.schema_created_at",
 			ValueType: pb.Posting_DATETIME,
+		}, &pb.SchemaUpdate{
+			Predicate: "dgraph.graphql.p_query",
+			ValueType: pb.Posting_STRING,
+			Directive: pb.SchemaUpdate_INDEX,
```
gja, line 695: no need for index here

edgraph/graphql.go
```diff
@@ -151,3 +155,103 @@ func UpdateSchemaHistory(ctx context.Context, schema string) error {
...
+func ProcessPersistedQuery(ctx context.Context, gqlReq *schema.Request) error {
+	query := gqlReq.Query
+	sha256Hash := gqlReq.Extensions.PersistedQuery.Sha256Hash
+
+	if len(sha256Hash) == 0 {
+		return nil
+	}
+
+	queryForSHA := fmt.Sprintf(`query{
+			me(func: eq(dgraph.graphql.p_sha256hash, %s)){
```
gja, line 179: don't interpolate here, use variables

edgraph/graphql.go
```diff
@@ -151,3 +155,103 @@ func UpdateSchemaHistory(ctx context.Context, schema string) error {
...
+	shaQueryRes := &shaQueryResponse{}
+	if err := json.Unmarshal(storedQuery.Json, shaQueryRes); err != nil {
+		return err
+	}
+
+	if len(shaQueryRes.Me) == 0 {
+		if len(query) == 0 {
+			return errors.New("PersistedQueryNotFound")
+		}
+		if !hashMatches(query, sha256Hash) {
```
gja, line 204: should we do this at the top of the function? Either way

graphql/schema/request.go
```diff
@@ -32,10 +32,18 @@ type Request struct {
 	Query         string                 `json:"query"`
 	OperationName string                 `json:"operationName"`
 	Variables     map[string]interface{} `json:"variables"`
+	Extensions    struct {
```
gja, line 35: Leave a comment here, that we should add explicit types for all extensions that we support in dgraph

graphql/web/http.go
```diff
@@ -253,6 +259,13 @@ func getRequest(ctx context.Context, r *http.Request) (*schema.Request, error) {
 		query := r.URL.Query()
 		gqlReq.Query = query.Get("query")
 		gqlReq.OperationName = query.Get("operationName")
+		if extensions, ok := query["extensions"]; ok {
```
gja, line 262: check length(extensions) before json decoding

gja (approved):

abhimanyusinghgaur (changes requested):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6788)* status: 0 of 6 files reviewed, 8 unresolved discussions (waiting on @anurags92, @gja, @manishrjain, @MichaelJCompton, @pawanrawal, and @vvbalaji-dgraph)

___
*[graphql/admin/admin.go, line 291 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/6788#-MKJJ88e18A2y4FLbhdz:-MKJJ88fBAGx1se1HWCv:b-7f4lyi) ([raw file](https://github.com/dgraph-io/dgraph/blob/5f02c6f3d1b1d8c68f14ef861fcd0f17e240fbc2/graphql/admin/admin.go#L291)):*
> ```Go
> 		getAllowedCORSOrigins: Cors
> 		querySchemaHistory(first: Int, offset: Int): [SchemaHistory]
> 		queryPersistedQuery: PersistedQuery
> ```

I am not sure, but should this query have IP whitelisting and Guardian only access (if ACL is enabled)?
___
*[graphql/schema/request.go, line 35 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/6788#-MKHETYYKR255qnEW61S-r1-35:-MKJK6JQ19cPXPCEPdb9:bwjvyl7) ([raw file](https://github.com/dgraph-io/dgraph/blob/5f02c6f3d1b1d8c68f14ef861fcd0f17e240fbc2/graphql/schema/request.go#L35)):*
<details><summary><i>Previously, gja (Tejas Dinkar) wrote…</i></summary><blockquote>

Leave a comment here, that we should add explicit types for all extensions that we support in dgraph
</blockquote></details>

This is the first extension that comes in GraphQL request, we already have explicit types for the ones which go in GraphQL response.
But, what we should do is mark this a separate type called `RequestExtensions` instead of saying `struct { ... }`
___
*[graphql/web/http.go, line 219 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/6788#-MKJLNrBDYNnB-TFb3gh:-MKJLNrBDYNnB-TFb3gi:b-l9s4it) ([raw file](https://github.com/dgraph-io/dgraph/blob/5f02c6f3d1b1d8c68f14ef861fcd0f17e240fbc2/graphql/web/http.go#L219)):*
<details><summary><i alt="summary">Quoted 9 lines of code…</i></summary>

> ```
> 	if err != nil {
> 		write(w, schema.ErrorResponse(err), strings.Contains(r.Header.Get("Accept-Encoding"), "gzip"))
> 		return
> 	}
> 	if err = edgraph.ProcessPersistedQuery(ctx, gqlReq); err != nil {
> 		write(w, schema.ErrorResponse(err), strings.Contains(r.Header.Get("Accept-Encoding"), "gzip"))
> 		return
> 	}
> 	res = gh.resolver.Resolve(ctx, gqlReq)
> ```
</details>

I believe we should just extended the `if else` logic instead of doing multiple returns and code duplication.
___
*[schema/schema.go, line 695 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/6788#-MKHETYYKR255qnEW61U-r1-695:-MKJNLV68AC9sBkM_ii5:b32v57) ([raw file](https://github.com/dgraph-io/dgraph/blob/5f02c6f3d1b1d8c68f14ef861fcd0f17e240fbc2/schema/schema.go#L695)):*
<details><summary><i>Previously, gja (Tejas Dinkar) wrote…</i></summary><blockquote>

no need for index here
</blockquote></details>

+1
___
*[x/keys.go, line 564 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/6788#-MKHETYYKR255qnEW61V-r1-564:-MKJMm1N1F9SOF7EiqBS:bo0duw5) ([raw file](https://github.com/dgraph-io/dgraph/blob/5f02c6f3d1b1d8c68f14ef861fcd0f17e240fbc2/x/keys.go#L564)):*
<details><summary><i>Previously, gja (Tejas Dinkar) wrote…</i></summary><blockquote>

this may be too much, check with pawan
</blockquote></details>

I think, this one should be there. There are various checks which happen for these pre-defined things at a lot of places.

PR #6833, opened 2020-11-03
chore(bulk): use --tmp directory for temporary buffers

NamanJain8:
Fixes DGRAPH-2617

The bulk loader uses temporary buffers for map-reduce logic. These temporary buffers are created inside `/tmp` by default. 
User may not have a that big `/tmp` directory and moreover we have a `--tmp` flag for temporary files in bulk loader. It is expected that we use that directory. So, we should use that instead.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6833)

 

Docs Preview: [Dgraph Preview](https://dgraph-a31868e640-107618.surge.sh)


jarifibrahim (commented):
We should delete the buffers after the bulk loader restarts.

pawanrawal (approved):

PR #6836, opened 2020-11-04
fix(query): Limit the results in hasquery

rahulgurnani:
---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6836)

### DIFF
--- a/worker/task.go
+++ b/worker/task.go
@@ -19,6 +19,7 @@ package worker
 import (
 	"bytes"
 	"context"
+	"math"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


worker/task.go
```diff
@@ -2344,12 +2346,16 @@ func (qs *queryState) handleHasFunction(ctx context.Context, q *pb.Query, out *p
...
 	}
-
+	nodesVisited := 0
 loop:
 	// This function could be switched to the stream.Lists framework, but after the change to use
 	// BitCompletePosting, the speed here is already pretty fast. The slowdown for @lang predicates
 	// occurs in filterStringFunction (like has(name) queries).
 	for it.Seek(startKey); it.Valid(); {
+		if nodesVisited > maxNodesToVisitForHasQuery {
+			break
```
rahulgurnani, line 2356: Note: We want to bound the time taken by the has query.

pawanrawal (approved):
LGTM though we should verify if this solves Emerson's issue.

pawanrawal (commented):
Another problem is that to find out the number of nodes that have a predicate, the user would now have to give a very large value which can be irritating.

PR #6838, opened 2020-11-04
fix(trie): Fix trie after ristretto update

ahsanbarkati:
The AllocateOffset API in z.Buffer used to return offset which included padding. This padding is internal to z.Buffer and should not be included in the offset returned by this API. In the latest ristretto, the AllocateOffset returns the logical offset (excluding the padding). This change make trie compatible to the change in the AllocateOffset. We can't keep the root node at zero offset because zero offset is used for the check for not-existent nodes. So, we pad the root node by 8 bytes.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6838)

 
 
 

Docs Preview: [Dgraph Preview](https://dgraph-db0bbd287c-107460.surge.sh)


jarifibrahim (approved):
Looks good to me.

PR #6870, opened 2020-11-09
fix(GraphQL): fix exclusion of filters in Query generation in case of just Has Filters.

minhaj-shakeel:
Fixes GRAPHQL-787.

This PR fixes the Bug in query and mutations generation in case of types that have just `has` filter and not any other kind of filter. For example for the given type: 
```
type Student {
   name: String
   age: Int
}
```
the corresponding query generated was 
```
queryStudent(order: StudentOrder, first: Int, offset: Int): [Student]
```
and only `add` mutation was being generated.
Whereas, the correct query should be: 
```
queryStudent(filter: StudentFilter, order: StudentOrder, first: Int, offset: Int): [Student]
```
and corresponding update and delete mutation should be:
```
updateStudent(input: UpdateStudentInput!): UpdateStudentPayload
deleteStudent(filter: StudentFilter!): DeleteStudentPayload
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6870)

### DIFF
--- a/graphql/resolve/query_test.yaml
+++ b/graphql/resolve/query_test.yaml
@@ -775,6 +775,20 @@
       }
     }
 
+- name: "Query Has Filter on type which has neither ID field nor any search argument"
+  gqlquery: |
+    query {
+      queryNode(filter: {has: name}){
+        name
+      }
+    }
+  dgquery: |-
+    query {
+      queryNode(func: type(Node)) @filter(has(Node.name)) {
+        name : Node.name
+        dgraph.uid : uid
+      }
+    }
 -
   name: "Filters in same input object implies AND"
   gqlquery: |
--- a/graphql/schema/gqlschema.go
+++ b/graphql/schema/gqlschema.go
@@ -892,7 +892,7 @@ func cleanupInput(sch *ast.Schema, def *ast.Definition, seen map[string]bool) {
 	}
 	def.Fields = def.Fields[:i]
 
-	if len(def.Fields) == 0 {
+	if len(def.Fields) == 0 || (strings.HasPrefix(def.Name, "Update") && len(def.Fields) == 1) {
 		delete(sch.Types, def.Name)
 	}
 }
@@ -902,7 +902,7 @@ func cleanSchema(sch *ast.Schema) {
 	// don't have field inside them.
 	for k := range sch.Types {
 		if strings.HasSuffix(k, "Ref") || strings.HasSuffix(k, "Patch") ||
-			(strings.HasPrefix(k, "Add") && strings.HasSuffix(k, "Input")) {
+			((strings.HasPrefix(k, "Add") || strings.HasPrefix(k, "Update")) && strings.HasSuffix(k, "Input")) {
 			cleanupInput(sch, sch.Types[k], map[string]bool{})
 		}
 	}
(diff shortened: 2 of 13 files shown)


pawanrawal (changes requested):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6870)* status: 0 of 13 files reviewed, 3 unresolved discussions (waiting on @MichaelJCompton, @minhaj-shakeel, and @pawanrawal)

___
*[graphql/schema/gqlschema.go, line 895 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/6870#-MMQAwRu5ha2p9E2k5dI:-MMQAwRu5ha2p9E2k5dJ:b-2yyfny) ([raw file](https://github.com/dgraph-io/dgraph/blob/d8d9dd040d9e4fa957d352a0443a3cba58d35110/graphql/schema/gqlschema.go#L895)):*
> ```Go
> 	def.Fields = def.Fields[:i]
> 
> 	if len(def.Fields) == 0 || (strings.HasPrefix(def.Name, "Update") && len(def.Fields) == 1) {
> ```

When does this happen, is it when Update only has filter but no set or remove? Can we just not generate the Update field in this case?
___
*[graphql/schema/gqlschema.go, line 1167 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/6870#-MMQB9bHBWjPR4v1XnBc:-MMQB9bI-J3rchW57DHC:b-j9q99m) ([raw file](https://github.com/dgraph-io/dgraph/blob/d8d9dd040d9e4fa957d352a0443a3cba58d35110/graphql/schema/gqlschema.go#L1167)):*
> ```Go
> func addFilterArgumentForField(schema *ast.Schema, fld *ast.FieldDefinition, fldTypeName string) {
> 	// Don't add filters for inbuilt types.
> 	if _, ok := inbuiltTypeToDgraph[fldTypeName]; ok {
> ```

This wasn't required before but is required now, why is that the case?
___
*[graphql/schema/gqlschema.go, line 1353 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/6870#-MMQASoB98D7tW83nJ0_:-MMQASoB98D7tW83nJ0a:b-nb1ggj) ([raw file](https://github.com/dgraph-io/dgraph/blob/d8d9dd040d9e4fa957d352a0443a3cba58d35110/graphql/schema/gqlschema.go#L1353)):*
> ```Go
> // hasFilterable Returns whether TypeFilter for a defn will be generated or not.
> // It returns true if any field have search arguments or it is an `ID` field or
> // HasFilter for that defn is generated.
> ```

or there is at least one non-custom filter which would be part of the HasFilter

pawanrawal (changes requested):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6870)* status: all files reviewed, 3 unresolved discussions (waiting on @MichaelJCompton and @minhaj-shakeel)

___
*[graphql/schema/gqlschema.go, line 895 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/6870#-MMQAwRu5ha2p9E2k5dI:-MMQCmSvCcD1MUS2kOvc:biad72j) ([raw file](https://github.com/dgraph-io/dgraph/blob/d8d9dd040d9e4fa957d352a0443a3cba58d35110/graphql/schema/gqlschema.go#L895)):*

When does this happen, is it when Update only has filter but no set or remove? Can we just not generate the Update field in this case?
</blockquote></details>

Add comments about when this occurs for Update and also check that the field name should be filter. 
___
*[graphql/schema/gqlschema.go, line 1167 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/6870#-MMQB9bHBWjPR4v1XnBc:-MMQD7TpAUsWmv0Z2THm:bkvvd8f) ([raw file](https://github.com/dgraph-io/dgraph/blob/d8d9dd040d9e4fa957d352a0443a3cba58d35110/graphql/schema/gqlschema.go#L1167)):*

This wasn't required before but is required now, why is that the case?
</blockquote></details>

inbuilt types like Point, MultiPolygon...

pawanrawal (approved):

PR #6903, opened 2020-11-17
chore(zero): remove unused flag cache_mb from dgraph zero

NamanJain8:
This PR removes the unused flag `--cache_mb` from zero. This flag is not needed as dgraph no longer uses badger for raftwal.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6903)


jarifibrahim (approved):
We're not using this flag anywhere. We can remove it.

PR #6909, opened 2020-11-18
fix(GraphQL): fix panic when querying custom field in fragment on interface (GRAPHQL-734)

abhimanyusinghgaur:
Fixes GRAPHQL-734.
This PR fixes a panic when a custom field in a type was queried via an interface using a fragment on that type. The fix is to return the field definition instead of just field name while constructing required fields for a custom field. So that one can find the field's Dgraph alias while constructing the child query during rewriting.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6909)

 

Docs Preview: [Dgraph Preview](https://dgraph-793e3ec41a-109474.surge.sh)


graphql/e2e/custom_logic/cmd/main.go
```diff
@@ -134,7 +134,11 @@ func check2(v interface{}, err error) {
 }
 
 func getError(key, val string) error {
-	return fmt.Errorf(`{ "errors": [{"message": "%s: %s"}] }`, key, val)
```
abhimanyusinghgaur, line 137: `strconv.Quote()` would not work for this use case.
See: https://play.golang.org/p/JOton9hhrww

The reason is, inside a JSON string any double quote should be escaped, `strconv.Quote(arg)` would escape the double quotes inside the `arg` but it would again put double quotes at the beginning and end of the `arg`,

pawanrawal (approved):

PR #6917, opened 2020-11-19
fix(GraphQL): fix exclusion of filters in Query generation in case of…

minhaj-shakeel:
… just Has Filters. (#6870)

Fixes GRAPHQL-787.

This PR fixes the Bug in query and mutations generation in case of types that have just `has` filter and not any other kind of filter. For example for the given type:
```
type Student {
   name: String
   age: Int
}
```
the corresponding query generated was
```
queryStudent(order: StudentOrder, first: Int, offset: Int): [Student]
```
and only `add` mutation was being generated.
Whereas, the correct query should be:
```
queryStudent(filter: StudentFilter, order: StudentOrder, first: Int, offset: Int): [Student]
```
and corresponding update and delete mutation should be:
```
updateStudent(input: UpdateStudentInput!): UpdateStudentPayload
deleteStudent(filter: StudentFilter!): DeleteStudentPayload
```

(cherry picked from commit 5e60ef190c3823671aaccd6bfde9486ddd9d75b0)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6917)

### DIFF
--- a/graphql/resolve/query_test.yaml
+++ b/graphql/resolve/query_test.yaml
@@ -792,6 +792,20 @@
       }
     }
 
+- name: "Query Has Filter on type which has neither ID field nor any search argument"
+  gqlquery: |
+    query {
+      queryNode(filter: {has: name}){
+        name
+      }
+    }
+  dgquery: |-
+    query {
+      queryNode(func: type(Node)) @filter(has(Node.name)) {
+        name : Node.name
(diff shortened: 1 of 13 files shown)


pawanrawal (commented):
A couple of tests seem to be failing

pawanrawal (approved):

PR #6919, opened 2020-11-19
fix(raftwal): Change the raftwal file size to 64MB

ahsanbarkati:
This PR changes the initial raftwal log-file size to 64MB. It also fixes a bug, we used to check the data offset in the raftwal and expected it to be less than `logfilesize`, but `logfilesize` denotes the initial size of the file. When the data is large enough to cause the logfile grow then that check won't be correct and would result in empty data in the returned raft entries.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6919)

 

Docs Preview: [Dgraph Preview](https://dgraph-86efe803a7-110233.surge.sh)


jarifibrahim (approved):
Add a test. Don't modify the existing test.

Merge after addressing the comments.

raftwal/encryption_test.go
```diff
@@ -80,9 +80,9 @@ func TestLogRotate(t *testing.T) {
 	const SEED = 1
 	rand.Seed(SEED)
 	makeEntry := func(i int) raftpb.Entry {
-		// Be careful when changing this value, as it could easily end up filling up
-		// the entire tmpfs. Currently, this writes ~1.5GB.
-		data := make([]byte, rand.Intn(1<<16))
+		// Write data of size 30000 * 5 KB, which is ~150MB. This will cause
+		// the log-file with initial size of 64MB to grow.
+		data := make([]byte, 5<<10)
```
jarifibrahim, line 85: Don't do this.

raftwal/log.go
```diff
@@ -171,7 +171,8 @@ func (lf *logFile) GetRaftEntry(idx int) raftpb.Entry {
 		Index: entry.Index(),
 		Type:  raftpb.EntryType(int32(entry.Type())),
 	}
-	if entry.DataOffset() > 0 && entry.DataOffset() < logFileSize {
+	if entry.DataOffset() > 0 {
+		y.AssertTrue(entry.DataOffset() < uint64(len(lf.Data)))
```
jarifibrahim, line 175: Use x.AssertTrue

PR #6928, opened 2020-11-20
chore: add increment to manual_tests

ajeetdsouza:
Currently, when a manual start of Dgraph is tested, a healthcheck is used to see if the start was successful. This PR also performs an increment operation on each started Alpha, as part of the test.

I've also added a manual start test for Dgraph with both internal and external TLS enabled:

```sh
./test.sh test::manual_start_tls2
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6928)

### DIFF
--- a/contrib/manual_tests/test.sh
+++ b/contrib/manual_tests/test.sh
@@ -64,27 +64,31 @@ function dgraph::killall() {
   done
 }
 
+function dgraph::start_zero() {
+  local -r i="$i"
+  log::debug "Starting Zero $i."
+
+  local grpc_port=$((5080 + i))
+  local http_port=$((6080 + i))
+
+  for port in "$grpc_port" "$http_port"; do
+    portkill "$port"
+  done
+
+  local zero_args_default=(--cwd "$DGRAPH_PATH/zero$i" --idx "$i" --port_offset "$i")
+
+  if [ "$i" -ne 1 ]; then
+    zero_args_default+=(--peer 'localhost:5081')
+  fi
+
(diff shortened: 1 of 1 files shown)


contrib/manual_tests/test.sh
```diff
@@ -277,6 +307,15 @@ function test::manual_start_acl() {
...
+  for i in $(seq "$n_alphas"); do
+    count="$(dgraph::increment "$i" --user groot --password password)"
+    if [ "$i" -ne "$count" ]; then
+      log::error "Expected increment: $i but got: $count"
+      return 1
+    fi
+  done
 }
 
 function test::manual_start_tls() {
```
jarifibrahim, line 321: Add a comment here as well about how is this different from `tls2()` test.

contrib/manual_tests/test.sh
```diff
@@ -297,6 +336,64 @@ function test::manual_start_tls() {
...
+      return 1
+    fi
+  done
+}
+
+# Test manual start with both internal and external TLS enabled.
+function test::manual_start_tls2() {
+  dgraph::generate_tls
+
+  local -r n_zeros=2
```
jarifibrahim, line 354: This is an odd configuration. Why 2 nodes? The nodes should always be in odd number, 3 or 5 so that they can form a quorum.

jarifibrahim (changes requested):
Got a comment.

jarifibrahim (approved):

PR #6941, opened 2020-11-23
fix(test): TestCrossGroupPermission (DGRAPH-2759)

abhimanyusinghgaur:
This test was adding a user to multiple groups, but after adding a user to a particular group, to check that the group should be present in the admin `updateUser` GraphQL response, it was comparing the exact response of the admin call to match only the name of the particular group. This resulted in the failure of this test because there can also be other groups present in the response.

This PR fixes this test to only check that the group which is required to be present in the response is present in the response, and ignore other groups.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6941)

### DIFF
--- a/ee/acl/acl_test.go
+++ b/ee/acl/acl_test.go
@@ -589,9 +589,40 @@ func addToGroup(t *testing.T, accessToken, userId, group string) {
 		},
 	}
 	b := makeRequest(t, accessToken, params)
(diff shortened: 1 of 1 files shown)


jarifibrahim:
Could you please link the original commit here? 
I usually do `git cherry-pick -x commtSHA` which adds the original commit to the cherry-picked commit message

jarifibrahim (approved):

PR #6944, opened 2020-11-23
Bring back --lru-mb and deprecate it

ajeetdsouza:
The `--lru-mb` flag has been removed in favor of `--cache-mb`. This PR brings it back in a backwards-compatible manner, but warns the user that the flag is deprecated.

Note that the [`viper`](https://github.com/spf13/viper) dependency had to be updated due to a bug that was fixed [here](https://github.com/spf13/viper/pull/331).

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6944)


jarifibrahim (approved):
Run it locally once and then it's good to merge.

PR #6961, opened 2020-11-24
chore(test): removing redundant mtls tests

aman-bansal:
This PR aims to fix redundant test cases of backup, loader etc from the mtls internal. This tests are mostly copy of the basic test with one change, that all ports communicate with tls encrypted. But these tests doesn't test backup or loader etc and thus making them redundant. 
One thing can be done is to add mtls in the basic test package but that is violating the idea on which systests are build. These test cases are now part of qa and release testing.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6961)


jarifibrahim (approved):
@aman-bansal I'm assuming there are other tests that are similar and test tls.

jarifibrahim (changes requested):
Let's enable TLS in the systest. Since TLS is a super set of basic systest, it should be okay to run them with TLS enabled.

jarifibrahim (approved):

PR #6962, opened 2020-11-24
chore: Add flag to test script to avoid building Dgraph

ajeetdsouza:
Added a `--rebuild-binary` flag to the `t` package that, when disabled, avoids building Dgraph before running tests.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6962)

### DIFF
--- a/t/main.go
+++ b/t/main.go
@@ -74,6 +74,8 @@ var (
 		"Clear all the test clusters.")
 	dry = pflag.BoolP("dry", "", false,
 		"Just show how the packages would be executed, without running tests.")
+	noBuild = pflag.BoolP("no-build", "", false,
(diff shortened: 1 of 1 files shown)


t/main.go
```diff
@@ -74,6 +74,8 @@ var (
 		"Clear all the test clusters.")
 	dry = pflag.BoolP("dry", "", false,
 		"Just show how the packages would be executed, without running tests.")
+	noBuild = pflag.BoolP("no-build", "", false,
```
jarifibrahim, line 77: Don't use double negatives.
Maybe call it `rebuild-binary` and set it to true by default.

jarifibrahim (approved):

PR #6967, opened 2020-11-24
test: don't compare counter in successive state calls (GRAPHQL-873)

abhimanyusinghgaur:
This PR fixes the flaky behavior of the test which verifies that the response of `/state` should be the same as the response of GraphQL admin state request. Previously it used to compare the counter returned in the state response, which can change between two consecutive state calls. The counter in state response is set based on the index of the last RAFT entry applied by Zero. If Zero has applied a RAFT entry between the two consecutive state calls, then the counter would have changed. So, let's not compare this in the test.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6967)

### DIFF
--- a/graphql/e2e/common/admin.go
+++ b/graphql/e2e/common/admin.go
@@ -741,7 +741,6 @@ func adminState(t *testing.T) {
 	queryParams := &GraphQLParams{
 		Query: `query {
 			state {
-				counter
 				groups {
 					id
 					members {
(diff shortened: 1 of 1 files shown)


manishrjain (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6967#-:-MMvqRD1DC4YFc6HH4ta:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @MichaelJCompton and @pawanrawal)

PR #6987, opened 2020-11-25
feat(Dgraph): Add badger.compression to Dgraph restore

ajeetdsouza:
`dgraph restore` does not currently use any form of compression. This PR adds a `--badger.compression` flag (similar to other Dgraph subcommands) and defaults all compression to `snappy`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6987)

### DIFF
--- a/dgraph/cmd/bulk/run.go
+++ b/dgraph/cmd/bulk/run.go
@@ -158,10 +158,6 @@ func run() {
 	if opt.Version {
 		os.Exit(0)
 	}
-	if opt.BadgerCompressionLevel < 0 {
-		fmt.Printf("Invalid compression level: %d. It should be non-negative",
-			opt.BadgerCompressionLevel)
-	}
 
 	totalCache := int64(Bulk.Conf.GetInt("badger.cache-mb"))
 	x.AssertTruef(totalCache >= 0, "ERROR: Cache size must be non-negative")
--- a/ee/backup/run.go
+++ b/ee/backup/run.go
@@ -39,6 +39,7 @@ var ExportBackup x.SubCommand
 
 var opt struct {
 	backupId    string
+	compression string
 	location    string
 	pdir        string
 	zero        string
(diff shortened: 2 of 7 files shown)


dgraph/cmd/bulk/run.go
```diff
@@ -158,10 +158,6 @@ func run() {
 	if opt.Version {
 		os.Exit(0)
 	}
-	if opt.BadgerCompressionLevel < 0 {
```
ajeetdsouza, line 161: This has been removed because this is derived from `x.ParseCompression`, which calls `glog.Fatal` in case of a negative compression level. Thus, there is no need to check over here.

ee/backup/run.go
```diff
@@ -111,6 +112,10 @@ $ dgraph restore -p . -l /var/backups/dgraph -z localhost:5080
 	}
 
 	flag := Restore.Cmd.Flags()
+	flag.StringVar(&opt.compression, "badger.compression", "snappy",
```
ajeetdsouza, line 115: This is the third place in the code where this exact description has been duplicated. Perhaps we could extract it out elsewhere.

worker/restore.go
```diff
@@ -35,7 +36,7 @@ import (
 )
 
 // RunRestore calls badger.Load and tries to load data into a new DB.
-func RunRestore(pdir, location, backupId string, key x.SensitiveByteSlice) LoadResult {
+func RunRestore(pdir, location, backupId string, key x.SensitiveByteSlice, ctype options.CompressionType, clevel int) LoadResult {
```
ajeetdsouza, line 39: Seems like too many arguments. Should this function be changed to take in the entire `opt` struct instead?

systest/backup/encryption/backup_test.go
```diff
@@ -354,7 +355,7 @@ func runFailingRestore(t *testing.T, backupLocation, lastDir string, commitTs ui
 	require.NotNil(t, k)
 	require.NoError(t, err)
 
-	result := worker.RunRestore("./data/restore", backupLocation, lastDir, k)
+	result := worker.RunRestore("./data/restore", backupLocation, lastDir, k, options.Snappy, 0)
```
ajeetdsouza, line 358: Note: these tests did not use any form of compression earlier, but I've changed them to use `snappy` by default.

jarifibrahim (approved):

PR #7013, opened 2020-11-27
fix(raftwal): Pass the encryption key instead of reading from WorkerConfig

ahsanbarkati:
We should not read the encryption key from the worker config, rather we can pass it as an argument to the Init() method which initializes the diskStorage for the raftwal. 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7013)

### DIFF
--- a/raftwal/encryption_test.go
+++ b/raftwal/encryption_test.go
@@ -24,16 +24,15 @@ import (
 	"os"
 	"testing"
 
-	"github.com/dgraph-io/dgraph/x"
 	"github.com/stretchr/testify/require"
 	"go.etcd.io/etcd/raft/raftpb"
 )
 
 func TestEntryReadWrite(t *testing.T) {
-	x.WorkerConfig.EncryptionKey = []byte("badger16byteskey")
+	key := []byte("badger16byteskey")
 	dir, err := ioutil.TempDir("", "raftwal")
 	require.NoError(t, err)
-	el, err := openWal(dir)
+	ds, err := Init(dir, key)
 	require.NoError(t, err)
 	defer os.RemoveAll(dir)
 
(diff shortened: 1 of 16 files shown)


danielmai (commented):
What if we had a func called `InitEncrypted(dir string, encKey x.SensitiveByteSlice) *DiskStorage` alongside the `Init` func? That way the change isn't as invasive as adding `nil` everywhere for `Init(dir, nil)`. And, given that all the error checks just panic/stop anyway if raftwal isn't init'd when returning an error value, is returning an error helpful?

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7013#-:-MNA02_S5rgeeOVOsygV:brwush5)* status: 0 of 10 files reviewed, all discussions resolved (waiting on @manishrjain and @vvbalaji-dgraph)

ahsanbarkati:
Yes, having `InitEncrypted` makes sense. The idea for returning the error was mainly because of the tests, we need to check against the errors in the tests.

raftwal/storage.go
```diff
@@ -85,26 +85,38 @@ type indexRange struct {
...
+// Init initializes an instance of DiskStorage without encryption.
 func Init(dir string) *DiskStorage {
+	ds, err := InitEncrypted(dir, nil)
+	x.Check(err)
+	return ds
+}
+
+// InitEncrypted initializes returns a properly initialized instance of DiskStorage.
+// To gracefully shutdown DiskStorage, store.Closer.SignalAndWait() should be called.
+func InitEncrypted(dir string, encKey x.SensitiveByteSlice) (*DiskStorage, error) {
```
jarifibrahim, line 97: There is no point in returning the error since we're doing x.Check(err) everywhere after calling `InitEncrypted`.

jarifibrahim (approved):
Remove the return parameter from `InitEncrypted` function and this PR should be good to merge.

raftwal/storage.go
ahsanbarkati: We needed to return the error so that we can test the `InitEncrypted` function for various cases. For example, calling `InitEncrypted(dir, wrongEncryptionKey)` will Fatal and there will be no way to do these tests. Had `x.Check()` been a panic instead of a Fatal, then we could have caught the panic by `require.PanicsWithError()`. Do you have some idea/workaround?

raftwal/encryption_test.go
```diff
@@ -24,47 +24,45 @@ import (
...
 	require.Equal(t, uint64(1), entries[0].Index)
 	require.Equal(t, uint64(1), entries[0].Term)
 	require.Equal(t, data, entries[0].Data)
 
 	// Opening it with a wrong key fails.
-	x.WorkerConfig.EncryptionKey = []byte("other16byteskeys")
-	_, err = openWal(dir)
+	wrongKey := []byte("other16byteskeys")
+	_, err = InitEncrypted(dir, wrongKey)
 	require.EqualError(t, err, "Encryption key mismatch")
```
ahsanbarkati, line 62: @jarifibrahim, we are checking for the error here.

jarifibrahim: But this is a test. The actual code isn't handling this error.

PR #7015, opened 2020-11-30
chore: Fix gql fuzz tests

ajeetdsouza:
Fixes some issues with the fuzz test script:
- Assuming the Dgraph folder is present in `$GOPATH`. This is not necessarily true, I've changed this to only run tests in the current (`gql`) directory.
- The Docker image being used for clang doesn't seem to be updated any more, and was causing problems with the latest builds. I've replaced it with the most popular clang image on Docker Hub. Note that I had to use `clang-10`, because the `clang-11` image is based on Ubuntu Bionic (glibc 2.39), which causes issues in our CI. This can be changed once we upgrade.
- Bash quoting/whitespace nits.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7015)


jarifibrahim (approved):
Make proper changes on teamCity as well.

PR #7039, opened 2020-12-01
feat(graphql):  adding auth token support for regexp, in and arrays

aman-bansal:
This PR aims to fix GRAPHQL-808. This enables multiple things.
1. support for regex matching in custom claims.
2. support for array in custom claims.
3. Claims match based on float, int64, int, structs, bool, string
4. in based custom claim match.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7039)

### DIFF
--- a/graphql/e2e/auth/auth_test.go
+++ b/graphql/e2e/auth/auth_test.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/dgrijalva/jwt-go/v4"
 	"os"
 	"strings"
 	"testing"
@@ -28,7 +29,6 @@ import (
 
 	"github.com/dgraph-io/dgraph/graphql/e2e/common"
 	"github.com/dgraph-io/dgraph/testutil"
-	"github.com/dgrijalva/jwt-go/v4"
 	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/require"
 )
@@ -1889,4 +1889,227 @@ func TestAuthWithSecretDirective(t *testing.T) {
 	gqlResponse = checkLogPassword(t, logID, newLog.Pwd, "USER")
 	require.JSONEq(t, `{"checkLogPassword": null}`, string(gqlResponse.Data))
 	deleteLog(t, logID)
+}
+
+var bookResponse = `{"queryBook":[{"bookId":"book1","name":"Introduction","desc":"Intro book"}]}`
+
+func TestAuthEqFilterWithObjectAsTokenVal(t *testing.T) {
+	getUserParams := &common.GraphQLParams{
+		Headers: common.GetJWT(t, map[string]interface{}{"a":"b"}, nil, metaInfo),
+		Query: `query {
+		  queryBook{
+			bookId
+			name
+			desc
+		  }
+		}`,
+	}
+
+	gqlResponse := getUserParams.ExecuteAsPost(t, common.GraphqlURL)
+	require.Nil(t, gqlResponse.Errors)
+	require.JSONEq(t, string(gqlResponse.Data), bookResponse)
+}
+
+func TestAuthEqFilterWithFloatTokenVal(t *testing.T) {
+	getUserParams := &common.GraphQLParams{
+		Headers: common.GetJWT(t, 123.12, nil, metaInfo),
+		Query: `query {
+		  queryBook{
+			bookId
+			name
+			desc
+		  }
+		}`,
+	}
+
+	gqlResponse := getUserParams.ExecuteAsPost(t, common.GraphqlURL)
+	require.Nil(t, gqlResponse.Errors)
+	require.JSONEq(t, string(gqlResponse.Data), bookResponse)
+}
+
+func TestAuthEqFilterWithInt64TokenVal(t *testing.T) {
+	getUserParams := &common.GraphQLParams{
+		Headers: common.GetJWT(t, 1237890123456, nil, metaInfo),
+		Query: `query {
+		  queryBook{
+			bookId
+			name
+			desc
+		  }
+		}`,
+	}
+
+	gqlResponse := getUserParams.ExecuteAsPost(t, common.GraphqlURL)
+	require.Nil(t, gqlResponse.Errors)
+	require.JSONEq(t, string(gqlResponse.Data), bookResponse)
+}
+
+func TestAuthEqFilterWithIntTokenVal(t *testing.T) {
+	getUserParams := &common.GraphQLParams{
+		Headers: common.GetJWT(t, 1234, nil, metaInfo),
+		Query: `query {
+		  queryBook{
+			bookId
+			name
+			desc
+		  }
+		}`,
+	}
+
+	gqlResponse := getUserParams.ExecuteAsPost(t, common.GraphqlURL)
+	require.Nil(t, gqlResponse.Errors)
+	require.JSONEq(t, string(gqlResponse.Data), bookResponse)
+}
+
+func TestAuthEqFilterWithBoolTokenVal(t *testing.T) {
+	getUserParams := &common.GraphQLParams{
+		Headers: common.GetJWT(t, true, nil, metaInfo),
+		Query: `query {
+		  queryBook{
+			bookId
+			name
+			desc
+		  }
+		}`,
+	}
+
+	gqlResponse := getUserParams.ExecuteAsPost(t, common.GraphqlURL)
+	require.Nil(t, gqlResponse.Errors)
+	require.JSONEq(t, string(gqlResponse.Data), bookResponse)
+}
+
+func TestAuthInFilterWithObjectAsTokenVal(t *testing.T) {
+	getUserParams := &common.GraphQLParams{
+		Headers: common.GetJWT(t, map[string]interface{}{"e":"f"}, nil, metaInfo),
+		Query: `query {
+		  queryBook{
+			bookId
+			name
+			desc
+		  }
+		}`,
+	}
+
+	gqlResponse := getUserParams.ExecuteAsPost(t, common.GraphqlURL)
+	require.Nil(t, gqlResponse.Errors)
+	require.JSONEq(t, string(gqlResponse.Data), bookResponse)
+}
+
+func TestAuthInFilterWithFloatTokenVal(t *testing.T) {
+	getUserParams := &common.GraphQLParams{
+		Headers: common.GetJWT(t, 312.124, nil, metaInfo),
+		Query: `query {
+		  queryBook{
+			bookId
+			name
+			desc
+		  }
+		}`,
+	}
+
+	gqlResponse := getUserParams.ExecuteAsPost(t, common.GraphqlURL)
+	require.Nil(t, gqlResponse.Errors)
+	require.JSONEq(t, string(gqlResponse.Data), bookResponse)
+}
+
+func TestAuthInFilterWithInt64TokenVal(t *testing.T) {
+	getUserParams := &common.GraphQLParams{
+		Headers: common.GetJWT(t, 1246879976444232435, nil, metaInfo),
+		Query: `query {
+		  queryBook{
+			bookId
+			name
+			desc
+		  }
+		}`,
+	}
+
+	gqlResponse := getUserParams.ExecuteAsPost(t, common.GraphqlURL)
+	require.Nil(t, gqlResponse.Errors)
+	require.JSONEq(t, string(gqlResponse.Data), bookResponse)
+}
+
+func TestAuthInFilterWithIntTokenVal(t *testing.T) {
+	getUserParams := &common.GraphQLParams{
+		Headers: common.GetJWT(t, 6872, nil, metaInfo),
+		Query: `query {
+		  queryBook{
+			bookId
+			name
+			desc
+		  }
+		}`,
+	}
+
+	gqlResponse := getUserParams.ExecuteAsPost(t, common.GraphqlURL)
+	require.Nil(t, gqlResponse.Errors)
+	require.JSONEq(t, string(gqlResponse.Data), bookResponse)
+}
+
+func TestAuthEqFilterFromTokenWithArrayVal(t *testing.T) {
+	getUserParams := &common.GraphQLParams{
+		Headers: common.GetJWT(t, []int{456,1234}, nil, metaInfo),
+		Query: `query {
+		  queryBook{
+			bookId
+			name
+			desc
+		  }
+		}`,
+	}
+
+	gqlResponse := getUserParams.ExecuteAsPost(t, common.GraphqlURL)
+	require.Nil(t, gqlResponse.Errors)
+	require.JSONEq(t, string(gqlResponse.Data), bookResponse)
+}
+
+func TestAuthInFilterFromTokenWithArrayVal(t *testing.T) {
+	getUserParams := &common.GraphQLParams{
+		Headers: common.GetJWT(t, []int{124324, 6872}, nil, metaInfo),
+		Query: `query {
+		  queryBook{
+			bookId
+			name
+			desc
+		  }
+		}`,
+	}
+
+	gqlResponse := getUserParams.ExecuteAsPost(t, common.GraphqlURL)
+	require.Nil(t, gqlResponse.Errors)
+	require.JSONEq(t, string(gqlResponse.Data), bookResponse)
+}
+
+func TestAuthRegexFilter(t *testing.T) {
+	getUserParams := &common.GraphQLParams{
+		Headers: common.GetJWT(t, "xyz@dgraph.io", nil, metaInfo),
+		Query: `query {
+		  queryBook{
+			bookId
+			name
+			desc
+		  }
+		}`,
+	}
+
+	gqlResponse := getUserParams.ExecuteAsPost(t, common.GraphqlURL)
+	require.Nil(t, gqlResponse.Errors)
+	require.JSONEq(t, string(gqlResponse.Data), bookResponse)
+}
+
+func TestAuthRegexFilterFromTokenWithArrayVal(t *testing.T) {
+	getUserParams := &common.GraphQLParams{
+		Headers: common.GetJWT(t, []string{"abc@def.com", "xyz@dgraph.io"}, nil, metaInfo),
+		Query: `query {
+		  queryBook{
+			bookId
+			name
+			desc
+		  }
+		}`,
+	}
+
+	gqlResponse := getUserParams.ExecuteAsPost(t, common.GraphqlURL)
+	require.Nil(t, gqlResponse.Errors)
+	require.JSONEq(t, string(gqlResponse.Data), bookResponse)
 }
--- a/graphql/e2e/auth/auth_test.go
+++ b/graphql/e2e/auth/auth_test.go
@@ -1893,7 +1893,7 @@ func TestAuthWithSecretDirective(t *testing.T) {
 	deleteLog(t, logID)
 }
 
-func TestAuth(t *testing.T) {
+func TestAuthRBACEvaluation(t *testing.T) {
 	tcs := []struct {
 		name   string
 		header http.Header
--- a/graphql/e2e/auth/auth_test.go
+++ b/graphql/e2e/auth/auth_test.go
@@ -20,10 +20,12 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/dgrijalva/jwt-go/v4"
+	"net/http"
 	"os"
 	"strings"
 	"testing"
+
+	"github.com/dgrijalva/jwt-go/v4"
 
 	"github.com/google/go-cmp/cmp/cmpopts"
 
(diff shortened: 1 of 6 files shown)


abhimanyusinghgaur (approved):
:lgtm:
Just some minor refactors

> Reviewed 5 of 6 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-:-MNXcwnPEKE8rBioB9ti:b-5kwqa3)* status: all files reviewed, 5 unresolved discussions (waiting on @aman-bansal, @MichaelJCompton, and @pawanrawal)

___
*[graphql/e2e/auth/auth_test.go, line 1903 at r2](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNXa_YR1F8pcKFfhVzB:-MNXa_YR1F8pcKFfhVzC:b-qg07lx) ([raw file](https://github.com/dgraph-io/dgraph/blob/fd8c8d89276639c5726dfdd72dcb9e9e5a63c7ce/graphql/e2e/auth/auth_test.go#L1903)):*
> ```
> require.Nil(t, gqlResponse.Errors)
> ```
`RequireNoGQLErrors` everywhere
___
*[graphql/e2e/auth/auth_test.go, line 1923 at r2](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNXb-Cb8bCzoxp_OUQG:-MNXb-Cb8bCzoxp_OUQH:bf2rhlm) ([raw file](https://github.com/dgraph-io/dgraph/blob/fd8c8d89276639c5726dfdd72dcb9e9e5a63c7ce/graphql/e2e/auth/auth_test.go#L1923)):*
> ```Go
> 	require.JSONEq(t, string(gqlResponse.Data), bookResponse)
> }
> ```

All the tests seem to be doing the same query and expect the same response, with a varying header. So, can we combine them into one single test which tests the query over the list of varying headers like:
```
testCases := []struct{
 name string
 headers: http.Header
}
for _, tcase := range testCases {
  t.Run(tcase.name, func(t *testing.T) {
    // query here with the current header and match the response
  }
}
```
___
*[graphql/schema/auth.go, line 464 at r2](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNXYwIa0oG-awkYD5-h:-MNXYwIb9QjJn1wwXlan:b-rr17g) ([raw file](https://github.com/dgraph-io/dgraph/blob/fd8c8d89276639c5726dfdd72dcb9e9e5a63c7ce/graphql/schema/auth.go#L464)):*
> ```
> objects and nil
> ```
nil
___
*[graphql/schema/auth.go, line 469 at r2](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNXZ7Lo54i_R3a4OJEn:-MNXZ7Lo54i_R3a4OJEo:b-b4ykn4) ([raw file](https://github.com/dgraph-io/dgraph/blob/fd8c8d89276639c5726dfdd72dcb9e9e5a63c7ce/graphql/schema/auth.go#L469)):*
<details><summary><i alt="summary">Quoted 17 lines of code…</i></summary>

> ```
> 	if op == nil {
> 		return nil, gqlerror.Errorf("Type %s: @auth: `%s` is not a valid GraphQL variable. "+
> 			"object or nil values aren't supported", typ.Name, rule[idx[0][2]:idx[0][3]])
> 	}
> 	query := RBACQuery{
> 		Variable: rule[idx[0][2]:idx[0][3]],
> 		Operator: rule[idx[0][4]:idx[0][5]],
> 		Operand:  op,
> 	}
> 	if !strings.HasPrefix(query.Variable, "$") {
> 		return nil, gqlerror.Errorf("Type %s: @auth: `%s` is not a valid GraphQL variable.",
> 			typ.Name, query.Variable)
> 	}
> 	query.Variable = query.Variable[1:]
> 	if err = validateRBACOperators(typ, query); err != nil {
> 		return nil, err
> 	}
> ```
</details>


can we have just one function which does all the validation on the `RBACQuery`, something like 
```
func Validate(typ *ast.Definition, query *RBACQuery) error {}
```
which will have all the validation related to variable, operator, and operand inside of it.
And if validate passes, then we call another function initialization function which can take care of removing the `$` from variable, and setting regex.
___
*[graphql/schema/auth.go, line 496 at r2](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNXRONH2mAciZA75Cg3:-MNXRONH2mAciZA75Cg4:b3idnlo) ([raw file](https://github.com/dgraph-io/dgraph/blob/fd8c8d89276639c5726dfdd72dcb9e9e5a63c7ce/graphql/schema/auth.go#L496)):*
> ```
> var ok = false
> ```
ok := false

pawanrawal (changes requested):
It's looking good as a change. Added some comments about the errors that are returned.

> Reviewed 1 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-:-MNxF_v50DeK59pDqEnl:b-4iimjk)* status: 3 of 6 files reviewed, 12 unresolved discussions (waiting on @abhimanyusinghgaur, @aman-bansal, and @MichaelJCompton)

___
*[graphql/schema/auth.go, line 98 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNwy7XR6HFWbCtiip1d:-MNwy7XS01IquqvML92v:b-prmel8) ([raw file](https://github.com/dgraph-io/dgraph/blob/8448f8063cfa5c92577a7f2522c530fdb0e4a044/graphql/schema/auth.go#L98)):*
> ```Go
> }
> 
> func checkIfRegexMatch(regex *regexp.Regexp, val1 interface{}) RuleResult {
> ```

instead of having this function can we use the above function for this case as well with just one element in `arr`?
___
*[graphql/schema/auth.go, line 120 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNwyfuW5TuE6uEmOeyw:-MNwyfuW5TuE6uEmOeyx:bzh5h2m) ([raw file](https://github.com/dgraph-io/dgraph/blob/8448f8063cfa5c92577a7f2522c530fdb0e4a044/graphql/schema/auth.go#L120)):*
> ```Go
> 		tokenValues, err := cast.ToSliceE(av[rq.Variable])
> 		if err != nil {
> 			// this means value for variable in token in not an array
> ```

interesting, so the function returns an error if you can't cast to slice? Any other cases in which an error is returned? Can't we just check the type of the value and then call the appropriate function?
___
*[graphql/schema/auth.go, line 126 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNxEKVu1uarqMRjcCHo:-MNxEKVu1uarqMRjcCHp:b-v7kcll) ([raw file](https://github.com/dgraph-io/dgraph/blob/8448f8063cfa5c92577a7f2522c530fdb0e4a044/graphql/schema/auth.go#L126)):*
> ```Go
> 	case "regexp":
> 		// if regexp, auth rule value should always be string and so as token values
> 		tokenValues, err := cast.ToSliceE(av[rq.Variable])
> ```

We do this again now. Why not just check the value type above outside this switch case first.
___
*[graphql/schema/auth_schemas_test.yaml, line 39 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNxEqOe4fR7UT5eC_8Q:-MNxEqOe4fR7UT5eC_8R:b-gylc5t) ([raw file](https://github.com/dgraph-io/dgraph/blob/8448f8063cfa5c92577a7f2522c530fdb0e4a044/graphql/schema/auth_schemas_test.yaml#L39)):*
> ```YAML
>       }
>     errlist: [
>       { "message": "Type X: @auth: `in` operator has invalid value in this rule." }
> ```

Should the error message say that the value should be an array?
___
*[graphql/schema/auth_schemas_test.yaml, line 62 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNxEz8B0W6b1JdRgsJu:-MNxEz8B0W6b1JdRgsJv:bmj29iw) ([raw file](https://github.com/dgraph-io/dgraph/blob/8448f8063cfa5c92577a7f2522c530fdb0e4a044/graphql/schema/auth_schemas_test.yaml#L62)):*
> ```YAML
>       }
>     errlist: [
>       { "message": "Type X: @auth: `$USER` is not a valid GraphQL variable." }
> ```

Strange error, shouldn't it say `nil` is an invalid value for eq`
___
*[graphql/schema/auth_schemas_test.yaml, line 74 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNxF6066VT4Q6iZyoIQ:-MNxF6066VT4Q6iZyoIR:bqhujmt) ([raw file](https://github.com/dgraph-io/dgraph/blob/8448f8063cfa5c92577a7f2522c530fdb0e4a044/graphql/schema/auth_schemas_test.yaml#L74)):*
> ```YAML
>       }
>     errlist: [
>       { "message": "Type X: @auth: `$USER` is not a valid GraphQL variable. object with nil values aren't supported" }
> ```

Why isn't `$USER` an invalid GraphQL variable. It seems to me that `eq: null` is an invalid check here and not the variable?
___
*[graphql/schema/auth_schemas_test.yaml, line 86 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNxFJENCPexkzZRhdys:-MNxFJENCPexkzZRhdyt:bmnp456) ([raw file](https://github.com/dgraph-io/dgraph/blob/8448f8063cfa5c92577a7f2522c530fdb0e4a044/graphql/schema/auth_schemas_test.yaml#L86)):*
> ```YAML
>       }
>     errlist: [
>       { "message": "Type X: @auth: `regexp` operator has invalid value in this rule." }
> ```

What is `this` rule? Please also add the invalid value using %v as part of the error and say that the value should be of type String.

aman-bansal (commented):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7039)* status: 2 of 6 files reviewed, 12 unresolved discussions (waiting on @abhimanyusinghgaur, @MichaelJCompton, and @pawanrawal)

___
*[graphql/e2e/auth/auth_test.go, line 1903 at r2](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNXa_YR1F8pcKFfhVzB:-MNxIRXi8qg-UE3i9lsm:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/fd8c8d89276639c5726dfdd72dcb9e9e5a63c7ce/graphql/e2e/auth/auth_test.go#L1903)):*

> ```
> require.Nil(t, gqlResponse.Errors)
> ```
`RequireNoGQLErrors` everywhere
</blockquote></details>

Done.
___
*[graphql/e2e/auth/auth_test.go, line 1923 at r2](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNXb-Cb8bCzoxp_OUQG:-MNxISDF7PzhxXDoYWbw:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/fd8c8d89276639c5726dfdd72dcb9e9e5a63c7ce/graphql/e2e/auth/auth_test.go#L1923)):*

All the tests seem to be doing the same query and expect the same response, with a varying header. So, can we combine them into one single test which tests the query over the list of varying headers like:
```
testCases := []struct{
 name string
 headers: http.Header
}
for _, tcase := range testCases {
  t.Run(tcase.name, func(t *testing.T) {
    // query here with the current header and match the response
  }
}
```
</blockquote></details>

Done.
___
*[graphql/schema/auth.go, line 464 at r2](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNXYwIa0oG-awkYD5-h:-MNxJ8Fy8kwyEK6qNW4d:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/fd8c8d89276639c5726dfdd72dcb9e9e5a63c7ce/graphql/schema/auth.go#L464)):*

> ```
> objects and nil
> ```
nil
</blockquote></details>

Done.
___
*[graphql/schema/auth.go, line 469 at r2](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNXZ7Lo54i_R3a4OJEn:-MNxJ8nZ1b9Cw-GqkOBH:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/fd8c8d89276639c5726dfdd72dcb9e9e5a63c7ce/graphql/schema/auth.go#L469)):*

> ```
> 	if op == nil {
> 		return nil, gqlerror.Errorf("Type %s: @auth: `%s` is not a valid GraphQL variable. "+
> 			"object or nil values aren't supported", typ.Name, rule[idx[0][2]:idx[0][3]])
> 	}
> 	query := RBACQuery{
> 		Variable: rule[idx[0][2]:idx[0][3]],
> 		Operator: rule[idx[0][4]:idx[0][5]],
> 		Operand:  op,
> 	}
> 	if !strings.HasPrefix(query.Variable, "$") {
> 		return nil, gqlerror.Errorf("Type %s: @auth: `%s` is not a valid GraphQL variable.",
> 			typ.Name, query.Variable)
> 	}
> 	query.Variable = query.Variable[1:]
> 	if err = validateRBACOperators(typ, query); err != nil {
> 		return nil, err
> 	}
> ```

can we have just one function which does all the validation on the `RBACQuery`, something like 
```
func Validate(typ *ast.Definition, query *RBACQuery) error {}
```
which will have all the validation related to variable, operator, and operand inside of it.
And if validate passes, then we call another function initialization function which can take care of removing the `$` from variable, and setting regex.
</blockquote></details>

Done.
___
*[graphql/schema/auth.go, line 496 at r2](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNXRONH2mAciZA75Cg3:-MNxJ9MHBAWWPK24Cuab:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/fd8c8d89276639c5726dfdd72dcb9e9e5a63c7ce/graphql/schema/auth.go#L496)):*

> ```
> var ok = false
> ```
ok := false
</blockquote></details>

Done.
___
*[graphql/schema/auth.go, line 98 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNwy7XR6HFWbCtiip1d:-MO-t5Y42WmT77bh72uH:b-3vavxl) ([raw file](https://github.com/dgraph-io/dgraph/blob/8448f8063cfa5c92577a7f2522c530fdb0e4a044/graphql/schema/auth.go#L98)):*

instead of having this function can we use the above function for this case as well with just one element in `arr`?
</blockquote></details>

I have made changes to this. This should make more sense now.
___
*[graphql/schema/auth.go, line 120 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNwyfuW5TuE6uEmOeyw:-MNxIq3s35DaOpkw65Nd:bhh8hqu) ([raw file](https://github.com/dgraph-io/dgraph/blob/8448f8063cfa5c92577a7f2522c530fdb0e4a044/graphql/schema/auth.go#L120)):*

interesting, so the function returns an error if you can't cast to slice? Any other cases in which an error is returned? Can't we just check the type of the value and then call the appropriate function?
</blockquote></details>

we are not returning error. There is no constraint over how the auth token looks. It could be array of values or a single value. So if we are not able to caste it into array, then it means its a single value and thus we check for equal match
___
*[graphql/schema/auth.go, line 126 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNxEKVu1uarqMRjcCHo:-MO-oW4r4qUkOYBY6186:b-a0yikf) ([raw file](https://github.com/dgraph-io/dgraph/blob/8448f8063cfa5c92577a7f2522c530fdb0e4a044/graphql/schema/auth.go#L126)):*

We do this again now. Why not just check the value type above outside this switch case first.
</blockquote></details>

good point. Thanks.
___
*[graphql/schema/auth_schemas_test.yaml, line 39 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNxEqOe4fR7UT5eC_8Q:-MO-oNpXAK8FGhp_ngxc:bg8nca9) ([raw file](https://github.com/dgraph-io/dgraph/blob/8448f8063cfa5c92577a7f2522c530fdb0e4a044/graphql/schema/auth_schemas_test.yaml#L39)):*

Should the error message say that the value should be an array?
</blockquote></details>

I have added this. it was making things bit messy thats why i didn't do it that time.
___
*[graphql/schema/auth_schemas_test.yaml, line 62 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNxEz8B0W6b1JdRgsJu:-MNxJH9-8s0_QEnnodBJ:b-erc26s) ([raw file](https://github.com/dgraph-io/dgraph/blob/8448f8063cfa5c92577a7f2522c530fdb0e4a044/graphql/schema/auth_schemas_test.yaml#L62)):*

Strange error, shouldn't it say `nil` is an invalid value for eq`
</blockquote></details>

It's because this error has been thrown by the gql parser itself. It fails before we get to check the value and validate it.
null from graphql translates into nil in go. but nil in graphql throws invalid error from the parser itself.
___
*[graphql/schema/auth_schemas_test.yaml, line 74 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNxF6066VT4Q6iZyoIQ:-MO-oJQO8mz52TX_YBVt:b7fxrc7) ([raw file](https://github.com/dgraph-io/dgraph/blob/8448f8063cfa5c92577a7f2522c530fdb0e4a044/graphql/schema/auth_schemas_test.yaml#L74)):*

Why isn't `$USER` an invalid GraphQL variable. It seems to me that `eq: null` is an invalid check here and not the variable?
</blockquote></details>


fixed the message
___
*[graphql/schema/auth_schemas_test.yaml, line 86 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNxFJENCPexkzZRhdys:-MO0-QEx5B6-Ri92LnHe:b-iofwca) ([raw file](https://github.com/dgraph-io/dgraph/blob/8448f8063cfa5c92577a7f2522c530fdb0e4a044/graphql/schema/auth_schemas_test.yaml#L86)):*

What is `this` rule? Please also add the invalid value using %v as part of the error and say that the value should be of type String.
</blockquote></details>

regex value provided in the auth rule should be of type string. This is needed to parse operand in regex. i have added value and reason in the message

pawanrawal (changes requested):
Mostly looking good, just some more changes and then I'll have another final look

> Reviewed 2 of 6 files at r1, 2 of 4 files at r3, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-:-MO0V3ojCA-iPsi8O2gq:b-ht0l11)* status: all files reviewed, 9 unresolved discussions (waiting on @abhimanyusinghgaur, @aman-bansal, @MichaelJCompton, and @pawanrawal)

___
*[graphql/e2e/auth/auth_test.go, line 1923 at r2](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNXb-Cb8bCzoxp_OUQG:-MO0TDYv3Fsx2k2MILTC:b6tpd9c) ([raw file](https://github.com/dgraph-io/dgraph/blob/fd8c8d89276639c5726dfdd72dcb9e9e5a63c7ce/graphql/e2e/auth/auth_test.go#L1923)):*
<details><summary><i>Previously, aman-bansal (aman bansal) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Also seems like the query is common across the test cases so maybe define it above and just use the same one?
___
*[graphql/schema/auth.go, line 146 at r4](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MO0TrSn45cp9ejrITBa:-MO0TrSn45cp9ejrITBb:bchirxo) ([raw file](https://github.com/dgraph-io/dgraph/blob/d587e13c0f69761242fafdc35620eda220ba60cd/graphql/schema/auth.go#L146)):*
> ```Go
> 		}
> 
> 		return rq.checkIfMatchInArray(tokenValues)
> ```

Do you need three separate cases now or can they all be clubbed into the same case with

```
case "eq", "in", "regexp":
...
```
___
*[graphql/schema/auth.go, line 515 at r4](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MO0U72CBWL_XrKCD-hh:-MO0U72CBWL_XrKCD-hi:b-w9nri1) ([raw file](https://github.com/dgraph-io/dgraph/blob/d587e13c0f69761242fafdc35620eda220ba60cd/graphql/schema/auth.go#L515)):*
> ```Go
> 		if isArray {
> 			return false, fmt.Sprintf("Type %s: @auth: `%s` operator has invalid value `%v` in "+
> 				"this rule. Reason: Array values in eq operator will not be supported.",
> ```

Remove `Reason: ` from all errors please. Also `this` rule doesn't tell which rule so best that it be removed. In the future we may also return the path of the rule.


The error should be like below.
```
.... operator has invalid value: %v. Array values for eq operator are not supported.
```
___
*[graphql/schema/auth_schemas_test.yaml, line 62 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNxEz8B0W6b1JdRgsJu:-MO0UkeZ82j3ZN8eZK5H:b32wtd) ([raw file](https://github.com/dgraph-io/dgraph/blob/8448f8063cfa5c92577a7f2522c530fdb0e4a044/graphql/schema/auth_schemas_test.yaml#L62)):*
<details><summary><i>Previously, aman-bansal (aman bansal) wrote…</i></summary><blockquote>

It's because this error has been thrown by the gql parser itself. It fails before we get to check the value and validate it.
null from graphql translates into nil in go. but nil in graphql throws invalid error from the parser itself.
</blockquote></details>

ok
___
*[graphql/schema/auth_schemas_test.yaml, line 86 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNxFJENCPexkzZRhdys:-MO0UrAT1lLsQoVc_DbZ:b-9yb8kz) ([raw file](https://github.com/dgraph-io/dgraph/blob/8448f8063cfa5c92577a7f2522c530fdb0e4a044/graphql/schema/auth_schemas_test.yaml#L86)):*
<details><summary><i>Previously, aman-bansal (aman bansal) wrote…</i></summary><blockquote>

regex value provided in the auth rule should be of type string. This is needed to parse operand in regex. i have added value and reason in the message
</blockquote></details>

We should still remove `in this rule` from all errors because if a user had say 20 rules, its not clear what this rule refers to.

aman-bansal (commented):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7039)* status: 3 of 6 files reviewed, 9 unresolved discussions (waiting on @abhimanyusinghgaur, @MichaelJCompton, and @pawanrawal)

___
*[graphql/e2e/auth/auth_test.go, line 1923 at r2](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNXb-Cb8bCzoxp_OUQG:-MOMFTFhBUolOQ2t7Fdv:bb6un1j) ([raw file](https://github.com/dgraph-io/dgraph/blob/fd8c8d89276639c5726dfdd72dcb9e9e5a63c7ce/graphql/e2e/auth/auth_test.go#L1923)):*

Also seems like the query is common across the test cases so maybe define it above and just use the same one?
</blockquote></details>

Done
___
*[graphql/schema/auth.go, line 146 at r4](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MO0TrSn45cp9ejrITBa:-MOMFM-W78BkkwcMEwvr:bnw9mh4) ([raw file](https://github.com/dgraph-io/dgraph/blob/d587e13c0f69761242fafdc35620eda220ba60cd/graphql/schema/auth.go#L146)):*

Do you need three separate cases now or can they all be clubbed into the same case with

```
case "eq", "in", "regexp":
...
```
</blockquote></details>

It's done.
___
*[graphql/schema/auth.go, line 515 at r4](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MO0U72CBWL_XrKCD-hh:-MOMEzHu--MYa43Mkioq:bck0x4p) ([raw file](https://github.com/dgraph-io/dgraph/blob/d587e13c0f69761242fafdc35620eda220ba60cd/graphql/schema/auth.go#L515)):*

Remove `Reason: ` from all errors please. Also `this` rule doesn't tell which rule so best that it be removed. In the future we may also return the path of the rule.


The error should be like below.
```
.... operator has invalid value: %v. Array values for eq operator are not supported.
```
</blockquote></details>

Got it. I have fixed this.
___
*[graphql/schema/auth_schemas_test.yaml, line 86 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7039#-MNxFJENCPexkzZRhdys:-MOMEvMs4GSqecFqeWls:b-kwc09c) ([raw file](https://github.com/dgraph-io/dgraph/blob/8448f8063cfa5c92577a7f2522c530fdb0e4a044/graphql/schema/auth_schemas_test.yaml#L86)):*

We should still remove `in this rule` from all errors because if a user had say 20 rules, its not clear what this rule refers to.
</blockquote></details>

oh Got it. I have fixed the message.

pawanrawal (approved):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7039)* status: 2 of 6 files reviewed, 5 unresolved discussions (waiting on @abhimanyusinghgaur, @MichaelJCompton, and @pawanrawal)

PR #7048, opened 2020-12-02
fix: Check for nil ServerCloser in shutdown handler

ajeetdsouza:
The shutdown handler was attempting to signal a possible nil `ServerCloser`, causing panics on early Ctrl-C. This PR fixes that.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7048)


jarifibrahim (approved):
Test it thoroughly

PR #7053, opened 2020-12-03
fix(GraphQL): This PR allows repetition of fields inside implementing type which are present in interface and also allow to inherit field of same name of type ID  from multiple interfaces.

JatinDevDG:
Fixes GRAPHQL-672
Currently, we don't allow the repetition of fields in implementing types that are in the interface. But according to GraphQl specs, we should allow them. Not allowing them can also result in unexpected behavior in some scenarios.
For example, in the below example `Banana` type doesn't have any fields apart from that are in the Fruit interface. And we don't allow empty types. So allowing repetitive fields solve this problem.

```
interface Fruit {
    id: ID!
    price: Int!
}

type Apple implements Fruit {
    id: ID!
    price: Int!
    color: String!
}

type Banana implements Fruit {
    id: ID!
    price: Int!
}

```
Also, currently, if we have two interfaces and a type that implements both then we don't allow a field of the same name in both of them. We are allowing it for `ID` type fields because it's required in some use cases and there is no problem with it as `ID` type field is not actually a predicate.

```
interface I1 {
  f1: ID!
}
interface I2 {
 f1: ID!
}
type T implements I1 & I2 {
 f2: Int!
}

```
 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7053)

PR #7074, opened 2020-12-07
fix(GraphQL): This PR allows repetition of fields inside implementing type which are present in interface and also allow to inherit field of same name of type ID  from multiple interfaces. (#7053)

JatinDevDG:
Fixes GRAPHQL-672
Currently, we don't allow the repetition of fields in implementing types that are in the interface. But according to GraphQl specs, we should allow them. Not allowing them can also result in unexpected behavior in some scenarios.
For example, in the below example Banana type doesn't have any fields apart from that are in the Fruit interface. And we don't allow empty types. So allowing repetitive fields solve this problem.

```
interface Fruit {
    id: ID!
    price: Int!
}

type Apple implements Fruit {
    id: ID!
    price: Int!
    color: String!
}

type Banana implements Fruit {
    id: ID!
    price: Int!
}
```

Also, currently, if we have two interfaces and a type that implements both then we don't allow a field of the same name in both of them. We are allowing it for ID type fields because it's required in some use cases and there is no problem with it as ID type field is not actually a predicate.

```
interface I1 {
  f1: ID!
}
interface I2 {
 f1: ID!
}
type T implements I1 & I2 {
 f2: Int!
}
```

(cherry picked from commit a6f35046c708c3ebfece76195653320ab07ec441)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7074)

### DIFF
--- a/graphql/e2e/directives/schema.graphql
+++ b/graphql/e2e/directives/schema.graphql
@@ -105,11 +105,20 @@ interface Character @dgraph(type: "performance.character") {
 }
 
 type Human implements Character & Employee {
+        id: ID!
+        name: String! @search(by: [exact])
+        appearsIn: [Episode!] @search
+        bio: String @lambda
+        ename: String!
         starships: [Starship]
         totalCredits: Float @dgraph(pred: "credits")
 }
 
 type Droid implements Character @dgraph(type: "roboDroid") {
+        id: ID!
+        name: String! @search(by: [exact])
+        appearsIn: [Episode!] @search
+        bio: String @lambda
         primaryFunction: String
 }
 
--- a/graphql/e2e/normal/schema.graphql
+++ b/graphql/e2e/normal/schema.graphql
@@ -105,11 +105,20 @@ interface Character {
 }
 
 type Human implements Character & Employee {
+        id: ID!
+        name: String! @search(by: [exact])
+        appearsIn: [Episode!] @search
+        bio: String @lambda
+        ename: String!
         starships: [Starship]
         totalCredits: Float
 }
 
 type Droid implements Character {
+        id: ID!
+        name: String! @search(by: [exact])
+        appearsIn: [Episode!] @search
+        bio: String @lambda
         primaryFunction: String
 }
 
--- a/graphql/schema/gqlschema.go
+++ b/graphql/schema/gqlschema.go
@@ -630,7 +630,7 @@ func copyAstFieldDef(src *ast.FieldDefinition) *ast.FieldDefinition {
 
 // expandSchema adds schemaExtras to the doc and adds any fields inherited from interfaces into
 // implementing types
-func expandSchema(doc *ast.SchemaDocument) {
+func expandSchema(doc *ast.SchemaDocument) *gqlerror.Error {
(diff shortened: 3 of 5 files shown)


pawanrawal (approved):
It's looking good but feels like the code could have been structured better. 

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7074#-:-MNwfFIV5bxMFqKviBb6:b-bevxnd)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @JatinDevDG, @MichaelJCompton, and @pawanrawal)

___
*[graphql/e2e/normal/schema.graphql, line 108 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7074#-MNwcbXGAT-r_V7zRWdO:-MNwcbXGAT-r_V7zRWdP:bvzajgm) ([raw file](https://github.com/dgraph-io/dgraph/blob/93ed1e9fc96f542f351e15005f965a184c60acb7/graphql/e2e/normal/schema.graphql#L108)):*
> ```GraphQL
> 
> type Human implements Character & Employee {
>         id: ID!
> ```

It would have been good to keep this as it is since we already added these for the directives e2e test above? This would mean we test with both scenarios.
___
*[graphql/schema/gqlschema.go, line 654 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7074#-MNwe9Zx7Hxil1ES9jSv:-MNwe9Zx7Hxil1ES9jSw:b-6t9fbt) ([raw file](https://github.com/dgraph-io/dgraph/blob/93ed1e9fc96f542f351e15005f965a184c60acb7/graphql/schema/gqlschema.go#L654)):*
> ```Go
> 			fieldSeen := make(map[string]string)
> 			// fieldSeen a map from field name to interface name in which the field was seen.
> 			defFields := make(map[string]int64)
> ```

Why is this not of type `map[string]bool` instead?
___
*[graphql/schema/gqlschema.go, line 680 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7074#-MNwf4oi5XNXzNrWCVJD:-MNwf4oi5XNXzNrWCVJE:b-6uzn9) ([raw file](https://github.com/dgraph-io/dgraph/blob/93ed1e9fc96f542f351e15005f965a184c60acb7/graphql/schema/gqlschema.go#L680)):*
> ```Go
> 						} else if field.Type.NamedType != IDType {
> 							// If field definition is already written,just add interface definition in type
> 							// It will later results in validation error because of repeated fields
> ```

Do we have a unit test for this scenario?
___
*[graphql/schema/gqlschema.go, line 683 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7074#-MNweYQ83IhrBS8VbKkf:-MNweYQ83IhrBS8VbKkg:b7zptct) ([raw file](https://github.com/dgraph-io/dgraph/blob/93ed1e9fc96f542f351e15005f965a184c60acb7/graphql/schema/gqlschema.go#L683)):*
> ```Go
> 							fields = append(fields, copyAstFieldDef(field))
> 						}
> 					} else if field.Type.NamedType == IDType && fieldSeen[field.Name] != "" {
> ```

This `else if` could have been clubbed in the else below to make code easier to understand.

PR #7082, opened 2020-12-07
fix(logs): Fix logs in xidmap flush

ahsanbarkati:
We should only log "Writing to DB" when it is actually being written.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7082)

### DIFF
--- a/xidmap/xidmap.go
+++ b/xidmap/xidmap.go
@@ -276,19 +276,19 @@ func (m *XidMap) AllocateUid() uint64 {
 
 // Flush must be called if DB is provided to XidMap.
 func (m *XidMap) Flush() error {
+	if m.writer == nil {
+		return nil
+	}
 	glog.Infof("Writing xid map to DB")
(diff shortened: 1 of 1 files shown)


NamanJain8 (approved):

xidmap/xidmap.go
```diff
@@ -276,19 +276,19 @@ func (m *XidMap) AllocateUid() uint64 {
...
 	defer func() {
 		glog.Infof("Finished writing xid map to DB")
 	}()
 
-	if m.writer != nil && len(m.kvBuf) > 0 {
+	if len(m.kvBuf) > 0 {
 		m.kvChan <- m.kvBuf
 	}
 	close(m.kvChan)
 	m.wg.Wait()
```
NamanJain8, line 291: The `wg` and `kvChan` are used for the purpose of db flush only. So it should be okay to leave the channel open.

PR #7105, opened 2020-12-10
perf(GraphQL): Part-2: Remove unnecessary `dgraph.uid : uid` selections for auth queries (GRAPHQL-854)

abhimanyusinghgaur:
We don't need to query `uid` for auth queries, as they always have at least one field in their selection set. This optimizes the number of `touched_uids` greatly, while time is not much improved here.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7105)

### DIFF
--- a/graphql/resolve/auth_add_test.yaml
+++ b/graphql/resolve/auth_add_test.yaml
@@ -162,11 +162,8 @@
         inProject : Column.inProject {
           roles : Project.roles @filter(eq(Role.permission, "ADMIN")) {
             assignedTo : Role.assignedTo @filter(eq(User.username, "user1"))
-            dgraph.uid : uid
           }
-          dgraph.uid : uid
         }
-        dgraph.uid : uid
       }
       Ticket(func: uid(Ticket3)) @filter(uid(TicketAuth4)) {
         uid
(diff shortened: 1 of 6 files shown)


pawanrawal (approved):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7105)* status: all files reviewed, 1 unresolved discussion (waiting on @abhimanyusinghgaur and @MichaelJCompton)

___
*[graphql/resolve/query_rewriter.go, line 562 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7105#-MOBTOnp32vlxXY00wb0:-MOBTOnq76cWRX9WRMlr:b-h5m17c) ([raw file](https://github.com/dgraph-io/dgraph/blob/3bd909314e5a4f50988614d9096128f08f9bdab8/graphql/resolve/query_rewriter.go#L562)):*
> ```Go
> 	addArgumentsToField(dgQuery[0], field)
> 	selectionAuth := addSelectionSetFrom(dgQuery[0], field, authRw)
> 	// we don't need to query uid for auth queries, as they always have at least one field in their
> ```

What happens if the auth rule itself queries for the `id`, is it added in that case?

abhimanyusinghgaur (commented):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7105)* status: all files reviewed, 1 unresolved discussion (waiting on @MichaelJCompton and @pawanrawal)

___
*[graphql/resolve/query_rewriter.go, line 562 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7105#-MOBTOnp32vlxXY00wb0:-MOBUlpx91JeTGhO4-FS:bj94z2n) ([raw file](https://github.com/dgraph-io/dgraph/blob/3bd909314e5a4f50988614d9096128f08f9bdab8/graphql/resolve/query_rewriter.go#L562)):*

What happens if the auth rule itself queries for the `id`, is it added in that case?
</blockquote></details>

Yes, that would be added as `id: uid`.
These `dgraph.uid : uid` queries were being added in case there was no `uid` query already existing in that subgraph. This is useful for non-null checks in GraphQL, but not useful for auth queries.

PR #7161, opened 2020-12-17
fix(bulk): Fix memory held by tree in reduce phase

ahsanbarkati:
This issue was observed earlier as well and there was a PR fixing it. This PR brings that change back.
https://github.com/dgraph-io/dgraph/pull/4738

We still need to find the exact cause of why the tree is not GCed without explicitly setting it to nil, despite the parent struct set to nil. 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7161)


NamanJain8 (approved):
Approved as per https://github.com/dgraph-io/dgraph/pull/4738.

PR #7167, opened 2020-12-18
fix(test) : http test to be more resilient and fix race

aman-bansal:
HTTP tests have race in terms of how they use a token variable. being a global variable it's being rewritten and read by multiple goroutines and thus leading to errors like "Token expired etc.". This PR is to make them more resilient and fix race Also fixes a few in other test cases.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7167)

### DIFF
--- a/systest/mutations_test.go
+++ b/systest/mutations_test.go
@@ -1715,7 +1715,7 @@ func CountIndexConcurrentSetDelUIDList(t *testing.T, c *dgo.Dgraph) {
 	err := c.Alter(ctx, op)
 	require.NoError(t, err)
 
-	r := rand.New(rand.NewSource(time.Now().Unix()))
+	rand.Seed(time.Now().Unix())
 	maxUID := 100
 	txnTotal := uint64(1000)
 	txnCur := uint64(0)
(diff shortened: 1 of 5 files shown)


systest/mutations_test.go
```diff
@@ -1733,7 +1733,7 @@ func CountIndexConcurrentSetDelUIDList(t *testing.T, c *dgo.Dgraph) {
 				if atomic.AddUint64(&txnCur, 1) > txnTotal {
 					break
 				}
-				id := 2 + int(r.Int31n(int32(maxUID))) // 1 id subject id.
+				id := 2 + int(rand.Int31n(int32(maxUID))) // 1 id subject id.
```
jarifibrahim, line 1736: You're using the global rand which has locks and it will cause contention.

worker/proposal_test.go
```diff
@@ -79,12 +79,14 @@ func TestLimiterDeadlock(t *testing.T) {
 	go func() {
 		now := time.Now()
 		for range ticker.C {
+			l.c.L.Lock()
```
jarifibrahim, line 82: Why do you need the lock? For `l.iou`?

aman-bansal: There is concurrent write which is leading to the race condition. This fixes that.

systest/mutations_test.go
aman-bansal: yeah I understand but rand.NewSource is not concurrent safe. And we were having race condition here. This is test case so I thought it would be fine.

worker/proposal_test.go
jarifibrahim: Where is the concurrent write?

jarifibrahim: Got it.

jarifibrahim (approved):

PR #7176, opened 2020-12-21
Fix(DQL): Fix Aggregate Functions on empty data

vmrajas:
Motivation:
Currently, in case aggregate functions like avg, sum, min, max are queried on empty data, the float value 0.0 is returned. This is particularly annoying as even min/max on a string returns 0.0 . This PR fixes this and returns NULL instead.

Testing:
1. Added e2e tests to graphql/common
2. Added e2e tests to quer/query1_test.go

Fixes DGRAPH-2843

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7176)


pawanrawal (approved):
:lgtm: The commit description should say that this is a breaking change and what cases would be breaking i.e. a different result would be returned. 

> Reviewed 3 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7176#-:-MQGDS5X7K-rTkR5kSV4:b-qxz4kn)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @MichaelJCompton, @pawanrawal, and @vmrajas)

___
*[query/query.go, line 1723 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7176#-MQG9R5DBWOmWthc9b_0:-MQG9R5DBWOmWthc9b_1:bht00gc) ([raw file](https://github.com/dgraph-io/dgraph/blob/0b97981da803822d00c78d889e937a53a153f7b0/query/query.go#L1723)):*
> ```Go
> 	}
> 
> 	if sg.DestUIDs.Size() > 0 {
> ```

`len(sg.DestUIDs.GetUids()) > 0` would be a better check and is also `O(1)` as you are just getting the length of a slice

PR #7168, opened 2020-12-18
fix: Migrate debug tool to new DiskStorage

ajeetdsouza:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7168)

PR #7196, opened 2020-12-23
fix: Migrate debug tool to new DiskStorage (more efficient)

ajeetdsouza:
This is an alternate implementation of https://github.com/dgraph-io/dgraph/pull/7168. This should be much faster, but leaves us with a fragmented logFile (entry data is not stored contiguously).

This should not break anything, as we never directly iterate through data fields.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7196)

### DIFF
--- a/dgraph/cmd/debug/run.go
+++ b/dgraph/cmd/debug/run.go
@@ -890,7 +890,6 @@ func run() {
 		return
 	}
 
-	// If this is a new format WAL, print and return.
 	if isWal {
 		store, err := raftwal.InitEncrypted(dir, opt.key)
 		x.Check(err)
--- a/dgraph/cmd/debug/run.go
+++ b/dgraph/cmd/debug/run.go
@@ -906,22 +906,9 @@ func run() {
 	if isWal && !opt.oldWalFormat {
 		store, err := raftwal.InitEncrypted(dir, opt.key)
 		x.Check(err)
-		fmt.Printf("RaftID: %+v\n", store.Uint(raftwal.RaftId))
-		isZero := store.Uint(raftwal.GroupId) == 0
-
-		// TODO: Fix the pending logic.
-		pending := make(map[uint64]bool)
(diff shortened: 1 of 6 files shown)


manishrjain (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7196#-:-MQI6JSf6qjGpaCDzGBb:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @ajeetdsouza, @jarifibrahim, and @vvbalaji-dgraph)

___
*[raftwal/log.go, line 84 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7196#-MQI6BHQ9cnIgyMH6vST:-MQI6BHQ9cnIgyMH6vSU:byp6rt5) ([raw file](https://github.com/dgraph-io/dgraph/blob/0fa4880a2114c510d46d5f9ee7385089a2e737ea/raftwal/log.go#L84)):*
> ```Go
> 	e := entry(b)
> 	e.SetTerm(term)
> 	e.SetIndex(index)
> ```

Revert
___
*[raftwal/wal.go, line 125 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7196#-MQI4B-18VNky84YII0b:-MQI4B-18VNky84YII0c:b-lqipnc) ([raw file](https://github.com/dgraph-io/dgraph/blob/0fa4880a2114c510d46d5f9ee7385089a2e737ea/raftwal/wal.go#L125)):*
> ```Go
> 			// Here, we set the length field of the entry data to zero.
> 			// As long as we never directly iterate through the data section,
> 			// this operation is safe.
> ```

Just comment about the system where we truncate all the entries in the file, and then need to write a new entry at one of these indices.

truncated: Index 1 - 100.
Comment about a scenario where you'd write at index 99.

PR #7203, opened 2020-12-24
fix(ludicrous): Fix data race in executor

ahsanbarkati:
The operations on `inDeg` field of `mutation` struct should be done atomically because it is being used to check if a mutation in dependent on any other mutation. And the mutations are run concurrently, so this update in `inDeg` should be protected. This change fixes the data race which results in alpha crash due to calling `Done` on watermark which is lesser than `doneUntil` 

Fixes: [DGRAPH-2805](https://dgraph.atlassian.net/browse/DGRAPH-2805)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7203)

### DIFF
--- a/worker/executor.go
+++ b/worker/executor.go
@@ -130,9 +130,9 @@ func generateConflictKeys(ctx context.Context, p *subMutation) map[uint64]struct
 }
 
 type mutation struct {
+	inDeg              int64
 	sm                 *subMutation
 	conflictKeys       map[uint64]struct{}
-	inDeg              int
 	dependentMutations map[uint64]*mutation
 	graph              *graph
 }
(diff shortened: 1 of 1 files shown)


worker/executor.go
```diff
@@ -185,8 +185,8 @@ func (e *executor) worker(mut *mutation) {
 
 	// Decrease inDeg of dependents. If this mutation unblocks them, queue them.
 	for _, dependent := range mut.dependentMutations {
-		dependent.inDeg -= 1
-		if dependent.inDeg == 0 {
+		atomic.AddInt64(&dependent.inDeg, -1)
+		if atomic.LoadInt64(&dependent.inDeg) == 0 {
```
jarifibrahim, line 189:
```suggestion
		
		if atomic.AddInt64(&dependent.inDeg, -1) == 0 {
```

jarifibrahim (approved):
Does this fix the crash we saw @ahsanbarkati ?

NamanJain8:
Can you please add a brief description in PR description about why this should be atomic?

worker/executor.go
ahsanbarkati: It has been running fine since last night. I think this will fix the crash. I'll let it run for a day or two.

PR #7204, opened 2020-12-24
chore(raft-migrate): make raft-migrate a no-op if already migrated

NamanJain8:
Slash wants the raft-migrate to be a no-op if it is already on the newer version. We currently don't store any version (like we store `magic number` in badger). So, we need to check the encoding of protobuf data.
 
This PR makes raft-migrate a No-op if the raftwal is already at new format.
Reference: [Protobuf encoding format](https://developers.google.com/protocol-buffers/docs/encoding#structure)
In the new format, we prepend Key in the first 8 bytes of the Marshalled proposal. 
https://github.com/dgraph-io/dgraph/blob/6884b151c355669d1387f30696566f458d0a905f/dgraph/cmd/zero/raft.go#L77-L79
So, the first byte will be 0 if the node ID is less than 1<<24, which is a valid assumption for a reasonable cluster.

If we were on older cluster, then the entry.Data[0] would be storing the `wireType` and `fieldNum` which has to be a non-zero number. 
Quoting from above, the first byte of stream is key (which further consist of wireType and fieldNum). So if its zero, then we can infer that we are on new format.
`You now know that the first number in the stream is always a varint key, and here it's 08, or (dropping the msb):`

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7204)

### DIFF
--- a/dgraph/cmd/raft-migrate/run.go
+++ b/dgraph/cmd/raft-migrate/run.go
@@ -82,6 +82,51 @@ func parseAndConvertSnapshot(snap *raftpb.Snapshot) {
 	x.Check(err)
 }
 
+func shouldMigrate(oldWal *raftwal.DiskStorage, oldEntries []raftpb.Entry) bool {
+	isOldFormat := func(entry raftpb.Entry) bool {
+		if entry.Data[0] == 0 {
+			return false
+		}
+		return true
+	}
+
+	groupID := oldWal.Uint(raftwal.GroupId)
+	if groupID == 0 {
+		// For zero we need to verify both the Snapshot as well as raft-entries.
+		snapshot, err := oldWal.Snapshot()
+		x.Checkf(err, "failed to read snaphot %s", err)
+		var ms pb.MembershipState
+		if err := ms.Unmarshal(snapshot.Data); err != nil {
+			// We are unmarshalling the ZeroSnapshot into MembershipState. If this fails, this means
+			// we are on newer version.
+			fmt.Println("Zero snapshot unmarshalling failed => New format")
+			return false
+		}
+		for _, entry := range oldEntries {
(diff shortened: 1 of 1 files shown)


dgraph/cmd/raft-migrate/run.go
```diff
@@ -82,6 +82,51 @@ func parseAndConvertSnapshot(snap *raftpb.Snapshot) {
...
+			return false
+		}
+		for _, entry := range oldEntries {
+			// Raft commits an empty entry on becoming leader.
+			if entry.Type == raftpb.EntryConfChange || len(entry.Data) == 0 {
+				continue
+			}
+			// In new format, we prepend key to entry.Data in first 8 bytes. First 4 bytes out of
+			// those contain a node ID. If we assume node ID to be less than 1<<24. Then, if
+			// entry.Data[0] = 0, we can conclude that we are on new format.
```
ahsanbarkati, line 112: rather than this, can we try Unmarshalling `entry.Data[8:]`? if it fails then we are on oldWal.

jarifibrahim: > if it fails then we are on oldWal

But that won't be conclusive. Mostly it would return an error but if you're unlucky, you might be able to unmarshal it.

jarifibrahim (approved):
Please run this tool locally with all possible combinations.
Use a WAL directory that has a decent amount of valid data.

manishrjain (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7204#-:-MPKJ7pd6ogrsf6geIBk:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @ahsanbarkati, @danielmai, and @vvbalaji-dgraph)

PR #7208, opened 2020-12-25
netlify: make loop = true so that branch deploy gets published

mbj36:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7208)


mbj36:
Closing since we don't need this

PR #7226, opened 2020-12-30
fix(ludicrous): Fix data race in executor (#7203)

danielmai:
Cherry-pick of #7203.

The operations on inDeg field of mutation struct should be done atomically because 
it is being used to check if a mutation in dependent on any other mutation. And the 
mutations are run concurrently, so this update/read on inDeg should be protected.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7226)


danielmai:
This change was merged in https://github.com/dgraph-io/dgraph/pull/7307.

PR #7230, opened 2020-12-30
perf(GraphQL): JSON Part-1: JSON is generated by Dgraph for simple queries (GRAPHQL-877)

abhimanyusinghgaur:
This PR moves the GraphQL JSON generation for simple queries inside Dgraph.

Previously, the GraphQL layer would get a JSON response from Dgraph in the form expected for DQL. That JSON response had to go through the following pipeline:
1. Unmarshal
2. Run GraphQL completion
3. Marshal

After this pipeline was executed, then the output JSON would be in the form as expected by GraphQL.

Now, the response from Dgraph would already be in the form expected by GraphQL. So, we no more need to run the above pipeline. Hence, saving valuable time and memory.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7230)


pawanrawal (approved):
> Reviewed 20 of 20 files at r1.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7230)* status: all files reviewed, 2 unresolved discussions (waiting on @abhimanyusinghgaur, @manishrjain, @MichaelJCompton, and @vvbalaji-dgraph)

___
*[edgraph/server.go, line 915 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7230#-MQBz5Xv1nhZTceUxcEF:-MQBz5Xv1nhZTceUxcEG:bpsjzwx) ([raw file](https://github.com/dgraph-io/dgraph/blob/2bc6b3e376d00d44670ecd8939edf39cbaf2d41f/edgraph/server.go#L915)):*
> ```Go
> 	// graphql indicates whether the given request is from graphql admin or not.
> 	graphql bool
> 	// gqlField stores the GraphQL field for which the query is being processed.
> ```

This would be nil in the case when the request is not coming from /graphql endpoint.
___
*[graphql/resolve/resolver.go, line 565 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7230#-MQC-jAp6EC-BUqJMuQi:-MQC-jAqCLYiTFt-vaxy:b-yv3mk9) ([raw file](https://github.com/dgraph-io/dgraph/blob/2bc6b3e376d00d44670ecd8939edf39cbaf2d41f/graphql/resolve/resolver.go#L565)):*
> ```Go
> 	// a path to the 2nd item in the f list would look like:
> 	// - [ "q", "f", 2, "g" ]
> 	if res.Data == nil && !res.Field.Type().Nullable() {
> ```

Add a comment about this that According to GraphQL spec if any one query returns null but expected return type is non-nullable then we set root data to null.

PR #7237, opened 2021-01-01
perf(GraphQL): JSON Part-2: JSON is generated by Dgraph for simple mutations (GRAPHQL-877)

abhimanyusinghgaur:
This PR is a continuation of #7230.

It makes standard GraphQL mutations work with the JSON changes.

Each mutation now returns the GraphQL JSON required for that mutation as a byte array from the MutationResolver itself, so no further processing of the JSON is required.
Delete mutations now query the `queryField` explicitly before executing the mutation, to make it work with the JSON changes.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7237)


pawanrawal (approved):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7237)* status: all files reviewed, 2 unresolved discussions (waiting on @abhimanyusinghgaur, @manishrjain, @MichaelJCompton, and @vvbalaji-dgraph)

___
*[graphql/resolve/mutation.go, line 240 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7237#-MQC8QoK3VuY3cS0HW1i:-MQC8QoK3VuY3cS0HW1j:b-k749wz) ([raw file](https://github.com/dgraph-io/dgraph/blob/9e7e43cfaa1567af2bc32d87e90acb1ce8f7b264/graphql/resolve/mutation.go#L240)):*
> ```Go
> 			Data:  []byte(`{"` + mutation.ResponseName() + `":null}`),
> 			Field: mutation,
> 			// there is no completion down the pipeline, so error's path should be prepended with
> ```

the error path only contains the query field, so we prepend the mutation response name
___
*[graphql/resolve/mutation.go, line 397 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7237#-MQC9Ft52ERn9FletIEh:-MQC9Ft52ERn9FletIEi:b-agezt7) ([raw file](https://github.com/dgraph-io/dgraph/blob/9e7e43cfaa1567af2bc32d87e90acb1ce8f7b264/graphql/resolve/mutation.go#L397)):*
> ```Go
> 			} else {
> 				// need to write only the value returned for query field
> 				x.Check2(buf.Write(qryResult[4+len(f.ResponseName()) : len(qryResult)-1]))
> ```

Add another comment here about the value 4

PR #7242, opened 2021-01-04
fix(rollup): Periodically rollup keys

jarifibrahim:
This PR adds support for periodic rollups since the existing implementation
is ineffective in rollups. We scan the entire DB and try to rollup 1000 deltas in each scan.

This PR depends upon https://github.com/dgraph-io/badger/pull/1633

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7242)

### DIFF
--- a/posting/mvcc.go
+++ b/posting/mvcc.go
@@ -18,8 +18,10 @@ package posting
 
 import (
 	"bytes"
+	"context"
 	"encoding/hex"
 	"math"
+	"math/rand"
 	"strconv"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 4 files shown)


posting/mvcc.go
```diff
@@ -86,6 +87,53 @@ func (ir *incrRollupi) rollUpKey(writer *TxnWriter, key []byte) error {
...
+	stream := pstore.NewStreamAt(math.MaxUint64)
+	stream.Prefix = []byte{x.DefaultPrefix}
+	stream.Logger = nil // We don't want any logs. This can get very noisy.
+	stream.ChooseKey = func(item *badger.Item) bool {
+		pk, err := x.Parse(item.Key())
+		if err != nil {
+			glog.Errorf("RollupDeltas Parse failed: err = %s", err)
+			return false
+		}
+		if pk.IsData() && item.UserMeta()&BitDeltaPosting > 0 {
```
jarifibrahim, line 105: Maybe we should try to randomize this. I think it might be possible that some keys would never get rolled up. @manishrjain is there a way to iterate over all data keys?

posting/mvcc.go
```diff
@@ -86,6 +88,61 @@ func (ir *incrRollupi) rollUpKey(writer *TxnWriter, key []byte) error {
...
+			atomic.StoreUint32(&failed, 0)
+			ir.addKeyToBatch(item.KeyCopy(nil))
+			wg.Done()
+		}
+		return false
+	}
+	go func() {
+		wg.Wait()
```
jarifibrahim, line 115: This should use a closer

jarifibrahim:
After discussion with @manishrjain , we have decided to not do this since we will be scanning the entire DB to find deltas and this will be very expensive.

PR #7143, opened 2020-12-15
[BREAKING] opt(raftwal):  store only membership state to disk in the zero snapshot

NamanJain8:
In zero's raftwal meta, we used to store `MembershipState` only. The changes in https://github.com/dgraph-io/dgraph/pull/7096 made it store `ZeroSnapshot`. `Index` and `CheckpointTs` actually need not be stored. 
This PR makes raftwal meta store only `MembershipState`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7143)

PR #7249, opened 2021-01-05
[UNDO-BREAKING] fix(zero-snapshot): Do not break the raftwal format due to ZeroSnapshot

NamanJain8:
We introduced a change in https://github.com/dgraph-io/dgraph/pull/7143 that would have been breaking for the next release v21.03. That change should have been a part of `release/v20.11` but was mistakenly missed. We are doing a revert so that `release/v21.03` is not breaking due to that change. 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7249)


jarifibrahim (approved):
Looks good to me. Thanks for fixing this @NamanJain8.

PR #7258, opened 2021-01-07
fix(restore): Handle MaxUid=0 appropriately

jarifibrahim:
At the end of the restore operation, we update the MaxUid in zero. This
MaxUid can be zero if the backup was created from an empty DB.

Fixes DGRAPH-2804

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7258)

### DIFF
--- a/ee/backup/run.go
+++ b/ee/backup/run.go
@@ -15,9 +15,10 @@ package backup
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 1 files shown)


NamanJain8 (approved):
It's a harmless good to have change.  LGTM

PR #7261, opened 2021-01-08
fix(dgraph): Fix dgraph crash on windows

jarifibrahim:
There were two issues on windows
1. We were trying to delete a file with open file descriptors. This
doesn't work on windows. The PR
https://github.com/dgraph-io/ristretto/pull/242 in ristretto fixes this.
2. We were using path.Join instead of filepath.Join. The path package is
supposed to be used only on unix systems. In windows the "\" is the path
separator instead of "/".

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7261)


NamanJain8 (changes requested):
Shouldn't we abandon usages of `path` then? For example at https://github.com/dgraph-io/dgraph/blob/d13e7440e053aac6126d854a3519d0c072d865d3/raftwal/log.go#L95

Also go.mod would need to be updated once ristretto PR is merged.

jarifibrahim:
@NamanJain8 nice catch. I'll update all the occurrences of `path.join`

NamanJain8 (approved):

PR #7266, opened 2021-01-08
fix(race): fixing race condition with closer use + assign uids + bgindex test fix

aman-bansal:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7266)


jarifibrahim (changes requested):
Reduce the number of changes.

dgraph/cmd/zero/assign.go
```diff
@@ -196,7 +196,7 @@ func (s *Server) AssignUids(ctx context.Context, num *pb.Num) (*pb.AssignedIds,
 
 	select {
 	case <-ctx.Done():
-		return reply, ctx.Err()
+		return &emptyAssignedIds, ctx.Err()
```
jarifibrahim, line 199: why do you need to change this?

graphql/admin/shutdown.go
```diff
@@ -28,7 +28,7 @@ import (
 var (
 	// ServerCloser is used to signal and wait for other goroutines to return gracefully after user
 	// requests shutdown.
-	ServerCloser *z.Closer
+	ServerCloser = z.NewCloser(0)
```
jarifibrahim, line 31:
```suggestion
	ServerCloser z.NewCloser
```

systest/bgindex/docker-compose.yml
```diff
@@ -19,109 +19,8 @@ services:
       source: ../../ee/acl/hmac-secret
       target: /secret/hmac
       read_only: true
-    command: /gobin/dgraph alpha --my=alpha1:7080 --zero=zero1:5080,zero2:5080,zero3:5080
```
jarifibrahim, line 22: So this PR has 3 changes?
1. closer fix
2. Assign UID fix
3. bgindex fix?

aman-bansal: yes, this PR has above three fix. let me change the description also

dgraph/cmd/zero/assign.go
aman-bansal: There is a race condition with the use of reply. This happens when ctx.Done was called but lease is still working to assign uids

jarifibrahim (approved):

PR #7272, opened 2021-01-12
[BREAKING] feat(Dgraph): Read-Only replicas

manishrjain:
This PR allows a Dgraph Alpha or Zero instance to join as a "learner" node. That allows a user to spin up a replica instance across the world, without paying for the cost of latency. This replica can be used to serve best-effort queries really fast. It can also do both reads and writes, but would incur the latency of the network calls to the main cluster.

The main code change is small. But, due to the flag changes, this PR touches lots of config files.

Breaking: The breaking change here is how we pass in Raft flags. Before this PR, we used to have `--idx` flag. That flag is now gone. Both Alpha and Zero now use a "raft" flag, which takes `idx=N; group=M; learner=true` string based options.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7272)


manishrjain (commented):
:lgtm: Tested and reviewed with @jarifibrahim .

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7272#-:-MRKxiO5EkmCexkK-ocb:b-ghrfk)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @martinmr and @vvbalaji-dgraph)

___
*[dgraph/cmd/zero/zero.go, line 539 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7272#-MRKwpda061QVH6K14gf:-MRKwpdbDc7cl9ddnODZ:b-9dttjb) ([raw file](https://github.com/dgraph-io/dgraph/blob/41d118e604b6d5f74f543a4fb8c21bd1ab805427/dgraph/cmd/zero/zero.go#L539)):*
> ```Go
> 				return proposal
> 			}
> 			// TODO: Allow each Alpha to specify which group it should be part of.
> ```

remove this.

jarifibrahim (approved):

PR #7283, opened 2021-01-12
perf(GraphQL): JSON Part-5: Change query rewriting for scalar/object fields asked multiple times at same level (GRAPHQL-938)

abhimanyusinghgaur:
Previously, for a GraphQL query like this:

```
query {
    queryThing {
      i1: id
      i2: id
      name
      n: name
      n1: name
    }
}
```

GraphQL layer used to generate following DQL query:

```
query {
    queryThing(func: type(Thing)) {
      id : uid
      name : Thing.name
    }
}
```

This can’t work with the new way of JSON encoding for GraphQL because it expects that everything asked for in GraphQL query is present in DQL response. So, Now, the GraphQL layer generates the following DQL to make it work:

```
query {
    queryThing(func: type(Thing)) {
      Thing.i1 : uid
      Thing.i2: uid
      Thing.name : Thing.name
      Thing.n : Thing.name
      Thing.n1 : Thing.name
    }
}
```

Basically, the `DgraphAlias()` is now generated using the `TypeCondition.GraphqlAlias` format. That works even with fragments as they have their specific type conditions. Here are related thoughts:
```
query {
	queryHuman {
		name
		n: name
	}
	# DQL: use field.name as DgAlias()
	# queryHuman(func: type(Human)) {
	#   name: Human.name
	#   name: Human.name
	# }
	# => just using field.name doesn't work => use field.alias
	# queryHuman(func: type(Human)) {
	#   name: Human.name
	#   n: Human.name
	# }
	# but there are interfaces and fragments
	queryCharacter {
		... on Human {
			n: name
		}
		... on Droid {
			n: bio
		}
	}
	# DQL: use field.alias as DgAlias()
	# queryCharacter(func: type(Character)) {
	#   n: Character.name
	#   n: Character.bio
	# }
	# => just using field.alias doesn't work => use field.name + field.alias
	# queryCharacter(func: type(Character)) {
	#   name.n: Character.name
	#   bio.n: Character.bio
	# }
	# but the implementing types may have fields with same name not inherited from the interface
	queryThing {
		... on ThingOne {
			c: color
		}
		... on ThingTwo {
			c: color
		}
	}
	# DQL: use field.name + field.alias as DgAlias()
	# queryThing(func: type(Thing)) {
	#   color.c: ThingOne.color
	#   color.c: ThingTwo.color
	# }
	# => even that doesn't work => use Typename + field.name + field.alias
	# queryThing(func: type(Thing)) {
	#   ThingOne.color.c: ThingOne.color
	#   ThingTwo.color.c: ThingTwo.color
	# }
	# nice! but then different frags explicitly ask for an inherited field with same name & alias
	queryCharacter {
		... on Human {
			n: name
		}
		... on Droid {
			n: name
		}
	}
	# DQL: use Typename + field.name + field.alias as DgAlias()
	# queryCharacter(func: type(Character)) {
	#   Character.name.n: Character.name
	#   Character.name.n: Character.name
	# }
	# => doesn't work => use typeCond + Typename + field.name + field.alias
	# queryCharacter(func: type(Character)) {
	#   Human.Character.name.n: Character.name
	#   Droid.Character.name.n: Character.name
	# }
	# but wait, wouldn't just typeCond + field.alias work?
	# queryCharacter(func: type(Character)) {
	#   Human.n: Character.name
	#   Droid.n: Character.name
	# }
	# yeah! that works even for all the above cases.
	#
	# OR
	#
	# just appending the count when encountering duplicate alias at the same level also works.
	# But, we need to maintain the count map every time we need DgAlias().
	# Also, this requires query pre-processing and the generated aliases are non-deterministic.
	#
	# So, we are going with the typeCond + field.alias approach. That will work with @custom(dql:...) too.
}
```

### DIFF
--- a/graphql/e2e/common/common.go
+++ b/graphql/e2e/common/common.go
@@ -463,6 +463,42 @@ func assertUpdateGqlSchemaUsingAdminSchemaEndpt(t *testing.T, authority, schema 
 	AssertSchemaUpdateCounterIncrement(t, authority, oldCounter)
 }
 
+// JSONEqGraphQL compares two JSON strings obtained from a /graphql response.
+// To avoid issues, don't use space for indentation in expected input.
+//
+// The comparison requirements for JSON reported by /graphql are following:
+//  * The key order matters in object comparison, i.e.
+//        {"hello": "world", "foo": "bar"}
+//    is not same as:
+//        {"foo": "bar", "hello": "world"}
+//  * A key missing in an object is not same as that key present with value null, i.e.
+//        {"hello": "world"}
+//    is not same as:
+//        {"hello": "world", "foo": null}
+//  * Integers that are out of the [-(2^53)+1, (2^53)-1] precision range supported by JSON RFC,
+//    should still be encoded with full precision. i.e., the number 9007199254740993 ( = 2^53 + 1)
+//    should not get encoded as 9007199254740992 ( = 2^53). This happens in Go's standard JSON
+//    parser due to IEEE754 precision loss for floating point numbers.
+//
+// The above requirements are not satisfied by the standard require.JSONEq or testutil.CompareJSON
+// methods.
+// In order to satisfy all these requirements, this implementation just requires that the input
+// strings be equal after removing `\r`, `\n`, `\t` whitespace characters from the inputs.
+// TODO:
+//  Find a better way to do this such that order isn't mandated in list comparison.
+//  So that it is actually usable at places it is not used at present.
+func JSONEqGraphQL(t *testing.T, expected, actual string) {
+	expected = strings.ReplaceAll(expected, "\r", "")
+	expected = strings.ReplaceAll(expected, "\n", "")
+	expected = strings.ReplaceAll(expected, "\t", "")
+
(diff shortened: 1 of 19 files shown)


pawanrawal (approved):
> Reviewed 20 of 20 files at r1.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7283)* status: all files reviewed, 1 unresolved discussion (waiting on @abhimanyusinghgaur, @manishrjain, and @vvbalaji-dgraph)

___
*[graphql/e2e/common/query.go, line 3104 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7283#-MQvLUi89FdPj6mOHZW2:-MQvLUi89FdPj6mOHZW3:br7fzcp) ([raw file](https://github.com/dgraph-io/dgraph/blob/66e4489f541231e9e966aad8746de4830ed3a15c/graphql/e2e/common/query.go#L3104)):*
> ```Go
> 			"queryCountry": [{
> 				"name": "India",
> 				"ag": null,
> ```

We can set count to 0 inside this and `nameMin` to null.

PR #7241, opened 2021-01-04
fix(sort): Make sort consistent for indexed and without indexed predicates

ahsanbarkati:
This PR makes the result of sort consistent for predicate with and without index. There was an issue that for predicate with index, we used to drop the nodes which doesn't contain the predicate used for sorting, while we retained them for that of without index. 

The nulls were dropped in thee case of `sortWithIndex` because in this case to calculate the result, we used to take intersection of the given list for sorting with the index of the predicate used for sorting, hence other predicates were dropped. In this change, we keep track of nodes which have null sort predicates and append them to the result as required.

Some benchmarks:
The dataset contain 1million UIDs, with half of them containing the predicate used for sorting. The time taken are in nano seconds.
```
No Index:
Query                                   Time
orderasc, first:100 ->                  7092526995
orderasc, offset:100000, first:100 ->   6809554104
orderdesc, first:100 ->                 6981851622
orderdesc, offset:100000, first:100 ->  6862242806

Index:
orderasc, first:100 ->                  1421648793
orderasc, offset:100000, first:100 ->   2044206032
orderdesc, first:100 ->                 1577750157
orderdesc, offset:100000, first:100 ->  1956177772
```

The results of the same queries on **Master**:
```
No Index:
orderasc, first:100 ->                  6881212989
orderasc, offset:100000, first:100 ->   7543954929
orderdesc, first:100 ->                 7327315514
orderdesc, offset:100000, first:100 ->  7262242806

Index:
orderasc, first:100 ->                  1410905825
orderasc, offset:100000, first:100 ->   1890659562
orderdesc, first:100 ->                 1404256406
orderdesc, offset:100000, first:100 ->  2106243131
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7241)

PR #7304, opened 2021-01-15
Release/v20.11- Revert "fix(sort): Make sort consistent for indexed and without index…

ahsanbarkati:
…ed predicates (#7241) (#7301)"

This reverts commit ea435fcc30aeff6c829bb501cbef7c7609761628.

We are reverting this change because it is a breaking change for the response of sort query. We plan to release it in 21.03.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7304)


jarifibrahim (approved):
@ahsanbarkati please mention why we're reverting this commit.

PR #7325, opened 2021-01-18
fix(GraphQl):This PR fix a panic when we pass a single ID as a integer and expected type is [ID].We now coerce that to type array of string. 

JatinDevDG:
The following query was giving panic because here we passed ID as an int which is expected to be a string. 

```
query allStories {
      queryUser(filter: {
        id: 22
      }) {
        stories {
          id
          text
        }
      }
    }


```
We now added input coercion so that the ID type value will be coerced to string type. And if we give a slice of integer values
to ID type variable then that will be coerced to slice of integer values.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7325)


pawanrawal (approved):
> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7325)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JatinDevDG)

PR #7253, opened 2021-01-06
opt(rollup): change the way rollups are done

jarifibrahim:
**ISSUE:**
In a load that has a lot of updations to a key, a lot of deltas accumulate over time. These keys are added to the batch for a rollup. Due to the lossy behavior of rollups, they get dropped while the dgraph is busy rolling up keys (that might have much smaller delta count).

**Fix:**
This PR changes the way rollups are done.
- We now add keys to rollupBatch just after pushing the deltas over to badger.
- We now have a priority inside the keys that we rollup. The keys for which the deltas have increased up to a certain limit (hard-coded 500), we add it to high priority `rollupKeyPool` so that it has high chances to get rolled up and does not get dropped due to lossy behavior of rollups.

**Results:**
With a workload where a set of values are frequently updated:
- On master, the p directory size roars up to 6-7GB within 7-8 hours, and the `deltaCount` increases to 7-9k for lots of keys.
- On this branch, p directory size stays bounded in range 1-2GB even after >24 hours of workload.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7253)

PR #7329, opened 2021-01-19
fix(GraphQL):This PR will fix the graphl tests that were failing because of recent change in dgraph. 

JatinDevDG:
we have changed in Dgraph the way rollups are done in this PR. https://github.com/dgraph-io/dgraph/pull/7253
As a result of it, we were having below unexpected behavior in graphql while updating the schema.
**issue:**
suppose we send below schema requests,
```
1.schema1
2.schema2

```
Badger is supposed to return the following updates to graphql

1.schema1
1.schema1
1.schema1
2.schema2
2.schema2
2.schema2

If we get a schema update that is same as our previous schema then we are already ignoring that 
in graphql.For example, below multiple updates for schema1,schema2 from badger  gets ignored
```
1.schema1  //accepted,new update
2.schema1  //discarded,same as previous
3.schema1  //discarded,same as previous
4.schema1  //discared,same as previous
5.schema2  //accepted,new update 
6.schema2  //discarded,same as previous
7.schema2  //discarded,same as previous
```
But because of the recent change we are now sometimes getting older updates in random order like below.
```
1.schema1  //version1
2.schema2  //version2
3.schema1  //version1
```

so in this case graphql was assuming that we got an new schema which actually was old schema.
And due to that multiple errors were broken.

**Solution**
The solution is simple.Every update from the badger also have version associated to it.
And that version number is strictly increasing. Now, if we get an schema with version <=  current schema version
then we don't update the graphql schema.And that solve the issue

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7329)

### DIFF
--- a/dgraph/cmd/alpha/run.go
+++ b/dgraph/cmd/alpha/run.go
@@ -714,7 +714,6 @@ func run() {
 	// schema before calling posting.Init().
 	schema.Init(worker.State.Pstore)
 	posting.Init(worker.State.Pstore, postingListCacheSize)
-	admin.Init(worker.State.Pstore)
 	defer posting.Cleanup()
 	worker.Init(worker.State.Pstore)
 
--- a/dgraph/cmd/alpha/run.go
+++ b/dgraph/cmd/alpha/run.go
@@ -714,6 +714,7 @@ func run() {
 	// schema before calling posting.Init().
 	schema.Init(worker.State.Pstore)
 	posting.Init(worker.State.Pstore, postingListCacheSize)
+	admin.Init(worker.State.Pstore)
 	defer posting.Cleanup()
 	worker.Init(worker.State.Pstore)
 
--- a/graphql/admin/admin.go
+++ b/graphql/admin/admin.go
@@ -506,6 +506,7 @@ func newAdminResolver(
 		// Last update contains the latest value. So, taking the last update.
 		lastIdx := len(kvs.GetKv()) - 1
 		kv := kvs.GetKv()[lastIdx]
+
 		glog.Infof("Updating GraphQL schema from subscription.")
 
 		// Unmarshal the incoming posting list.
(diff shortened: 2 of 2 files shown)


pawanrawal (approved):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7329)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JatinDevDG)

___
*[graphql/admin/admin.go, line 540 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7329#-MRPPzMX83ha4VuxaVSk:-MRPPzMX83ha4VuxaVSl:bbrzzb2) ([raw file](https://github.com/dgraph-io/dgraph/blob/c60615a066a89b38bbb8d59873e1c21983925838/graphql/admin/admin.go#L540)):*
> ```Go
> 		if newSchema.Version <= server.schema.Version || newSchema.Schema == server.schema.Schema {
> 			glog.Infof("Skipping GraphQL schema update because either we got same schema or schema with older" +
> 				" version than version of current schema ")
> ```

Skipping GraphQL schema update, new badger key version is %d, the old version was %d.

JatinDevDG (commented):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7329)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @manishrjain, @pawanrawal, and @vvbalaji-dgraph)

___
*[graphql/admin/admin.go, line 540 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7329#-MRPPzMX83ha4VuxaVSk:-MRQIlVuEVbMGPE-bw8B:b-7rl0jd) ([raw file](https://github.com/dgraph-io/dgraph/blob/c60615a066a89b38bbb8d59873e1c21983925838/graphql/admin/admin.go#L540)):*

Skipping GraphQL schema update, new badger key version is %d, the old version was %d.
</blockquote></details>

done.

PR #7349, opened 2021-01-21
fix(subscriptions): fix subscription to use the kv with the max version

NamanJain8:
**Issue:**
The last KV from the list received from subscription need not be the latest update. This is because of the KVs written at the top due to value log rewrites or due to rollups. 
We were using the last KV from the updates received via subscription.

**Fix:**
Iterate over the KVs to get the KV with the latest version.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7349)

### DIFF
--- a/dgraph/cmd/alpha/run.go
+++ b/dgraph/cmd/alpha/run.go
@@ -808,9 +808,15 @@ func listenForCorsUpdate(closer *z.Closer) {
 	// Remove uid from the key, to get the correct prefix
 	prefix = prefix[:len(prefix)-8]
 	worker.SubscribeForUpdates([][]byte{prefix}, func(kvs *badgerpb.KVList) {
-		// Last update contains the latest value. So, taking the last update.
-		lastIdx := len(kvs.GetKv()) - 1
-		kv := kvs.GetKv()[lastIdx]
+		// Iterate over kvs to get the KV with the latest version. It is not necessary that the last
+		// KV contain the latest value.
+		var kv badgerpb.KV
+		for _, Kv := range kvs.GetKv() {
+			if Kv.GetVersion() > kv.GetVersion() {
+				kv = *Kv
+			}
+		}
+
 		glog.Infof("Updating cors from subscription.")
 		// Unmarshal the incoming posting list.
 		pl := &pb.PostingList{}
--- a/dgraph/cmd/alpha/run.go
+++ b/dgraph/cmd/alpha/run.go
@@ -808,15 +808,8 @@ func listenForCorsUpdate(closer *z.Closer) {
 	// Remove uid from the key, to get the correct prefix
 	prefix = prefix[:len(prefix)-8]
 	worker.SubscribeForUpdates([][]byte{prefix}, func(kvs *badgerpb.KVList) {
-		// Iterate over kvs to get the KV with the latest version. It is not necessary that the last
-		// KV contain the latest value.
-		var kv badgerpb.KV
-		for _, Kv := range kvs.GetKv() {
-			if Kv.GetVersion() > kv.GetVersion() {
-				kv = *Kv
-			}
-		}
 
+		kv := x.KvWithMaxVersion(kvs)
 		glog.Infof("Updating cors from subscription.")
 		// Unmarshal the incoming posting list.
 		pl := &pb.PostingList{}
--- a/dgraph/cmd/alpha/run.go
+++ b/dgraph/cmd/alpha/run.go
@@ -809,7 +809,7 @@ func listenForCorsUpdate(closer *z.Closer) {
 	prefix = prefix[:len(prefix)-8]
 	worker.SubscribeForUpdates([][]byte{prefix}, func(kvs *badgerpb.KVList) {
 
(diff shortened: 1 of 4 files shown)


edgraph/access_ee.go
```diff
@@ -351,7 +351,13 @@ func RefreshAcls(closer *z.Closer) {
...
 			return
 		}
-		if err := retrieveAcls(kvs.Kv[0].Version); err != nil {
+		var maxVs uint64
+		for _, kv := range kvs.GetKv() {
+			if kv.Version > maxVs {
+				maxVs = kv.Version
+			}
+		}
+		if err := retrieveAcls(maxVs); err != nil {
```
NamanJain8, line 360: Why were we using the `KV[0]` ? Confirm if we needed the `kv` with the minimum version instead.

edgraph/access_ee.go
```diff
@@ -351,7 +351,13 @@ func RefreshAcls(closer *z.Closer) {
 		if kvs == nil || len(kvs.Kv) == 0 {
 			return
 		}
```
NamanJain8, line 353: Is it even possible that we get an empty list? If yes, then we should have this check at other places as well.

dgraph/cmd/alpha/run.go
```diff
@@ -808,9 +808,15 @@ func listenForCorsUpdate(closer *z.Closer) {
...
 	prefix = prefix[:len(prefix)-8]
 	worker.SubscribeForUpdates([][]byte{prefix}, func(kvs *badgerpb.KVList) {
-		// Last update contains the latest value. So, taking the last update.
-		lastIdx := len(kvs.GetKv()) - 1
-		kv := kvs.GetKv()[lastIdx]
+		// Iterate over kvs to get the KV with the latest version. It is not necessary that the last
+		// KV contain the latest value.
+		var kv badgerpb.KV
+		for _, Kv := range kvs.GetKv() {
+			if Kv.GetVersion() > kv.GetVersion() {
```
jarifibrahim, line 815: Check the key as well.

NamanJain8: Why do we need to check key?

edgraph/access_ee.go
NamanJain8: We want the latest version.

dgraph/cmd/alpha/run.go
NamanJain8: I don't think there will be any benefit in comparing the key, unless we suspect that badger might have sent the key which does not match the subscribed prefixes.

pawanrawal (approved):
> Reviewed 4 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7349)* status: all files reviewed, 4 unresolved discussions (waiting on @jarifibrahim, @manishrjain, @NamanJain8, and @vvbalaji-dgraph)

___
*[x/x.go, line 1219 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7349#-MRdtxYb-R9I6IagPBM5:-MRdtxYb-R9I6IagPBM6:b-p68g8e) ([raw file](https://github.com/dgraph-io/dgraph/blob/be6fe3e8f5ecf9502c5fa368c2603a5ea82062ae/x/x.go#L1219)):*
> ```Go
> 	// Iterate over kvs to get the KV with the latest version. It is not necessary that the last
> 	// KV contain the latest value.
> 	var maxKv badgerpb.KV
> ```

Why not define it like 
```
var maxKv *badgerpb.KV
```

and then below you can do
maxKv = kv

...
return maxKv

NamanJain8 (commented):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7349)* status: 3 of 4 files reviewed, 4 unresolved discussions (waiting on @jarifibrahim, @manishrjain, @pawanrawal, and @vvbalaji-dgraph)

___
*[dgraph/cmd/alpha/run.go, line 815 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7349#-MR_CZLuUcUeSCsKo6M4-r1-815:-MReJBgR9n3HsYCX0Bop:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/0593906e1809e621a337f99af67c496ba23f70d3/dgraph/cmd/alpha/run.go#L815)):*
<details><summary><i>Previously, NamanJain8 (Naman Jain) wrote…</i></summary><blockquote>

I don't think there will be any benefit in comparing the key, unless we suspect that badger might have sent the key which does not match the subscribed prefixes.
</blockquote></details>

Done.
___
*[x/x.go, line 1219 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7349#-MRdtxYb-R9I6IagPBM5:-MReJDeb5zCN4E4kIc8X:b-i21bbo) ([raw file](https://github.com/dgraph-io/dgraph/blob/be6fe3e8f5ecf9502c5fa368c2603a5ea82062ae/x/x.go#L1219)):*

Why not define it like 
```
var maxKv *badgerpb.KV
```

and then below you can do
maxKv = kv

...
return maxKv
</blockquote></details>

Thanks. Done.

PR #7356, opened 2021-01-22
Avoid printing decoder version info and page size

OmarAyo:
As of now all dgraph processes will pring this two lines at the very begining:
```
[Decoder]: Using assembly version of decoder
Page Size: 4096
```
This PR will avoid printing these two lines that break the shell script that the `completion` command will generate

As of now the completion command will generate the following shell script:
```
[Decoder]: Using assembly version of decoder
Page Size: 4096
# bash completion for dgraph                               -*- shell-script -*-

__dgraph_debug()
{
    if [[ -n ${BASH_COMP_DEBUG_FILE} ]]; then
        echo "$*" >> "${BASH_COMP_DEBUG_FILE}"
    fi
}
...
..
.
```
As you can see it has the first two lines reporting info about decoder and pagesize. In this case, one has to edit the generated shell script and remove these two line.

This PR should remove these lines in order to get a functional shell script

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7356)


danielmai (changes requested):
I agree that this shouldn't be printed for all subcommands. If this info is still needed though, we can print these out for each subcommand where it's meaningful (e.g., Zero and Alpha)

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7356#-:-MRzMPCc7p1puwTsoY--:b-miv2ii)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @manishrjain and @vvbalaji-dgraph)

PR #7392, opened 2021-02-03
perf(dgraph) - Use badger sinceTs in backups

jarifibrahim:
This PR improves incremental backups.
We set the sinceTs in the stream framework to skip data that is below the sinceTs. Since the schema  and type are written at timestamp 1, we need to separately process and store them.

Time taken to backup 21 million -> 1.11 minutes
**Time taken to create an incremental backup with 2 new nquads**
Master - 5.8 seconds
This PR - 1 second

Fixes DGRAPH-2958

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7392)


worker/backup_processor.go
```diff
@@ -157,14 +181,8 @@ func (pr *BackupProcessor) WriteBackup(ctx context.Context) (*pb.BackupResponse,
...
 		}
 
-		// Backup type keys in every group.
-		if parsedKey.IsType() {
-			return true
-		}
-
-		// Only backup schema and data keys for the requested predicates.
-		_, ok := predMap[parsedKey.Attr]
-		return ok
```
NamanJain8, line 167: We should have this back. Otherwise, we will also backup the stale data. 
Also, handle this case in schema.

jarifibrahim: Nice catch. Fixed it.

worker/backup_processor.go
```diff
@@ -246,8 +328,7 @@ func (tl *threadLocal) toBackupList(key []byte, itr *badger.Iterator) (
 	var dropOp *pb.DropOperation
 
 	item := itr.Item()
-	if item.UserMeta() != posting.BitSchemaPosting &&
-		(item.Version() < tl.Request.SinceTs || item.IsDeletedOrExpired()) {
+	if item.Version() < tl.Request.SinceTs || item.IsDeletedOrExpired() {
```
jarifibrahim, line 331: Log an error if this happens.

PR #7478, opened 2021-02-24
perf(txn): de-duplicate the context keys and predicates

NamanJain8:
The zero raft loop got stuck because the proposals took a long time to process. Hence, the goroutine which checks the quorum gets struck leading to leadership change (even in single zero single alpha). 

**Update:** The real issue was the Keys in txnContext. While merging the conflict keys, deduplication was not done. Do deduplication before sending the request to zero. 

The issue is due to lock contention.
https://github.com/dgraph-io/dgraph/blob/c66e86985b642eb6b563014ed05389e7b3170f72/dgraph/cmd/zero/oracle.go#L132-L151 is holding the lock, while this function tries to send update https://github.com/dgraph-io/dgraph/blob/c66e86985b642eb6b563014ed05389e7b3170f72/dgraph/cmd/zero/oracle.go#L273-L283.

The processing of the keyCommit map is slow and the locks for o.commits and o.keyCommit can be separated by using `sync.Map` for o.commits.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7478)

### DIFF
--- a/dgraph/cmd/zero/oracle.go
+++ b/dgraph/cmd/zero/oracle.go
@@ -367,12 +367,7 @@ func (s *Server) commit(ctx context.Context, src *api.TxnContext) error {
 
 	checkPreds := func() error {
 		// Check if any of these tablets is being moved. If so, abort the transaction.
-		preds := make(map[string]struct{})
-
-		for _, k := range src.Preds {
-			preds[k] = struct{}{}
-		}
-		for pkey := range preds {
+		for _, pkey := range src.Preds {
 			splits := strings.Split(pkey, "-")
 			if len(splits) < 2 {
 				return errors.Errorf("Unable to find group id in %s", pkey)
--- a/dgraph/cmd/zero/oracle.go
+++ b/dgraph/cmd/zero/oracle.go
@@ -83,7 +83,7 @@ func (o *Oracle) hasConflict(src *api.TxnContext) bool {
 	if src.StartTs < o.startTxnTs {
 		return true
 	}
-	for k := range src.Keys {
+	for _, k := range src.Keys {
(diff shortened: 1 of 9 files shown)


dgraph/cmd/zero/oracle.go
```diff
@@ -130,8 +143,8 @@ func (o *Oracle) purgeBelow(minTs uint64) {
 }
 
 func (o *Oracle) commit(src *api.TxnContext) error {
-	o.Lock()
-	defer o.Unlock()
+	o.keyCommitLock.Lock()
```
jarifibrahim, line 146: This might not work.

manishrjain (commented):
Move the proposal application out of the Zero Raft loop, into an applyCh style application, like we do in Alpha. That would tackle the slowness issues.

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7478#-:-MUiNRJa0R8l_5xFBd3Y:b-mz586f)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @NamanJain8 and @vvbalaji-dgraph)

NamanJain8 (commented):
As discussed the real issue was with txnContext.Keys containing duplicates. So, we don't need that Raft loop change.

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7478#-:-MUoApHJ0veXH5yHBOab:b-b6iaau)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @martinmr, @NamanJain8, and @vvbalaji-dgraph)

jarifibrahim (approved):
Nice work @NamanJain8 :tada:

PR #7480, opened 2021-02-24
Fix(Dgraph): Throttle number of files to open while schema update

vmrajas:
Motivation:
Currently, while doing GraphQL schema update, many files are opened simultaneously for the purpose of indexing predicates. This may create a problem and throw error, "Too many open files" if ulimit is not set properly.
This PR fixes this and puts a limit on the number of threads which create and open files.

Note that this is only meant to address this problem on an empty database. The error may still be thrown (with lower frequency) if the schema update is done on a large dataset.

Testing:
1. Added e2e test.
2. Tested locally with a large schema (around 4000 indexed fields)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7480)

### DIFF
--- a/dgraph/cmd/bulk/run.go
+++ b/dgraph/cmd/bulk/run.go
@@ -372,7 +372,7 @@ func maxOpenFilesWarning() {
 		yellow = "\x1b[33m"
 		reset  = "\x1b[0m"
 	)
-	maxOpenFiles, err := queryMaxOpenFiles()
+	maxOpenFiles, err := worker.QueryMaxOpenFiles()
 	if err != nil || maxOpenFiles < 1e6 {
 		fmt.Println(green + "\nThe bulk loader needs to open many files at once. This number depends" +
 			" on the size of the data set loaded, the map file output size, and the level" +
--- a/dgraph/cmd/bulk/run.go
+++ b/dgraph/cmd/bulk/run.go
@@ -372,7 +372,7 @@ func maxOpenFilesWarning() {
 		yellow = "\x1b[33m"
 		reset  = "\x1b[0m"
 	)
-	maxOpenFiles, err := worker.QueryMaxOpenFiles()
+	maxOpenFiles, err := x.QueryMaxOpenFiles()
 	if err != nil || maxOpenFiles < 1e6 {
 		fmt.Println(green + "\nThe bulk loader needs to open many files at once. This number depends" +
 			" on the size of the data set loaded, the map file output size, and the level" +
--- a/worker/mutation.go
+++ b/worker/mutation.go
@@ -24,8 +24,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"golang.org/x/sys/unix"
-
 	"github.com/dgraph-io/badger/v3/y"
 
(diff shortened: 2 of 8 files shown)


dgraph/cmd/bulk/run.go
```diff
@@ -372,7 +372,7 @@ func maxOpenFilesWarning() {
 		yellow = "\x1b[33m"
 		reset  = "\x1b[0m"
 	)
-	maxOpenFiles, err := queryMaxOpenFiles()
+	maxOpenFiles, err := worker.QueryMaxOpenFiles()
```
NamanJain8, line 375: Move this to x package maybe.

worker/mutation.go
```diff
@@ -195,6 +200,19 @@ func runSchemaMutation(ctx context.Context, updates []*pb.SchemaUpdate, startTs
...
 	defer wg.Done()
+	// This throttle allows is used to limit the number of files which are opened simultaneously
+	// by badger while building indexes for predicates in background.
+	maxOpenFileLimit, err := QueryMaxOpenFiles()
+	if err != nil {
+		// Setting to default value on unix systems
+		maxOpenFileLimit = 1024
+	}
+	glog.Infof("Max open files limit: %d", maxOpenFileLimit)
+	// Badger opens around 7 files for indexing per predicate.
```
NamanJain8, line 211: How did you calculate the number 7?

worker/mutation.go
```diff
@@ -195,6 +200,19 @@ func runSchemaMutation(ctx context.Context, updates []*pb.SchemaUpdate, startTs
...
+	// The throttle limit is set to maxOpenFileLimit/8 to ensure that indexing does not throw
+	// "Too many open files" error.
+	throttle := y.NewThrottle(maxOpenFileLimit / 8)
```
jarifibrahim, line 214: This won't work. Each badger instance could open X number of files. We have no way of knowing how many files would badger open.

If you set throttle to 1024/8 = 128, you are assuming each badger instance will open at max 8 files which is wrong. This wouldn't work.

This isn't easy to fix. A simpler thing to do would be to ask the user to increate the ulimit on the system.

jarifibrahim (changes requested):
This wouldn't solve the problem. Please see my comment.

dgraph/cmd/bulk/run.go
vmrajas: Done with the latest update.

worker/mutation.go
vmrajas: This PR to address the error of "Too many open files" on empty database. I arrived at this number to account for the files, MANIFEST, LOCK, KEYREGISTRY, DISCARD, 00001.mem and 00001.vlog

worker/mutation.go
vmrajas: Had a discussion with @jarifibrahim today afternoon. He has agreed that the solution would work in case of an empty database. I have addressed Naman's comments and fixed tests. @jarifibrahim can you take a look at the revised PR.

jarifibrahim (approved):

abhimanyusinghgaur (approved):

PR #7481, opened 2021-02-24
chore(test): making create namespace in test more robust

aman-bansal:
---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7481)

### DIFF
--- a/systest/backup/multi-tenancy/backup_test.go
+++ b/systest/backup/multi-tenancy/backup_test.go
@@ -60,7 +60,7 @@ func TestBackupMultiTenancy(t *testing.T) {
 	galaxyToken := testutil.Login(t, galaxyCreds)
 
 	// Create a new namespace
-	ns, err := testutil.CreateNamespace(t, galaxyToken)
+	ns, err := testutil.CreateNamespaceWithRetry(t, galaxyToken)
 	require.NoError(t, err)
 	dg1 := testutil.DgClientWithLogin(t, "groot", "password", ns)
 
--- a/testutil/multi_tenancy.go
+++ b/testutil/multi_tenancy.go
@@ -24,6 +24,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/golang/glog"
+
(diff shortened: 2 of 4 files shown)


testutil/multi_tenancy.go
```diff
@@ -74,10 +76,19 @@ func CreateNamespace(t *testing.T, token *HttpToken) (uint64, error) {
 	params := GraphQLParams{
 		Query: createNs,
 	}
-	resp := MakeRequest(t, token, params)
-	if len(resp.Errors) > 0 {
-		return 0, errors.Errorf(resp.Errors.Error())
+	var resp *GraphQLResponse
+	for {
+		resp = MakeRequest(t, token, params)
```
ahsanbarkati, line 81: Lets add a small sleep between two retries.

ahsanbarkati (approved):
Looks good to me

NamanJain8 (changes requested):
I think we should instead add retries here. https://github.com/dgraph-io/dgraph/blob/2a43fa3251c41aacb92e80a84e9cfe9786da3daa/edgraph/multi_tenancy_ee.go#L115-L117

Else we will have created lots of namespaces and `/state` endpoint will be polluted.

systest/backup/multi-tenancy/backup_test.go
```diff
@@ -60,7 +60,7 @@ func TestBackupMultiTenancy(t *testing.T) {
 	galaxyToken := testutil.Login(t, galaxyCreds)
 
 	// Create a new namespace
-	ns, err := testutil.CreateNamespace(t, galaxyToken)
+	ns, err := testutil.CreateNamespaceWithRetry(t, galaxyToken)
```
NamanJain8, line 63: We don't need `CreateNamespaceWithRetry` now.

aman-bansal: we actually do. even if fails after retry we don't want tests to fail. this will retry to get a successful response

NamanJain8 (approved):

PR #7400, opened 2021-02-05
feat(multi-tenancy): Make GraphQL work with multiple namespaces

pawanrawal:
This PR also removes the schema history feature. If needed it can be reimplemented in Slash GraphQL.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7400)

PR #7489, opened 2021-02-26
fix(GraphQL): Fix flaky tests caused by multiple increments of schemaUpdateCounter.

JatinDevDG:
Fixes GRAPHQL-1050
There were a couple of tests that were behaving flakily because for a single schema update we were incrementing `schemaUpdateCounter` multiple times, in case the schema update was for a non-Galaxy namespace. This behavior was caused by a bug in schema lazy loading which got added recently with multi-tenancy in PR #7400.

Consider a scenario when we don't have any schema stored in the admin GraphQL server for a non-Galaxy namespace.
Sending a schema POST request at` /admin/schema` gives a badger subscription update for this schema. Since we don't have any schema in the admin GraphQL server, we update the `schemaUpdateCounter` but don't update the schema in both the admin and the main GraphQL server, because this schema should be loaded lazily when the main GraphQL server receives a GraphQL request.

Now, what happens is, badger gives another update from the subscription, which contains the same schema with a higher version. Since the admin server didn't store any information about this schema previously, so it is not able to decide whether to reject this update. Resulting in the `schemaUpdateCounter` being incremented again. Yet, nothing stored in the admin server about the schema.

This results in the `schemaUpdateCounter` being incremented multiple times for a single schema update request.
This behavior persists until the main GraphQL server doesn't get a request for that namespace resulting in the lazy loading of the schema, which both stores the schema in the admin server and loads it into the main server. Once that happens, updating schema once would only update the `schemaUpdateCounter` once as well.

So, to fix this bug we introduced a new variable `loaded`  in `gqlSchema` struct. This flag is set to true whenever a schema is loaded in the main GraphQL server. If there is no schema stored in the admin server or if this flag is false and we get an update from badger subscription, then we just store the `gqlSchema` on the admin server with this flag set to false, but don't load it into the main GraphQL server. Only if this flag were true, then the schema would be loaded in the main GraphQL server. This mechanism lets the admin server know correctly when to increment the `schemaUpdateCounter`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7489)


abhimanyusinghgaur (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7489#-:-MUSp3nL49xI91R3WChO:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @JatinDevDG and @pawanrawal)

___
*[graphql/admin/admin.go, line 579 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7489#-MUSoVhnCP-p2gLoDypk:-MUSoVhnCP-p2gLoDypl:b-pb4vzi) ([raw file](https://github.com/dgraph-io/dgraph/blob/2afaf4f03067fc4ea0b0fb3f9324c8307c4b9982/graphql/admin/admin.go#L579)):*
> ```
>  ns
> ```
it isn't used in the message, should be removed.
___
*[graphql/admin/admin.go, line 584 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7489#-MUSorBZ5eAEOjY1JyHs:-MUSorBZ5eAEOjY1JyHt:bupwbmp) ([raw file](https://github.com/dgraph-io/dgraph/blob/2afaf4f03067fc4ea0b0fb3f9324c8307c4b9982/graphql/admin/admin.go#L584)):*
> ```
> newSchema.loaded = currentSchema.loaded
> ```
lets explicitly set it to `true` as `currentSchema.loaded` would be `true` if it reached here

JatinDevDG (commented):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7489)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @abhimanyusinghgaur and @pawanrawal)

___
*[graphql/admin/admin.go, line 579 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7489#-MUSoVhnCP-p2gLoDypk:-MUSqv6v0fnSf-EjGWEe:b-7rl0jd) ([raw file](https://github.com/dgraph-io/dgraph/blob/2afaf4f03067fc4ea0b0fb3f9324c8307c4b9982/graphql/admin/admin.go#L579)):*

> ```
>  ns
> ```
it isn't used in the message, should be removed.
</blockquote></details>

done.
___
*[graphql/admin/admin.go, line 584 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7489#-MUSorBZ5eAEOjY1JyHs:-MUSqwW60j0lIFvb72Cc:b-7rl0jd) ([raw file](https://github.com/dgraph-io/dgraph/blob/2afaf4f03067fc4ea0b0fb3f9324c8307c4b9982/graphql/admin/admin.go#L584)):*

> ```
> newSchema.loaded = currentSchema.loaded
> ```
lets explicitly set it to `true` as `currentSchema.loaded` would be `true` if it reached here
</blockquote></details>

done.

PR #7505, opened 2021-03-02
fix(multi-tenancy): fix live loader for case when namespace does not exist for data

NamanJain8:
Fixes DGRAPH-3075

Earlier, when live-loading data through Guardians of Galaxy, the NQuads make their way to the DB even if the namespace does not exist for them. This might lead to very interesting bugs.

This PR fixes it by extracting the namespaces by querying the schema. If we encounter a Nquad or schema for which namespace does not exist, we error out and live loader stops.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7505)

### DIFF
--- a/dgraph/cmd/live/run.go
+++ b/dgraph/cmd/live/run.go
@@ -259,6 +259,7 @@ func (l *loader) processSchemaFile(ctx context.Context, file string, key x.Sensi
 	op := &api.Operation{}
 	op.Schema = string(b)
 	if opt.namespaceToLoad == math.MaxUint64 {
+		// Verify schema if we are loding into multiple namespaces.
 		if err := validateSchema(op.Schema, l.namespaces); err != nil {
 			return err
 		}
@@ -733,7 +734,7 @@ func run() error {
 		}
 	}()
 	ctx := context.Background()
-	galaxyOp := false
+	var galaxyOp bool
 	if len(creds.GetString("user")) > 0 && creds.GetUint64("namespace") == x.GalaxyNamespace &&
 		opt.namespaceToLoad != x.GalaxyNamespace {
 		galaxyOp = true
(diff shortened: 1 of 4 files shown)


dgraph/cmd/live/run.go
```diff
@@ -618,6 +652,33 @@ func setup(opts batchMutationOptions, dc *dgo.Dgraph, conf *viper.Viper) *loader
 	return l
 }
 
+// populateNamespace fetches the schema and extracts the information about the existing namespaces.
+func (l *loader) populateNamespaces(ctx context.Context, dc *dgo.Dgraph, isGalaxyOp bool) error {
+	if !isGalaxyOp {
+		l.namespaces[opt.namespaceToLoad] = struct{}{}
```
jarifibrahim, line 658: why do we do this? Please add a comment.
Also, why do this only for the GalaxyOp?

NamanJain8: Done. Thanks.

dgraph/cmd/live/run.go
```diff
@@ -703,8 +771,20 @@ func run() error {
 	l := setup(bmOpts, dg, Live.Conf)
 	defer l.zeroconn.Close()
 
+	if err := l.populateNamespaces(ctx, dg, galaxyOp); err != nil {
+		fmt.Printf("Error while populating namespaces %s\n", err)
+		return err
+	}
+
+	if opt.namespaceToLoad != math.MaxUint64 {
```
jarifibrahim, line 779: Please create a constant for math.MaxUint64 and give it a better name :)

dgraph/cmd/live/run.go
```diff
@@ -673,8 +737,12 @@ func run() error {
 		}
 	}()
 	ctx := context.Background()
+	var galaxyOp bool
```
jarifibrahim, line 740: Rename this `singleNSOp`. And add a comment to explain what it means.

jarifibrahim (approved):

dgraph/cmd/live/run.go
NamanJain8: Made `preserveNs` a part of `opt`.

PR #7519, opened 2021-03-04
Fix(Dgraph): Change max open file descriptor dividing factor.

vmrajas:
Currently, to reduce frequency of "Too many open files" error on schema update, we have a throttle which allows a maximum of maxOpenFiles / 8 files to open at a time.
After having an internal discussion, it was found out that a factor of 8 could be insufficient to prevent such errors. Moreover, a factor of 64 also found out to not change latency of updates much.

Benchmarks: (Update on empty schema with factors )
1. maxOpenfiles / 8 : 11.040 secs.
4. maxOpenFiles / 64: 8.673 secs.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7519)

### DIFF
--- a/worker/mutation.go
+++ b/worker/mutation.go
@@ -209,9 +209,9 @@ func runSchemaMutation(ctx context.Context, updates []*pb.SchemaUpdate, startTs 
 	}
 	glog.Infof("Max open files limit: %d", maxOpenFileLimit)
 	// Badger opens around 8 files for indexing per predicate.
-	// The throttle limit is set to maxOpenFileLimit/8 to ensure that indexing does not throw
+	// The throttle limit is set to maxOpenFileLimit/64 to ensure that indexing does not throw
 	// "Too many open files" error.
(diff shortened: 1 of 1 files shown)


jarifibrahim (commented):
@vmrajas You're reducing the number of parallel indexing operations that can run. How is size/8 faster than size/64?
I would expect size/64 to be slower since you'll have fewer indexing operations running at the same time.

vmrajas:
@jarifibrahim , Yes, that was the expectation. I was trying to keep the factor as low as possible.
My best guess is that this is due to the overhead of managing more threads.

jarifibrahim:
> @jarifibrahim , Yes, that was the expectation. I was trying to keep the factor as low as possible.
> My best guess is that this is due to the overhead of managing more threads.

I don't think that's the case. Golang can handle thousands of goroutines easily at the same time.

vmrajas:
Closing the PR after having an internal discussion that this will make updateSchema much slower and may also not benefit much in reducing the frequency of "Too many open files" error.

PR #7522, opened 2021-03-08
Chore(GraphQL): Fix flaky behaviour of TestLargeSchemaUpdate

vmrajas:
Motivatoin:
TestLargeSchemaUpdate test was failing when race flag is set to true in "Periodic Test All" tests. This PR reduce the size of large schema from 1000 to 250 fields. It also changes the time out of update gql schema request from 50 seconds to 200 seconds.

It took around 80 seconds for the test to run locally with race = true. The timeout has been set to 200 seconds to keep a buffe on slower machinesr.

Testing:
1. Tested locally with race flag set to true.
2. Tested without the throttling of indexing code (the test fails in this case as expected)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7522)

### DIFF
--- a/graphql/e2e/common/common.go
+++ b/graphql/e2e/common/common.go
@@ -1138,7 +1138,7 @@ func (params *GraphQLParams) createApplicationGQLPost(url string) (*http.Request
 
 // RunGQLRequest runs a HTTP GraphQL request and returns the data or any errors.
 func RunGQLRequest(req *http.Request) ([]byte, error) {
-	client := &http.Client{Timeout: 50 * time.Second}
+	client := &http.Client{Timeout: 200 * time.Second}
 	resp, err := client.Do(req)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


abhimanyusinghgaur (approved):

anurags92:
This reduces the size of the schema and increases the time-out limit. Wondering if it solves the core issue of having a race condition (if at all that's the core issue being identified by --race=true flag). I am afraid that with these changes we might still have the underlying issue affecting the codebase silently. What's your take @abhimanyusinghgaur @vmrajas ?

vmrajas:
@anurags92 , There is no race in the test. The reason of failure with --race=true flag is that with the flag set, the test was running slowly causing the updateGQLSchema request to timeout (due to the client timeout set to 50 seconds). As this increases the client timeout to 200 seconds, no such failure is anticipated.

abhimanyusinghgaur:
Yeah! that makes sense because the more the number of fields in the schema with indexes, the more time schema update will take, and the race flag makes it take even more time.
So, it seems fine to increase the timeout.

anurags92 (approved):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7522)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pawanrawal)

vmrajas:
Periodic Test All and CI passes for this test on mrjn-workstation. https://teamcity.dgraph.io/buildConfiguration/Dgraph_PeriodicTestAll/125268 and https://teamcity.dgraph.io/buildConfiguration/Dgraph_Ci/125278

PR #7525, opened 2021-03-08
Fix(Rollups): Don't try splitting a posting list with cardinality less than 2.

ahsanbarkati:
This PR fixes the issue of rollups getting hung. it happened because there was no check on the cardinality of the posting.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7525)

### DIFF
--- a/posting/list.go
+++ b/posting/list.go
@@ -1088,11 +1088,14 @@ func (ro *rollupOutput) getRange(uid uint64) (uint64, uint64) {
 }
 
 func shouldSplit(plist *pb.PostingList) (bool, error) {
-	r := roaring64.New()
-	if err := codec.FromPostingList(r, plist); err != nil {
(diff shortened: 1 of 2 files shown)


NamanJain8 (approved):

posting/list.go
```diff
@@ -1087,10 +1087,22 @@ func (ro *rollupOutput) getRange(uid uint64) (uint64, uint64) {
 	return 1, math.MaxUint64
 }
 
+func shouldSplit(plist *pb.PostingList) (bool, error) {
+	r := roaring64.New()
```
NamanJain8, line 1091: Do the cardinality check only when the first condition succeed.

ahsanbarkati: Done.

PR #7535, opened 2021-03-09
fix(Roaring): Remove pack from posting list

ahsanbarkati:
We no longer user group varint based UidPack in posting lists. This PR removes it from the proto message and also removes the corresponding stale codes.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7535)


worker/restore.go
```diff
@@ -194,7 +194,7 @@ func loadFromBackup(db *badger.DB, in *loadBackupInput) (uint64, uint64, error)
 					return 0, 0, errors.Wrapf(err, "while reading backup posting list")
 				}
 				pl := posting.FromBackupPostingList(backupPl)
-				shouldSplit := pl.Size() >= (1<<20)/2 && len(pl.Pack.Blocks) > 1
+				shouldSplit := pl.Size() >= (1<<20)/2
```
jarifibrahim, line 197: should we check the len(bitmap) over here? Similar to how we were checking len(pl.Pack.Blocks)?

ahsanbarkati: It is not strictly required. We call rollups if `shouldSplit` is `true`, which will then be a noop as it checks for the cardinality. 
Plus, checking for cardinality of a bitmap is expensive (we will have to de-serialize the bitmap). So I think that the check on posting list size can act as a preliminary filter here.

ahsanbarkati: Done

NamanJain8: Also add this check in bulk loader.

NamanJain8 (approved):

PR #7542, opened 2021-03-10
fix(GraphQL): fix order of entities query result 

minhaj-shakeel:
Fixes GRAPHQL-1083.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7542)


abhimanyusinghgaur (changes requested):
> Reviewed 4 of 4 files at r1, 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7542)* status: 4 of 6 files reviewed, 12 unresolved discussions (waiting on @abhimanyusinghgaur, @minhaj-shakeel, and @pawanrawal)

___
*[graphql/resolve/query.go, line 59 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVR9gdt6YNqIgpfhBbD:-MVR9gdt6YNqIgpfhBbE:b-kz6yti) ([raw file](https://github.com/dgraph-io/dgraph/blob/0a7e9f99ea779cb4049b41acad0c101538c14ba0/graphql/resolve/query.go#L59)):*
> ```Go
> // 1) rewrite the query using qr (return error if failed)
> // 2) execute the rewritten query with ex (return error if failed)
> func NewQueryResolver(qr QueryRewriter, ex DgraphExecutor) QueryResolver {
> ```

add a comment about the 3rd step of completion too from previous code
___
*[graphql/resolve/query.go, line 64 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVR9p2D36O4I1w_aRGY:-MVR9p2D36O4I1w_aRGZ:b-98jx2k) ([raw file](https://github.com/dgraph-io/dgraph/blob/0a7e9f99ea779cb4049b41acad0c101538c14ba0/graphql/resolve/query.go#L64)):*
> ```Go
> 
> // NewEntitiesQueryResolver ...
> func NewEntitiesQueryResolver(qr QueryRewriter, ex DgraphExecutor) QueryResolver {
> ```

proper comment
___
*[graphql/resolve/query_rewriter.go, line 164 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVR9vpnBAtpqhSC9Meq:-MVR9vpnBAtpqhSC9Mer:b-8eb39h) ([raw file](https://github.com/dgraph-io/dgraph/blob/0a7e9f99ea779cb4049b41acad0c101538c14ba0/graphql/resolve/query_rewriter.go#L164)):*
> ```
> string, string, bool, []interface{}
> ```
make a struct and put these in that struct. Not a good practice to return so many results from a func.
The signature should be like:
```
type parsedRepresentations struct {
 ...
}

func parseRepresentationsArgument(field schema.Query) (parsedRepresentations, error) { ... }
```
___
*[graphql/resolve/query_rewriter.go, line 288 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVRAt2v464hPfjTmZ_F:-MVRAt2v464hPfjTmZ_G:b-s6or73) ([raw file](https://github.com/dgraph-io/dgraph/blob/0a7e9f99ea779cb4049b41acad0c101538c14ba0/graphql/resolve/query_rewriter.go#L288)):*
<details><summary><i alt="summary">Quoted 4 lines of code…</i></summary>

> ```
> 	// Add the  ascending Order of the keyField in the query.
> 	// The result will be converted into the exact in the resultCompletion step.
> 	dgQuery.Order = append(dgQuery.Order,
> 		&pb.Order{Attr: typeDefn.DgraphPredicate(keyFieldName)})
> ```
</details>

we need to do this only when the key field is `@id` and not for `ID`. For, `ID` you won't get a `DgraphPredicate` as that maps to `uid` and also that for uid the result is by default sorted in the order of uid, so no need to sort for them.

So, put this in the `else` case above.
___
*[graphql/resolve/query_test.yaml, line 3301 at r2](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVRBka63tAnNDwtywBD:-MVRBka7A3icArzxdgGd:bpnp1vn) ([raw file](https://github.com/dgraph-io/dgraph/blob/9b7d6bab21f62423028e6bb4f652754e50f18790/graphql/resolve/query_test.yaml#L3301)):*
> ```
>  orderasc: Astronaut.id
> ```
so this should be removed as there is nothing called `Astronaut.id` stored in dgraph, it is only `uid`
___
*[graphql/resolve/resolver.go, line 88 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVR9E6QC-VqYvpaOCDW:-MVR9E6QC-VqYvpaOCDX:bobwepe) ([raw file](https://github.com/dgraph-io/dgraph/blob/0a7e9f99ea779cb4049b41acad0c101538c14ba0/graphql/resolve/resolver.go#L88)):*
> ```Go
> }
> 
> // A ResultCompleter ....
> ```

add in the full comments from the previous code
___
*[graphql/resolve/resolver.go, line 328 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVRD3Pg5tBpL1r2H5pZ:-MVRD3Pg5tBpL1r2H5p_:btqp0uh) ([raw file](https://github.com/dgraph-io/dgraph/blob/0a7e9f99ea779cb4049b41acad0c101538c14ba0/graphql/resolve/resolver.go#L328)):*
> ```
> entitiesCompletion
> ```
```
entitiesQueryCompletion
```
as it is only for queries
___
*[graphql/resolve/resolver.go, line 329 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVRCrht5xtgkQ4SJ1t8:-MVRCrht5xtgkQ4SJ1t9:b-1b5rff) ([raw file](https://github.com/dgraph-io/dgraph/blob/0a7e9f99ea779cb4049b41acad0c101538c14ba0/graphql/resolve/resolver.go#L329)):*
> ```
> id
> ```
if
___
*[graphql/resolve/resolver.go, line 349 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVRE4la3znu8sWk8pwl:-MVRE4la3znu8sWk8pwm:b2ii4va) ([raw file](https://github.com/dgraph-io/dgraph/blob/0a7e9f99ea779cb4049b41acad0c101538c14ba0/graphql/resolve/resolver.go#L349)):*
> ```
> map[interface{}]int
> ```
add a comment about why the key in this map is an interface? (as there are multiple types like String, Int, Int64 allowed as `@id`)
___
*[graphql/resolve/resolver.go, line 356 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVREXT13unagyAz7Up_:-MVREXT13unagyAz7Upa:bsd4d9t) ([raw file](https://github.com/dgraph-io/dgraph/blob/0a7e9f99ea779cb4049b41acad0c101538c14ba0/graphql/resolve/resolver.go#L356)):*
> ```
> .(string)
> ```
not necessary that they are string. Use a type switch and do the comparison accordingly.
Possible types are:
* string
* int/json.Number (one of these)
* float/json.Number (one of these)

We need to check json.Number case.

Also, add one e2e test case for the case of `@id` with `Int` in entities query.
___
*[graphql/resolve/resolver.go, line 362 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVRGO0s-I859Bnx5ZPQ:-MVRGO0s-I859Bnx5ZPR:b-hadhxe) ([raw file](https://github.com/dgraph-io/dgraph/blob/0a7e9f99ea779cb4049b41acad0c101538c14ba0/graphql/resolve/resolver.go#L362)):*
> ```
> []interface{}(data["_entities"])
> ```
let's save this bit in a variable `entitiesQryResp` prior to the for loop and use that.
also, can there be a case that data for a particular key was missing in dgraph, so that dgraph returns null for that index?
That should be handled by this right?
we should add this scenario in the e2e as well.
___
*[graphql/resolve/resolver.go, line 378 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVRBxy46CdhY1P0rFQR:-MVRBxy46CdhY1P0rFQS:b-bmbxcv) ([raw file](https://github.com/dgraph-io/dgraph/blob/0a7e9f99ea779cb4049b41acad0c101538c14ba0/graphql/resolve/resolver.go#L378)):*
> ```
> func entitiesQueryCompletion() CompletionFunc {
> 	return entitiesCompletion
> }
> ```
not required

abhimanyusinghgaur (changes requested):
> Reviewed 1 of 2 files at r2, 6 of 7 files at r3.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7542)* status: 8 of 9 files reviewed, 14 unresolved discussions (waiting on @abhimanyusinghgaur, @minhaj-shakeel, and @pawanrawal)

___
*[graphql/resolve/resolver.go, line 350 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVWRAQu-SbxNUyEbFsB:-MVWRAQu-SbxNUyEbFsC:b-wmpxk9) ([raw file](https://github.com/dgraph-io/dgraph/blob/3cf3dab337692e8bb6528681927bd74adde8609f/graphql/resolve/resolver.go#L350)):*
> ```
> key in the map as 
> ```
key in the map is of type interface because
___
*[graphql/resolve/resolver.go, line 367 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVWRZHx66_u7aBP6MAI:-MVWRZHx66_u7aBP6MAJ:b-qp6dk7) ([raw file](https://github.com/dgraph-io/dgraph/blob/3cf3dab337692e8bb6528681927bd74adde8609f/graphql/resolve/resolver.go#L367)):*
> ```
> switch uniqueKeyList[i].(type)
> ```
`switch val := uniqueKeyList[i].(type)`

Then you can just use `val` in each case without having to explicitly cast it.

The order of cases can be optimized to:
* string
* json.Number
* int64
* float64
___
*[graphql/resolve/resolver.go, line 375 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVWSnxF3QrUK33An6RW:-MVWSnxF3QrUK33An6RX:b-efsog4) ([raw file](https://github.com/dgraph-io/dgraph/blob/3cf3dab337692e8bb6528681927bd74adde8609f/graphql/resolve/resolver.go#L375)):*
> ```
>  uniqueKeyList[i].(json.Number) < uniqueKeyList[j].(json.Number)
> ```
json.Number is basically a string.
Doing this would end up in a lexicographic comparison, resulting in
`[12, 100, 1]` being sorted as `[1, 100, 12]`. Which is not what we want.

So, we should use val.Int64() or val.Float64() depending on the type of the key field, and then compare.

minhaj-shakeel (commented):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7542)* status: 7 of 11 files reviewed, 14 unresolved discussions (waiting on @abhimanyusinghgaur, @minhaj-shakeel, and @pawanrawal)

___
*[graphql/resolve/query.go, line 64 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVR9p2D36O4I1w_aRGY:-MVX-ebJ0_2A9n-hYjaH:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/0a7e9f99ea779cb4049b41acad0c101538c14ba0/graphql/resolve/query.go#L64)):*

proper comment
</blockquote></details>

Done.
___
*[graphql/resolve/query_rewriter.go, line 164 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVR9vpnBAtpqhSC9Meq:-MVX-fHzDcA8t6l3tz6T:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/0a7e9f99ea779cb4049b41acad0c101538c14ba0/graphql/resolve/query_rewriter.go#L164)):*

> ```
> string, string, bool, []interface{}
> ```
make a struct and put these in that struct. Not a good practice to return so many results from a func.
The signature should be like:
```
type parsedRepresentations struct {
 ...
}

func parseRepresentationsArgument(field schema.Query) (parsedRepresentations, error) { ... }
```
</blockquote></details>

Done.
___
*[graphql/resolve/query_rewriter.go, line 288 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVRAt2v464hPfjTmZ_F:-MVX-fi--oY5Z25K-Zq6:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/0a7e9f99ea779cb4049b41acad0c101538c14ba0/graphql/resolve/query_rewriter.go#L288)):*

> ```
> 	// Add the  ascending Order of the keyField in the query.
> 	// The result will be converted into the exact in the resultCompletion step.
> 	dgQuery.Order = append(dgQuery.Order,
> 		&pb.Order{Attr: typeDefn.DgraphPredicate(keyFieldName)})
> ```
we need to do this only when the key field is `@id` and not for `ID`. For, `ID` you won't get a `DgraphPredicate` as that maps to `uid` and also that for uid the result is by default sorted in the order of uid, so no need to sort for them.

So, put this in the `else` case above.
</blockquote></details>

Done.
___
*[graphql/resolve/resolver.go, line 88 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVR9E6QC-VqYvpaOCDW:-MVX-gxMCZg3UzxtjsX2:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/0a7e9f99ea779cb4049b41acad0c101538c14ba0/graphql/resolve/resolver.go#L88)):*

add in the full comments from the previous code
</blockquote></details>

Done.
___
*[graphql/resolve/resolver.go, line 328 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVRD3Pg5tBpL1r2H5pZ:-MVX-hUm5OlwWMRc5a3N:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/0a7e9f99ea779cb4049b41acad0c101538c14ba0/graphql/resolve/resolver.go#L328)):*

> ```
> entitiesCompletion
> ```
```
entitiesQueryCompletion
```
as it is only for queries
</blockquote></details>

Done.
___
*[graphql/resolve/resolver.go, line 329 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVRCrht5xtgkQ4SJ1t8:-MVX-hr45qlsfaxg7EAr:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/0a7e9f99ea779cb4049b41acad0c101538c14ba0/graphql/resolve/resolver.go#L329)):*

> ```
> id
> ```
if
</blockquote></details>

Done.
___
*[graphql/resolve/resolver.go, line 349 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVRE4la3znu8sWk8pwl:-MVX-jTVE7lds5yAPiTn:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/0a7e9f99ea779cb4049b41acad0c101538c14ba0/graphql/resolve/resolver.go#L349)):*

> ```
> map[interface{}]int
> ```
add a comment about why the key in this map is an interface? (as there are multiple types like String, Int, Int64 allowed as `@id`)
</blockquote></details>

Done.
___
*[graphql/resolve/resolver.go, line 356 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVREXT13unagyAz7Up_:-MVX-k2T3BU_kYS2Qq1n:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/0a7e9f99ea779cb4049b41acad0c101538c14ba0/graphql/resolve/resolver.go#L356)):*

> ```
> .(string)
> ```
not necessary that they are string. Use a type switch and do the comparison accordingly.
Possible types are:
* string
* int/json.Number (one of these)
* float/json.Number (one of these)

We need to check json.Number case.

Also, add one e2e test case for the case of `@id` with `Int` in entities query.
</blockquote></details>

Done.
___
*[graphql/resolve/resolver.go, line 362 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVRGO0s-I859Bnx5ZPQ:-MVX-oyD05hdXdiJ1S1u:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/0a7e9f99ea779cb4049b41acad0c101538c14ba0/graphql/resolve/resolver.go#L362)):*

> ```
> []interface{}(data["_entities"])
> ```
let's save this bit in a variable `entitiesQryResp` prior to the for loop and use that.
also, can there be a case that data for a particular key was missing in dgraph, so that dgraph returns null for that index?
That should be handled by this right?
we should add this scenario in the e2e as well.
</blockquote></details>

Done.
___
*[graphql/resolve/resolver.go, line 378 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVRBxy46CdhY1P0rFQR:-MVX-pWw7jOxShdirinm:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/0a7e9f99ea779cb4049b41acad0c101538c14ba0/graphql/resolve/resolver.go#L378)):*

> ```
> func entitiesQueryCompletion() CompletionFunc {
> 	return entitiesCompletion
> }
> ```
not required
</blockquote></details>

Done.
___
*[graphql/resolve/resolver.go, line 350 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVWRAQu-SbxNUyEbFsB:-MVX-iZF03cXvwTbb0RM:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/3cf3dab337692e8bb6528681927bd74adde8609f/graphql/resolve/resolver.go#L350)):*

> ```
> key in the map as 
> ```
key in the map is of type interface because
</blockquote></details>

Done.
___
*[graphql/resolve/resolver.go, line 367 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVWRZHx66_u7aBP6MAI:-MVX-kmG3VD2QDr4kYoM:b-jrvdre) ([raw file](https://github.com/dgraph-io/dgraph/blob/3cf3dab337692e8bb6528681927bd74adde8609f/graphql/resolve/resolver.go#L367)):*

> ```
> switch uniqueKeyList[i].(type)
> ```
`switch val := uniqueKeyList[i].(type)`

Then you can just use `val` in each case without having to explicitly cast it.

The order of cases can be optimized to:
* string
* json.Number
* int64
* float64
</blockquote></details>

Even if I use the `val`,  the second value will be typecasted explicitly. Can we do it in a single statement for both of the values? 
___
*[graphql/resolve/resolver.go, line 375 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVWSnxF3QrUK33An6RW:-MVX-l_P71ZYqKKkz59M:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/3cf3dab337692e8bb6528681927bd74adde8609f/graphql/resolve/resolver.go#L375)):*

> ```
>  uniqueKeyList[i].(json.Number) < uniqueKeyList[j].(json.Number)
> ```
json.Number is basically a string.
Doing this would end up in a lexicographic comparison, resulting in
`[12, 100, 1]` being sorted as `[1, 100, 12]`. Which is not what we want.

So, we should use val.Int64() or val.Float64() depending on the type of the key field, and then compare.
</blockquote></details>

Done.

abhimanyusinghgaur (approved):
:lgtm:

Just got one more comment, and then good to go.

> Reviewed 4 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-:-MVZxa2W29GjOPOv1qIH:b-uakdz1)* status: all files reviewed, 1 unresolved discussion (waiting on @minhaj-shakeel and @pawanrawal)

___
*[graphql/resolve/resolver.go, line 375 at r4](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVZv15C1Rah2xGcXS0z:-MVZv15DBkM9t6s3vvzm:b6g6u1k) ([raw file](https://github.com/dgraph-io/dgraph/blob/35465acff990c24e2d23f4119349f1350d86ad3b/graphql/resolve/resolver.go#L375)):*
> ```
> case "Int"
> ```
`case "Int","Int64":`

minhaj-shakeel (commented):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7542)* status: 10 of 11 files reviewed, 1 unresolved discussion (waiting on @abhimanyusinghgaur and @pawanrawal)

___
*[graphql/resolve/resolver.go, line 375 at r4](https://reviewable.io/reviews/dgraph-io/dgraph/7542#-MVZv15C1Rah2xGcXS0z:-MV_54EL4wCcpRO2PQO3:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/35465acff990c24e2d23f4119349f1350d86ad3b/graphql/resolve/resolver.go#L375)):*

> ```
> case "Int"
> ```
`case "Int","Int64":`
</blockquote></details>

Done.

PR #7546, opened 2021-03-11
fix(GRAPHQL): fix duplicate xid error for multiple xid fields.

JatinDevDG:
We were getting duplicate xid error for the multiple xid fields if `typeName+xidName+xidValue `is same for those fields. The reason for this is that we need to  generate unique query variables for the xids field while doing mutation rewriting and for that we use a map to variable names. 
we were using below key for that map 
`key = typ.FieldOriginatedFrom(xidName)  + xidName + xidVal`
This was generating same key for different xid fields. For example in below type we have two fields with id fields

```
type ABC {
    ab: String! @id
    abc: String! @id
}
```
And below mutation try to set them such that `key = typ.FieldOriginatedFrom(xidName)  + xidName + xidVal` will be same for both of them i.e. `ABC+ab+cd=ABC+abc+d=ABCabcd`
```
mutation {
		addABC(input: [{ ab: "cd", abc: "d" }]) {
		  aBC {
		        ab
		        abc
		       }
		  }
	 }
```
This error also occur in multiple types.
We now added "." between values to separate typename and xidname , and the xidValue.This will ensure unique keys for xid fields across types.
`key = typ.FieldOriginatedFrom(xidName) + "." + xidName + "." + xidVal`

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7546)

### DIFF
--- a/graphql/resolve/mutation_rewriter.go
+++ b/graphql/resolve/mutation_rewriter.go
@@ -148,7 +148,7 @@ func (v *VariableGenerator) Next(typ schema.Type, xidName, xidVal string, auth b
 		//    }
 		//   }
 		// }
-		// The two generated kes for this case will be
+		// The two generated keys for this case will be
 		// ABC.ab.cd and ABC.abc.d
 		// It also ensures that xids from different types gets different variable names
 		// here we are using the assertion that field name or type name can't have "." in them
--- a/graphql/resolve/mutation_rewriter.go
+++ b/graphql/resolve/mutation_rewriter.go
@@ -139,7 +139,20 @@ func (v *VariableGenerator) Next(typ schema.Type, xidName, xidVal string, auth b
 	if xidName == "" || xidVal == "" {
 		key = typ.Name()
 	} else {
-		key = typ.FieldOriginatedFrom(xidName) + xidName + xidVal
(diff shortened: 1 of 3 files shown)


graphql/resolve/mutation_rewriter.go
```diff
@@ -139,7 +139,20 @@ func (v *VariableGenerator) Next(typ schema.Type, xidName, xidVal string, auth b
...
+		// We add "." between values while generating key to removes duplicate xidError from below type of cases
+		// mutation {
+		//  addABC(input: [{ ab: "cd", abc: "d" }]) {
+		//    aBC {
+		//      ab
+		//      abc
+		//    }
+		//   }
+		// }
+		// The two generated kes for this case will be
```
vmrajas, line 151: nit: keys

vmrajas (approved):
LGTM.
Nicely written tests covering a wide range of cases.

graphql/resolve/mutation_rewriter.go
JatinDevDG: fixed.

abhimanyusinghgaur (approved):

PR #7552, opened 2021-03-11
Fix(multi-tenancy): Format namespace to human readable form

ahsanbarkati:
This PR changes the logs to print namespace in human-readable form. This change doesn't fix all of them. 
We have left the cases where we print a struct and the fields in the struct contains predicate with namespace. 
We have also skipped logs of `SubscribeForUpdates` because we can't assume the prefix being logged will always contain namespace. 

This is not exhaustive, it fixes the things which are apparent on basic usage of dgraph. 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7552)

### DIFF
--- a/posting/index.go
+++ b/posting/index.go
@@ -592,7 +592,7 @@ func (r *rebuilder) Run(ctx context.Context) error {
 
 	glog.V(1).Infof(
 		"Rebuilding index for predicate %s: Starting process. StartTs=%d. Prefix=\n%s\n",
-		r.attr, r.startTs, hex.Dump(r.prefix))
+		x.FormatNsAttr(r.attr), r.startTs, hex.Dump(r.prefix))
 
 	// Counter is used here to ensure that all keys are committed at different timestamp.
 	// We set it to 1 in case there are no keys found and NewStreamAt is called with ts=0.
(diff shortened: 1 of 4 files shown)


posting/index.go
```diff
@@ -592,15 +592,16 @@ func (r *rebuilder) Run(ctx context.Context) error {
 
 	glog.V(1).Infof(
 		"Rebuilding index for predicate %s: Starting process. StartTs=%d. Prefix=\n%s\n",
-		r.attr, r.startTs, hex.Dump(r.prefix))
+		x.FormatNsAttr(r.attr), r.startTs, hex.Dump(r.prefix))
```
jarifibrahim, line 595: Print namespace as well.

posting/index.go
```diff
@@ -592,15 +592,16 @@ func (r *rebuilder) Run(ctx context.Context) error {
...
 
 	// Counter is used here to ensure that all keys are committed at different timestamp.
 	// We set it to 1 in case there are no keys found and NewStreamAt is called with ts=0.
 	var counter uint64 = 1
 
 	tmpWriter := tmpDB.NewManagedWriteBatch()
 	stream := pstore.NewStreamAt(r.startTs)
-	stream.LogPrefix = fmt.Sprintf("Rebuilding index for predicate %s (1/2):", r.attr)
+	stream.LogPrefix = fmt.Sprintf("Rebuilding index for predicate %s (1/2):",
```
jarifibrahim, line 603: Print namespace and attr

posting/index.go
```diff
@@ -662,19 +663,21 @@ func (r *rebuilder) Run(ctx context.Context) error {
 		return err
 	}
 	glog.V(1).Infof("Rebuilding index for predicate %s: building temp index took: %v\n",
-		r.attr, time.Since(start))
+		x.FormatNsAttr(r.attr), time.Since(start))
```
jarifibrahim, line 666: here as well. Print namespace everywhere. Without the namespace, we wouldn't know which namespace is doing this operation.

jarifibrahim (commented):
Print the namespace information also.

posting/index.go
ahsanbarkati: It is already printing namespace information. 
`FormatNsAttr(attr)` formats them to this format: `x-predicate` 

predicate `name` from namespace `1` will be printed as `1-name` rather than current thing which is `/000/000/000/000/000/000/000/001name`

jarifibrahim (approved):

posting/index.go
jarifibrahim: My bad. I didn't realize the formatNSAttr was adding the namespace as well.

PR #7553, opened 2021-03-11
chore(live): fixing live loader abort errors when using force-namespace

aman-bansal:
This PR fixes the issue of multiple txn aborts while live loading data with force-namespace enabled. The problem was if it's a galaxy operation, the schema will already have predicates with a namespace prefix. But we were appending it again in the schema.init function thus causing two namespaces in the predicate Name. This results in the wrong generation of conflict keys and thus txn aborts later

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7553)

### DIFF
--- a/dgraph/cmd/live/run.go
+++ b/dgraph/cmd/live/run.go
@@ -191,8 +191,7 @@ func init() {
 		"specific namespace. Setting it to negative value will preserve the namespace.")
 }
 
-func getSchema(ctx context.Context, dgraphClient *dgo.Dgraph, ns uint64,
-	galaxyOperation bool) (*schema, error) {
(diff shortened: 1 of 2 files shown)


dgraph/cmd/live/run.go
```diff
@@ -191,7 +191,8 @@ func init() {
 		"specific namespace. Setting it to negative value will preserve the namespace.")
 }
 
-func getSchema(ctx context.Context, dgraphClient *dgo.Dgraph, ns uint64) (*schema, error) {
+func getSchema(ctx context.Context, dgraphClient *dgo.Dgraph, ns uint64,
+	galaxyOperation bool) (*schema, error) {
```
NamanJain8, line 195: 5 parametes are too much. Let's not pass `ns` and use it from `opt` directly.

NamanJain8 (approved):
Awesome 🎉

PR #7554, opened 2021-03-11
fix(learner nodes): Reconnect to learner nodes after restart

jarifibrahim:
When an alpha would restart, it wasn't reconnecting to the learner
nodes. This PR fixes that.
Fixes DGRAPH-3125

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7554)

### DIFF
--- a/worker/draft.go
+++ b/worker/draft.go
@@ -1760,6 +1760,12 @@ func (n *node) InitAndStartNode() {
 					n.Connect(id, m.Addr)
 				}
 			}
+			for _, id := range sp.Metadata.ConfState.Learners {
+				m, ok := members[id]
+				if ok {
+					n.Connect(id, m.Addr)
+				}
+			}
 		}
 		n.SetRaft(raft.RestartNode(n.Cfg))
 		glog.V(2).Infoln("Restart node complete")


worker/draft.go
```diff
@@ -1760,6 +1760,12 @@ func (n *node) InitAndStartNode() {
 					n.Connect(id, m.Addr)
 				}
 			}
+			for _, id := range sp.Metadata.ConfState.Learners {
+				m, ok := members[id]
```
jarifibrahim, line 1764: I am not sure why this check is necessary. The loop above had the check and so I've added it.

NamanJain8: I think this might be needed because of the following reason. Not sure though.
Say an alpha node has a stale snapshot and the cluster has progressed (some nodes added/deleted). When we restart that node, it first gets the snapshot from the WAL and verifies that it actually exists  in latest membership state. If yes, then it connects to it.

This change will also mean that learner nodes will connect among themselves. Do we want that?

jarifibrahim: >This change will also mean that learner nodes will connect among themselves. Is that expected behaviour?

That's expected. All alphas connect to all other alphas.

PR #7555, opened 2021-03-11
opt(getNew): use txn.GetValues() instead of KeyIterator in getNew function

NamanJain8:
TODO

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7555)


NamanJain8:
This is no longer needed.

PR #7557, opened 2021-03-12
fix(GRAPHQL): fix empty remove in update mutation patch, that remove all the data for nodes in filter.

JatinDevDG:
We have this behavior of update mutation where we are deleting all the data for nodes that are selected in the filter when we give empty remove {} in the update patch.
For example in the below case , we delete everything for node with id `0x123`
```
mutation updateAuthor($patch: UpdateAuthorInput!) {
      updateAuthor(input: $patch) {
        author {
          id
        }
      }
    }
```
  variables: 
 ```
   {
      "patch": {
        "filter": {
          "id": ["0x123"]
        },
        "remove": {}
      }
```
    }

This can be more dangeours if we apply filter also empty like below, in that case it will delete every node of type Author.
```
   {
      "patch": {
        "filter": {
          "id": ["0x123"]
        },
        "remove": {}
      }
```
We are now changing this behaviour , such that empty remove{}, doesn't have any effect.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7557)

### DIFF
--- a/graphql/resolve/mutation_rewriter.go
+++ b/graphql/resolve/mutation_rewriter.go
@@ -330,16 +330,18 @@ func (urw *UpdateRewriter) RewriteQueries(
 	// Write existence queries for remove
 	if delArg != nil {
 		obj := delArg.(map[string]interface{})
-		queries, errs := existenceQueries(ctx, mutatedType, nil, urw.VarGen, obj, urw.XidMetadata)
-		if len(errs) > 0 {
-			var gqlErrors x.GqlErrorList
-			for _, err := range errs {
-				gqlErrors = append(gqlErrors, schema.AsGQLErrors(err)...)
+		if len(obj) != 0 {
+			queries, errs := existenceQueries(ctx, mutatedType, nil, urw.VarGen, obj, urw.XidMetadata)
+			if len(errs) > 0 {
+				var gqlErrors x.GqlErrorList
+				for _, err := range errs {
+					gqlErrors = append(gqlErrors, schema.AsGQLErrors(err)...)
+				}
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: graphql/resolve/mutation_rewriter.go [hunk 1]
```diff
@@ -587,14 +589,14 @@ func (urw *UpdateRewriter) Rewrite(
 
 	upsertQuery := RewriteUpsertQueryFromMutation(m, authRw, MutationQueryVar, m.Name(), "")
 	srcUID := MutationQueryVarUID
-
-	if setArg == nil && delArg == nil {
+	objDel, ok := delArg.(map[string]interface{})
+	if setArg == nil && (delArg == nil || (len(objDel) == 0 && ok)) {
```
vmrajas, line 593: I believe we can return at this stage even for an empty set. Let's compute objSet here and return if len(objSet) == 0 as well.

graphql/resolve/update_mutation_test.yaml
```diff
@@ -1891,3 +1891,100 @@
...
+  dgquerysec: |-
+    query {
+      x as updateAuthor(func: type(Author)) {
+        uid
+      }
+    }
+  dgmutations:
+    - setjson: |
+        { "uid": "uid(x)" }
+      cond: "@if(gt(len(x), 0))"
```
vmrajas, line 1990: nit: Extra line at end of file.

vmrajas (changes requested):
I believe the PR should be raised on 21.03 and not master. 
Nicely written test cases. 
We can also add similar conditions for setArg. Currently when an empty set is supplied, we generate a harmless (but also useless) setjson. We could simply return from function earlier in this case as well.

PR #7563, opened 2021-03-15
fix(GRAPHQL): fix empty remove in update mutation patch, that remove all the data for nodes in filter.

JatinDevDG:
We have this behavior of update mutation where we are deleting all the data for nodes that are selected in the filter when we give empty remove {} in the update patch.
For example in the below case , we delete everything for node with id 0x123
```

mutation updateAuthor($patch: UpdateAuthorInput!) {
      updateAuthor(input: $patch) {
        author {
          id
        }
      }
    }
```
variables:

 ```
 {
     "patch": {
       "filter": {
         "id": ["0x123"]
       },
       "remove": {}
     }
}
```
This can be more dangeours if we apply filter also empty like below, in that case it will delete every node of type Author.

  ```
 {
      "patch": {
        "filter": {
          "id": ["0x123"]
        },
        "remove": {}
      }
```
We are now changing this behaviour , such that empty remove{}, doesn't have any effect.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7563)


graphql/resolve/update_mutation_test.yaml
```diff
@@ -1891,3 +1891,71 @@
...
+      x as updateAuthor(func: uid(0x123)) @filter(type(Author)) {
+        uid
+      }
+    }
+  dgmutations:
+    - deletejson: |
+        { "uid": "uid(x)",
+          "Author.name": "Alice"
+        }
+      cond: "@if(gt(len(x), 0))"
```
vmrajas, line 1961: nit: Add an extra line at end of file.

graphql/resolve/mutation_rewriter.go
```diff
@@ -600,46 +602,48 @@ func (urw *UpdateRewriter) Rewrite(
 
 	upsertQuery := RewriteUpsertQueryFromMutation(m, authRw, MutationQueryVar, m.Name(), "")
 	srcUID := MutationQueryVarUID
-
-	if setArg == nil && delArg == nil {
+	objDel, okDelArg := delArg.(map[string]interface{})
+	objSet, okSetArg := setArg.(map[string]interface{})
+	if (setArg == nil || (len(objSet) == 0 && okSetArg)) && (delArg == nil || (len(objDel) == 0 && okDelArg)) {
```
vmrajas, line 607: May be add a small comment here on why this is done.

graphql/resolve/mutation_rewriter.go
```diff
@@ -343,16 +343,18 @@ func (urw *UpdateRewriter) RewriteQueries(
 	// Write existence queries for remove
 	if delArg != nil {
 		obj := delArg.(map[string]interface{})
-		queries, errs := existenceQueries(ctx, mutatedType, nil, urw.VarGen, obj, urw.XidMetadata)
-		if len(errs) > 0 {
-			var gqlErrors x.GqlErrorList
-			for _, err := range errs {
-				gqlErrors = append(gqlErrors, schema.AsGQLErrors(err)...)
+		if len(obj) != 0 {
```
vmrajas, line 346: Also add corresponding if condition on setArg near line 330.

graphql/resolve/mutation_rewriter.go
```diff
@@ -600,46 +602,48 @@ func (urw *UpdateRewriter) Rewrite(
 
 	upsertQuery := RewriteUpsertQueryFromMutation(m, authRw, MutationQueryVar, m.Name(), "")
 	srcUID := MutationQueryVarUID
-
-	if setArg == nil && delArg == nil {
+	objDel, okDelArg := delArg.(map[string]interface{})
```
vmrajas, line 605: This statement is copy of line 345. I believe this statement could be written before line 344 to avoid statement on line 345. Similarly, for setArg.

vmrajas (approved):

graphql/resolve/mutation_rewriter.go
JatinDevDG: added.

graphql/resolve/mutation_rewriter.go
JatinDevDG: statement at line 345  is in different function.

graphql/resolve/mutation_rewriter.go
JatinDevDG: added.

graphql/resolve/update_mutation_test.yaml
JatinDevDG: added.

PR #7567, opened 2021-03-15
fix(audit): logs not getting deleted after N days

ajeetdsouza:
Fixes DGRAPH-3139.

There were two bugs fixed here:
- The wrong parameter was being passed to `processOldLogFiles` (`l.MaxSize` instead of `l.MaxAge`).
- When formatting the closing timestamp of an audit log file, we use the local timezone. However, when parsing it back, we were assuming a UTC timezone. I changed this to assume local timezone.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7567)

### DIFF
--- a/x/log_writer.go
+++ b/x/log_writer.go
@@ -373,8 +373,8 @@ func (l *LogWriter) manageOldLogs() {
 	}
 }
 
-// prefixAndExt extracts the filename (without the extension) as well its
-// extension (including the `.`) from a filepath.
+// prefixAndExt extracts the filename and extension from a filepath.
+// eg. prefixAndExt("/home/foo/file.ext") would return ("file", ".ext").
 func prefixAndExt(file string) (prefix, ext string) {
 	filename := filepath.Base(file)
 	ext = filepath.Ext(filename)
@@ -395,8 +395,8 @@ func processOldLogFiles(fp string, maxAge int64) ([]string, []string, error) {
 	toRemove := make([]string, 0)
 	toKeep := make([]string, 0)
 
-	diff := time.Duration(int64(24*time.Hour) * maxAge)
-	cutoff := time.Now().Add(-1 * diff)
+	diff := 24 * time.Hour * time.Duration(maxAge)
+	cutoff := time.Now().Add(-diff)
 
 	for _, f := range files {
 		if f.IsDir() || // f is directory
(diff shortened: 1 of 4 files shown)


x/log_writer.go
```diff
@@ -404,7 +406,8 @@ func processOldLogFiles(fp string, maxAge int64) ([]string, []string, error) {
 		}
 
 		_, e := prefixAndExt(fp)
-		ts, err := time.Parse(backupTimeFormat, f.Name()[len(defPrefix):len(f.Name())-len(e)])
+		tsString := f.Name()[len(defPrefix) : len(f.Name())-len(e)]
+		ts, err := time.ParseInLocation(backupTimeFormat, tsString, time.Local)
```
jarifibrahim, line 410: @ajeetdsouza can you please point me to the code that creates the file? I can't find that code.

jarifibrahim: https://github.com/dgraph-io/dgraph/blob/4e4842624d26bb24caa08019d7be834bc59e282b/x/log_writer.go#L302-L307

I don't think we're writing the time in Local format.

ajeetdsouza: @jarifibrahim `time.Now()` returns time using the local timezone, but `backupTimeFormat` does not include timezone information, which is what is causing the problem.

Sample code:

```go
package main

import (
	"fmt"
	"time"
)

func main() {
	const backupTimeFormat = "2006-01-02T15-04-05.000"
	now := time.Now()
	tsLocal := now.Format(backupTimeFormat)
	tsUTC := now.UTC().Format(backupTimeFormat)
	fmt.Printf("tsLocal: %s\ntsUTC: %s\n", tsLocal, tsUTC) // different values
}
```

There are three alternate approaches here:
- Change `backupTimeFormat` to include timezone.
- Use UTC when formatting to `backupTimeFormat`.
- Both of the above.

I think these would be better than the current approach, since there is no ambiguity that way.

jarifibrahim: >Change backupTimeFormat to include timezone.
Use UTC when formatting to backupTimeFormat.
Both of the above.

This would change the file name and we might not be able to parse it again. Is this backupTimeFormat used only in audit logs? @aman-bansal why is it called backupTimeFormat?

If this is used only in audit logs, we can use time.UTC() everywhere for simplicity.

ajeetdsouza: @jarifibrahim `backupTimeFormat` is only used within `LogWriter`, so I think it's safe to switch over to UTC timestamps. I just changed it to use UTC everywhere.

aman-bansal (approved):

jarifibrahim (approved):

PR #7577, opened 2021-03-16
chore(test): fixing flaky tests

aman-bansal:
This PR fixes multiple test cases like TestLiveLoadMultiTenancy, Error in CreateNamespace, TestCountIndex, TestParallelIndex generally in case of tests when enabled with race flag. 
This also fixes one error in t.go where we are printing dgraph logs whenever a test fails.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7577)

### DIFF
--- a/systest/bgindex/parallel_test.go
+++ b/systest/bgindex/parallel_test.go
@@ -102,24 +102,35 @@ func TestParallelIndexing(t *testing.T) {
 	}
 
 	fmt.Println("building indexes in background for int and string data")
-	if err := dg.Alter(context.Background(), &api.Operation{
-		Schema: `
+	// Wait until previous indexing is complete.
+	for {
+		if err := dg.Alter(context.Background(), &api.Operation{
+			Schema: `
 			balance_int: int @index(int) .
 			balance_str: string @index(fulltext, term, exact) .
 		`,
-		RunInBackground: true,
-	}); err != nil && !strings.Contains(err.Error(), "errIndexingInProgress") {
-		t.Fatalf("error in adding indexes :: %v\n", err)
+			RunInBackground: true,
+		}); err != nil && !strings.Contains(err.Error(), "errIndexingInProgress") {
+			t.Fatalf("error in adding indexes :: %v\n", err)
+		} else if err == nil {
+			break
+		}
+		time.Sleep(time.Second)
 	}
-
-	if err := dg.Alter(context.Background(), &api.Operation{
-		Schema: `
+	// Wait until previous indexing is complete.
+	for {
+		if err := dg.Alter(context.Background(), &api.Operation{
+			Schema: `
 			balance_int: int @index(int) .
 			balance_str: string @index(fulltext, term, exact) .
 		`,
-		RunInBackground: true,
-	}); err != nil && !strings.Contains(err.Error(), "errIndexingInProgress") {
-		t.Fatalf("error in adding indexes :: %v\n", err)
+			RunInBackground: true,
+		}); err != nil && !strings.Contains(err.Error(), "errIndexingInProgress") {
+			t.Fatalf("error in adding indexes :: %v\n", err)
+		} else if err == nil {
+			break
+		}
+		time.Sleep(time.Second)
 	}
 
 	// Wait until previous indexing is complete.
--- a/systest/bgindex/parallel_test.go
+++ b/systest/bgindex/parallel_test.go
@@ -27,6 +27,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/dgraph-io/dgraph/x"
+	"github.com/stretchr/testify/require"
+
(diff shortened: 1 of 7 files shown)


systest/bgindex/parallel_test.go
```diff
@@ -96,24 +102,35 @@ func TestParallelIndexing(t *testing.T) {
 	}
 
 	fmt.Println("building indexes in background for int and string data")
-	if err := dg.Alter(context.Background(), &api.Operation{
-		Schema: `
+	// Wait until previous indexing is complete.
+	for {
```
ahsanbarkati, line 106: This will cause the test to be hung forever if there is some issue. Better add `RetryUntilSuccess` here as well so that it is easy to figure out what was the error after `n` tries.

systest/multi-tenancy/basic_test.go
```diff
@@ -351,12 +354,33 @@ func TestLiveLoadMulti(t *testing.T) {
...
+		var o2 interface{}
+
+		var err error
+		expected := `{"me": [{"name":"ns alice"}, {"name": "ns bob"},{"name":"ns chew"},
 		{"name": "ns dan"},{"name":"ns eon"}, {"name": "ns free"},{"name":"ns gary"},
-		{"name": "ns hola"}]}`, string(resp))
+		{"name": "ns hola"}]}`
+		err = json.Unmarshal([]byte(expected), &o1)
+		if err != nil {
+			return fmt.Errorf("Error mashalling string 1 :: %s", err.Error())
```
ahsanbarkati, line 369: Error unmarshaling string 1 ...

systest/multi-tenancy/basic_test.go
```diff
@@ -351,12 +354,33 @@ func TestLiveLoadMulti(t *testing.T) {
...
+		}
+		err = json.Unmarshal([]byte(string(resp)), &o2)
+		if err != nil {
+			return fmt.Errorf("Error mashalling string 2 :: %s", err.Error())
```
ahsanbarkati, line 373: Here as well. Error unmarshaling string 2 ...

t/t.go
```diff
@@ -251,7 +254,7 @@ func hasTestFiles(pkg string) bool {
 
 var _threadId int32
 
-func runTests(taskCh chan task, closer *z.Closer) error {
+func runTests(taskCh chan task, closer *z.Closer) (err error) {
```
ahsanbarkati, line 257: Can we avoid named returns? It is confusing

systest/bgindex/parallel_test.go
aman-bansal: This will only retry if the error is of type Indexing in Progress in all other cases it will fail. That part is fine i believe. Indexing can take much longer to complete especially when race is enabled. Anyways we have checks in place to timeout tests itself after 10 minutes so that will take care of this part

ahsanbarkati: Hmm. Makes sense

ahsanbarkati (approved):
Looks good

systest/multi-tenancy/basic_test.go
```diff
@@ -351,12 +354,33 @@ func TestLiveLoadMulti(t *testing.T) {
...
+		var o1 interface{}
+		var o2 interface{}
+
+		var err error
+		expected := `{"me": [{"name":"ns alice"}, {"name": "ns bob"},{"name":"ns chew"},
 		{"name": "ns dan"},{"name":"ns eon"}, {"name": "ns free"},{"name":"ns gary"},
-		{"name": "ns hola"}]}`, string(resp))
+		{"name": "ns hola"}]}`
+		err = json.Unmarshal([]byte(expected), &o1)
+		if err != nil {
```
jarifibrahim, line 368: if err := ....; err != nil ....

systest/multi-tenancy/basic_test.go
```diff
@@ -351,12 +354,33 @@ func TestLiveLoadMulti(t *testing.T) {
...
+			return fmt.Errorf("error unmarshalling string 1 :: %s", err.Error())
+		}
+		err = json.Unmarshal([]byte(string(resp)), &o2)
+		if err != nil {
```
jarifibrahim, line 372: if err := ....; err != nil ....

systest/multi-tenancy/basic_test.go
```diff
@@ -351,12 +354,33 @@ func TestLiveLoadMulti(t *testing.T) {
...
+			return fmt.Errorf("error unmarshalling string 2 :: %s", err.Error())
+		}
 
+		if reflect.DeepEqual(o1, o2) {
```
jarifibrahim, line 376: why do you have to do this instead of CompareJson?

aman-bansal: i actaully misinterpret this failure. I have removed this code part

PR #7582, opened 2021-03-16
Don't start a learner node with no peers

ahsanbarkati:
When a learner node starts and there is no alpha peer then it promotes itself to a normal mode.
This PR makes it crash itself in this case.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7582)

### DIFF
--- a/worker/draft.go
+++ b/worker/draft.go
@@ -1759,6 +1759,7 @@ func (n *node) InitAndStartNode() {
 			// zero-member Raft group.
 			n.SetConfState(&sp.Metadata.ConfState)
 
+			// TODO: Making connections here seems unnecessary, evaluate.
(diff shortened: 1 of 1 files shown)


worker/draft.go
```diff
@@ -1742,6 +1743,10 @@ func (n *node) InitAndStartNode() {
 		}
 	}
 
+	if n.RaftContext.IsLearner && !hasPeer {
```
ahsanbarkati, line 1746: For the case of restart, should we check again after making connections to nodes in ConfState? I don't know why it is required.

jarifibrahim: I suggest we keep this here for now.

jarifibrahim (approved):

PR #7584, opened 2021-03-16
[Breaking] fix(GRAPHQL): fix input coercing for integers and make them GraphQL spec complaint.

JatinDevDG:
GraphQL spec doesn't allow any value apart from `integer` to be passed to a variable of type `Int`. So we are disallowing strings to be passed as `Int` variables. Int64 is our own scalar data type, so we continue to accept `string` and `Integer` values for it.
This PR allows passing `strings` to `Int64` values when given directly in query/mutations and disallows `string` to be passed to `Int` in variables.

`So we will have below final behavior: `
```
`Int32` will accept only `Integer` values, and otherwise will give a coercion error.
`Int64` will accept `Integer` as well as `String` value and will return an error if any other value. And if we can only pass Integer in a string like "1234" otherwise it will give a coercion error.
```

`Int coercion rules in GraphQL spec:`
https://spec.graphql.org/June2018/#sec-Int

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7584)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/dgraph-io/badger/v3 v3.0.0-20210309075542-2245c18dfd1f
 	github.com/dgraph-io/dgo/v200 v200.0.0-20210212152539-e0a5bde40ba2
 	github.com/dgraph-io/gqlgen v0.13.2
-	github.com/dgraph-io/gqlparser/v2 v2.1.8
+	github.com/dgraph-io/gqlparser/v2 v2.1.9
 	github.com/dgraph-io/graphql-transport-ws v0.0.0-20210223074046-e5b8b80bb4ed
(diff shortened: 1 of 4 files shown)


abhimanyusinghgaur (approved):
:lgtm:
Nice change in test.

In the PR title and description, we should correct some typos:
* Graphql -> GraphQL
* specs -> spec
* apart from `Int` to be -> apart from integer to be (basically, `Int` anythwere -> integer)
* variable of type `Int32` -> variable of type `Int` (basically, `Int32` anywhere -> Int)

> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7584#-:-MVu0AYSBQky8ONusRsd:b-8ljzmx)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @manishrjain, @pawanrawal, and @vvbalaji-dgraph)

PR #7609, opened 2021-03-18
Fix(rollups): Fix splits in roll-up

ahsanbarkati:
There was an issue with split creation in roll-up. When a part `p` is split 
into two parts say `p1` and `p2` then the first UID of `p2` should be greater
than the last UID of p1. This was being violated because of removing UIDs 
until `maxInt64` instead of `maxUint64` from `p` in order to create `p1`. 

This PR fixes this issue and also removes some unwanted codes. 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7609)

### DIFF
--- a/posting/list.go
+++ b/posting/list.go
@@ -37,7 +37,6 @@ import (
 	"github.com/dgraph-io/dgraph/x"
 	"github.com/dgraph-io/ristretto/z"
 	"github.com/dgraph-io/roaring/roaring64"
-	"github.com/golang/glog"
 	"github.com/golang/protobuf/proto"
 )
 
@@ -1133,7 +1132,7 @@ func (ro *rollupOutput) split(startUid uint64) error {
 	newpl.Postings = pl.Postings[idx:]
 
 	// Update pl as well. Keeps the lower UIDs.
-	codec.RemoveRange(r, uid, math.MaxInt64)
+	codec.RemoveRange(r, uid, math.MaxUint64)
 	pl.Bitmap = codec.ToBytes(r)
 	pl.Postings = pl.Postings[:idx]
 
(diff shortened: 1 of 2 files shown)


posting/list.go
```diff
@@ -1191,6 +1193,8 @@ func (l *List) encode(out *rollupOutput, readTs uint64, split bool) error {
 		out.parts[startUid] = plist
 	}
 
+	out.plist.Bitmap = codec.ToBytes(bm)
```
ahsanbarkati, line 1196: Need to figure out why things were working without this change.

ahsanbarkati: Add a comment here, we don't set plist here. It is set by the finalize() function at the end.

posting/list.go
```diff
@@ -963,12 +962,13 @@ func (l *List) ToBackupPostingList(bl *pb.BackupPostingList, alloc *z.Allocator)
 	// out is only nil when the list's minTs is greater than readTs but readTs
 	// is math.MaxUint64 so that's not possible. Assert that's true.
 	x.AssertTrue(out != nil)
-	// defer out.free()
 
 	ol := out.plist
 	bm := roaring64.New()
-	if err := bm.UnmarshalBinary(ol.Bitmap); err != nil {
-		return nil, errors.Wrapf(err, "failed when unmarshal binary bitmap")
+	if ol.Bitmap != nil {
```
jarifibrahim, line 968: Is this nil check necessary?

jarifibrahim (approved):
The PR looks good to me. Can we please add a test for this @ahsanbarkati ?

posting/list.go
ahsanbarkati: Yes. In case of empty list `ol.Bitmap` turns out to be `nil` and `bm.UnmarshalBinray()` panics.

PR #7630, opened 2021-03-22
fix(GraphQL): Fix panic when no schema exists for a new namespace

abhimanyusinghgaur:
This PR fixes the issue where if you create a new namespace and run a GraphQL query for the newly created namespace without first applying any GraphQL schema, then you would get a `panic` in alpha.
Now, such a query will not result in `panic`, but instead will give back a similar response:
```
{
  "errors": [
    {
      "message": "Not resolving getAuthor. There's no GraphQL schema in Dgraph.  Use the /admin API to add a GraphQL schema"
    }
  ]
}
``` 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7630)

### DIFF
--- a/graphql/admin/admin.go
+++ b/graphql/admin/admin.go
@@ -39,7 +39,7 @@ import (
 const (
 	errMsgServerNotReady = "Unavailable: Server not ready."
 
-	errNoGraphQLSchema = "Not resolving %s. There's no GraphQL schema in Dgraph.  " +
+	errNoGraphQLSchema = "Not resolving %s. There's no GraphQL schema in Dgraph. " +
 		"Use the /admin API to add a GraphQL schema"
 	errResolverNotFound = "%s was not executed because no suitable resolver could be found - " +
 		"this indicates a resolver or validation bug. Please let us know by filing an issue."
--- a/graphql/admin/admin.go
+++ b/graphql/admin/admin.go
@@ -814,15 +814,19 @@ func resolverFactoryWithErrorMsg(msg string) resolve.ResolverFactory {
 	return resolve.NewResolverFactory(qErr, mErr)
 }
 
-func (as *adminServer) incrementSchemaUpdateCounter(ns uint64) {
-	// Increment the Epoch when you get a new schema. So, that subscription's local epoch
-	// will match against global epoch to terminate the current subscriptions.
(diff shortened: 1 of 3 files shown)


graphql/admin/admin.go
```diff
@@ -885,15 +892,18 @@ func (as *adminServer) lazyLoadSchema(namespace uint64) {
 		return
 	}
 
+	var generatedSchema schema.Schema
 	if sch.Schema == "" {
+		// if there was no schema stored in Dgraph, we still need to attach resolvers to the main
+		// graphql server which should just return errors for any incoming request.
+		// generatedSchema will be nil in this case
 		glog.Infof("No GraphQL schema in Dgraph; serving empty GraphQL API")
```
vmrajas, line 900: Why has the return from here being removed ?

graphql/e2e/multi_tenancy/schema_test.go
```diff
@@ -175,17 +175,37 @@ func TestGraphQLResponse(t *testing.T) {
...
+		queryAuthor {
+			name
+		}
+	}`
+	resp0 := (&common.GraphQLParams{Query: query, Headers: header}).ExecuteAsPost(t,
+		groupOneGraphQLServer)
+	resp1 := (&common.GraphQLParams{Query: query, Headers: header1}).ExecuteAsPost(t,
+		groupOneGraphQLServer)
+	expectedErrs := x.GqlErrorList{{Message: "Not resolving queryAuthor. " +
+		"There's no GraphQL schema in Dgraph.  Use the /admin API to add a GraphQL schema"}}
```
vmrajas, line 197: nit: There are two spaces between the two sentences. I believe this is because there are two spaces between the sentences in `errNoGraphQLSchema` . Can you also fix that with this.

vmrajas (approved):

graphql/admin/admin.go
abhimanyusinghgaur: If we return from here, then there will be no resolver attached to the main graphql server for the given namespace. Which is what results into panic when a query is sent for that namespace.

graphql/e2e/multi_tenancy/schema_test.go
abhimanyusinghgaur: Nice catch! Done

PR #7631, opened 2021-03-22
fix(Query): [Breaking] Return error for illegal math operations.

rohanprasad:
If an operand is passed to math operations like log (ln, logbase), sqrt, pow  which might result in illegal operation (resulting in NaN) we return 0 instead of an error. This PR fixes that.

Fixes DGRAPH-3150

Operation | Data type | Current behavior | New behavior
-- | -- | -- | -- 
Add | Int64 | Overflow/Underflow | Return error
Sub | Int64 | Overflow/Underflow | Return error
Multiply | Int64 | Overflow/Underflow | Return error | Return error
Pow | Float64 | Fractional power of -ve number returns 0 | Return error
Logbase | Int64 | -ve number returns 0 | Return error
Logbase | Int64 | Log with base 1 returns large float | Return error
Logbase | Float64 | -ve number return 0 | Return error
Logbase | Float64 | Log with base 1 returns large float | Return error
Ln | Int64 | -ve number returns 0 | Return error
Ln | Float64 | -ve number returns 0 | Return error
Negative | Int64 | -ve of MinInt64 remains same | Return error
Sqrt | Int64 | -ve number return 0 | Return error
Sqrt | Float64 | -ve number return 0 | Return error

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7631)

### DIFF
--- a/query/aggregator.go
+++ b/query/aggregator.go
@@ -118,28 +118,15 @@ func applyAdd(a, b, c *types.Val) error {
 	case INT:
 		aVal, bVal := a.Value.(int64), b.Value.(int64)
 
-		if aVal > 0 && bVal > math.MaxInt64-aVal {
-			return ErrorIntOverflow
-		}
-
-		if aVal < 0 && bVal < math.MinInt64-aVal {
+		if (aVal > 0 && bVal > math.MaxInt64-aVal) ||
+			(aVal < 0 && bVal < math.MinInt64-aVal) {
 			return ErrorIntOverflow
 		}
 
 		c.Value = aVal + bVal
 
 	case FLOAT:
-		aVal, bVal := a.Value.(float64), b.Value.(float64)
-
-		if aVal > 0 && bVal > math.MaxFloat64-aVal {
-			return ErrorFloatOverflow
-		}
-
-		if aVal < 0 && bVal < math.MaxFloat64-aVal {
-			return ErrorFloatOverflow
-		}
-
-		c.Value = aVal + bVal
+		c.Value = a.Value.(float64) + b.Value.(float64)
 
 	case DEFAULT:
 		return errors.Errorf("Wrong type %v encountered for func +", a.Tid)
@@ -153,28 +140,15 @@ func applySub(a, b, c *types.Val) error {
 	case INT:
 		aVal, bVal := a.Value.(int64), b.Value.(int64)
 
-		if bVal < 0 && aVal > math.MaxInt64+bVal {
-			return ErrorIntOverflow
-		}
-
-		if bVal > 0 && aVal < math.MinInt64+bVal {
+		if (bVal < 0 && aVal > math.MaxInt64+bVal) ||
+			(bVal > 0 && aVal < math.MinInt64+bVal) {
 			return ErrorIntOverflow
 		}
 
 		c.Value = aVal - bVal
 
 	case FLOAT:
-		aVal, bVal := a.Value.(float64), b.Value.(float64)
-
-		if bVal < 0 && aVal > math.MaxFloat64+bVal {
-			return ErrorFloatOverflow
-		}
-
-		if bVal > 0 && aVal < -math.MaxFloat64+bVal {
-			return ErrorFloatOverflow
-		}
-
-		c.Value = aVal - bVal
+		c.Value = a.Value.(float64) - b.Value.(float64)
 
 	case DEFAULT:
 		return errors.Errorf("Wrong type %v encountered for func -", a.Tid)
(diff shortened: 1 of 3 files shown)


query/aggregator.go
```diff
@@ -162,13 +218,13 @@ func applyDiv(a, b, c *types.Val) error {
...
 		if b.Value.(int64) == 0 {
-			return errors.Errorf("Division by zero")
+			return ErrorDivisionByZero
 		}
 		c.Value = a.Value.(int64) / b.Value.(int64)
 
 	case FLOAT:
 		if b.Value.(float64) == 0 {
-			return errors.Errorf("Division by zero")
+			return ErrorDivisionByZero
```
ajeetdsouza, line 227: IEEE 754 says that floating point values divided by zero should return infinity. Clarify what we want to do here.

query/math.go
```diff
@@ -30,6 +30,16 @@ type mathTree struct {
 	Child []*mathTree
 }
 
+var (
+	ErrorIntOverflow     = errors.New("Integer overflow")
+	ErrorFloatOverflow   = errors.New("Float overflow")
```
ajeetdsouza, line 35: Floats don't overflow. Since IEEE 754 is already best-effort and widely used, I think it should be fine to remove all floating-point overflow checks.

jarifibrahim (approved):
Looks good to me. Defer to @ajeetdsouza for the final approval.

query/math.go
```diff
@@ -30,6 +30,16 @@ type mathTree struct {
...
+	ErrorDivisionByZero  = errors.New("Division by zero")
+	ErrorModuloByZero    = errors.New("Modulo by zero")
```
ajeetdsouza, line 37: Combine these two. Python, for example, would return a `ZeroDivisionError` in both cases. Also, ensure that we only return this for integers.

query/math.go
ajeetdsouza, line 35: Don't return errors for floats - stick to IEEE 754 standards. If there's an invalid operation, return `NaN`, etc. When encoding this as JSON, encode these special values as strings:
- `"NaN"`
- `"+Inf"`
- `"-Inf"`

query/math.go
```diff
@@ -30,6 +30,16 @@ type mathTree struct {
...
+	ErrorFractionalPower = errors.New("Fractional power of negative number")
+	ErrorNegativeLog     = errors.New("Log of negative number")
+	ErrorNegativeRoot    = errors.New("Root of negative number")
```
ajeetdsouza, line 40: All of these operations promote integer values to float, AFAIK. If we're returning floats here, we can just return `NaN` instead of an error.

ajeetdsouza (changes requested):
Create a table with the old and new behaviours of all the math operations we've changed, and notify the docs team.

query/math.go
rohanprasad: As discussed in the follow-up call, returning an error instead as there's no provision for NaN in JSON.

query/math.go
rohanprasad: As we can't handle NaN and Infinity in JSON, falling back to errors.

query/math.go
```diff
@@ -30,6 +30,20 @@ type mathTree struct {
...
+)
+
+const (
+	NotANumber       = "NaN"
+	PositiveInfinity = "+Inf"
+	NegativeInfinity = "-Int"
+)
```
ajeetdsouza, line 45: Do we still need these?

query/aggregator.go
```diff
@@ -146,7 +160,14 @@ func applyMul(a, b, c *types.Val) error {
 	vBase := getValType(a)
 	switch vBase {
 	case INT:
-		c.Value = a.Value.(int64) * b.Value.(int64)
+		aVal, bVal := a.Value.(int64), b.Value.(int64)
+		c.Value = aVal * bVal
+
+		if aVal == 0 || bVal == 0 || aVal == 1 || bVal == 1 {
+			return nil
+		} else if aVal == math.MinInt64 || bVal == math.MinInt64 || c.Value.(int64)/bVal != aVal {
```
ajeetdsouza, line 168: Why do we check for `math.MinInt64` here? This doesn't overflow when multiplying by `0` or `1`, for example.

query/math.go
rohanprasad: Removed

query/aggregator.go
rohanprasad: 0 and 1 are handled in the first if block. Otherwise, it can overflow for any other integer value.

query/aggregator.go
```diff
@@ -146,7 +160,14 @@ func applyMul(a, b, c *types.Val) error {
 	vBase := getValType(a)
 	switch vBase {
 	case INT:
-		c.Value = a.Value.(int64) * b.Value.(int64)
+		aVal, bVal := a.Value.(int64), b.Value.(int64)
+		c.Value = aVal * bVal
+
+		if aVal == 0 || bVal == 0 || aVal == 1 || bVal == 1 {
```
NamanJain8, line 166: Remove checks for 1 and minInt64 as they are handled in `c.Value.(int64)/bVal != aVal` case.

NamanJain8 (approved):
Looks good to me.

PR #7637, opened 2021-03-23
fix(login): fix login based on refresh token logic

NamanJain8:
We were fetching the namespace from the `api.LoginRequest.Namespace` even if the user has sent the refresh token. We should not do that and extract the namespace from the refresh token itself.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7637)

### DIFF
--- a/edgraph/access_ee.go
+++ b/edgraph/access_ee.go
@@ -584,16 +584,12 @@ func upsertGroot(ctx context.Context, passwd string) error {
 }
 
 // extract the userId, groupIds from the accessJwt in the context
-func extractUserAndGroups(ctx context.Context) ([]string, error) {
+func extractUserAndGroups(ctx context.Context) (*userData, error) {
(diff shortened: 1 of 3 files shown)


edgraph/access_ee.go
```diff
@@ -579,7 +589,11 @@ func extractUserAndGroups(ctx context.Context) ([]string, error) {
 	if err != nil {
 		return nil, err
 	}
-	return validateToken(accessJwt)
+	userData, err := validateToken(accessJwt)
+	if err != nil {
+		return nil, err
+	}
+	return append([]string{userData.userId}, userData.groupIds...), nil
```
ahsanbarkati, line 596: We can return `userData` here.

ahsanbarkati (approved):

edgraph/access_ee.go
NamanJain8: Done.

PR #7646, opened 2021-03-24
Chore(GraphQL): Fix query logging for mutations in case of panic

vmrajas:
This adds logging for mutations in case of panic.

Testing:
Tested Locally

```
=== RUN   TestRunAll_Normal/panic_catcher/mutation
E0324 18:20:07.423360  955317 panics.go:34] panic: 
****
this test should trap this panic.
It's working as expected if this message is logged with a stack trace
****.
 query: mutation {
						addCountry(input: [{ name: "A Country" }]) { country { id } }
					}
 trace: goroutine 5272 [running]:
```

Fixes GRAPHQL-1062

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7646)

### DIFF
--- a/graphql/resolve/resolver.go
+++ b/graphql/resolve/resolver.go
@@ -480,15 +480,15 @@ func (r *RequestResolver) Resolve(ctx context.Context, gqlReq *schema.Request) (
 			},
 		},
 	}
-	defer func() {
-		// Panic Handler for mutation. This ensures that the mutation which causes panic
-		// gets logged in Alpha logs. This panic handler overrides the default Panic Handler
-		// used in recoveryHandler in admin/http.go
-		api.PanicHandler(
-			func(err error) {
-				resp.Errors = schema.AsGQLErrors(schema.AppendGQLErrs(resp.Errors, err))
-			}, gqlReq.Query)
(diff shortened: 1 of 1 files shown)


graphql/resolve/resolver.go
```diff
@@ -480,15 +480,15 @@ func (r *RequestResolver) Resolve(ctx context.Context, gqlReq *schema.Request) (
...
-		// gets logged in Alpha logs. This panic handler overrides the default Panic Handler
-		// used in recoveryHandler in admin/http.go
-		api.PanicHandler(
-			func(err error) {
-				resp.Errors = schema.AsGQLErrors(schema.AppendGQLErrs(resp.Errors, err))
-			}, gqlReq.Query)
+	// Panic Handler for mutation. This ensures that the mutation which causes panic
+	// gets logged in Alpha logs. This panic handler overrides the default Panic Handler
+	// used in recoveryHandler in admin/http.go
+	defer api.PanicHandler(
```
abhimanyusinghgaur, line 486: Maybe we should also add a comment about why we can't merge this defer with the other defer. That would be really helpful for anyone else to get context.

abhimanyusinghgaur (approved):

PR #7668, opened 2021-03-30
fix(query): Fix pagination with match functions

ahsanbarkati:
Due to narrowing down the query to only fetch N results from the
postings, some of the queries got broken. Some queries like 
`allofterms` requires fetching all the uids/postings from the posting list 
corresponding to each term relevant to the queries. And then do an
intersection of the results corresponding to each token. 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7668)

### DIFF
--- a/query/common_test.go
+++ b/query/common_test.go
@@ -328,6 +328,9 @@ gender                         : string .
 indexpred                      : string @index(exact) .
 pred                           : string .
 pname                          : string .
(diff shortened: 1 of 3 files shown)


NamanJain8 (approved):
Awesome 🎉

NamanJain8:
Also, please cherry-pick it to release/v20.11

PR #7643, opened 2021-03-24
release/v20.11-slash: perf(compression): Use gzip with BestSpeed in export and backup

jarifibrahim:
The export and backup code were using BestCompression and Default
gzip compression levels respectively. We see a 4x speed improvement when
switch from BestCompression to BestSpeed compression level.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7643)

PR #7683, opened 2021-04-05
release/v21.03 - perf(compression): Use gzip with BestSpeed in export and backup (#7643)

jarifibrahim:
The export and backup code were using BestCompression and Default
gzip compression levels respectively. We see a 4x speed improvement when
switching from BestCompression to BestSpeed compression level.

(cherry picked from commit 5676462f3601c641e75576acc45e491bddb06dac)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7683)


NamanJain8 (approved):
We will be using snappy instead of gzip in backup. Anyways, it should be fine to make these incremental improvements in performance, in case the snappy one doesn't make it into release.

PR #7688, opened 2021-04-06
chore: upgrade etcd/raft to v3

NamanJain8:


jarifibrahim (commented):
This PR needs a description. 
@rohanprasad Isn't this a breaking change for us? Is this change backward compatible?

harshil-goel (dismissed):

worker/draft.go
```diff
@@ -1839,7 +1839,7 @@ func (n *node) InitAndStartNode() {
 			// snapshot as needed after joining the group, instead of us forcing one upfront.
 			glog.Infoln("Trying to join peers.")
 			n.retryUntilSuccess(n.joinPeers, time.Second)
-			n.SetRaft(raft.StartNode(n.Cfg, nil))
+			n.SetRaft(raft.RestartNode(n.Cfg))
```
all-seeing-code, line 1842: @mangalaman93 I think we should take a closer look here because this case is not of `restart`. It is possible that the API is same and start/restart is just syntactic sugar but this is counter intuitive. At the very least we should add a comment on top of this as well.

mangalaman93: Done

mangalaman93 (dismissed):

worker/draft.go
```diff
@@ -1813,7 +1813,7 @@ func (n *node) InitAndStartNode() {
 
 			// TODO: Making connections here seems unnecessary, evaluate.
 			members := groups().members(n.gid)
-			for _, id := range sp.Metadata.ConfState.Nodes {
+			for _, id := range sp.Metadata.ConfState.Voters {
```
all-seeing-code, line 1816: @mangalaman93 apologies for not highlighting this earlier, if we could verify that this change in API is legit in a sense that Node -> Voters in v3, then I'd feel more comfortable approving this. Do we have a changelog which points to this?

worker/draft.go
mangalaman93: Tests are not happy when I switch to using StartNode. Let me find an explanation for it.

worker/draft.go
mangalaman93: it comes from this PR https://github.com/etcd-io/etcd/pull/10914. Mentioned in the changelog too https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.4.md

harshil-goel (approved):

harshil-goel (approved):

worker/draft.go
all-seeing-code: This is resolved?

worker/draft.go
all-seeing-code: Exact changelog for the above: https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.4.md#package-raft

worker/draft.go
mangalaman93: Yes, you should see a comment in the diff now.

all-seeing-code (approved):

PR #7699, opened 2021-04-07
Remove LastBackupStatus

ajeetdsouza:
Users can check logs to know the status of backups.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7699)


testutil/client.go
```diff
@@ -532,3 +533,35 @@ func GetAlphaClientConfig(t *testing.T) *tls.Config {
 	require.NoError(t, err)
 	return tlsConf
 }
+
+func WaitForTask(t *testing.T, task string) {
+	healthUrl := "http://" + SockAddrHttp + "/health"
+	for {
+		time.Sleep(5 * time.Second)
+
+		health := func() []pb.HealthInfo {
```
jarifibrahim, line 542: Why do you need a function here?
Also, I see health variable inside the function as well. 
Please use a different variable name to avoid confusion.

ajeetdsouza: I'm using a function so that `defer response.Body.Close()` runs immediately, so that the connection can be re-used in the next loop. I've fixed the variable name, could you take another look?

jarifibrahim (approved):

PR #7700, opened 2021-04-07
fix(pagination): Fix after for regexp, match functions

rohanprasad:
This change fixes the pagination functioning that used "after" to specify a UID as offset.

Fixes DGRAPH-3222

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7700)

### DIFF
--- a/query/query1_test.go
+++ b/query/query1_test.go
@@ -2545,3 +2545,31 @@ func TestExpandAll_empty_panic(t *testing.T) {
 	js := processQueryNoErr(t, query)
 	require.JSONEq(t, `{"data":{"me":[]}}`, js)
 }
+
+func TestMatchFuncWithAfter(t *testing.T) {
+	query := `
+		{
+			q(func: match(name, Ali, 5), after: 0x2710) {
+				uid
+				name
+			}
+		}
+	`
+
+	js := processQueryNoErr(t, query)
+	require.JSONEq(t, `{"data": {"q": [{"name": "Alice", "uid": "0x2712"}, {"name": "Alice", "uid": "0x2714"}]}}`, js)
+}
+
+func TestCompareFuncWithAfter(t *testing.T) {
+	query := `
+		{
+			q(func: eq(name, Alice), after: 0x2710) {
+				uid
+				name
+			}
+		}
+	`
+
+	js := processQueryNoErr(t, query)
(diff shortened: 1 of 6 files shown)


ahsanbarkati (changes requested):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7700)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @manishrjain, @pawanrawal, @rohanprasad, and @vvbalaji-dgraph)

___
*[worker/task.go, line 1547 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7700#-MXgV4h4EL8kwJwUfnjC:-MXgV4h4EL8kwJwUfnjD:b-ohkv8d) ([raw file](https://github.com/dgraph-io/dgraph/blob/3596e860acc6f92d790f5876b21b3ad0695b1fd0/worker/task.go#L1547)):*
> ```Go
> 			}
> 			var tv pb.TaskValue
> 			err = pl.Iterate(arg.q.ReadTs, 0, func(p *pb.Posting) error {
> ```

The second argument in `Iterate()` function is for afterUid. You can just pass it there, so no need for the if block below.

NamanJain8 (commented):
I see a [bunch of places](https://github.com/dgraph-io/dgraph/blob/ef382b18c365fcf06dfa317f15c64ef7fae372ab/worker/task.go#L824-L846) where we are setting AfterUid to `0`. Can you check what might be the reason for that? Or do we actually need to add this `AfterUid` check over there as well.

rohanprasad:
___
*[worker/task.go, line 1547 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7700#-MXgV4h4EL8kwJwUfnjC:-MXm3dTo0LsSKxPjDioD:b-32uup2) ([raw file](https://github.com/dgraph-io/dgraph/blob/3596e860acc6f92d790f5876b21b3ad0695b1fd0/worker/task.go#L1547)):*
<details><summary><i>Previously, ahsanbarkati (Ahsan Barkati) wrote…</i></summary><blockquote>

The second argument in `Iterate()` function is for afterUid. You can just pass it there, so no need for the if block below.
</blockquote></details>

Verified, this works fine without the change.

NamanJain8 (approved):

PR #7703, opened 2021-04-08
opt(predMove): hot tablet move

NamanJain8:
Currently, the writes are blocked on the tablet that is being moved. Shard movement can take a while (say 30 mins) to be moved completely. Then, its not a good idea to block the writes on that predicate.

**Solution:  Hot Tablet Move**
1. Don't block initially. Do majority of the move first.
2. Block commits. Do incremental move. Resume commits.

TODO: Fix the issues found while working on this.
1. If a drop operation is triggered while the tablet move is in progress, the drop succeeds and we end up in an inconsistent state.
2. There is a corner-case, where an indexing operation might succeed before the tablet has moved completely. Once, the tablet move is complete, the indexing goes away, leading to a bad user experience. 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7703)


manishrjain (commented):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7703)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @NamanJain8)

___
*[dgraph/cmd/zero/tablet.go, line 179 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7703#-MXmV5VGB8qSEhkRj7aS:-MXmV5VGB8qSEhkRj7aT:btgi370) ([raw file](https://github.com/dgraph-io/dgraph/blob/855867894f51fdbeb64f28adf5720a8361562d9a/dgraph/cmd/zero/tablet.go#L179)):*
> ```Go
> 		SourceGid: srcGroup,
> 		DestGid:   dstGroup,
> 		TxnTs:     ids.StartId,
> ```

Rename it to ReadTs.
___
*[worker/predicate_move.go, line 97 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7703#-MXmWUOyCp4kKksVpiY6:-MXmWUOyCp4kKksVpiY7:b-whber5) ([raw file](https://github.com/dgraph-io/dgraph/blob/855867894f51fdbeb64f28adf5720a8361562d9a/worker/predicate_move.go#L97)):*
> ```Go
> 						return errors.Errorf("Expecting clean predicate instruction with schema key: %+v", kv)
> 					}
> 					kv.Meta = nil
> ```

Use stream_id.

if stream_id == 0, then do the clean up.
If stream_id == 1, then don't.

You can then create an enum or something to set the stream id, so it's clear to the code reader.

ahsanbarkati (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7703#-:-MY4Hdea7RQ7qCYQYJRy:bnfp4nl)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @NamanJain8 and @vvbalaji-dgraph)

___
*[dgraph/cmd/zero/tablet.go, line 57 at r2](https://reviewable.io/reviews/dgraph-io/dgraph/7703#-MY4DqmE-KA9t5ais45O:-MY4DqmE-KA9t5ais45P:b-owfsdo) ([raw file](https://github.com/dgraph-io/dgraph/blob/194af85940d9951541ef6acc75770c1ef23c5d76/dgraph/cmd/zero/tablet.go#L57)):*
> ```Go
> • Zero would tell G1 to move rest of the P's data (with T1<=ts<=T2) to G2 (Endpoint: Zero → G1)
> • G2 acceps the data, but this time does not clean the current keys for P.
> • G1 should tell Zero whetjer it succeeded or failed. (Endpoint G1 → Zero)
> ```

whetjer -> whether

dgraph/cmd/zero/tablet.go
```diff
@@ -176,12 +176,37 @@ func (s *Server) movePredicate(predicate string, srcGroup, dstGroup uint32) erro
 		Predicate: predicate,
 		SourceGid: srcGroup,
 		DestGid:   dstGroup,
-		TxnTs:     ids.StartId,
+		ReadTs:    ids.StartId,
+		SinceTs:   0,
+	}
+
+	// Move the predicate. Commits on this predicate is not blocked yet. Any data after TxnTs
```
jarifibrahim, line 183: nit: Commits on this predicate _are_ not blocked yet. Any data above the _read_ts_

jarifibrahim (approved):
Nice change! :100:

simpl:
@NamanJain8 

Rather than just doing one pass of phase 1 and one of phase 2, could the downtime due to phase 2 be minimized by doing possibly multiple cycles of phase 1 bringing it up to present (or almost present) before doing a final phase 2 that blocks only for a few seconds at most?

What I mean is something like this:

1) Phase 1 as current
2) At end of phase 1, check to see if any commits have happened during (1) or bring the checks up to the current (or very recent) timestamp, restarting phase 1 if necessary (i.e. still non-blocking for writes)
3) Repeat 1+2 as many times as necessary to bring it up to present or nearly present
4) Go to phase 2 when it should be possible to only have a few seconds of downtime for commits at most

PR #7713, opened 2021-04-09
fix(ee): GetKeys should return an error

ajeetdsouza:
Currently, `GetKeys` exits immediately upon error. While this works for most uses, it ends up shutting down the Alpha when a restore request is made with bad input. This PR fixes that.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7713)

### DIFF
--- a/dgraph/cmd/decrypt/decrypt.go
+++ b/dgraph/cmd/decrypt/decrypt.go
@@ -57,9 +57,7 @@ func init() {
 }
 func run() {
 	keys, err := ee.GetKeys(Decrypt.Conf)
-	if err != nil {
-		glog.Fatal(err)
-	}
+	x.Check(err)
 	if len(keys.EncKey) == 0 {
 		glog.Fatal("Error while reading encryption key: Key is empty")
 	}
--- a/dgraph/cmd/decrypt/decrypt.go
+++ b/dgraph/cmd/decrypt/decrypt.go
@@ -18,15 +18,14 @@ package decrypt
 
 import (
 	"compress/gzip"
-	"fmt"
 	"io"
-	"log"
 	"os"
 	"strings"
 
 	"github.com/dgraph-io/dgraph/ee"
 	"github.com/dgraph-io/dgraph/ee/enc"
 	"github.com/dgraph-io/dgraph/x"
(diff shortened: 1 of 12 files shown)


dgraph/cmd/decrypt/decrypt.go
```diff
@@ -57,17 +56,23 @@ func init() {
 	ee.RegisterEncFlag(flag)
 }
 func run() {
-	opts := options{
-		file:   Decrypt.Conf.GetString("file"),
-		output: Decrypt.Conf.GetString("out"),
+	keys, err := ee.GetKeys(Decrypt.Conf)
+	if err != nil {
```
jarifibrahim, line 60: Use x.Check everywhere.

dgraph/cmd/live/run.go
```diff
@@ -697,8 +697,11 @@ func run() error {
 	}
 
 	creds := z.NewSuperFlag(Live.Conf.GetString("creds")).MergeAndCheckDefault(x.DefaultCreds)
+	keys, err := ee.GetKeys(Live.Conf)
+	if err != nil {
```
jarifibrahim, line 701: use x.Check

ee/keys_ee.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/dgraph-io/ristretto/z"
+	"github.com/spf13/viper"
+)
+
+// GetKeys returns the ACL and encryption keys as configured by the user
+// through the --acl, --encryption, and --vault flags. On OSS builds,
+// this function always returns an error.
+func GetKeys(config *viper.Viper) (*Keys, error) {
+	keys := &Keys{}
+	var err error
```
jarifibrahim, line 28: can this be removed?

jarifibrahim (approved):
Some minor comments.

ee/keys_ee.go
ajeetdsouza: The first use of `err` assigns directly to a struct:

```go
keys.AclKey, err = ioutil.ReadFile(aclKeyFile)
```

So we can't remove the error declaration, unfortunately.

PR #7718, opened 2021-04-13
feat(GraphQL): add selection set of query to the body of lambda request.

minhaj-shakeel:
Fixes DGRAPH-3180

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7718)

### DIFF
--- a/graphql/e2e/common/common.go
+++ b/graphql/e2e/common/common.go
@@ -915,6 +915,7 @@ func RunAll(t *testing.T) {
 	t.Run("lambda on type field", lambdaOnTypeField)
 	t.Run("lambda on interface field", lambdaOnInterfaceField)
 	t.Run("lambda on query using dql", lambdaOnQueryUsingDql)
+	t.Run("lambda on query with filters", lambdaOnQueryWithFilters)
 	t.Run("lambda on mutation using graphql", lambdaOnMutationUsingGraphQL)
 	t.Run("lambda on query with no unique parents", lambdaOnQueryWithNoUniqueParents)
 	t.Run("query lambda field in a mutation with duplicate @id", lambdaInMutationWithDuplicateId)
--- a/graphql/e2e/common/common.go
+++ b/graphql/e2e/common/common.go
@@ -915,7 +915,6 @@ func RunAll(t *testing.T) {
 	t.Run("lambda on type field", lambdaOnTypeField)
 	t.Run("lambda on interface field", lambdaOnInterfaceField)
 	t.Run("lambda on query using dql", lambdaOnQueryUsingDql)
-	t.Run("lambda on query with filters", lambdaOnQueryWithFilters)
 	t.Run("lambda on mutation using graphql", lambdaOnMutationUsingGraphQL)
 	t.Run("lambda on query with no unique parents", lambdaOnQueryWithNoUniqueParents)
 	t.Run("query lambda field in a mutation with duplicate @id", lambdaInMutationWithDuplicateId)
--- a/graphql/e2e/common/lambda.go
+++ b/graphql/e2e/common/lambda.go
@@ -135,32 +135,6 @@ func lambdaOnInterfaceField(t *testing.T) {
 }
 
 func lambdaOnQueryUsingDql(t *testing.T) {
-	query := `
-		query {
-			authorsByName(name: "Ann Author") {
-				name
-				dob
-				reputation
-			}
-		}`
-	params := &GraphQLParams{Query: query}
-	resp := params.ExecuteAsPost(t, GraphqlURL)
-	RequireNoGQLErrors(t, resp)
-
-	expectedResponse := `{
-		"authorsByName": [
-			{
-				"name":"Ann Author",
-				"dob":"2000-01-01T00:00:00Z",
-				"reputation":6.6
-			}
-		]
-	}`
-	testutil.CompareJSON(t, expectedResponse, string(resp.Data))
-}
-
-func lambdaOnQueryWithFilters(t *testing.T) {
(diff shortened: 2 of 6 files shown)


abhimanyusinghgaur (changes requested):
> Reviewed 2 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7718)* status: 2 of 4 files reviewed, 7 unresolved discussions (waiting on @minhaj-shakeel and @pawanrawal)

___
*[graphql/e2e/common/lambda.go, line 162 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7718#-MYA8G2dEnSaV1shtGhZ:-MYA8G2dEnSaV1shtGh_:btk484q) ([raw file](https://github.com/dgraph-io/dgraph/blob/c856000681921c9345cea330d508975dc0062cc4/graphql/e2e/common/lambda.go#L162)):*
> ```
> lambdaOnQueryUsingDql
> ```
1. seems we can merge this into the above `lambdaOnQueryUsingDql` test. Also, need to update the lambda script files to make use of the `info` for this query.
2. We should also add a test for a lambda field that returns an object, there is no such test yet. So that we can test that nested filters are propagated for lambda fields as well.
___
*[graphql/schema/custom_http.go, line 197 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7718#-MYAHfGr3y9wYrTlD14i:-MYAHfGr3y9wYrTlD14j:byw2gli) ([raw file](https://github.com/dgraph-io/dgraph/blob/c856000681921c9345cea330d508975dc0062cc4/graphql/schema/custom_http.go#L197)):*
> ```
> *fldValue
> ```
maybe just make it an `interface{}` because there can be cases when the argument value comes from a variable, then the `Raw` would contain `$varName` and not the actual value of the `$varName`.
Also, since `Raw` is a string so even `Int` values will be serialized as string, resulting in issues for the users.
By using `interface{}` we can avoid all that,

OR we can go the way we saw in the Apollo code.

OR just an extension of the `interface{}` approach, we can use `f.ArgumentMap()` for the root lambda field, and then don't need to create any struct for args here. `fldArgumentList` will just be a `[]map[string]interface{}` then. Maybe in a similar manner, we won't need any other structs as well.

I prefer the 3rd one, but let's discuss it tomorrow.
___
*[graphql/schema/custom_http.go, line 220 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7718#-MYAL3V4Ctdl10SFd-HJ:-MYAL3V5A0iYiOlUPYVF:b-sg6h17) ([raw file](https://github.com/dgraph-io/dgraph/blob/c856000681921c9345cea330d508975dc0062cc4/graphql/schema/custom_http.go#L220)):*
> ```
> type fldSelectionSet []fldSelection
> type fldSelection interface{}
> ```
shouldn't this just be?
```
type fldSelectionSet []*selectionField
```

Also, let's cleanup the `fld` prefix from every struct. We can name them something better. For example, this one can just be `selectionSet`.
And, add JSON tags within every struct, which ensures that if we rename a field, we don't change the output JSON.
___
*[graphql/schema/custom_http.go, line 246 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7718#-MYAMdc64B26j-OSlMkF:-MYAMdc64B26j-OSlMkG:bfdz3kx) ([raw file](https://github.com/dgraph-io/dgraph/blob/c856000681921c9345cea330d508975dc0062cc4/graphql/schema/custom_http.go#L246)):*
> ```Go
> }
> 
> func convert(f ast.Field) selectionField {
> ```

1. maybe rename this function as `getInfo()`, and it returns an `Info` object.
2. Just thinking along what we saw in the apollo resolvers, maybe we should nest the field inside the info:
```
{
  "info": {
    "field": <the current output returned by convert here>
  }
}
```
so that if we later want to add the `returnType` in info, we can do that.
___
*[graphql/schema/custom_http.go, line 253 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7718#-MYACZZl5ighQgriQRwp:-MYACZZl5ighQgriQRwq:b-1x6nvl) ([raw file](https://github.com/dgraph-io/dgraph/blob/c856000681921c9345cea330d508975dc0062cc4/graphql/schema/custom_http.go#L253)):*
> ```
> outField.SelectionSet = make(fldSelectionSet, len(f.SelectionSet))
> ```
can be moved within the initializer
___
*[graphql/schema/custom_http.go, line 255 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7718#-MYAFWq_7KcOLBl4znG7:-MYAFWq_7KcOLBl4znG8:b-cc7tnb) ([raw file](https://github.com/dgraph-io/dgraph/blob/c856000681921c9345cea330d508975dc0062cc4/graphql/schema/custom_http.go#L255)):*
> ```
> *sel.(*ast.Field)
> ```
convert can directly accept a pointer, instead of dereferencing, creating a copy and passing that every time.
___
*[graphql/schema/wrappers.go, line 205 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7718#-MYADZg28QgmoeAk0gvG:-MYADZg28QgmoeAk0gvH:b-xf4l71) ([raw file](https://github.com/dgraph-io/dgraph/blob/c856000681921c9345cea330d508975dc0062cc4/graphql/schema/wrappers.go#L205)):*
> ```
> GetField() ast.Field
> ```
let's not expose the `ast.Field` to other packages, rather move the `convert` logic to wrappers. And actually, as the convert logic is within the same package, we shouldn't even be needing this method. So, rather remove this.

PR #7719, opened 2021-04-13
fix(raftwal): take snapshot after restore

NamanJain8:
Fixes DGRAPH-3202

We should propose a snapshot immediately after the restore to prevent the restore from being replayed.
Earlier we were trying to that too. Snapshot happens after the restore but that does not actually include the restore proposal as it has not been applied yet (`n.Applied.Done()` is not yet called for that proposal). 
Now, we will first wait for the raft entries to be marked as done and then try to propose the snapshot.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7719)

### DIFF
--- a/worker/draft.go
+++ b/worker/draft.go
@@ -752,6 +752,8 @@ func (n *node) applyCommitted(proposal *pb.Proposal) error {
 		}
 		defer closer.Done()
 
+		glog.Infof("Restoring proposal at Index:%d, ReadTs:%d",
+			proposal.Index, proposal.Restore.RestoreTs)
 		if err := handleRestoreProposal(ctx, proposal.Restore); err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


ahsanbarkati (approved):
Looks good to me. I was wondering if the snapshot should be proposed only by the leader, but seems like this is also fine considering only one proposal will be applied. And all other proposals after it will be no-op.

worker/draft.go
```diff
@@ -867,9 +869,18 @@ func (n *node) processApplyCh() {
 			ms := x.SinceMs(start)
 			ostats.RecordWithTags(context.Background(), tags, x.LatencyMs.M(ms))
 		}
-
 		n.Proposals.Done(prop.Key, perr)
 		n.Applied.Done(prop.Index)
+		if !ok && prop.Restore != nil && perr == nil {
```
jarifibrahim, line 874: why do we check for `!ok`? Can we get duplicate restore proposals?

NamanJain8: Yes, we can get duplicate proposals (not specific to restore proposals). Here, https://github.com/dgraph-io/dgraph/blob/b7e7dad2791496800cc1e9fbe40e0d89214672a4/worker/proposal.go#L293-L297 we retry the proposal 3 times. And it might happen that we retried it even if it made to raft log on any server. So for safety, we use the same unique key for the identification of duplicates.

jarifibrahim (approved):
Looks good to me. @NamanJain8 we could verify the snapshot TS in the online restore test.

You can query the /state endpoint and check if the snapshot Ts was different before the restore request was sent. 
You don't need a new test. Just use one of the existing tests.

ahsanbarkati (changes requested):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7719)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @jarifibrahim, @manishrjain, @NamanJain8, and @vvbalaji-dgraph)

ahsanbarkati (approved):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7719)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @jarifibrahim, @manishrjain, and @vvbalaji-dgraph)

Looks good to me

PR #7721, opened 2021-04-14
fix(cdc): Show namespace info in event meta

jarifibrahim:
This PR sets the namespace in the cdc event metadata.

I've also changed the type of key from `[]byte` to `uint64` because it didn't need to be a `[]byte`

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7721)

### DIFF
--- a/worker/cdc_ee.go
+++ b/worker/cdc_ee.go
@@ -360,7 +360,7 @@ type CDCEvent struct {
 
 type EventMeta struct {
 	RaftIndex uint64 `json:"-"`
-	Namespace []byte `json:"-"`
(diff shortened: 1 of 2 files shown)


aman-bansal (approved):

NamanJain8:
Looks good. 🎉  
Btw, this would be a breaking change, right?

jarifibrahim:
>Would this be a breaking change?

@NamanJain8 The JSON would have an additional field called namespace. It shouldn't be breaking.
See this https://play.golang.org/p/g4WEr-a5SSv . The "Age" in json wasn't part of the struct

PR #7725, opened 2021-04-14
fix(dgraph): Prevent proposal from being dropped accidentally

jarifibrahim:
Each proposal in alpha has a unique key but this unique key is not
guaranteed to be unique. We could end up creating duplicate keys and
these duplicate keys would get ignored in the raft loop.

To prevent this, we use the entry.Size(). This doesn't work sometimes 
because the user could be sending the exact same proposal multiple
times the entry.Size() would be equal.

This PR proposes using z.Memhash to detect collision when a duplicate
proposal is detected.

Fixes - DGRAPH-3012

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7725)

### DIFF
--- a/worker/draft.go
+++ b/worker/draft.go
@@ -672,7 +672,7 @@ func (n *node) processApplyCh() {
 
 	type P struct {
 		err  error
-		size int
+		hash uint64
 		seen time.Time
 	}
 	previous := make(map[uint64]*P)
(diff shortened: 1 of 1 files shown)


jarifibrahim:
Closing this PR in favour of https://github.com/dgraph-io/dgraph/pull/7741

PR #7726, opened 2021-04-15
Fix(GraphQL): Fix GraphQL encoding in case of empty list

vmrajas:
Motivation:
Current GraphQL encoding had a bug which led to timing out in case there is an empty list in dataset and it is queried before any non-null field. This fixes this behaviour.

Fixes DGRAPH-3257

Testing:
1. The test times out (as expected) without the change to outputnode_graphql.go
2. The test passes with the change to outputnode_graphql.go 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7726)

### DIFF
--- a/query/outputnode_graphql.go
+++ b/query/outputnode_graphql.go
@@ -449,7 +449,8 @@ func (genc *graphQLEncoder) encode(encInp encodeInput) bool {
 		// 2. We are at the end of json encoding process and there is no fastjson node ahead (next == nil)
 		// 3. We are at the end of list writing and the type of next fastJSON node is not equal to
 		//    type of curr fastJSON node.
-		// 4. The current selection set which we are encoding is not equal to the type of current fastJSON node.
+		// 4. The current selection set which we are encoding is not equal to the type of
+		//    current fastJSON node.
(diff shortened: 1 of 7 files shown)


query/outputnode_graphql.go
```diff
@@ -444,7 +444,15 @@ func (genc *graphQLEncoder) encode(encInp encodeInput) bool {
 		}
 
 		// Step-3: Update counters and Write closing ] for JSON arrays
-		if !curSelectionIsDgList || next == nil || genc.getAttr(cur) != genc.getAttr(next) {
+		// We perform this step in any of the 4 conditions is satisfied.
+		// 1. The current selection is not a Dgraph List (It's of custom type or a single JSON object)
+		// 2. We are at the end of json encoding process and there is no fastjson node ahead (next == nil)
+		// 3. We are at the end of list writing and the type of next fastJSON node is not equal to
+		//    type of curr fastJSON node.
+		// 4. The current selection set which we are encoding is not equal to the type of current fastJSON node.
```
abhimanyusinghgaur, line 452: nit: 100 chars

abhimanyusinghgaur (approved):

PR #7731, opened 2021-04-15
[BREAKING] fix(usability): make force-namespace flag compulsory in live loader for galaxy user 

NamanJain8:
**This change will be breaking because `--force-namespace` flag was optional for galaxy users earlier.** 
There might be an issue where a user may unknowingly load the data into a default namespace (when she instead wants to preserve a namespace). This can be dangerous. So, this change makes the live loader expects the user to explicitly set the `--force-namespace`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7731)


ahsanbarkati (approved):
Looks good

jarifibrahim (approved):

PR #7741, opened 2021-04-21
fix: Prevent proposal from being dropped accidentally

jarifibrahim:
Each proposal in alpha has a unique key but this unique key is not
guaranteed to be unique. We could end up creating duplicate keys and
these duplicate keys would get ignored in the raft loop.

This PR proposes a different way of creating the proposal keys so that
they are always unique.

Fixes - DGRAPH-3012

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7741)


NamanJain8 (approved):

NamanJain8 (commented):
Should we make this change in zero's unique key too?

worker/proposal.go
```diff
@@ -108,9 +109,18 @@ func (rl *rateLimiter) decr(retry int) {
 	rl.c.Broadcast()
 }
 
+var proposalKey uint64
+
+// {2 bytes Node ID} {4 bytes for random} {2 bytes zero}
+func initProposalKey(id uint64) {
+	x.AssertTrue(id != 0)
+	proposalKey = uint64(groups().Node.Id)<<48 | uint64(uint64(z.FastRand())<<16)
```
NamanJain8, line 117:
```suggestion
	proposalKey = uint64(groups().Node.Id)<<48 | uint64(z.FastRand())<<16
```

ahsanbarkati (approved):
Looks good

jarifibrahim:
>Should we make this change in zero's unique key too?

@NamanJain8 zero doesn't get that many proposals so we wouldn't have the problem of duplicate keys.

PR #7757, opened 2021-04-27
opt(GraphQL): filter existence queries on GraphQL side instead of using @filter(type)

NamanJain8:
Fixes GraphQL-1124

A simple GraphQL mutation like:
```
 mutation($cuisineName: String!) {
  addCuisine(
    input: [
      {
        name: $cuisineName
      }
    ]
  ) {
    cuisine {
      id
    }
  }
}
```
gets translated to the following **existence query**
```
 query {\n  Cuisine_1(func: eq(Cuisine.name, \"yaMqnmHeov\")) @filter(type(Cuisine)) {\n uid \n  } \n}
```
The filter operation can be very heavy on `dgraph.type` as it may store millions of nodes. A better query could be like this:
```
 query {\n  Cuisine_1(func: eq(Cuisine.name, \"yaMqnmHeov\")) {\n uid dgraph.type \n  } \n}
```
This PR removes the above @filter operation from the **existence query** and uses the suggested alternate query which is lightweight. 
The existence check operation is path critical and this fix is expected to add some performance improvement along with preventing OOM. 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7757)


graphql/e2e/common/mutation.go
```diff
@@ -3003,7 +3003,7 @@ func addMultipleMutationWithOneError(t *testing.T) {
 	require.NoError(t, err)
 
 	require.Contains(t, gqlResponse.Errors[0].Error(),
-		`because ID "0x0" isn't a Author`)
+		`Invalid UID with value 0 for key`)
```
vmrajas, line 3006: Why did we need to change this error. This does not seem right. The error seems to be coming from x/keys.go: Parse function. The error should not change.

graphql/resolve/mutation.go
```diff
@@ -98,7 +98,7 @@ type MutationRewriter interface {
 	//     }
 	// These query will be created in case of Add or Update Mutation which references node
 	// 0x127 or Computer of name "computer1"
-	RewriteQueries(ctx context.Context, m schema.Mutation) ([]*gql.GraphQuery, error)
+	RewriteQueries(ctx context.Context, m schema.Mutation) ([]*gql.GraphQuery, []schema.Type, error)
```
vmrajas, line 101: I don't understand the need for returning []schema.Type from RewriteQueries. As far as I can see only the name of the type is used in mutation.go. Let's only return an array of type names from here.
Also update the comment above which describes what RewriteQueries does.

vmrajas (changes requested):

graphql/resolve/mutation.go
```diff
@@ -297,20 +298,24 @@ func (mr *dgraphResolver) rewriteAndExecute(
...
 	// 			}
 	//		],
-	//	Column2 :
+	//	Column_2 :
 	//		[
 	//			{
 	//				"uid": "0x234"
 	// 			}
 	//		]
 	// }
```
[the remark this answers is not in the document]
NamanJain8: Done

graphql/resolve/mutation.go
```diff
@@ -320,22 +325,32 @@ func (mr *dgraphResolver) rewriteAndExecute(
 		return emptyResult(gqlErr), resolverFailed
 	}
 
+	x.AssertTrue(len(filterTypes) == len(queries))
+	qNameToFilterType := make(map[string]string)
+	for i, typ := range filterTypes {
+		qNameToFilterType[queries[i].Attr] = typ.DgraphName()
```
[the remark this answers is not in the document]
NamanJain8: Done

graphql/resolve/mutation.go
NamanJain8: Yeah, correct. Thanks.

graphql/e2e/common/mutation.go
NamanJain8: Earlier as we were using `@filter(type(Author))`, it was querying the index key and then returning `because ID "0x0" isn't a Author` error because the UID=0 does not exist in the index (Value of the index key does not contain UID 0).
But now, we will be querying `dgraph.type`, it will construct a key that has UID=0 set, hence that key will fail at the time of parsing the key.

vmrajas: As discussed over call, lets keep the error message as before while changing the invalid author ID.

vmrajas (approved):
LGTM

PR #7758, opened 2021-04-27
chore(tests): Add basic test for k-random nodes

ahsanbarkati:
Add some sanity tests for k-random nodes argument.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7758)

### DIFF
--- a/query/query0_test.go
+++ b/query/query0_test.go
@@ -3487,23 +3487,23 @@ func TestMatchingWithPagination(t *testing.T) {
 func TestKRandomNodes(t *testing.T) {
 	q := `{
 		data(func: uid(61, 62, 63, 64, 65, 66, 67, 68, 69, 70), random: 2) @filter(has(connects)) {
-			name
+			kname
 			connects(random:2){
-				name
+				kname
 			}
 		}
 	}`
 	result := processQueryNoErr(t, q)
 	expected := `{"data":{"data":[{
-		"name":"can_be_picked",
+		"kname":"can_be_picked",
 		"connects":[
-			{"name":"yes"},
-			{"name":"yes"}
+			{"kname":"yes"},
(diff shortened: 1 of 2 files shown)


query/query0_test.go
```diff
@@ -3484,6 +3484,38 @@ func TestMatchingWithPagination(t *testing.T) {
 	}
 }
 
+func TestKRandomNodes(t *testing.T) {
+	q := `{
+		data(func: uid(61, 62, 63, 64, 65, 66, 67, 68, 69, 70), random: 2) @filter(has(connects)) {
```
NamanJain8, line 3489: Node 69,70 exist?

query/query0_test.go
```diff
@@ -3484,6 +3484,38 @@ func TestMatchingWithPagination(t *testing.T) {
...
+			]},
+		{"kname":"can_be_picked",
+		"connects":[
+			{"kname":"yes"},
+			{"kname":"yes"}
+		]}]}}`
+	require.JSONEq(t, expected, result)
+
+	q = `{
+		data(func: uid(61, 62, 63, 64, 65, 66, 67, 68, 69, 70), random: 10) @filter(has(connects)) {
```
NamanJain8, line 3511: What if number of input uids is less than 10?

NamanJain8 (approved):
Looks good to me. Just a couple of questions.

query/query0_test.go
ahsanbarkati: Yes, they exist but It doesn't matter, but for clarity, I have removed them from the query.

query/query0_test.go
ahsanbarkati: It was just by chance that, number of nodes in the query and the value of random was chosen to be the same. No specific reason.

PR #7759, opened 2021-04-28
fix(bulk): upsert guardian/groot for all existing namespaces

NamanJain8:
**Issue:**
If the data was loaded into some namespace using bulk loader with `--force-namespace`, that data wasn't actually accessible.
This is because the users/groups were not created for it, if they were not present originally in the RDF. 

**Solution:**
This PR fixes that by upserting the default user and password for that namespace, when the alpha starts (to be precise, whenever ACL is reset).

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7759)


ahsanbarkati (commented):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7759#-:-MZRF3Sp2GiF6lkch8ZH:bg7igkv)* status: 0 of 7 files reviewed, all discussions resolved (waiting on @manishrjain and @vvbalaji-dgraph)

jarifibrahim (approved):
LGTM. It would be good to have a test.

PR #7760, opened 2021-04-28
release/slash: opt(GraphQL): filter existence queries on GraphQL side instead of using @filter(type) (#7757)

NamanJain8:
For the existence queries we used to do:
 query {\n  Cuisine_1(func: eq(Cuisine.name, \"yaMqnmHeov\")) @filter(type(Cuisine)) {\n uid \n  } \n}

The filter operation can be very heavy on dgraph.type as it may store millions of nodes. A better query could be like this:
 query {\n  Cuisine_1(func: eq(Cuisine.name, \"yaMqnmHeov\")) {\n uid dgraph.type \n  } \n}

This PR removes the above @filter operation from the existence query and uses the suggested alternate query which is lightweight.
The existence check operation is path critical and this fix is expected to add some performance improvement along with preventing OOM.

(cherry picked from commit 690fbbe05386a1234cb4b67807767651aa07ecac)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7760)


vmrajas (approved):
LGTM

PR #7776, opened 2021-05-04
Fix(GraphQL): Fix duplicate XID error in case of interface XIDs

vmrajas:
Implementation:
This PR adds the implementing type name to the key. This ensures that different variable names are generated for different implementing types of the same interface with same XID (name and value). It also adds an error condition to give error in case two nodes of different types with same XID(with interface=true) flag are added.

It also fixes other bug related to referencing same node with multiple IDs in same mutation.

Testing:
Added yaml and e2e tests.
Note that two tests from yaml tests have been removed as they were wrong.

Fixes GRAPHQL-1164

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7776)

### DIFF
--- a/graphql/resolve/add_mutation_test.yaml
+++ b/graphql/resolve/add_mutation_test.yaml
@@ -661,8 +661,10 @@
     }
   explanation: "A reference must be a valid UID"
   error:
-    { "message":
-        "failed to rewrite mutation payload because ID argument (HI!) was not able to be parsed" }
+    {
+      "message":
+        "failed to rewrite mutation payload because ID argument (HI!) was not able to be parsed"
+    }
 
 -
   name: "Add mutation with inverse reference"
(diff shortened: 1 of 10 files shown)


graphql/resolve/add_mutation_test.yaml
```diff
@@ -2741,43 +2743,6 @@
     message: |-
       failed to rewrite mutation payload because duplicate XID found: S1
 
-- name: "Duplicate XIDs in single mutation for Interface"
```
vmrajas, line 2744: Removed this test as it was wrong. As xid does not have interface=true flag. It is perfectly fine to have Student and Teacher with same xid.

graphql/resolve/update_mutation_test.yaml
```diff
@@ -1457,48 +1457,6 @@
     message: |-
       failed to rewrite mutation payload because duplicate XID found: T1
 
-- name: "Duplicate XIDs in single mutation for Interface"
```
vmrajas, line 1460: Removed this test as it was wrong. As xid does not have interface=true flag. It is perfectly fine to have Student and Teacher with same xid.

abhimanyusinghgaur (approved):

PR #7786, opened 2021-05-05
fix(backup-map): Free the UidPack after use

ahsanbarkati:
We were not freeing the UidPack which was causing a memory leak. This PR fixes this issue.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7786)

### DIFF
--- a/worker/restore_map.go
+++ b/worker/restore_map.go
@@ -31,6 +31,7 @@ import (
 
 	bpb "github.com/dgraph-io/badger/v3/pb"
 	"github.com/dgraph-io/badger/v3/y"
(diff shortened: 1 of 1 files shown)


jarifibrahim (approved):

NamanJain8:
Nice catch. @ahsanbarkati can you just confirm that we are not making such a mistake at some other place too?

ahsanbarkati:
> Nice catch. @ahsanbarkati can you just confirm that we are not making such a mistake at some other place too?

Yep checked it. We do not do such mistake anywhere else for uidpacks

PR #7788, opened 2021-05-06
perf(dgraph): Avoid redundant sorting

ajeetdsouza:
Repeatedly sending mutations under a single transaction in Dgraph are slow. The reason for this is:

- We sort Preds upon every mutation to remove duplicates.
- We sort Preds again upon committing.
- This is slow and redundant, since the Zero only needs to iterate over Preds once, and the duplicates won't matter.

Downsides of removing these sorts are:
- Any client using Dgraph will now get duplicates in their `response.Txn.Preds` field.
- The `/mutate` endpoint returns duplicate Preds too. I added logic here to do deduplication in the `mutationHandler`.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7788)

### DIFF
--- a/dgraph/cmd/alpha/http.go
+++ b/dgraph/cmd/alpha/http.go
@@ -426,7 +426,7 @@ func mutationHandler(w http.ResponseWriter, r *http.Request) {
 		Latency: resp.Latency,
 	}
 	sort.Strings(e.Txn.Keys)
-	sort.Strings(e.Txn.Preds)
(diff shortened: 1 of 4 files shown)


ajeetdsouza:
The pros don't seem to outweigh the cons here - we do optimize this particular case, but users are expected to batch mutations rather than send large numbers of mutations under a single transaction. On the flip side, we might end up sending a lot of duplicate preds to the user in the general case - we should not do this.

PR #7810, opened 2021-05-12
BREAKING: fix json marshal unmarshal for namespace > 127

NamanJain8:
We used to store predicate as `<namespace>|<attribute>` (pipe `|` signifies concatenation). We store this as a string. `<namespace>` is 8 bytes uint64, which when marshaled to JSON bytes mess up the predicate. This is because for the namespace greater than 127, the UTF-8 encoding might take up several bytes (also if the mapping does not exist, then it replaces it with some other rune). This affects three identified places in Dgraph:
- Live loader
- Backup and List Backup
- Http clients and Ratel

**Fix:**
New Format: `<anmespace>-<attribute>`  (`-` is the hyphen literal)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7810)

### DIFF
--- a/dgraph/cmd/alpha/http_test.go
+++ b/dgraph/cmd/alpha/http_test.go
@@ -428,7 +428,7 @@ func TestTransactionBasic(t *testing.T) {
 	require.Equal(t, 2, len(mr.preds))
 	var parsedPreds []string
 	for _, pred := range mr.preds {
-		p := strings.Split(pred, "-")[1]
+		p := strings.SplitN(pred, "-", 2)[1]
 		parsedPreds = append(parsedPreds, x.ParseAttr(p))
 	}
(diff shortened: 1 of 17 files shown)


manishrjain (approved):
:lgtm: Looks good. One thing to consider is if we can use hex for representing namespaces to be consistent, instead of decimal. Don't need `0x` in the front. Just the hex should be sufficient.

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7810#-:-M_X77PL8DLFVVRII0Sw:b-3mcifz)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pawanrawal and @vvbalaji-dgraph)

PR #7817, opened 2021-05-13
BREAKING: cherry: fix json marshal unmarshal for namespace > 127 (#7810)

NamanJain8:

Live loader
Backup and List Backup
Http clients and Ratel
Fix:
New Format: <anmespace>-<attribute> (- is the hyphen literal)

(cherry picked from commit 90d77f337a8509d21e94288fa2ea0e0c8542404e)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7817)


NamanJain8:
This is no longer needed. Issue fixed by: https://github.com/dgraph-io/dgraph/pull/7848

PR #7828, opened 2021-05-18
fix(updatemanifest): update the version of manifest after upgrade

NamanJain8:
We were not updating the manifest version after the updation. This PR fixes that.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7828)

### DIFF
--- a/ee/updatemanifest/run.go
+++ b/ee/updatemanifest/run.go
@@ -47,7 +47,7 @@ var opt struct {
 func init() {
 	UpdateManifest.Cmd = &cobra.Command{
 		Use:   "update_manifest",
-		Short: "Run the Dgraph update tool to update the manifest from v21.03 to latest.",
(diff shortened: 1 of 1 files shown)


ee/updatemanifest/run.go
```diff
@@ -101,6 +102,7 @@ func run() error {
 			for _, pred := range preds {
 				ns, attr, err := parseNsAttr(pred)
 				if err != nil {
+					parsedPreds = append(parsedPreds, pred)
```
ahsanbarkati, line 105: Why is this required?

NamanJain8: As we are overwriting the manifest, I think it should be a good idea to only fix what the tool can fix. It is better to leave the erroneous predicate in the manifest so that the user can fix/remove it, rather than the tool removing it.

ahsanbarkati (approved):

PR #7838, opened 2021-05-19
fix(restore): update the schema and type from 2103

NamanJain8:
With https://github.com/dgraph-io/dgraph/pull/7810 change, we changed the format of the predicate. We missed updating the schema and predicate. This PR fixes it.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7838)

### DIFF
--- a/worker/restore_map.go
+++ b/worker/restore_map.go
@@ -392,6 +392,9 @@ func (m *mapper) processReqCh(ctx context.Context) error {
 				return err
 			}
 			changeFormat := func() error {
+				// In the backup taken on 2103, we have the schemaUpdate.Predicate in format
+				// <namespace 8 bytes>|<attribute>. That had issues with JSON marshalling.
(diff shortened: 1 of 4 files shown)


worker/restore_map.go
```diff
@@ -391,11 +391,52 @@ func (m *mapper) processReqCh(ctx context.Context) error {
...
+						if sch.Predicate, err = x.AttrFrom2103(sch.Predicate); err != nil {
+							return err
+						}
+					}
+					kv.Value, err = update.Marshal()
+					return err
+				}
+				return nil
+			}
+			switch in.version {
```
ahsanbarkati, line 425: Please write a comment here about what was the problem and 20.11 and 21.03 respectively. Mention the reason why we are doing this update.

ahsanbarkati (approved):

PR #7841, opened 2021-05-21
fix(Snapshot): Improve snapshot handling

manishrjain:
Monitor for stuck snapshots in both directions, sender and receiver. If any of them is stuck for a while, cancel snapshot transfer.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7841)

### DIFF
--- a/worker/draft.go
+++ b/worker/draft.go
@@ -912,13 +912,19 @@ func (n *node) retrieveSnapshot(snap pb.Snapshot) error {
 	var pool *conn.Pool
 	addr := snap.Context.GetAddr()
 	glog.V(2).Infof("Snapshot.RaftContext.Addr: %q", addr)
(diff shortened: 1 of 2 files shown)


worker/draft.go
```diff
@@ -912,13 +912,19 @@ func (n *node) retrieveSnapshot(snap pb.Snapshot) error {
 	var pool *conn.Pool
 	addr := snap.Context.GetAddr()
 	glog.V(2).Infof("Snapshot.RaftContext.Addr: %q", addr)
-	if len(addr) > 0 {
+	if addr == n.MyAddr {
```
mangalaman93, line 915: won't we always steam from leader to follower? And what happens when we skip it?

mangalaman93 (changes requested):

PR #7852, opened 2021-05-25
feat(schema): do schema versioning and make backup non-blocking for indexing

NamanJain8:
**Issue:**
We used to write schema/types at `timestamp=1`. That single bit violates the Snapshot Isolation. Hence, we couldn't run schema updates and other operations like backup concurrently. This is an issue in a multi-tenant cluster because if the backup is running we couldn't update the schema. 
**Fix:**
This PR writes the schema/types at the `startTs` of the transaction and enables the schema updates to be made even if the backup is running.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7852)

### DIFF
--- a/posting/index.go
+++ b/posting/index.go
@@ -561,7 +561,6 @@ func (r *rebuilder) Run(ctx context.Context) error {
 		WithNumVersionsToKeep(math.MaxInt32).
 		WithLogger(&x.ToGlog{}).
 		WithCompression(options.None).
-		WithEncryptionKey(x.WorkerConfig.EncryptionKey).
 		WithLoggingLevel(badger.WARNING).
 		WithMetricsEnabled(false)
 
@@ -1233,24 +1232,24 @@ func DeleteData(ns uint64) error {
 
 // DeletePredicate deletes all entries and indices for a given predicate. The delete may be logical
 // based on DB options set.
-func DeletePredicate(ctx context.Context, attr string) error {
+func DeletePredicate(ctx context.Context, attr string, ts uint64) error {
 	glog.Infof("Dropping predicate: [%s]", attr)
 	prefix := x.PredicatePrefix(attr)
 	if err := pstore.DropPrefix(prefix); err != nil {
 		return err
 	}
-	return schema.State().Delete(attr)
+	return schema.State().Delete(attr, ts)
 }
 
 // DeletePredicateBlocking deletes all entries and indices for a given predicate. It also blocks the
 // writes.
-func DeletePredicateBlocking(ctx context.Context, attr string) error {
+func DeletePredicateBlocking(ctx context.Context, attr string, ts uint64) error {
 	glog.Infof("Dropping predicate: [%s]", attr)
 	prefix := x.PredicatePrefix(attr)
 	if err := pstore.DropPrefixBlocking(prefix); err != nil {
 		return err
 	}
-	return schema.State().Delete(attr)
+	return schema.State().Delete(attr, ts)
 }
 
 // DeleteNamespace bans the namespace and deletes its predicates/types from the schema.
--- a/systest/backup/filesystem/backup_test.go
+++ b/systest/backup/filesystem/backup_test.go
@@ -98,7 +98,6 @@ func TestBackupOfOldRestore(t *testing.T) {
 	require.NoError(t, err)
 
 	testutil.DropAll(t, dg)
-	time.Sleep(2 * time.Second)
 
 	_ = runBackup(t, 3, 1)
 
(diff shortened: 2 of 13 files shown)


posting/index.go
```diff
@@ -561,7 +561,6 @@ func (r *rebuilder) Run(ctx context.Context) error {
 		WithNumVersionsToKeep(math.MaxInt32).
 		WithLogger(&x.ToGlog{}).
 		WithCompression(options.None).
-		WithEncryptionKey(x.WorkerConfig.EncryptionKey).
```
jarifibrahim, line 564: Why do we need to remove this?

systest/backup/filesystem/backup_test.go
```diff
@@ -113,7 +112,6 @@ func TestBackupOfOldRestore(t *testing.T) {
 
 	// Clean the cluster and try restoring the backups created above.
 	testutil.DropAll(t, dg)
-	time.Sleep(2 * time.Second)
```
jarifibrahim, line 116: Don't we need these anymore? The DropAll could take a few minutes to be effective.

worker/draft.go
```diff
@@ -606,7 +623,7 @@ func (n *node) applyMutations(ctx context.Context, proposal *pb.Proposal) (rerr
 		}
 
 		for _, tupdate := range proposal.Mutations.Types {
-			if err := runTypeMutation(ctx, tupdate); err != nil {
+			if err := runTypeMutation(ctx, tupdate, startTs); err != nil {
```
jarifibrahim, line 626: The startTs is actually the `MaxAssigned`. See line 605. You might end up writing data at the wrong timestamp.

jarifibrahim (commented):
Got some comments.

posting/index.go
NamanJain8: We are already setting the encryption key a few lines below. This was redundant.

systest/backup/filesystem/backup_test.go
NamanJain8: I think this should not be needed. Why would DropAll take few minutes to be effective?

manishrjain (approved):
:lgtm: Do test for predicate moves and snapshots, etc.

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7852#-:-MaYz1QlCiH5b5WjnV3P:bj9eq8e)* status: all files reviewed, 2 unresolved discussions (waiting on @NamanJain8 and @vvbalaji-dgraph)

___
*[worker/draft.go, line 230 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7852#-MaYy8fg4CCDuHjhyPwp:-MaYy8fg4CCDuHjhyPwq:b-bx0fi3) ([raw file](https://github.com/dgraph-io/dgraph/blob/f4122592e432a4922dc40f63d8d2d7ff1b0d7462/worker/draft.go#L230)):*
> ```Go
> 	case opIndexing:
> 		for otherId, otherCloser := range n.ops {
> 			if otherId == opBackup {
> ```

Ideally, we should look at the backup timestamp. If indexing has a higher ts, then it's OK. If not, then not.

systest/backup/filesystem/backup_test.go
NamanJain8: I will remove these sleep for now. If they are needed, or tests become flaky, then we can bring them back.

NamanJain8 (commented):
> Dismissed @jarifibrahim and @manishrjain from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7852)* status: 10 of 13 files reviewed, all discussions resolved (waiting on @manishrjain and @vvbalaji-dgraph)

___
*[worker/draft.go, line 230 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7852#-MaYy8fg4CCDuHjhyPwp:-MabWHKl0z0mVaocFNMr:b-896fix) ([raw file](https://github.com/dgraph-io/dgraph/blob/f4122592e432a4922dc40f63d8d2d7ff1b0d7462/worker/draft.go#L230)):*

Ideally, we should look at the backup timestamp. If indexing has a higher ts, then it's OK. If not, then not.
</blockquote></details>

Done.
___
*[worker/draft.go, line 626 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/7852#-MaXr_c-AvzY9Ly2FV1r-r1-626:-MabWKY37eIIhpRCAcVc:b-ybdfpz) ([raw file](https://github.com/dgraph-io/dgraph/blob/f4122592e432a4922dc40f63d8d2d7ff1b0d7462/worker/draft.go#L626)):*
<details><summary><i>Previously, jarifibrahim (Ibrahim Jarif) wrote…</i></summary><blockquote>

The startTs is actually the `MaxAssigned`. See line 605. You might end up writing data at the wrong timestamp.
</blockquote></details>

We should use the same timestamp as the one used for index rebuild (`MaxAssigned`). Max assigned is used here to index all the data that has been committed up until this point.

PR #7865, opened 2021-05-27
Compose file changes

kaushik691:
compose.go file changes:
Added 2 new flags:
1. `--alias` - alias for the containers
2. `--custom_alpha_flags` Custom alpha flags for specific alphas, following {"1:custom_flags", "2:custom_flags"}, 
eg: {"2: -p <bulk_path>"}
3. `--custom_groups` configure groups to alphas by {"2:2", "3:2"}
  
` go run compose.go -a=2 -z=1 --alias=test   --custom_alpha_flags={"2:-p <bulk_path>"}  --custom_groups="2:2"`

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7865)


compose/compose.go
```diff
@@ -158,10 +164,17 @@ func getOffset(idx int) int {
 	return idx
 }
 
-func initService(basename string, idx, grpcPort int) service {
+func initService(basename string, idx, grpcPort int, container_name string) service {
```
NamanJain8, line 167: In Go, camelcase is the idiomatic way for defining variables. Please don't use camelcase.

compose/compose.go
```diff
@@ -269,11 +282,11 @@ func getZero(idx int, raft string) service {
 	return svc
 }
 
-func getAlpha(idx int, raft string) service {
+func getAlpha(idx int, raft string, custom_flag string) service {
```
NamanJain8, line 285: Use camelcase.

compose/compose.go
```diff
@@ -292,10 +305,15 @@ func getAlpha(idx int, raft string) service {
 		maxZeros = opts.NumZeros
 	}
 
-	zeroHostAddr := fmt.Sprintf("zero%d:%d", 1, zeroBasePort+opts.PortOffset)
+	zero_name := "zero"
```
NamanJain8, line 308: Use camelcase

compose/compose.go
```diff
@@ -362,11 +380,21 @@ func getAlpha(idx int, raft string) service {
 			svc.Volumes = append(svc.Volumes, getVolume(vol))
 		}
 	}
+
+	// Configure Alpha
```
NamanJain8, line 384: Remove commented code.

compose/compose.go
```diff
@@ -647,11 +679,32 @@ func main() {
...
+	}
+	cust_index := 0
 	for i := 1; i <= opts.NumAlphas; i++ {
 		gid := int(math.Ceil(float64(i) / float64(opts.NumReplicas)))
 		rs := fmt.Sprintf("idx=%d; group=%d", i, gid)
-		svc := getAlpha(i, rs)
+
+		// Handling alpha customizations
+		custom_flags := ""
+		alpha_index := 0
```
NamanJain8, line 693: Use camelcase everywhere in variable declaration.

compose/compose.go
```diff
@@ -647,11 +679,32 @@ func main() {
...
+	custom_alphas := ""
+	if len(opts.CustomAlphaOptions) > 0 {
+		custom_alphas = opts.CustomAlphaOptions[0]
+	}
+	cust_index := 0
 	for i := 1; i <= opts.NumAlphas; i++ {
 		gid := int(math.Ceil(float64(i) / float64(opts.NumReplicas)))
 		rs := fmt.Sprintf("idx=%d; group=%d", i, gid)
-		svc := getAlpha(i, rs)
+
```
NamanJain8, line 690: Instead of parsing the custom flag in the for `loop`. Let's create a map upfront map[uint64]string -> Map from `alphaIndex` to `customFlag`. So we will not need any repetitive parsing in for loop and the code will look much simpler..

NamanJain8 (changes requested):
Got few comments. Also, please mention the use case of the flags.

compose/compose.go
kaushik691: okay.

compose/compose.go
kaushik691: okay

compose/compose.go
```diff
@@ -158,10 +165,17 @@ func getOffset(idx int) int {
 	return idx
 }
 
-func initService(basename string, idx, grpcPort int) service {
+func initService(basename string, idx, grpcPort int, aliasName string) service {
 	var svc service
 
-	svc.name = name(basename, idx)
+	cont_name := func() string {
```
NamanJain8, line 171: This is a simple if-else, you don't need a function for this. Also, don't use snake case.

compose/compose.go
```diff
@@ -118,6 +119,12 @@ type options struct {
 	MinioPort      uint16
 	MinioEnvFile   []string
 
+	// Alpha Configurations
+	CustomAlphaOptions []string
+
+	// Container Alias
+	ContainerAlias string
```
NamanJain8, line 126: Maybe rename it to `ContainerPrefix`.

compose/compose.go
```diff
@@ -226,7 +240,7 @@ func getZero(idx int, raft string) service {
 	basePort := zeroBasePort + opts.PortOffset
 	grpcPort := basePort + getOffset(idx)
 
-	svc := initService(basename, idx, grpcPort)
+	svc := initService(basename, idx, grpcPort, opts.ContainerAlias)
```
NamanJain8, line 243: Probably you don't need to pass in the opts.ContainerAlias. `opts` is global, so you can directly use `opts.ContainerAlias` in this function.

compose/compose.go
```diff
@@ -648,16 +678,66 @@ func main() {
 		services[svc.name] = svc
 	}
 
+	// Alpha Customization
+	customAlphas := make(map[int]string)
+	if len(opts.CustomAlphaOptions) > 0 {
```
NamanJain8, line 683: You don't need this length check. If the list is empty, the for loop will not be executed by itself.

compose/compose.go
```diff
@@ -648,16 +678,66 @@ func main() {
...
+			custIdx, err := strconv.Atoi(splits[0])
+			if err != nil {
+				fatal(errors.Errorf(" --custom_alpha_flags captured erros while parsing index value %v", err))
+			}
+			customAlphas[custIdx] = splits[1]
+		}
+	}
+
+	alphaGroups := make(map[int]int)
+	if len(opts.CustomGroups) > 0 {
```
NamanJain8, line 698: Same here.

compose/compose.go
```diff
@@ -648,16 +678,66 @@ func main() {
 		services[svc.name] = svc
 	}
 
+	// Alpha Customization
+	customAlphas := make(map[int]string)
```
NamanJain8, line 682: Maybe rename it to customOptions?

compose/compose.go
```diff
@@ -648,16 +678,66 @@ func main() {
...
+		}
+	}
+
+	gid := 0
+	idxNum := 0
 	for i := 1; i <= opts.NumAlphas; i++ {
-		gid := int(math.Ceil(float64(i) / float64(opts.NumReplicas)))
-		rs := fmt.Sprintf("idx=%d; group=%d", i, gid)
-		svc := getAlpha(i, rs)
+		if val, ok := alphaGroups[i]; ok {
```
NamanJain8, line 716: Why do you need this logic? Say if we want 2 groups each with 1 alpha. Then we can set the replicas=1, alphas=2, right?

NamanJain8 (changes requested):

kaushik691:
updated with this PR: https://github.com/dgraph-io/dgraph/pull/7872

PR #7867, opened 2021-05-27
feat(header): add content length to GRPC/HTTP response

NamanJain8:
---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7867)


jarifibrahim (commented):
@NamanJain8 Are we using `Dgraph-Content-Length` instead of `Content-Length` because of https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding ?

PR #7871, opened 2021-05-31
perf: Speed up parsing of a huge query with a lot of conditional mutations

Phill240:
Recently I faced a situation when my queries containing a lot of conditional mutations (+5000) took a long time to process. According to the servers reply the slowest part of the query handling was parsing. For example:
parsing_ns:835833349 processing_ns:157596840 encoding_ns:5105536 assign_timestamp_ns:454011 total_ns:1033841289 
So the parsing took several times longer that the processing itself!
The reason of the issue is in the ineffective implementation of buildUpsertQuery method. I suggest using String Builder instead of just simple string concatenation. After these changes the parsing speed increased significantly:
parsing_ns:134053025 processing_ns:114263143 encoding_ns:4992320 assign_timestamp_ns:495987 total_ns:286159717 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7871)

### DIFF
--- a/edgraph/server.go
+++ b/edgraph/server.go
@@ -829,7 +829,7 @@ func buildUpsertQuery(qc *queryContext) string {
 	qc.condVars = make([]string, len(qc.req.Mutations))
 
 	var b strings.Builder
-	b.WriteString(strings.TrimSuffix(qc.req.Query, "}"))
+	x.Check2(b.WriteString(strings.TrimSuffix(qc.req.Query, "}")))
 
 	for i, gmu := range qc.gmuList {
 		isCondUpsert := strings.TrimSpace(gmu.Cond) != ""
@@ -855,11 +855,11 @@ func buildUpsertQuery(qc *queryContext) string {
 			// The variable __dgraph_0__ will -
 			//      * be empty if the condition is true
 			//      * have 1 UID (the 0 UID) if the condition is false
-			b.WriteString(qc.condVars[i] + ` as var(func: uid(0)) ` + cond + `
-			 `)
+			x.Check2(b.WriteString(qc.condVars[i] + ` as var(func: uid(0)) ` + cond + `
+			 `))
 		}
 	}
-	b.WriteString(`}`)
+	x.Check2(b.WriteString(`}`))
 
 	return b.String()
 }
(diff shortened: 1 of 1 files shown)


NamanJain8:
Hey @Phill240 , this looks good. Thanks for it. Do we have any numbers for this change made?

edgraph/server.go
```diff
@@ -827,7 +827,11 @@ func buildUpsertQuery(qc *queryContext) string {
 	}
 
 	qc.condVars = make([]string, len(qc.req.Mutations))
-	upsertQuery := strings.TrimSuffix(qc.req.Query, "}")
+
+        // We use String Builder for effective string concatenation
```
NamanJain8, line 831: In my opinion, this comment is not needed. StringBuilder should be used wherever the performance is a concern.
Shall we remove the comment please?

edgraph/server.go
```diff
@@ -827,7 +827,11 @@ func buildUpsertQuery(qc *queryContext) string {
...
+        var upsertQueryBuilder strings.Builder
```
NamanJain8, line 832: Let's keep the variable name short. Maybe just rename it to `builder` or just `b`. In Go, short local variables are idiomatic. Also can you please check if the indentation of text is fine?

NamanJain8 (commented):
Thanks for the PR. Nice change. 🎉 
Please address the comments and then we can proceed with merging this,

edgraph/server.go
Phill240: Done!

edgraph/server.go
```diff
@@ -827,7 +827,10 @@ func buildUpsertQuery(qc *queryContext) string {
 	}
 
 	qc.condVars = make([]string, len(qc.req.Mutations))
-	upsertQuery := strings.TrimSuffix(qc.req.Query, "}")
+
+	var b strings.Builder
+	b.WriteString(strings.TrimSuffix(qc.req.Query, "}"))
```
NamanJain8, line 832: Thanks for addressing the comments. One more small comment.
`WriteString` returns an error. Please add x.Check2(..) that validates the error as well. We can merge this after that.

NamanJain8 (approved):
Hey @Phill240 , thanks for addressing the comment.

edgraph/server.go
NamanJain8: @Phill240 please address this too. ^

Phill240: Done. Now all the WriteString methods are wrapped in Check2.

NamanJain8:
Thanks, @Phill240 ! 🎉

PR #7880, opened 2021-06-03
fix(restore): append galaxy namespace to type name

NamanJain8:
When restoring from backup taken on version `<v21.03`, we should append the galaxy namespace to the typename. 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7880)


ahsanbarkati (approved):
LGTM

PR #7884, opened 2021-06-04
fix(GraphQL): fix JWT token error when Authorization is not enabled.

minhaj-shakeel:
Fixes DGRAPH-3392.
This has been manually tested. This issue does not exist in `21.03` and is only present in previous versions.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7884)


abhimanyusinghgaur (approved):

vmrajas (approved):
LGTM
Can you add more details to the PR like about its testing (if it has been tested manually) and that this problem is not present in 21.03 and later (if that is the case). It will help in getting context later.

PR #7887, opened 2021-06-04
fix(multi-tenancy): Check namespace existence before banning namespace

ahsanbarkati:
Before banning namespace, we should check if it exists or not.


NamanJain8 (dismissed):
LGTM. Please add a unit test for this.

all-seeing-code (approved):

harshil-goel (approved):

PR #7890, opened 2021-06-07
fix(export): exported output should be sane even if errored out

NamanJain8:
---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7890)


NamanJain8:
Need not be merged, most probably.

PR #7926, opened 2021-06-25
fix(ACL): The Acl cache should be updated on restart and restore

ahsanbarkati:
---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7926)


manishrjain (approved):
:lgtm: Ensure that the licenses are correctly applied.

> Reviewed 1 of 3 files at r1, 6 of 7 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7926#-:-MdwNMkZ6FjVWkylINYO:b-ve11oh)* status: all files reviewed, 2 unresolved discussions (waiting on @ahsanbarkati)

___
*[worker/acl_cache.go, line 1 at r3](https://reviewable.io/reviews/dgraph-io/dgraph/7926#-MdwMtgwEBFv95AoIhns:-MdwMtgwEBFv95AoIhnt:b-360akv) ([raw file](https://github.com/dgraph-io/dgraph/blob/517b64101002b0f6b458bbb93a01b4029f25d6e6/worker/acl_cache.go#L1)):*
> ```Go
> package worker
> ```

Keep the file under the same license. So, this file should stay as enterprise.

Also, if we were to compile with oss flag, would this then build?

edgraph/access_ee.go
```diff
@@ -327,38 +363,20 @@ func RefreshAcls(closer *z.Closer) {
...
-		queryResp, err := (&Server{}).doQuery(ctx, req)
-		if err != nil {
-			return errors.Errorf("unable to retrieve acls: %v", err)
-		}
-		groups, err := acl.UnmarshalGroups(queryResp.GetJson(), "allAcls")
-		if err != nil {
-			return err
+
+		if !worker.AclCachePtr.Loaded() {
+			updaters := z.NewCloser(1)
```
ahsanbarkati, line 375: use `closer.Ctx()`

edgraph/access_ee.go
```diff
@@ -406,8 +424,8 @@ var aclPrefixes = [][]byte{
 	x.PredicatePrefix(x.GalaxyAttr("dgraph.xid")),
 }
 
-// clears the aclCachePtr and upserts the Groot account.
-func ResetAcl(closer *z.Closer) {
+// upserts the Groot account.
+func InitializeAcl(closer *z.Closer) {
 	defer func() {
 		glog.Infof("ResetAcl closed")
```
ahsanbarkati, line 430: update the log message

NamanJain8 (commented):
Got a few comments. Rest looks good to me. Please address them before merging. 

> Reviewed 4 of 11 files at r4.
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7926#-:-MePkX7z7psMCNATbThm:bwy2d8m)* status: 7 of 14 files reviewed, 8 unresolved discussions (waiting on @ahsanbarkati and @manishrjain)

___
*[edgraph/access_ee.go, line 372 at r4](https://reviewable.io/reviews/dgraph-io/dgraph/7926#-MePho6184enOwAL9Hcz:-MePho6184enOwAL9Hd-:bbp540f) ([raw file](https://github.com/dgraph-io/dgraph/blob/d01a8fde21102b33544e3a64d36723bf31de0f66/edgraph/access_ee.go#L372)):*
> ```Go
> 		}
> 		maxRefreshTs = refreshTs
> 		ctx := x.AttachNamespace(closer.Ctx(), ns)
> ```

Seems like you probably don't need to attach namespace over here as we are passing the namespace in `refreshAclCache`.
___
*[edgraph/access_ee.go, line 376 at r4](https://reviewable.io/reviews/dgraph-io/dgraph/7926#-MePi6567-0pNDDp8-9V:-MePi6567-0pNDDp8-9W:b-x0xo4t) ([raw file](https://github.com/dgraph-io/dgraph/blob/d01a8fde21102b33544e3a64d36723bf31de0f66/edgraph/access_ee.go#L376)):*
> ```Go
> 		if !worker.AclCachePtr.Loaded() {
> 			updaters := z.NewCloser(1)
> 			RefreshACLs(updaters.Ctx())
> ```

Why do we need to call this? This will load the ACLs for all the namespaces :thinking: 
___
*[worker/acl_cache.go, line 43 at r4](https://reviewable.io/reviews/dgraph-io/dgraph/7926#-MePiwER86zP6FtZmuB-:-MePiwES3Tu3Cq1AQlt0:b-hpqdtw) ([raw file](https://github.com/dgraph-io/dgraph/blob/d01a8fde21102b33544e3a64d36723bf31de0f66/worker/acl_cache.go#L43)):*
> ```Go
> }
> 
> func (cache *AclCache) Set() {
> ```

Don't we need lock over here and in the above functions?
___
*[worker/acl_cache.go, line 54 at r4](https://reviewable.io/reviews/dgraph-io/dgraph/7926#-MePjjBn3rvzzLv93rE3:-MePjjBn3rvzzLv93rE4:bxw850n) ([raw file](https://github.com/dgraph-io/dgraph/blob/d01a8fde21102b33544e3a64d36723bf31de0f66/worker/acl_cache.go#L54)):*
> ```Go
> 
> func (cache *AclCache) GetUserPredPerms(userId string) map[string]int32 {
> 	return cache.userPredPerms[userId]
> ```

I think it would be good to move the lock/unlock from the caller to `GetUserPredPerms`.

edgraph/access_ee.go
minhaj-shakeel: Done.

edgraph/access_ee.go
minhaj-shakeel: Done.

PR #7930, opened 2021-06-25
fix(restore): Set kv version to restoreTs for rolled up keys and schema keys

ahsanbarkati:
The kv version should be set to restore timestamp for rolled-up keys and schema keys as well.

NOTE: I need to double-check if writing schema to restoreTs is right thing to do.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7930)


NamanJain8 (approved):
LGTM

PR #7944, opened 2021-07-09
perf: Multiple `append` combined into a single call

iamrajiv:
## Description

The `append` built-in function appends elements to the end of a slice. If it has sufficient capacity, the destination is resliced to accommodate the new elements, but if capacity is not enough, then `append` will allocate a new underlying array and return the updated slice. Therefore, it is necessary to store the result of append, often in the variable holding the slice itself. Appending calls in a single call of `append` allocates memory just once to accommodate all the elements to be appended. Whereas multiple calls to `append` introduce many overheads, most notably being the possibility of more calls for memory allocation because the total number of elements to be appended over multiple calls of `append` is unknown beforehand, resulting in inaccurate preallocation.

## 

This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7944)

### DIFF
--- a/dgraph/cmd/alpha/http_test.go
+++ b/dgraph/cmd/alpha/http_test.go
@@ -187,8 +187,7 @@ func queryWithTsForResp(inp queryInp) (string, *tsInfo, *http.Response, error) {
 		params = append(params, "debug="+inp.debug)
 	}
 	if inp.ts != 0 {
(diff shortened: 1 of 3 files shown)


NamanJain8 (commented):
I think that multiple append calls are more readable here. Moreover, they are in tests so does not impact dgraph's performance as such.

I think that the weight of readability is more here.

PR #7957, opened 2021-07-22
fix(rollups): Write rollup keys at ts+1

ahsanbarkati:
We should write rolled up keys at (maxTs of the deltas + 1) because if we write the rolled up keys at the same ts as
that of the delta, then in case of WAL replay the rolled-up key would get over-written by the
delta which can bring DB to an invalid state.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7957)

### DIFF
--- a/posting/list.go
+++ b/posting/list.go
@@ -827,7 +827,7 @@ func (l *List) Rollup(alloc *z.Allocator) ([]*bpb.KV, error) {
 
 	var kvs []*bpb.KV
 	kv := MarshalPostingList(out.plist, alloc)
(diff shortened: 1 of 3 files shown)


NamanJain8 (approved):
:lgtm: Nice work @ahsanbarkati !!

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7957#-:-MfIU3H3CoSwn0WAjMBl:bf3gg6u)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @manishrjain)

PR #7963, opened 2021-07-29
fix(pool): use write lock when getting health info

NamanJain8:
Fixes: https://discuss.dgraph.io/t/bug-mutex-lock-while-health-checking/14940

There was a race condition when accessing HealthInfo. This PR fixes that.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7963)

### DIFF
--- a/conn/pool.go
+++ b/conn/pool.go
@@ -344,10 +344,11 @@ func (p *Pool) IsHealthy() bool {
 
 // HealthInfo returns the healthinfo.
 func (p *Pool) HealthInfo() pb.HealthInfo {
-	p.RLock()
-	defer p.RUnlock()
+	ok := p.IsHealthy()
+	p.Lock()
+	defer p.Unlock()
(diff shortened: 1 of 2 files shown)


conn/pool.go
```diff
@@ -344,10 +344,11 @@ func (p *Pool) IsHealthy() bool {
 
 // HealthInfo returns the healthinfo.
 func (p *Pool) HealthInfo() pb.HealthInfo {
-	p.RLock()
-	defer p.RUnlock()
+	ok := p.IsHealthy()
```
ahsanbarkati, line 347: reading before acquiring lock?

ahsanbarkati (approved):

conn/pool.go
NamanJain8: p.IsHealthy() takes the read lock internally.

PR #7965, opened 2021-07-29
cloud: fix(snapshot): update last snapshot time across members

NamanJain8:
Addresses DO-79.
We see frequent snapshots in an HA cluster because of the leadership re-election because of leader crash/shutdown. The local variable that we use for snapshot frequency is not synchronized across alphas. So each alpha thinks differently about the time the last snapshot was taken.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7965)

### DIFF
--- a/worker/draft.go
+++ b/worker/draft.go
@@ -636,6 +636,7 @@ func (n *node) applyCommitted(proposal *pb.Proposal, key uint64) error {
 			}
 			glog.Warningf("Error while calling CreateSnapshot: %v. Retrying...", err)
 		}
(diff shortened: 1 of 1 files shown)


manishrjain (commented):
Do this for master first. Then we can choose to cherrypick in release. :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7965#-:-MfmO4qw5EcUea86wDhR:bxiwkk8)* status: 0 of 1 files reviewed, all discussions resolved

PR #7978, opened 2021-08-10
fix(acl): filter out the results based on type

NamanJain8:
We store the groupId and userId in a predicate named `dgraph.xid`.There was a subtle bug where if a we create the group with same name as that of any user, then the user is not able to log in. This happens because we were not applying a filter by type. 

This PR fixes that.


ahsanbarkati (approved):
Nice! LGTM

PR #8002, opened 2021-08-26
Use rank() API from sroar and some cleanup

ahsanbarkati:
IndexOf was converting the bitmap to array and then used binary search to find the rank.
The conversion to array is extremely inefficient. This PR uses newly implemented
Rank API from sroar which finds the rank in an efficient way. This PR also does some
cleanup.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/8002)

### DIFF
--- a/algo/uidlist.go
+++ b/algo/uidlist.go
@@ -17,8 +17,6 @@
 package algo
 
 import (
-	"sort"
-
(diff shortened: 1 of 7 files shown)


NamanJain8 (approved):
Looks good to me.

PR #8007, opened 2021-08-31
fix: add validation of null values with correct order of graphql rule validation

aman-bansal:
This PR fixes the bug mentioned in https://discuss.dgraph.io/t/no-input-field-null-validation-on-non-nullable-fields/14085.
 Issue is suppose a schema is defined type Task { name: String! }.
Now dgraph generates the mutation schema as follows:
```
type mutation {
 addTask(input: [AddTaskInput]!): Result!
}
```

Now while executing mutation, graphql accept both arrays and single object as input. This means:
```
M1: 

mutation {
 addTask(input: {name: string})
}

m2:
mutation {
 addTask(input: [{name: string}])
}

```
both the above mutation are valid. But dgraph after the validation assumes that it will always be an list. This leads to some panics in the past and hence to make it compliant we added listInputCoercion which will change the value to List if needed so that there wont be any issue afterwards. But this leads to other bugs where few of the validations were getting skipped. For example this mutation will fail
```
mutation MyMutation {
  addTask(input: {}) {
    numUids
    test {
      name
    }
  }
}
```
because listInputCoercion changes the {} to [{}] and this will be skipped by the data type validation because of arrays. Hence
an empty data object will be inserted leading to a corrupted database.

This PR adds the order to the execution of rule validations and listInputCoercion is placed in the end so that when the rules are getting executed, all validations are executed first and value change is executed last so that it wont mess up the queries again 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/8007)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/dgraph-io/badger/v3 v3.0.0-20210707084205-c40b2e9af902
 	github.com/dgraph-io/dgo/v210 v210.0.0-20210407152819-261d1c2a6987
 	github.com/dgraph-io/gqlgen v0.13.2
-	github.com/dgraph-io/gqlparser/v2 v2.2.0
+	github.com/dgraph-io/gqlparser/v2 v2.2.1
 	github.com/dgraph-io/graphql-transport-ws v0.0.0-20210511143556-2cef522f1f15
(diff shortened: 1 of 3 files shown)


NamanJain8 (approved):
LGTM. It would be good to add an example in the PR description.

PR #8013, opened 2021-09-03
fix(lambda): fix race condition in lambda server spin up

NamanJain8:
Fixes the race condition while spinning up the lambda server.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/8013)

### DIFF
--- a/dgraph/cmd/alpha/run.go
+++ b/dgraph/cmd/alpha/run.go
@@ -34,6 +34,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


dgraph/cmd/alpha/run.go
```diff
@@ -544,23 +549,32 @@ func setupLambdaServer(closer *z.Closer) {
 			case <-closer.HasBeenClosed():
 				return
 			case <-ticker.C:
-				timestamp := time.Now().UnixNano()
-				for _, l := range lambdas {
+				healthCheck := func(l *lambda) {
```
ahsanbarkati, line 552: We should move this function definition outside the for...select

ahsanbarkati (approved):

dgraph/cmd/alpha/run.go
NamanJain8: Done

PR #8021, opened 2021-09-09
fix(admin): prevent unresponsive /admin if bad GraphQL schema is set

NamanJain8:
In case a bad GraphQL schema is inserted in a namespace, the `/admin` endpoint becomes unavailable. The user is not able to even reset the schema.
Moreover, the admin schema for all the namespaces is common. So, it should be okay to use GalaxyNamespace's admin schema.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/8021)

### DIFF
--- a/dgraph/cmd/alpha/run.go
+++ b/dgraph/cmd/alpha/run.go
@@ -684,10 +684,13 @@ func setupServer(closer *z.Closer) {
 	baseMux.Handle("/probe/graphql", graphqlProbeHandler(gqlHealthStore, globalEpoch))
 
 	baseMux.HandleFunc("/admin", func(w http.ResponseWriter, r *http.Request) {
-		r.Header.Set("resolver", "0")
 		// We don't need to load the schema for all the admin operations.
 		// Only a few like getUser, queryGroup require this. So, this can be optimized.
-		if err := admin.LazyLoadSchema(x.ExtractNamespaceHTTP(r)); err != nil {
+		// All the namespaces share the same admin schema. So, it is okay to use Galaxy namespace
+		// over here. This also helps prevent cases where a bad GraphQL schema makes /admin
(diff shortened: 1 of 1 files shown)


dgraph/cmd/alpha/run.go
```diff
@@ -684,10 +684,13 @@ func setupServer(closer *z.Closer) {
 	baseMux.Handle("/probe/graphql", graphqlProbeHandler(gqlHealthStore, globalEpoch))
 
 	baseMux.HandleFunc("/admin", func(w http.ResponseWriter, r *http.Request) {
-		r.Header.Set("resolver", "0")
 		// We don't need to load the schema for all the admin operations.
```
NamanJain8, line 687: @aman-bansal  I am not very sure about this comment. I couldn't understand the purpose of lazy loading the schema here.

NamanJain8: Hmm.. possible. Could it be related to auth rules? Because I believe, to resolve getUser/getGroup/getRule, etc queries, GraphQL schema should not be needed.

aman-bansal: get group will need permissions. this permission set might be filtered using the current schema. i am not sure. just speculating

aman-bansal: there is /admin/schema api. do you think we should expose that api to override the schema if needed?

NamanJain8: Users can access `/admin/schema` even now. That can be used right now as well to fix the schema.

NamanJain8: Then this change is not needed.

NamanJain8:
This is not needed.

PR #8024, opened 2021-09-14
fix(snapshot): Fix snapshot calculation after restore

ahsanbarkati:
If there is no delta proposal with a valid commit then the snapshot was
skipped. For the purpose of snapshot calculation, we should also
consider the restore timestamp. 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/8024)

### DIFF
--- a/worker/draft.go
+++ b/worker/draft.go
@@ -2039,13 +2039,19 @@ func (n *node) calculateSnapshot(startIdx, lastIdx, minPendingStart uint64) (*pb
 				}
 			}
 
-			// If there is a restore proposal then consider the restoreTs as commitTs for the
-			// purpose of snapshot calculation.
+			// If we encounter a restore proposal, we can immediately truncate the WAL and create
+			// a snapshot. This is to avoid the restore happening again if the server restarts.
 			if proposal.Restore != nil {
 				restoreTs := proposal.Restore.GetRestoreTs()
-				maxCommitTs = x.Max(maxCommitTs, restoreTs)
 				glog.Infof("Found restore proposal with restoreTs: %d", restoreTs)
 				span.Annotatef(nil, "Found restore proposal with restoreTs: %d", restoreTs)
+
+				return &pb.Snapshot{
+					Context:     n.RaftContext,
+					Index:       entry.Index,
+					ReadTs:      restoreTs,
+					MaxAssigned: restoreTs,
+				}, nil
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


worker/draft.go
```diff
@@ -2038,6 +2038,15 @@ func (n *node) calculateSnapshot(startIdx, lastIdx, minPendingStart uint64) (*pb
 					maxCommitTs = x.Max(maxCommitTs, txn.CommitTs)
 				}
 			}
+
+			// If there is a restore proposal then consider the restoreTs as commitTs for the
```
NamanJain8, line 2042: This looks good to me. Also, it would be better to add a comment why the `restoreTs` can be taken as `commitTs` in this case. Because this is against the normal case scenario where commits come via delta proposals.

As per my understanding, this is because restoreTs provides the guarantee that all proposals before the restoreTs would be committed already.

manishrjain (changes requested):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/8024)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @ahsanbarkati)

___
*[worker/draft.go, line 2045 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/8024#-MjcmpSL4yGJ5uOBobtw:-MjcmpSL4yGJ5uOBobtx:b-1mwck6) ([raw file](https://github.com/dgraph-io/dgraph/blob/489874c3f9467dd0dffb29ec135066e6c6e33257/worker/draft.go#L2045)):*
> ```Go
> 			// purpose of snapshot calculation.
> 			if proposal.Restore != nil {
> 				restoreTs := proposal.Restore.GetRestoreTs()
> ```

If we did a restore, then we should just overwrite the timestamp to use the restore timestamp. But, does the restore TS get correctly picked up?

worker/draft.go
```diff
@@ -2038,6 +2038,21 @@ func (n *node) calculateSnapshot(startIdx, lastIdx, minPendingStart uint64) (*pb
 					maxCommitTs = x.Max(maxCommitTs, txn.CommitTs)
 				}
 			}
+
+			// If we encounter a restore proposal, we can immediately truncate the WAL and create
+			// a snapshot. This is to avoid the restore happening again if the server restarts.
+			if proposal.Restore != nil {
+				restoreTs := proposal.Restore.GetRestoreTs()
+				glog.Infof("Found restore proposal with restoreTs: %d", restoreTs)
```
NamanJain8, line 2046: Add a log like `calculated snapshot from the restore proposal` and maybe dump the snapshot as well in the logs?

NamanJain8 (approved):
Looks good to me.

PR #8032, opened 2021-09-16
feat(magicNumber): Introduce magic number

ahsanbarkati:
Magic number is a unique identifier for the version of Dgraph. In 21.09 we
have changed the data format of posting lists by bringing in sroar. Running
Dgraph with sroar change on older p directory can cause data corruption.
This magic number prevents DB to start up if the version is not compatible.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/8032)

### DIFF
--- a/worker/config.go
+++ b/worker/config.go
@@ -26,7 +26,7 @@ import (
 const (
 	// magicVersion is a unique uint16 number. Badger won't start if this magic number doesn't match
 	// with the one present in the manifest. It prevents starting up dgraph with new data format
-	// (using roaring bitmap) on older p directory.
+	// (eg. the change in 21.09 by using roaring bitmap) on older p directory.
 	magicVersion = 1
 
 	// AllowMutations is the mode allowing all mutations.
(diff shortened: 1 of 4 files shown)


worker/config.go
```diff
@@ -24,6 +24,11 @@ import (
 )
 
 const (
+	// magicVersion is a unique uint16 number. Badger won't start if this magic number doesn't match
+	// with the one presnet in the manifest. It prevents starting up dgraph with new data format
```
NamanJain8, line 28:
```suggestion
	// with the one present in the manifest. It prevents starting up dgraph with new data format
```

aman-bansal (approved):

worker/config.go
```diff
@@ -24,6 +24,11 @@ import (
...
+	// (using roaring bitmap) on older p directory.
```
NamanJain8, line 29: I think comments can be generalized. Roaring bitmap thing can act as an example

NamanJain8 (approved):
Looks good to me.

PR #8039, opened 2021-09-18
fix(json-encoding): remove duplicate uids from normalized result

NamanJain8:

### DIFF
--- a/query/outputnode.go
+++ b/query/outputnode.go
@@ -949,27 +949,23 @@ func (enc *encoder) normalize(fj fastJsonNode) ([]fastJsonNode, error) {
 	}
 
 	for i, slice := range parentSlice {
-		// sort the fastJson list
-		// This will ensure that nodes with same attribute name comes together in response
+		// Sort the fastJson list.
+		// This will ensure that nodes with same attribute name comes together in response.
 		enc.MergeSort(&parentSlice[i])
 		// From every list we need to remove node with attribute "uid".
-		var prev, cur fastJsonNode
-		cur = slice
+		var first, last fastJsonNode
+		cur := slice
 		for cur != nil {
 			if enc.getAttr(cur) == enc.uidAttr {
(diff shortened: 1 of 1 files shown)


NamanJain8:
This change reverts the behaviour of `normalize` to the one present in [`release/v20.07`](https://github.com/dgraph-io/dgraph/blob/b3e416e4c2e54c2282f0d5bf5337744f2096af52/query/outputnode.go#L703-L723). Basically, it removes all the UID nodes but keep the one. 

In the ideal scenario, the use of UID field without alias should be avoided.

NamanJain8 (commented):
> *[Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/8039)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[query/outputnode.go, line 967 at r1](https://reviewable.io/reviews/dgraph-io/dgraph/8039#-Mk8UiCRCewjcQyaKeSt:-Mk8UiCRCewjcQyaKeSu:b-d6s4xd) ([raw file](https://github.com/dgraph-io/dgraph/blob/2c9594ec230b8f64b90511ba1599a7d6fd38827a/query/outputnode.go#L967)):*
> ```Go
> 			cur = cur.next
> 		}
> 		if first != nil && first != last {
> ```

This PR reverts back the behavior to the one being used https://github.com/dgraph-io/dgraph/blob/b3e416e4c2e54c2282f0d5bf5337744f2096af52/query/outputnode.go#L703-L723. 
In my opinion, all the UID nodes should be removed and I don't understand why one UID was kept in `v20.07`. I think that the behavior of JSON encoding made the user use the bad behavior as a feature.

Moreover, as mentioned in the docs, normalize should be used with aliases.

NamanJain8:
This PR is not needed as mentioned in the description. The normalize operation should be used along with the aliases.

PR #8048, opened 2021-09-21
fix: fixing audit logs for websocket connections

aman-bansal:
Audits were made to operate as a middleware for all the requests over the 8080 port of Dgraph alpha. This works great for grpc and http requests but not for graphQL subscriptions over websockets. websocket connections works in different way. They start as normal Http connection and then gets upgraded to the websocket connection. Till the upgrade is over, there is no transfer of metadata and headers across the network. This makes auditing impossible at the middleware layer for websocket connections.

This PR separates out the audit for websockets connection to the http.go#subscriber which will handle the auditing in the right manner.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/8048)

### DIFF
--- a/ee/audit/interceptor_ee.go
+++ b/ee/audit/interceptor_ee.go
@@ -145,7 +145,7 @@ func AuditWebSockets(ctx context.Context, req *schema.Request) {
 		ClientHost:  ip,
 		Endpoint:    "/graphql",
 		ReqType:     WebSocket,
-		Req:         req.Query,
+		Req:         truncate(req.Query, maxReqLength),
 		Status:      http.StatusText(http.StatusOK),
 		QueryParams: nil,
 	})
--- a/ee/audit/interceptor_ee.go
+++ b/ee/audit/interceptor_ee.go
@@ -95,7 +95,7 @@ func AuditRequestHttp(next http.Handler) http.Handler {
 			return
 		}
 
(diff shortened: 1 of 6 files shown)


ee/audit/interceptor_ee.go
```diff
@@ -102,6 +117,40 @@ func AuditRequestHttp(next http.Handler) http.Handler {
...
+
+	auditor.Audit(&AuditEvent{
+		User:        user,
+		Namespace:   namespace,
+		ServerHost:  x.WorkerConfig.MyAddr,
+		ClientHost:  ip,
+		Endpoint:    "/graphql",
+		ReqType:     WebSocket,
+		Req:         req.Query,
+		Status:      http.StatusText(http.StatusOK),
```
NamanJain8, line 149: Seems like we are setting the response status to OK in the audit event even before processing the request. Is it okay?
For example, It may be possible that the request may not have a valid token, even then we are setting an audit event to `ok`.

NamanJain8 (approved):
This looks good to me. Make sure to test it out.

ee/audit/interceptor_ee.go
aman-bansal: graphql endpoints will always return 200 OK. thats why i put that

PR #8047, opened 2021-09-21
fix(restore-test): Use separate map directory for each group in offline restore

ahsanbarkati:
Offline restore was using the same map directory for all the  groups. This causes the same map files to get reduced in multiple groups. 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/8047)

PR #8051, opened 2021-09-23
cherrypick - cloud: fix(restore-test): Make offline restore use separate map directory fo…

ahsanbarkati:
…r each group (#8047)

Fix offline restore by making offline restore use a separate map directory for all the groups.

(cherry picked from commit ccca73761794b25edbf76b58933aac3ce1a7f9ca)

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/8051)

### DIFF
--- a/dgraph/cmd/debug/run.go
+++ b/dgraph/cmd/debug/run.go
@@ -71,6 +71,7 @@ type flagOptions struct {
 	namespace     uint64
 	key           x.Sensitive
 	onlySummary   bool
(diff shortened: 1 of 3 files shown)


NamanJain8 (approved):

NamanJain8 (changes requested):
I don't think you need dgraph magic version in dgraph debug tool. It should use the magic version being used in alpha. Flag is not needed for that,

PR #8066, opened 2021-10-02
feat(auth): Allow none signing for development environments

Bob-Thomas:
"TL;DR"
> Enable the "none" signing algorithm in the supported algorithms list in dgraph to allow the usage of unsigned tokens during development.
(https://discuss.dgraph.io/t/accepting-the-none-algorithm-in-dgraph-graphql-auth/15813)

I was developing dgraph locally and using the firebase emulator as an auth provider.
But I had to disable the authentication layer on my local environment because dgraph doesn't support the JWT none signing method.

This method is being used in the firebase emulator to generate unsigned tokens to use during development and is only able to generate these using the emulator and cannot be adjusted to created actual signed tokens.

But I would love to actually test authentication in dgraph locally as well using the emulated firebase authentication flow. So to solve this I am making this pull request to enable the "none" signing algorithm in the supported algorithm list in dgraph auth.

I haven't tested the implementation yet will do that later this week. But feel free to review and let me know if this is worth pursuing or requires a totally different approach.|

Thanks in advance for the review

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/8066)

### DIFF
--- a/graphql/authorization/auth.go
+++ b/graphql/authorization/auth.go
@@ -45,6 +45,8 @@ const (
 
 var (
 	supportedAlgorithms = map[string]jwt.SigningMethod{
+		// Allow none signing for development environments
+		jwt.SigningMethodNone.Alg(): jwt.SigningMethodNone,
 		jwt.SigningMethodRS256.Name: jwt.SigningMethodRS256,
 		jwt.SigningMethodRS384.Name: jwt.SigningMethodRS384,
 		jwt.SigningMethodRS512.Name: jwt.SigningMethodRS512,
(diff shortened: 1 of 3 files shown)


Bob-Thomas:
Ah also found the auth_test.go files so I will add test cases for this as well. After approval that this is something dgraph actually wants to support

Bob-Thomas:
Updated the tests and added the ability to change the verificationKey in the graphql testutils with a new function.
`AppendAuthInfoWithAlgoAndVerificationkey`

MichelDiz:
@Bob-Thomas  Master doesn't exist anymore. Can you open a new PR pointing to Main? Thank you. 
Closing and waiting a New PR.

PR #8069, opened 2021-10-04
fix(race): fix multiple race conditions 

NamanJain8:
Fixes 2 race conditions:
1. Txn's `MaxAssignedSeen` and `AppliedIndexSeen` were accessed without locks, this leads to a read-write race.
2. Rollups are not supposed to modify the List.plist. But it was modifying the plist by assigning it in `out.plist` that further modifies it by running splits.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/8069)


posting/list.go
```diff
@@ -1226,7 +1223,7 @@ func (l *List) rollup(readTs uint64, split bool) (*rollupOutput, error) {
 	} else {
 		// We already have a nicely packed posting list. Just use it.
 		x.VerifyPack(l.plist)
-		out.plist = l.plist
+		out.plist = proto.Clone(l.plist).(*pb.PostingList)
```
ahsanbarkati, line 1226: Do we need to copy it? Can't we avoid copy by using locks

NamanJain8: Any modifications made to `plist` over here get reflected in the cache. So if the plist is split by the rollup then the item that was added with a cost (say 100) gets to a size of cost (say 20) and the cache would not admit more items.

So basically, we would be underusing the cache.

My current approach adds an overhead of a clone operation, I agree.

ahsanbarkati (approved):

PR #8092, opened 2021-11-10
perf(rollup): use NSplit API from sroar to improve rollup performance

NamanJain8:
This PR improves the performance of the rollups. Also, it fixes memory issues of the bulk loader.
- Use the optimized `Split` API from the sroar to split the bitmap (while doing rollup). This is an improvement over the recursive binary split that was very slow.
- In bulk loader, use `HandoverSkipList` API instead of `WriteBatch` to avoid the memory constraints of doing a huge batch write.

**Also, it introduces `BitForbidPosting` that limits the capability to store large posting lists that generate splits of length greater than the `--limit max-splits` flag. Once a posting list is marked as Forbidden it cannot be recovered.**

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/8092)

### DIFF
--- a/dgraph/cmd/alpha/run.go
+++ b/dgraph/cmd/alpha/run.go
@@ -218,7 +218,8 @@ they form a Raft group and provide synchronous replication.
 			"It expects the access JWT to be constructed outside dgraph for those users as even "+
 			"login is denied to them. Additionally, this disables access to environment variables"+
 			"for minio, aws, etc.").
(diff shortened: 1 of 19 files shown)


aman-bansal:
In alpha max splits default value is probably missing. Secondly can we make default value of max-splits equal to nquad-limits. So that it wont affect currently running workflows for customers. They can set this variable to enhance the performance

aman-bansal (approved):

PR #7775, opened 2021-05-03
feat(GraphQL): Support auth with custom DQL

minhaj-shakeel:
Fixes GRAPHQL-1171.
This PR adds support for @auth on custom DQL queries.
This PR also adds the parsing step to the DQL query.
The Algorithm tries to figure out the `queried type` at the root of the query either from the `root func` or `filters`.
Suppose if the root func is `func: type(Person)` or `func: has(Person.name)` or `func: eq(Person.age, 10)` then 
it infers that the queried type is `Person`. 
However, there are certain DQL queries on which @auth rules are not added currently since the queried type is not clear. 
for eg: 
```
me(func: uid(0x1, 0x2)) {
 ...
}
```
or the query:
```
me(func: has(name@en) {
 ...
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7775)

PR #8103, opened 2021-12-01
fix(graphql): Checkpwd func  and val inside eq func now work in custom dql

Schartey:
The checkpwd func does not work in custom dql queries since #7775 . The query rewriter for custom dql did not handle the checkpwd function. Also it did not allow to use val inside of the eq function.

Example:
```
type Query {
    authenticate(username: String, email: String, password: String!): [User] @custom(dql: """
	query q($username: string, $email: string, $password: string) {
          var(func: type(User)) @filter(eq(User.username, $username) OR eq(User.email, $email)) {
              check as checkpwd(User.password, $password)
          }
              
          authenticate(func: eq(val(check), 1)) {
              id: uid
              username: User.username
              email: User.email
          }
      }
    """
    )
}
```
This query uses checkpwd in the first query, which would be rewritten to 
```
check as User.password)
```
and it uses func: eq(val(check), 1) in the second query, which would result in
```
func: eq(check, "1")
``` 

I added handling for both of these issues.
The issue has also been documented in [discuss](https://discuss.dgraph.io/t/checkpwd-in-custom-dql-supported/16200/7)

Please let me know if I missed something. I would be glad if this could make it into the next release, as it is a minor change.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/8103)

### DIFF
--- a/gql/parser.go
+++ b/gql/parser.go
@@ -228,6 +228,9 @@ func (f *Function) IsAggregator() bool {
 
 // IsPasswordVerifier returns true if the function name is "checkpwd".
 func (f *Function) IsPasswordVerifier() bool {
+	if f == nil {
+		return false
+	}
 	return f.Name == "checkpwd"
 }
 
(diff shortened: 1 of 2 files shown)


MichelDiz:
@Schartey  Master doesn't exist anymore. Can you open a new PR pointing to Main? 
Closing and waiting a New PR.

Thank you.

PR #8109, opened 2021-12-07
feat(lambda) add docker image support for lambda.

Schartey:
We are currently restricted to usage of a node sandbox as lambda server.

With this pull request I want to add support for docker containers to run as lambda server. This allows developers to use any server language they want to develop lambdas.

How does it work:
- Rebased the Dockerfile on a dind ubuntu image, so we can start docker images on alpha.
- Added multiple flags to lambda: docker-image, docker-user, docker-password, docker-registry
- If docker-image flag is set, alpha tries to download docker image and runs image as container instead of node

Example:
I have created an example repository that uses golang as [lambda server](https://github.com/Schartey/dgraph-lambda-example).
[Here](https://github.com/Schartey/dgraph-lambda-example/blob/main/prod.docker-compose.yml) you can find an example docker-compose file on how to start the alpha server with docker image.

What's missing:
- admin api update
Allow update of docker image through admin api
- Refactoring
- Tests
- Your opinion

About alternatives:
I thought a lot about alternatives and tried to implement WebAssembly as runtime, but it is still very restricted. I could only make it work with Rust and Typescript, but I think Docker is a better and simpler alternative. 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/8109)


Schartey:
After some discussions in [discuss](https://discuss.dgraph.io/t/develop-lambdas-with-any-programming-language/16257), I will close this pull request and create a new one for WASM support in lambdas in due time.

Issue #8120, opened 2022-01-11
[Bug] Entities Response for Apollo Federation is out of order


JLaferri:
Unfortunately I don't know the minimal amount of data required to trigger this behavior but basically the data returned in `_entities` is not in the same order as the objects in my `representations` input. This causes the data to mismatch with the ID when using Dgraph as a subgraph.

Here is an example:
![image](https://user-images.githubusercontent.com/1534726/148892575-0fdcd993-cee4-415f-bbfb-3ff1823f14eb.png)

It does look like the logic attempts to order the entities correctly as per here: https://github.com/dgraph-io/dgraph/blob/master/graphql/resolve/resolver.go#L407

Seems for some reason this fails for me. I suppose I will tag @minhaj-shakeel as it seems this was written by them.

Additionally, I will try to get a local dev environment set up for Dgraph tomorrow to see if I can fix this bug though I have a feeling I may struggle getting that going.

tomhollingworth:
What version are you running? I see that v21.12.0 *might* have a fix for this from https://github.com/dgraph-io/dgraph/pull/7542

PR #8123, opened 2022-01-12
fix(federation): sort hex number ID type correctly

JLaferri:
Fixes issue: https://github.com/dgraph-io/dgraph/issues/8120

The short of it is that when the key is of type ID, dgraph uses a string to represent a hex value. The results from dgraph used in this function come back sorted by the hex value. The function, however, was sorting the expected key strings as if they were strings. So if you ever hit a boundary such as: 0x1, 0x10, 0x2, 0x3, your results would be out of order and the Apollo gateway would mismatch the data and key.

This was a pretty critical bug in my application as it destroyed any sorting I had done in a different subgraph and also cached data to the wrong ID.

Certainly would like some input as to whether this change is safe. Specifically:
1. Is it safe to check keyFieldType and assume we have a hex string?
2. Is it safe to use `len(string)` or should I use `len([]rune(string))`?
3. Would it be better to parse the ID as a u64 and sort based on that? I thought this might be risky in the case where the u64 max is hit (though this seemed unlikely?)

Additionally, would it be beneficial to write a test which caught this error case?

There are also a few other concerns I have, perhaps founded or unfounded, which I figured I would ask:
1. Is it truly safe to expect there will only ever be a single type passed into representations? https://github.com/dgraph-io/dgraph/blob/master/graphql/schema/wrappers.go#L1781-L1782
2. Is it safe to expect dgraph will always return results sorted by key when used in the function?

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/8123)

### DIFF
--- a/graphql/resolve/resolver.go
+++ b/graphql/resolve/resolver.go
@@ -374,7 +374,19 @@ func entitiesQueryCompletion(ctx context.Context, resolved *Resolved) {
 	sort.Slice(uniqueKeyList, func(i, j int) bool {
 		switch val := uniqueKeyList[i].(type) {
 		case string:
-			return val < uniqueKeyList[j].(string)
+			val2 := uniqueKeyList[j].(string)
+			switch keyFieldType {
+			case "ID":
+				// If the type is ID, we actually have a hex number as a string, so we can't sort by string
+				// naively or we end up with the wrong sort example: (0x1, 0x10, 0x2, 0x3, 0x4)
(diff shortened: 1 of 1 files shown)


MichelDiz:
@JLaferri Master doesn't exist anymore. Can you open a new PR pointing to Main? Thank you.

BTW, please Sign the CLA.

MichelDiz:
Closing and waiting a New PR.

PR #7740, opened 2021-04-21
fix(GraphQL): fix auth query rewriting with ID filter

minhaj-shakeel:
Fixes GRAPHQL-1159.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7740)

PR #8157, opened 2022-08-19
fix(GraphQL) fix auth query rewriting with ID filter (cherry-pick-7740)

joshua-goldstein:
Refers to [this](https://github.com/dgraph-io/dgraph/pull/7740) PR.  Cherry pick 7740.

### DIFF
--- a/graphql/resolve/query_rewriter.go
+++ b/graphql/resolve/query_rewriter.go
@@ -627,15 +627,6 @@ func addCommonRules(
 		return []*gql.GraphQuery{dgQuery}, rbac
 	}
 
-<<<<<<< HEAD
-	if authRw != nil && (authRw.isWritingAuth || authRw.filterByUid) &&
-		(authRw.varName != "" || authRw.parentVarName != "") {
-		// When rewriting auth rules, they always start like
-		// Todo2 as var(func: uid(Todo1)) @cascade {
-		// Where Todo1 is the variable generated from the filter of the field
-		// we are adding auth to.
-
-=======
 	// When rewriting auth rules, they always start like
 	// Todo2 as var(func: uid(Todo1)) @cascade {
 	// Where Todo1 is the variable generated from the filter of the field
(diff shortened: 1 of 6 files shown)


meghalims (approved):
Looks good 👍

skrdgraph:
@joshua-goldstein - the tests seem to have failed, I am re-running them (because I suspect this must have happened when one of us was mucking with the runners)

matthewmcneely:
Please don't merge this yet. Getting a panic when testing this locally.

joshua-goldstein:
> Please don't merge this yet. Getting a panic when testing this locally.

Hi Matthew, any idea why this might be happening locally?  Do you have a stack trace of the panic?  I reran the test on our runners three times and we see the test passing there.

graphql/resolve/query_rewriter.go
```diff
@@ -1567,7 +1573,18 @@ func idFilter(filter map[string]interface{}, idField schema.FieldDefinition) []u
 	if idsFilter == nil {
 		return nil
 	}
-	idsSlice := idsFilter.([]interface{})
```
matthewmcneely, line 1570: This was panic-ing if the type coming in from the wild was a simple string—which I'd say is the typical scenario for IDs when encoded in JWTs (taken from some session ID or something).

matthewmcneely:
@joshua-goldstein, @skrdgraph Could you guys review my recent commits to this branch?

matthewmcneely:
> > Please don't merge this yet. Getting a panic when testing this locally.
> 
> Hi Matthew, any idea why this might be happening locally? Do you have a stack trace of the panic? I reran the test on our runners three times and we see the test passing there.

@joshua-goldstein The issue was that the tests were not creating all the conditions in which an ID-assigned JWT variable was being passed—namely, the tests covered arrays (slices), but not single values. I uncovered this by actually deploying a graph with auth protected types and running queries with actual JWTs in the requests.

So the fix was to check for that single value, not just the slice. I added another test to cover this condition.

skrdgraph (approved):
thnx for explaining this, it was easy to understand this @matthewmcneely

nswamy (approved):

PR #8163, opened 2022-08-23
fix(bug): use copy of loop variable in Go routines

renatolabs:
This fixes cases where loop variables are incorrectly captured in
asynchronous code (parallel tests or Go routines). In the case of a
parallel test, the most likely scenario is that only the last test
case will be executed. In the case of Go routines that directly
access loop variables without any synchronization, the value stored in
the loop variable will likely not be the same as what it was when the
Go routine is created.

### DIFF
--- a/ee/acl/acl_test.go
+++ b/ee/acl/acl_test.go
@@ -1204,6 +1204,7 @@ func TestQueryRemoveUnauthorizedPred(t *testing.T) {
 	}
 
 	for _, tc := range tests {
+		tc := tc // capture range variable
 		t.Run(tc.description, func(t *testing.T) {
 			t.Parallel()
 			resp, err := userClient.NewTxn().Query(ctx, tc.input)
@@ -1889,6 +1890,7 @@ func TestNewACLPredicates(t *testing.T) {
 	}
 
 	for _, tc := range queryTests {
+		tc := tc // capture range variable
 		t.Run(tc.description, func(t *testing.T) {
 			t.Parallel()
 			resp, err := userClient.NewTxn().Query(ctx, tc.input)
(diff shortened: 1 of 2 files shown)


matthewmcneely (commented):
@skrdgraph I recommend that this get merged, good catch by @renatolabs here

matthewmcneely (approved):

skrdgraph:
@renatolabs thanks for the PR. 👍 to merge this (it does make sense to get this in asap)

I guess we also have similar issues in our `t.go`.

skrdgraph:
@renatolabs - could you rebase with latest changes and push again please? We would like to get this in once the tests pass.

renatolabs:
> @renatolabs - could you rebase with latest changes and push again please? We would like to get this in once the tests pass. The rebase should pull in our latest test CI setup.

Done!

skrdgraph:
> > @renatolabs - could you rebase with latest changes and push again please? We would like to get this in once the tests pass. The rebase should pull in our latest test CI setup.
> 
> Done!

Thanks, I have approved a CI run to happen on this PR now. Ill merge it one once it completes.

skrdgraph (approved):

skrdgraph:
@kevinmingtarja we need to look at why the COVERALLS token was not passed. The tests did pass - I am going to merge this, and if this continues to be an issue, we should look into this.

joshua-goldstein:
For future reference I want to add a link to a reference [here](https://go.dev/doc/faq#closures_and_goroutines) on Go.dev that mentions this issue.  It seems like a common point of confusion in Go, and this PR is indeed idiomatic Go.  One could equivalently pass the variables as arguments to the closure but this also works.

PR #8180, opened 2022-08-26
`fix(build)`: Update Makefiles

joshua-goldstein:
ARM64 builds on Apple M1 silicon are currently not functional due to rpc errors.  For example,

```
make image-local
make up
make drop-all
curl -X POST localhost:8080/alter -d '{"drop_all": true}'
{"errors":[{"message":"error occurred updating GraphQL schema, please retry: rpc error: code = Unavailable desc = transport is closing","extensions":{"code":"Error"}}]}%  
```

The issue was resolved by running `GOOS=linux GOARCH=amd64 make dgraph`

It would be nice to have uniformity when testing Dgraph locally.  The situation is confounding now because running 

```
make image-local
make up
make drop-all
```

on an Apple M1  Mac will fail, but it will work on Intel Macs.  To this end I am proposing to modify the Makefile in the root directory to default to this behavior.  At some point arm64 architecture should be supported but we want to focus on x86 first.  This PR also adds some explanatory comments to the Makefile.


Makefile
```diff
@@ -22,15 +22,16 @@ BUILD_VERSION  ?= $(shell git describe --always --tags)
 
 MODIFIED = $(shell git diff-index --quiet HEAD || echo "-mod")
 
-SUBDIRS = dgraph
+#SUBDIRS = dgraph
```
MichelDiz, line 25: if this variable is not going to be used, it must be removed. It is not good practice to keep parts of code commented out.

joshua-goldstein: I agree.  This part of the Makefile

```
install:
	@(set -e;for i in $(SUBDIRS); do \
		echo Installing $$i ...; \
		$(MAKE) -C $$i install; \
	done)
```

looks like it at some point went through multiple subdirectories when installing.  Now it is only visiting the dgraph subdirectory.  If this logic is really not necessary anymore I can change this to just `$(MAKE) -C dgraph install`.

joshua-goldstein: I removed the for loop and the SUBDIRS variable.

skrdgraph: the test `t.go` binary & other local binaries (like `compose`) possibly get built through that `install`. I don't know if it's a good idea to remove it.

Makefile
```diff
@@ -44,38 +43,38 @@ version:
...
 uninstall:
-	@(set -e;for i in $(SUBDIRS); do \
-		echo Uninstalling $$i ...; \
-		$(MAKE) -C $$i uninstall; \
-	done)
+	@echo "Uninstalling dgraph..."
+	@$(MAKE) -C dgraph uninstall
 
+# ./test.sh is defunct
+# use t.go in t directory
```
skrdgraph, line 57: add a `TODO` at the start of the comments in one 56-57

joshua-goldstein: Done

Makefile
joshua-goldstein: As we discussed `make install` is not being removed, this change only removes the unnecessary for loop.  (Posting this comment in case someone reads this exchange later).

joshua-goldstein:
Rebasing on main (CI tests should now pass).

joshua-goldstein:
Merged in [this PR](https://github.com/dgraph-io/dgraph/pull/8304)

PR #8185, opened 2022-08-29
fix(test):Unskip wiki test suite

meghalims:
## Problem
wiki tests were skiipped 
 

## Solution
Unskipping wiki tests


skrdgraph (approved):
looks good 👍

PR #8186, opened 2022-08-29
fix(test):Unskip compose test suite

meghalims:
## Problem
 compose test suite was skipped
 

## Solution
remove compose from skip list as it has no tests and hence doesn't need to be skipped
 <!-
 Please add a description with these things:
 1. Explain the solution to make it easier to review the PR.
 2. Make it easier for the reviewer by describing complex sections with comments.
 -->


skrdgraph (approved):
👍

PR #8187, opened 2022-08-30
fix(test): adding lint ignore

skrdgraph:
## Problem
 
- `lint` runs on certain test-folders 
- `lint` doesn't exclude certain test-files
- `lint` `gosimple` is a good to have vs. a necessity

## Solution
 
- introduce skip-dirs (for folder skipping)
- introduce skip-files (for file skipping)
- comment out `gosimple` 
- introduce line limits for `lll` to `120`

#### Results 

`lint` error count reduces by 210+ with this change

- ##### before
  ```
   > golangci-lint run | wc -l
  WARN [runner] The linter 'golint' is deprecated (since v1.41.0) due to: The repository of the linter has been archived by the owner.  Replaced by revive. 
       507
  ```
- ##### after
  ```
   > golangci-lint run | wc -l
  WARN [runner] The linter 'golint' is deprecated (since v1.41.0) due to: The repository of the linter has been archived by the owner.  Replaced by revive. 
       291
  ```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -14,7 +14,7 @@ run:
 
 linters-settings:
   lll:
-    line-length: 100
+    line-length: 120
 
 linters:
   disable-all: true
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,5 +1,16 @@
 run:
   tests: false
+  skip-dirs:
+    - contrib
+    - compose
+    - graphql/bench
+    - graphql/e2e
+    - graphql/test
+    - graphql/testdata
+    - systest
+    - t
+  skip-files:
+    - ".*test.go$"
 
 linters-settings:
   lll:
@@ -13,7 +24,7 @@ linters:
     - gas
     - gofmt
     - golint
-    - gosimple
+    #- gosimple
     - govet
     - lll
     - varcheck


skrdgraph:
@matthewmcneely @MichelDiz @joshua-goldstein @dshekhar95 - pls do take a look (I'm making these changes as per our discussion yesterday)

.golangci.yml
```diff
@@ -1,9 +1,20 @@
...
+    - contrib
+    - compose
+    - graphql/bench
+    - graphql/e2e
+    - graphql/test
+    - graphql/testdata
+    - systest
+    - t
+  skip-files:
+    - ".*test.go$"
```
matthewmcneely, line 13: I think this should be ".*_test.go$"

From [https://go.dev/doc/tutorial/add-a-test](https://go.dev/doc/tutorial/add-a-test): "Ending a file's name with _test.go tells the go test command that this file contains test functions."

matthewmcneely (approved):

.golangci.yml
skrdgraph: actually they have this flag set here for what you are referring to https://github.com/dgraph-io/dgraph/pull/8187/files#diff-6179837f7df53a6f05c522b6b7bb566d484d5465d9894fb04910dd08bb40dcc9R2, but the issue I noticed was some test-files don't follow that pattern for that flag to take effect. Hence, I introduced the `*test.go` vs. `*_test.go`.

skrdgraph: I'm resolving this, because the `regex` `*test.go` is a superset to `*_test.go`. I hope that is okay.

PR #8194, opened 2022-08-30
chore(test): unskip codec

kevinmingtarja:
Problem
We were skipping codec suite. Tests pass.

Solution
Unskip codec suite.


kevinmingtarja:
Rebase done. All tests run and succeed in this package.

PR #8204, opened 2022-08-31
chore(test): unskip algo test suite

kevinmingtarja:
### Problem
We were skipping algo suite. Tests pass.

### Solution
Unskip algo suite.


meghalims (approved):
Looks good 👍

PR #8217, opened 2022-09-02
chore (test): unskip lex suite

SashaDgraph:
## Problem
 
Lex was skipped.
## Solution
 
Lex was unskipped.


SashaDgraph:
In both runs of push tests, query suite is failing. Same passes on pull request.

meghalims (approved):
Looks Good. Query suite failing on push, but passing on pull-request. Approving because Query is a known failure

PR #8219, opened 2022-09-02
fix(testing): parallelism and binary build

joshua-goldstein:
## Problem

Changing some default parameters in t.go for testing purposes.  We have had issues with concurrency flag and build.

## Solution

Change j=3 to j=1 and set rebuild-binary to false.

### DIFF
--- a/t/t.go
+++ b/t/t.go
@@ -63,7 +63,8 @@ var (
 		"Run only custom cluster tests.")
 	count = pflag.IntP("count", "c", 0,
 		"If set, would add -count arg to go test.")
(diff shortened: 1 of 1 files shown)


joshua-goldstein:
We are observing that dgraph-tests (push) ran in 6 minutes, which is about 2 minutes faster than we were seeing before.

skrdgraph (approved):
Something to keep in mind as we do this, we are changing the defaults for the `t.go` script - and this could affect TC in some ways. From what I am seeing, this won't affect anything - but this is something to remember.

PR #8225, opened 2022-09-06
fix(test): unskip contrib and t pkg

skrdgraph:
## Problem
 
contrib & t pig's are skipped 

## Solution
 
unskip `contrib` & `t`


joshua-goldstein (approved):
Looks good 👍

PR #8226, opened 2022-09-06
unskip tok test suite

kevinmingtarja:
## Problem
`tok` was skipped, but all tests pass

## Solution
Unskip `tok`


skrdgraph:
I resolved the conflict after the tests had passed for `Tok`. It's rerunning because of conflict resolution. @dshekhar95 could you check the test run log to make sure `Tok` passed in the previous run?

dshekhar95 (approved):
Looks fine, tests passing

PR #8234, opened 2022-09-06
possible fix for t pkg flag

dshekhar95:
## Problem

when using the --pkg flag running tests thru t locally, currently it takes any suffix match.
ex. if i set --pkg=auth, it will run tests not only for auth but also for admin_auth
 

## Solution
Add a "/" before the suffix check to ensure only full pkg name works


skrdgraph (approved):
looks good 👍 

PS: I'll merge this - after other PRs have gotten in (working on some listing PRs, will merge this one later in the evening)

PR #8239, opened 2022-09-07
added compatibility flag to docker compose in t.go

dshekhar95:
## Problem
docker container names are not matching those generated in t.go, preventing tests from running properly
 

## Solution
Add the --compatibility flag to the t.go docker compose command to ensure consistency with container names (prefix and container label should be concatenated with _, not -)


joshua-goldstein (commented):
Looks good to me, I just ran into the same issue.  For future reference, the container name convention change is mentioned [here](https://docs.docker.com/compose/cli-command-compatibility/).  Might also be a good idea to just change that one line (and not the spacing of the comments).

t/t.go
```diff
@@ -109,7 +109,7 @@ func command(args ...string) *exec.Cmd {
 
 func startCluster(composeFile, prefix string) error {
 	cmd := command(
-		"docker-compose", "-f", composeFile, "-p", prefix,
+		"docker-compose", "--compatibility", "-f", composeFile, "-p", prefix,
```
skrdgraph, line 112: I think we should add this everywhere in `t.go`, even inside `stopCluster` (I'm guessing this may affect the `./t -r` cleanup process.

skrdgraph (changes requested):

t/t.go
dshekhar95: added in latest commit

skrdgraph:
@dshekhar95 could you rebase?

Once the tests pass - we can merge this in.

dshekhar95:
> @dshekhar95 could you rebase?
> 
> Once the tests pass - we can merge this in.

did i do this right?

PR #8241, opened 2022-09-08
fix(ci): modify ci trigger events

skrdgraph:
## Problem
 
The current CI is causing double triggers for tests (it's setup for `pull_request` & `push` events)
Also we don't have continuous scheduled triggers on `default` main branch (which should run every 30mins)

## Solution
 
- modify the event triggers to avoid double runs
- add schedule triggers for every 30mins on `main` branch (this should also help us identify test-flakiness)

### DIFF
--- a/.github/workflows/ci-dgraph-tests.yml
+++ b/.github/workflows/ci-dgraph-tests.yml
@@ -1,5 +1,13 @@
 name: ci-dgraph-tests
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+  schedule:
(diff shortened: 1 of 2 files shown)


skrdgraph:
fyi @matthewmcneely @meghalims .. this PR is addressing these double triggers.

And because we still have test-flakiness, the `schedule` run on default branch will open up the can of worms.

Screen Shot 2022-09-08 at 06 05 52

skrdgraph:
This is for your reference @meghalims https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onschedule ... there are more options we should explore here.

PR #8266, opened 2022-09-13
fix(test) open tests with gogc setting

skrdgraph:


skrdgraph:
@meghalims @joshua-goldstein @dshekhar95 @kevinmingtarja - `GOGC` does make the test very stable, but takes extra time. 

We are closing this PR because this is not the ideal way to run our tests.

PR #8271, opened 2022-09-13
fix(test): to execute on self-hosted runner

skrdgraph:
## Problem
 
GH-Hosted Runners have limited resources for our tests. This is causing certain tests to fail because of mem issues.

## Solution
 
We are switching to self-hosted runners.


joshua-goldstein (approved):
Looks good 👍  This is the same skip list that was in TeamCity.  This self-hosted machine should have enough resources to pass all tests consistently.

PR #8298, opened 2022-09-15
fix(test): fix docker layers & test deps for local image

skrdgraph:
## Problem
 
Current problem is as follows: 
- we are referencing stale docker images for our tests
- the image we are using comes from our repository, which was built 1+yr ago
- this means our test environment is running tests on a stale docker environment
- we should move away from this model for obvious reasons because:
  - we are testing `environment`
  - we also testing our `build` in that `environment` (they are coupled*) - and we should treat them this way!

## Solution
 
- I am building the fresh local docker image
- I am changing the reference spec for this and we are moving out of `dgraph/dgraph:latest` -> `:local`
- I am also changing the Docker RUN cmd to layer it correctly to speed up the build steps

### DIFF
--- a/.github/workflows/ci-dgraph-tests.yml
+++ b/.github/workflows/ci-dgraph-tests.yml
@@ -29,6 +29,8 @@ jobs:
           go mod tidy
           make regenerate
           git diff --exit-code -- .
+      - name: Make Docker Image
+        run: make image-local
       - name: Make Linux Build
         run: |
           #!/bin/bash
--- a/contrib/Dockerfile
+++ b/contrib/Dockerfile
@@ -7,16 +7,16 @@
 FROM ubuntu:20.04
 LABEL maintainer="Dgraph Labs <contact@dgraph.io>"
 
-RUN apt-get update && \
-  apt-get install -y --no-install-recommends \
-    ca-certificates \
-    curl \
-    htop \
-    iputils-ping \
-    jq \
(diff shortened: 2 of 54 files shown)


sanjayk-github-dev (approved):
Nice change! Now we can be sure that we are indeed testing the latest environment.

.github/workflows/ci-dgraph-tests.yml
```diff
@@ -29,6 +29,8 @@ jobs:
           go mod tidy
           make regenerate
           git diff --exit-code -- .
+      - name: Make Docker Image
+        run: make image-local
```
matthewmcneely, line 33: One thing I'm working about: for the Github Action that's not a custom runner, are we sure there's a local docker registry that this is getting tagged in? I guess, the answer is "yes" because dgraph/dgraph:local is never pushed to a remote image registry.

matthewmcneely (commented):
Look good

.github/workflows/ci-dgraph-tests.yml
skrdgraph: Yes is the answer. 

I validated this on the `runner` to make sure this gets referenced correctly. And it does! I also went one step above, I cleaned up all the `stale` docker images pointing to `latest` on the runner. And since the merge, the machine has not pulled in `dgraph/dgraph:latest`

PR #8301, opened 2022-09-16
fix(ci): add cleanup step

joshua-goldstein:
## Problem

Dgraph CI tests were failing overnight during the Github actions/checkout setup.  This looks like a small, common issue.  See [here](https://github.com/actions/checkout/issues/211) and [here](https://github.com/Gamify-IT/issues/issues/60).  

## Solution
 
Add a cleanup job before running the actual tests.


skrdgraph:
@joshua-goldstein don't close this PR - till we are very sure everything is consistent. This PR may come in handy.

PR #8302, opened 2022-09-16
feat(testing): automate local testing workflow

joshua-goldstein:
## Problem

Currently in order to run tests locally we are mirroring the workflow [here](https://github.com/dgraph-io/dgraph/blob/main/.github/workflows/ci-dgraph-tests.yml#L32-L65).  On a local machine this requires running (from root dgraph directory):

```
make image-local
GOOS=linux GOARCH=amd64 make dgraph
cp dgraph/dgraph $GOPATH/bin
cd t; go build t
./t -r
go clean -testcache
./t --coverage=true --skip systest/backup,systest/online-restore,systest/loader,tlstest
./t -r
```

This is tedious to do repeatedly.

## Solution

I modified the root Makefile and added a Makefile in the t directory to automate this workflow.  If you want to change which tests you run, simply modify t/Makefile appropriately.  This workflow was written with an Apple M1 in mind but it should also work on Linux.  To use this workflow now, all you need to do is run `make test`.

### DIFF
--- a/t/Makefile
+++ b/t/Makefile
@@ -28,6 +28,6 @@ test:
 #	clean go testcache
 	go clean -testcache
 #	run the tests
-	@./t --pkg=posting
+	@./t --skip systest/backup,systest/online-restore,systest/loader,tlstest
 #	clean up docker containers after test execution
 	./t -r
--- a/t/Makefile
+++ b/t/Makefile
@@ -1,0 +1,45 @@
+#
+# Copyright 2018 Dgraph Labs, Inc. and Contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


t/Makefile
```diff
@@ -0,0 +1,33 @@
...
+
+all: test
+
+test: 
+# 	build the t.go binary
+	@go build .
+#	clean go testcache
+	go clean -testcache
+#	run the tests
+	@./t --pkg=posting
```
skrdgraph, line 31: the idea here is ... 

```
make test <content>
```
should get passed into this area
```
@./t <content>
```

this way everything will be clean

skrdgraph (approved):

PR #7895, opened 2021-06-09
fix(GraphQL): optimize eq filter queries

minhaj-shakeel:
Fixes GQLSAAS-1236.
This PR optimizes `eq` filter GraphQL queries.
For ex:
The below DQL query:
```
query {
  queryPerson(filter: {nick: {eq: "Dgraph"}}){
    id
    name
    nick
  }
}
```
will now be written into:
```
query {
  queryPerson(func: eq(Person.nick, "Dgraph")) @filter(type(Person)) {
    Person.id : uid
    Person.name : Person.name
    Person.nick : Person.nick
  }
}
```
which was earlier written into:-
```
query {
  queryPerson(func: type(Person)) @filter(eq(Person.nick, "Dgraph")){
    Person.id : uid
    Person.name : Person.name
    Person.nick : Person.nick
  }
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7895)

PR #8303, opened 2022-09-16
fix(GraphQL): optimize eq filter queries (#7895)

dshekhar95:
Fixes GQLSAAS-1236.
This PR optimizes `eq` filter GraphQL queries.
For ex:
The below DQL query:
```
query {
  queryPerson(filter: {nick: {eq: "Dgraph"}}){
    id
    name
    nick
  }
}
```
will now be written into:
```
query {
  queryPerson(func: eq(Person.nick, "Dgraph")) @filter(type(Person)) {
    Person.id : uid
    Person.name : Person.name
    Person.nick : Person.nick
  }
}
```
which was earlier written into:-
```
query {
  queryPerson(func: type(Person)) @filter(eq(Person.nick, "Dgraph")){
    Person.id : uid
    Person.name : Person.name
    Person.nick : Person.nick
  }
}
```

### DIFF
--- a/graphql/resolve/query_rewriter.go
+++ b/graphql/resolve/query_rewriter.go
@@ -671,6 +671,17 @@ func rewriteAsQuery(field schema.Field, authRw *authRewriter) []*gql.GraphQuery 
 		return append(dgQuery, selectionAuth...)
 	}
 
+	dgQuery = rootQueryOptimization(dgQuery)
+	return dgQuery
+}
+
+func rootQueryOptimization(dgQuery []*gql.GraphQuery) []*gql.GraphQuery {
+	if dgQuery[0].Filter != nil && dgQuery[0].Filter.Func != nil &&
+		dgQuery[0].Filter.Func.Name == "eq" && dgQuery[0].Func.Name == "type" {
+		rootFunc := dgQuery[0].Func
+		dgQuery[0].Func = dgQuery[0].Filter.Func
+		dgQuery[0].Filter.Func = rootFunc
+	}
 	return dgQuery
 }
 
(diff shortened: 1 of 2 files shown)


matthewmcneely:
We should point out in the release notes that the optimization is not realized when complex filters (AND, NOT, OR) are present in the GraphQL query. For instance see the "Filter with nested 'and'" test in query_test.yaml

matthewmcneely (approved):

meghalims (approved):
Looks Good 👍 . Comment from Matthew will be added to Release Notes

skrdgraph:
@dshekhar95 @matthewmcneely @MichelDiz - `dgraph` docs tells me that these are basic types https://dgraph.io/tour/basic/3/

The test cases test for `eq` for `string`, `date time` ... I am wondering if we should add cases for other types. 

Another question here, how would this `eq` work for complex types like arrays / objects? Does this PR address those conditions?

nosql22 (approved):

matthewmcneely:
> The test cases test for `eq` for `string`, `date time` ... I am wondering if we should add cases for other types.

As this PR strictly deals with adding @filter(type(XXX)) when it can, I don't think totally covering the other types is necessary.

matthewmcneely:
> Another question here, how would this `eq` work for complex types like arrays / objects? Does this PR address those conditions?

I don't think raw _Objects_ are a valid type of argument for any comparison in GraphQL itself. Dgraph _does_ have the `in` filter parameter which allows searching for values in array (really unordered sets) edges. There seems to be a test for that as it relates to this PR [here](https://github.com/dgraph-io/dgraph/blob/f3395e3cc0b9f332cc1d21405183c91817fc922e/graphql/resolve/query_test.yaml#L34)

nswamy (approved):

skrdgraph (approved):

PR #8307, opened 2022-09-16
fix(test) add cleanup at start

skrdgraph:
## Problem
 
so the current issue is the cleanup step does not work because crufts left by `load-test` are in sudo user group. The action runner cannot clean those because of user perm issues. This was the original PR https://github.com/dgraph-io/dgraph/pull/8301 which I closed, but in hindsight had the right fix. The fix details are referenced in the solution section.

## Solution
 
this is an elegant way to clean it, because the underlying container runs in sudo-ers group.

### DIFF
--- a/.github/workflows/ci-dgraph-load-tests.yml
+++ b/.github/workflows/ci-dgraph-load-tests.yml
@@ -9,12 +9,17 @@ on:
   schedule:
     - cron: "*/30 * * * *"
 jobs:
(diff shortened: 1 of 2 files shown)


joshua-goldstein (approved):
Looks good.  This is a fix for the issue described in [this PR](https://github.com/dgraph-io/dgraph/pull/8301).

PR #8309, opened 2022-09-16
Chore (test): unskip tlstest 

SashaDgraph:
## Problem
 
Some crts were expired.
## Solution
 
Expired crts have been updated


.github/workflows/ci-dgraph-tests.yml
```diff
@@ -60,7 +60,7 @@ jobs:
           # build the test binary
           cd t; go build .
           # run the tests
-          ./t --coverage=true --skip systest/backup,systest/online-restore,systest/loader,tlstest
+          ./t --skip systest,dgraph/cmd,graphql/e2e
```
skrdgraph, line 63: @SashaDgraph your rebase seems to incorrect here, we have opened up all the tests. And we have left out specific subfolders in `systest`. You will have to align this file correctly to point to this https://github.com/dgraph-io/dgraph/blob/main/.github/workflows/ci-dgraph-tests.yml#L63

This should ideally get resolved with a rebase* - could you please see if you can fix this up?

skrdgraph (changes requested):

.github/workflows/ci-dgraph-tests.yml
SashaDgraph: OK. I will look into it and update in a few minutes.

SashaDgraph: Updated. Please review when tests are done running.

skrdgraph: Thanks 🙏

skrdgraph:
@SashaDgraph things look good, you can make another PR for your readme Monday. I would ideally suggest this gets automated into our test-go code, so we don't have to update these certs because of cert expiry* issue we noticed. Lets see if we can spend sometime on that next week.

skrdgraph (approved):
@SashaDgraph things looks good with your changes. The tests pass consistently on `tastiest`. 

Next set of PRs (needed here):
- we need to make one readme for the `cert gen steps`
- we need to see if we can automate this into the tests, so that there are no dependencies to externally created certificates

Lets talk about this early next week. I am merging this.

PR #8313, opened 2022-09-18
fix(ci) load test environment cleanup

skrdgraph:
## Problem
 
environment cleanup is not happening correctly, there are crufts by docker - which are  owned by root. This causes the next runner job to get impacted in the cleanup step.

## Solution
 
switching to a cron based ownership changer, to handle the issue at infrastructure layer.


skrdgraph:
fyi @joshua-goldstein @dshekhar95

PR #8314, opened 2022-09-18
fix(test) CI test data fix badges & increase frequency for cruft cleanup

skrdgraph:
## Problem
 
badges don't reflect the true state of our repo
frequency of cruft cleanup could cause issues

## Solution
 
update badges
increasing cleanup frequency


skrdgraph:
@MichelDiz - I will need some help next week, we should start cleaning this landing page doc. Could you setup a 30min discussion on my calendar for this.

PR #8315, opened 2022-09-19
fix(groot): do not upsert groot for all namespaces on restart (#7917)

dshekhar95:
Earlier, whenever the alpha starts(or restarts), we were upserting guardian and groot for all the namespaces. This is not actually needed. The change was made in the PR #7759 to fix a bulk loader edge case. This PR fixes that by generating the required RDFs in the bulk loader itself. Essentially, it inserts the ACL RDFs when force loading into non-Galaxy namespace.

### DIFF
--- a/dgraph/cmd/bulk/mapper.go
+++ b/dgraph/cmd/bulk/mapper.go
@@ -31,6 +31,7 @@ import (
 
 	"github.com/dgraph-io/dgo/v210/protos/api"
 	"github.com/dgraph-io/dgraph/chunker"
+	"github.com/dgraph-io/dgraph/ee/acl"
 	"github.com/dgraph-io/dgraph/gql"
 	"github.com/dgraph-io/dgraph/posting"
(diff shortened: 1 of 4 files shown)


matthewmcneely:
There doesn't seem to be much test coverage for this. I found a thread that discusses issues with testing ACLs when they were first added: https://discuss.dgraph.io/t/acl-testing-findings/4920/3

@dshekhar95 Can you describe the testing that went into this?

dshekhar95:
> There doesn't seem to be much test coverage for this. I found a thread that discusses issues with testing ACLs when they were first added: https://discuss.dgraph.io/t/acl-testing-findings/4920/3
> 
> @dshekhar95 Can you describe the testing that went into this?

Hi Matthew... there is no testing for this. This is not a cherry-pick that I have high confidence in given the lack of a test here and lack of broader coverage as you pointed out

skrdgraph:
@dshekhar95 @meghalims & I decided to punt this for next release. Going to move this draft

PR #8320, opened 2022-09-19
fix: add validation of null values with correct order of graphql rule…

dshekhar95:
… validation (#8007)

* fix: add validation of null values with correct order of graphql rule validation

### DIFF
--- a/graphql/schema/rules.go
+++ b/graphql/schema/rules.go
@@ -47,7 +47,11 @@ func init() {
 	validator.AddRule("Check variable type is correct", variableTypeCheck)
 	validator.AddRule("Check arguments of cascade directive", directiveArgumentsCheck)
 	validator.AddRule("Check range for Int type", intRangeCheck)
-	validator.AddRule("Input Coercion to List", listInputCoercion)
+	// Graphql accept both single object and array of objects as value when the schema is defined
+	// as an array. listInputCoercion changes the value to array if the single object is provided.
+	// Changing the value can mess up with the other data validation rules hence we are setting
+	// up the order to a high value so that it will be executed last.
+	validator.AddRuleWithOrder("Input Coercion to List", 100, listInputCoercion)
 	validator.AddRule("Check filter functions", filterCheck)
 
 }
--- a/graphql/schema/rules.go
+++ b/graphql/schema/rules.go
@@ -31,6 +31,11 @@ import (
 	"github.com/dgraph-io/gqlparser/v2/validator"
 )
 
+const (
+	baseRules         = 0
+	listCoercionRules = 1000
+)
+
 func init() {
 	schemaDocValidations = append(schemaDocValidations, typeNameValidation,
 		customQueryNameValidation, customMutationNameValidation)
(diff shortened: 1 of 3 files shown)


skrdgraph:
@dshekhar95 you may have to resign CLA license

graphql/schema/rules.go
```diff
@@ -47,7 +47,11 @@ func init() {
 	validator.AddRule("Check variable type is correct", variableTypeCheck)
 	validator.AddRule("Check arguments of cascade directive", directiveArgumentsCheck)
 	validator.AddRule("Check range for Int type", intRangeCheck)
-	validator.AddRule("Input Coercion to List", listInputCoercion)
+	// Graphql accept both single object and array of objects as value when the schema is defined
+	// as an array. listInputCoercion changes the value to array if the single object is provided.
+	// Changing the value can mess up with the other data validation rules hence we are setting
+	// up the order to a high value so that it will be executed last.
+	validator.AddRuleWithOrder("Input Coercion to List", 100, listInputCoercion)
```
nosql22, line 54: We should create a GitHub issue for fixing this correctly. Number 100 should not hardcoded. What if we have more than 100 rules?

nosql22 (approved):

graphql/schema/rules.go
MichelDiz: Limiting is a good practice, but also left a note for the user is good.

If you don't limit some things, abuse can happen. Especially in shared clusters. Some user might run a very wide complex query and bring down a cluster.

matthewmcneely: @MichelDiz The 100 value is not a size limit on the array, rather the _position_ of the `rule` in the list. At present there are only 4 other rules in the gqlparser validator, so 100 definitely places it at the bottom.

Ideally the gqlparser's API would allow access to the length of the list, so one could make a dynamic decision, but at present that data structure to private to the package.

I don't like seeing hardcoded stuff like this either, but in this case it's not as bad as it looks.

matthewmcneely (approved):
Looks good

skrdgraph (changes requested):
@dshekhar95 @matthewmcneely - I was trying looking at the cherry-pick PR https://github.com/dgraph-io/dgraph/pull/8007 summary - and I have a few thoughts here: 
- from what is explained there, it looks like `listInputCoercion changes the {} to [{}]`
- leading to `empty data object will be inserted leading to a corrupted database`

the `listInputCoercion` is placed at the end by setting to `100` to solve this issue. This is a VERY DIRTY hack! I think we can do a little better here. 

Here are my thoughts: 
- I did look into `gqlparser`, this is a custom package written by old-team and its located here 
- the specific function call they are using is this one https://github.com/dgraph-io/gqlparser/blob/master/validator/validator.go#L29-L36
- notice the `order int`, this is of type `int` (we should use the max-value of int for our use-case)
- I am of the strong opinion that we should change `100` to `const MaxInt = int(MaxUint >> 1) `

fyi @joshua-goldstein @nosql22 @meghalims ^

Pls let me know if this is acceptable? 

----------------------------------------------------

[Good To Also Do] Also, it would be good to add a negated test-case here: 
- the fundamental issue here is the rule-engine ordering where validations should be made to happen first, and then insertions.
- we should add a simple test-case to test this fix, or a negated condition

graphql/schema/rules.go
skrdgraph: @matthewmcneely what are your thoughts here https://github.com/dgraph-io/dgraph/pull/8320#pullrequestreview-1121706742 ... I think we should do it right.

I understand there are only 4 other rules in the gqlparser, but the number of rules is capped by MaxInt value. Can we pin this to `MaxInt`. This way we are also being explicit (vs. implicit) in the code.

dshekhar95:
> @dshekhar95 @matthewmcneely - I was trying looking at the cherry-pick PR #8007 summary - and I have a few thoughts here:
> 
> * from what is explained there, it looks like `listInputCoercion changes the {} to [{}]`
> * leading to `empty data object will be inserted leading to a corrupted database`
> 
> the `listInputCoercion` is placed at the end by setting to `100` to solve this issue. This is a VERY DIRTY hack! I think we can do a little better here.
> 
> Here are my thoughts:
> 
> * I did look into `gqlparser`, this is a custom package written by old-team and its located here
> * the specific function call they are using is this one https://github.com/dgraph-io/gqlparser/blob/master/validator/validator.go#L29-L36
> * notice the `order int`, this is of type `int` (we should use the max-value of int for our use-case)
> * I am of the strong opinion that we should change `100` to `const MaxInt = int(MaxUint >> 1) `
> 
> fyi @joshua-goldstein @nosql22 @meghalims ^
> 
> Pls let me know if this is acceptable?
> 
> [Good To Also Do] Also, it would be good to add a negated test-case here:
> 
> * the fundamental issue here is the rule-engine ordering where validations should be made to happen first, and then insertions.
> * we should add a simple test-case to test this fix, or a negated condition

Will add your fix Sudhish

graphql/schema/rules.go
matthewmcneely: Maybe MaxInt isn't the best solution... what was wanted is that for that rule to be at the end _of where the set of rules currently stands_, MaxInt definitely does that but what about the next Rule that someone adds that wants to go just before?

An alternative: create a few consts that can be used to place items in logical order, say

```
const (
    baseRules = 0
    listCoercionRules = 1000
    fooRules = 2000
)
```

With this, subsequent authors can decide where in the logical order they want their rules to go.

skrdgraph (approved):

PR #8321, opened 2022-09-20
WIP: Add memory profiling in t.go

kevinmingtarja:
## Problem
 
Some heavy tests are still failing with exit code 1 in CI, which might be due to the machine running out of memory. We want to investigate whether or not that is the case.

## Solution
 
To investigate this, we need to be able to get the memory profile of the tests we run.


skrdgraph:
@kevinmingtarja just to add further data point as to why we need this, even 64GB is not sufficient for `21 million` bulk loader test.

PR #8323, opened 2022-09-20
security(logging): fix aes implementation in audit logging

joshua-goldstein:
## Background

Audit logging is an enterprise feature with an option for encryption using AES.  An AES cipher has three ingredients: a private key, an iv (or nonce), and plaintext.  The nonce is public but it must be unique.

## Problem

Currently we use a [16]byte nonce.  The first 12 bytes come from a baseIv which is initialized when an audit log is created.  The last 4 bytes come from the length of the log line being encrypted.  This is problematic because two log lines will often have the same length, so due to these collisions we are reusing the same nonce many times.  AES requires a unique nonce and without a unique nonce it is trivial to break.  Many thanks to @HakuPiku for pointing this out to us.

## Solution

Use crypto/rand to generate a unique nonce every time we encrypt a log line.  Previously we used what was essentially a 16 byte header (12 byte baseIv and 4 byte length of line).  We now use a [20] byte header, consisting of a unique nonce (16 bytes) and length of line (4 bytes).

## Remarks

- For audit logging, _x/log_writer.go_ controls all the audit logging (i.e., writing) while  _ee/audit/run_ee.go_ does all the decrypting.  There are other encryption/decryption tools (in testutil, e.g.) that are used for other purposes (e.g., exports).
- Customers may have old audit logs that will have to be decrypted using the deprecated method.  Now the decrypt method will determine whether the audit log is old or new and use the appropriate scheme.
- This PR also adds several integration tests to check if we can appropriately encrypt and decrypt audit logs (previously we had no integration tests for this, only a small unit test)
- We also add tests to check if an audit log in the old format can be decrypted correctly.  This is the reason there is an `.enc` file checked into the PR.
- Finally we add an extra unit test to verify that encrypted audit logs can be truncated and still decrypted successfully (up to the line that was truncated)

## References
- https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation
- https://pkg.go.dev/crypto/cipher#example-NewCTR


HakuPiku:
Any updates on when this will be merged and a security advisory created?

ee/audit/run_ee.go
```diff
@@ -107,37 +107,116 @@ func run() error {
...
 
-	for {
-		// if its the end of data. finish decrypting
-		if iterator >= stat.Size() {
-			break
-		}
-		x.Check2(file.ReadAt(iv[12:], iterator))
-		iterator = iterator + 4
+	// [12]byte baseIV + [4]byte len(x.VerificationText_Deprecated) + [11]byte x.VerificationText_Deprecated
+	decryptHeader_Deprecated := func() ([]byte, int64, error) {
```
mangalaman93, line 133: no underscore in function names

x/log_writer.go
```diff
@@ -41,7 +41,9 @@ const (
 	backupTimeFormat = "2006-01-02T15-04-05.000"
 	bufferSize       = 256 * 1024
 	flushInterval    = 10 * time.Second
-	VerificationText = "Hello World"
+	//  old logs before https://github.com/dgraph-io/dgraph/pull/8323 contain deprecated verification text in header
+	VerificationText_Deprecated = "Hello World"
+	VerificationText            = "dlroW olloH"
```
mangalaman93, line 46: is it really supposed to be olloH? or olleH?

ee/audit/run_ee.go
```diff
@@ -107,37 +107,116 @@ func run() error {
...
+		var iterator int64 = 0
+		iv := make([]byte, aes.BlockSize)
+		x.Check2(file.ReadAt(iv, iterator))
+		iterator = iterator + aes.BlockSize
+
+		t := make([]byte, len(x.VerificationText_Deprecated))
+		x.Check2(file.ReadAt(t, iterator))
+		iterator = iterator + int64(len(x.VerificationText_Deprecated))
```
mangalaman93, line 141: same for variable name, no underscore

ee/audit/run_ee.go
```diff
@@ -107,37 +107,116 @@ func run() error {
...
-		x.Check2(file.ReadAt(iv[12:], iterator))
-		iterator = iterator + 4
+	// [12]byte baseIV + [4]byte len(x.VerificationText_Deprecated) + [11]byte x.VerificationText_Deprecated
+	decryptHeader_Deprecated := func() ([]byte, int64, error) {
+		var iterator int64 = 0
+		iv := make([]byte, aes.BlockSize)
+		x.Check2(file.ReadAt(iv, iterator))
+		iterator = iterator + aes.BlockSize
+
+		t := make([]byte, len(x.VerificationText_Deprecated))
```
mangalaman93, line 139: better name than `t`

ee/audit/run_ee.go
```diff
@@ -107,37 +107,116 @@ func run() error {
...
-		if iterator >= stat.Size() {
-			break
-		}
-		x.Check2(file.ReadAt(iv[12:], iterator))
-		iterator = iterator + 4
+	// [12]byte baseIV + [4]byte len(x.VerificationText_Deprecated) + [11]byte x.VerificationText_Deprecated
+	decryptHeader_Deprecated := func() ([]byte, int64, error) {
+		var iterator int64 = 0
+		iv := make([]byte, aes.BlockSize)
+		x.Check2(file.ReadAt(iv, iterator))
```
mangalaman93, line 136: does it make sense to crash when we can't read from the file?

ee/audit/run_ee.go
```diff
@@ -107,37 +107,116 @@ func run() error {
...
+		iterator = iterator + int64(len(x.VerificationText))
 
-	stream := cipher.NewCTR(block, iv)
-	stream.XORKeyStream(t, t)
-	if string(t) != x.VerificationText {
-		return errors.New("invalid encryption key provided. Please check your encryption key")
+		text := make([]byte, len(x.VerificationText))
+		stream := cipher.NewCTR(block, iv)
+		stream.XORKeyStream(text, t)
+		if string(text) != x.VerificationText {
```
mangalaman93, line 126: can we definer the VerificationText of type []byte and use bytes.Compare instead?

systest/audit_encrypted/audit_test.go
```diff
@@ -0,0 +1,245 @@
...
+	out, err := decrypt.CombinedOutput()
+	if err != nil {
+		fmt.Printf("Error %v\n", err)
+		fmt.Printf("Output %v\n", string(out))
+		t.Fatal(err)
+	}
+
+	verifyLogs(t, fmt.Sprintf("audit_dir/za/zero_audit_0_%s.log", nId), msgs)
+}
+func TestAlphaAuditEncrypted(t *testing.T) {
```
mangalaman93, line 83: leave a newline before this

systest/audit_encrypted/audit_test.go
```diff
@@ -0,0 +1,245 @@
...
+		fmt.Printf("Output %v\n", string(out))
+		t.Fatal(err)
+	}
+
+	verifyLogs(t, fmt.Sprintf("audit_dir/aa/alpha_audit_1_%s.log", nId), msgs)
+}
+
+// deprecated logs generated with dgraph v22.0.2
+func TestZeroAuditDecryptDeprecated(t *testing.T) {
+	if len(zeroMsgs) == 0 {
```
mangalaman93, line 166: I think this is not necessary, the test could run in either order. Each test should be independent.

systest/audit_encrypted/audit_test.go
```diff
@@ -0,0 +1,245 @@
...
+
+	var args []string
+	args = append(args, "audit", "decrypt",
+		"--encryption_key_file=../../ee/enc/test-fixtures/enc-key",
+		"--in", "audit_dir_deprecated/za/zero_audit_0_1.log.enc",
+		"--out", "audit_dir_deprecated/za/zero_audit_0_1.log")
+
+	decrypt := exec.Command(testutil.DgraphBinaryPath(), args...)
+
+	fmt.Println("Running: ", decrypt.Args)
```
mangalaman93, line 179: use t.Log instead

systest/audit_encrypted/docker-compose.yml
```diff
@@ -0,0 +1,46 @@
...
+    working_dir: /data/alpha1
+    ports:
+      - "8080"
+      - "9080"
+    volumes:
+      - type: bind
+        source: $GOPATH/bin
+        target: /gobin
+        read_only: true
+      - type: bind
```
mangalaman93, line 14: use a volume instead

x/log_writer.go
```diff
@@ -236,6 +306,109 @@ func (l *LogWriter) open() error {
...
+			}
+		}
+	}
+
+	l.writer = bufio.NewWriterSize(l.file, bufferSize)
+	l.size = size()
+	return nil
+}
+
+func (l *LogWriter) open_deprecated() error {
```
mangalaman93, line 395: Have we added these functions?

mangalaman93 (changes requested):
1. Do we need the .gitkeep files?

x/log_writer.go
joshua-goldstein: this is the code that existed there previously - I just appended _deprecated to the function name.  If you feel it is not necessary to keep this function in the code I can remove it

x/log_writer.go
joshua-goldstein: Good catch :)  all we need is a string of length 11 that is different from "Hello World".  I can change it but it wouldn't make a difference

systest/audit_encrypted/docker-compose.yml
joshua-goldstein: > 1. Do we need the .gitkeep files?

I believe so, since we want to keep the audit_dir directory structure intact.  See e.g. systest/audit.  I copied over the directories from there.

x/log_writer.go
mangalaman93: we didn't add the replacement code here then? and I think we can remove the code that is not used, we can always find in the history.

systest/audit_encrypted/docker-compose.yml
mangalaman93: You can use a volume (not volume mount) and probably you wouldn't need those folders.

ee/audit/run_ee.go
```diff
@@ -107,37 +107,116 @@ func run() error {
...
+	decryptHeader := func() ([]byte, int64, error) {
+		var iterator int64 = 0
+		iv := make([]byte, aes.BlockSize)
+		x.Check2(file.ReadAt(iv, iterator))     // get first iv
+		iterator = iterator + aes.BlockSize + 4 // length of verification text encoded in uint32
 
-	t := make([]byte, len(x.VerificationText))
-	x.Check2(file.ReadAt(t, iterator))
-	iterator = iterator + int64(len(x.VerificationText))
+		t := make([]byte, len(x.VerificationText))
```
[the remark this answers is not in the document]
joshua-goldstein: done

ee/audit/run_ee.go
joshua-goldstein: done

ee/audit/run_ee.go
joshua-goldstein: done

ee/audit/run_ee.go
joshua-goldstein: done

systest/audit_encrypted/audit_test.go
joshua-goldstein: done

systest/audit_encrypted/audit_test.go
joshua-goldstein: done

x/log_writer.go
joshua-goldstein: sounds reasonable to me - I remove the unused code and kept only the functions necessary to decrypt old audit logs

ee/audit/run_ee.go
joshua-goldstein: Good catch...  this is indeed an issue because we may try using multiple decrypt methods and don't want to crash and should handle the error more gracefully.

systest/audit_encrypted/docker-compose.yml
joshua-goldstein: done

systest/audit_encrypted/audit_test.go
joshua-goldstein: done

systest/audit_encrypted/docker-compose.yml
joshua-goldstein: moving back to a bind mount - as we discussed using a mount volume created other issues and is not necessary here since we are not persisting data beyond the life of the test.  The only "issue" is that .gitkeep directories are aesthetically unpleasing, but we already do this elsewhere in the repo (e.g. in systest/audit package) so we can revisit this later in the context of making a broader change (to other test packages as well)

ee/audit/run_ee.go
joshua-goldstein: Currently we declare a string constant in `x/log_writer.go` .  I think this is reasonable since this should be a constant, and we can't declare an array/slice to be a constant.  We could cast the x.VerificationText as a byte slice and compare the bytes but I don't think this makes a difference at all

ee/audit/run_ee.go
```diff
@@ -107,37 +107,161 @@ func run() error {
...
+	// #################################################################
+	decryptBody := func() {
+		for {
+			count++
+			// if its the end of data. finish decrypting
+			if iterator >= stat.Size() {
+				break
+			}
+			n, err := file.ReadAt(iv, iterator)
+			if err != nil {
```
mangalaman93, line 193: shouldn't we break the loop or something?

ee/audit/run_ee.go
```diff
@@ -107,37 +107,161 @@ func run() error {
...
+	// #####   [16]byte IV + [4]byte uint32(len(p)) + [:]byte p    #####
+	// #################################################################
+	decryptBody := func() {
+		for {
+			count++
+			// if its the end of data. finish decrypting
+			if iterator >= stat.Size() {
+				break
+			}
+			n, err := file.ReadAt(iv, iterator)
```
mangalaman93, line 192: we should also check whether n is same as iterator probably here. I think there is a better way to write this whole code. You could take some inspiration from this code maybe https://dave.cheney.net/2019/01/27/eliminate-error-handling-by-eliminating-errors

ee/audit/run_ee.go
```diff
@@ -107,37 +107,161 @@ func run() error {
...
+			if err != nil {
+				glog.Warningf("received %v while decrypting audit log\n", err)
+				glog.Warningf("read %v bytes, expected %v\n", n, len(iv))
+			}
+			iterator = iterator + 16
+			length := make([]byte, 4)
+			n, err = file.ReadAt(length, iterator)
```
mangalaman93, line 199: same here

mangalaman93 (changes requested):
I think we should consider writing more robust functions and test. This needs more work.

ee/audit/run_ee.go
```diff
@@ -107,37 +107,161 @@ func run() error {
...
 
-	for {
-		// if its the end of data. finish decrypting
-		if iterator >= stat.Size() {
-			break
+	// [12]byte baseIV + [4]byte len(x.VerificationTextDeprecated) + [11]byte x.VerificationTextDeprecated
+	decryptHeaderDeprecated := func() ([]byte, int64, error) {
+		var iterator int64 = 0
+
+		iv := make([]byte, aes.BlockSize)
```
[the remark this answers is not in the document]
joshua-goldstein: The length of the blocksize is 16 in both cases.  Previously we were using a baseIV (12 bytes) appended with the length of the cipher text (4 bytes, uint32) as the IV.  Now we randomly generate a 16 byte IV each time, followed by the length of the cipher text (4 bytes, uint32).  This is the point of the PR, because what we were doing before (baseIV + len(ciphertext)) introduces the possibility of collisions.

ee/audit/run_ee.go
```diff
@@ -107,37 +107,161 @@ func run() error {
...
 		return nil
 	}
-	var iterator int64 = 0
 
-	iv := make([]byte, aes.BlockSize)
-	x.Check2(file.ReadAt(iv, iterator))
-	iterator = iterator + aes.BlockSize
+	// decrypt header in audit log to verify encryption key
+	// [16]byte IV + [4]byte len(x.VerificationText) + [11]byte x.VerificationText
+	decryptHeader := func() ([]byte, int64, error) {
```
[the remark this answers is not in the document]
joshua-goldstein: What is the purpose of making them the same function?  For the updated audit logs the header looks like
```
[16]byte IV + [4]byte len(x.VerificationText) + [11]byte x.VerificationText
```

whereas for the old audit logs the header is
```
[12]byte baseIV + [4]byte len(x.VerificationTextDeprecated) + [11]byte x.VerificationTextDeprecated
```

and therefore they need to be parsed differently.  It is possible that we would have to try both functions to decrypt the header, so it would be very confusing to try to put the parsing logic of both inside one function.

ee/audit/run_ee.go
```diff
@@ -107,37 +107,161 @@ func run() error {
...
+		var iterator int64 = 0
+		iv := make([]byte, aes.BlockSize)
+		_, err := file.ReadAt(iv, iterator) // get first iv
+		if err != nil {
+			return nil, 0, err
```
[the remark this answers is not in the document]
joshua-goldstein: what kind of information?  this error is propagated up, and if we are also unable to decrypt the audit log header with `decryptHeaderDeprecated` then we return `errors.New("invalid encryption key provided. Please check your encryption key")` to the caller, the run() function

ee/audit/run_ee.go
```diff
@@ -107,37 +107,161 @@ func run() error {
...
+		}
+		iterator = iterator + aes.BlockSize + 4 // length of verification text encoded in uint32
 
-	t := make([]byte, len(x.VerificationText))
-	x.Check2(file.ReadAt(t, iterator))
-	iterator = iterator + int64(len(x.VerificationText))
+		ct := make([]byte, len(x.VerificationText))
+		_, err = file.ReadAt(ct, iterator)
+		if err != nil {
+			return nil, 0, err
```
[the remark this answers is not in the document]
joshua-goldstein: again, what kind of context?  this will be propagated up and return the invalid encryption key error to the user

ee/audit/run_ee.go
```diff
@@ -107,37 +107,161 @@ func run() error {
...
+		}
+		iterator = iterator + int64(len(x.VerificationText))
 
-	stream := cipher.NewCTR(block, iv)
-	stream.XORKeyStream(t, t)
```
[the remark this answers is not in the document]
joshua-goldstein: done

ee/audit/run_ee.go
```diff
@@ -107,37 +107,161 @@ func run() error {
...
-		iterator = iterator + 4
+		iterator = iterator + aes.BlockSize
 
-		content := make([]byte, binary.BigEndian.Uint32(iv[12:]))
-		x.Check2(file.ReadAt(content, iterator))
-		iterator = iterator + int64(binary.BigEndian.Uint32(iv[12:]))
+		ct := make([]byte, len(x.VerificationTextDeprecated))
+		_, err = file.ReadAt(ct, iterator)
+		if err != nil {
+			return nil, 0, err
```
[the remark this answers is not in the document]
joshua-goldstein: again, what kind of context? this will be propagated up and return the invalid encryption key error to the user

ee/audit/run_ee.go
joshua-goldstein: n is not the same as the iterator; it is the increment by which we need to increment the iterator.  I made a few changes to make this more clear

ee/audit/run_ee.go
joshua-goldstein: yes, added break statement here

ee/audit/run_ee.go
```diff
@@ -107,37 +107,161 @@ func run() error {
...
+	// #####   [4]byte uint32(len(p)) + [:]byte p         #####
+	// ########################################################
+	decryptBodyDeprecated := func() {
+		for {
+			// if its the end of data. finish decrypting
+			if iterator >= stat.Size() {
+				break
+			}
+			n, err := file.ReadAt(iv[12:], iterator)
+			if err != nil {
```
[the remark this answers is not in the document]
joshua-goldstein: done - added break statements

ee/audit/run_ee.go
joshua-goldstein: yes, good catch - added break statement here

joshua-goldstein:
> I think we should consider writing more robust functions and test. This needs more work.

I have already added 4 integration tests, and we have a unit test to check basic parsing.  What additional tests are necessary?

ee/audit/run_ee.go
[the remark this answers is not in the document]
joshua-goldstein: Additionally the functions `decryptBody` and `decryptBodyDeprecated` have very different parsing logic (the way they increment and get the length of text is fundamentally different).  They should not be written in one function.

ee/audit/run_ee.go
```diff
@@ -107,37 +107,161 @@ func run() error {
...
 
-	iv := make([]byte, aes.BlockSize)
-	x.Check2(file.ReadAt(iv, iterator))
-	iterator = iterator + aes.BlockSize
+	// decrypt header in audit log to verify encryption key
+	// [16]byte IV + [4]byte len(x.VerificationText) + [11]byte x.VerificationText
+	decryptHeader := func() ([]byte, int64, error) {
+		var iterator int64 = 0
+		iv := make([]byte, aes.BlockSize)
+		_, err := file.ReadAt(iv, iterator) // get first iv
```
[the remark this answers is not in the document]
joshua-goldstein: we now return the number of bits written `n` and use this to increment the iterator

ee/audit/run_ee.go
```diff
@@ -107,37 +107,161 @@ func run() error {
...
+		if err != nil {
+			return nil, 0, err
+		}
+		iterator = iterator + aes.BlockSize + 4 // length of verification text encoded in uint32
 
-	t := make([]byte, len(x.VerificationText))
-	x.Check2(file.ReadAt(t, iterator))
-	iterator = iterator + int64(len(x.VerificationText))
+		ct := make([]byte, len(x.VerificationText))
+		_, err = file.ReadAt(ct, iterator)
```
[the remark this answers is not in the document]
joshua-goldstein: we now return the number of bits written `n` and use this to increment the iterator

ee/audit/run_ee.go
```diff
@@ -107,37 +107,161 @@ func run() error {
...
-	for {
-		// if its the end of data. finish decrypting
-		if iterator >= stat.Size() {
-			break
+	// [12]byte baseIV + [4]byte len(x.VerificationTextDeprecated) + [11]byte x.VerificationTextDeprecated
+	decryptHeaderDeprecated := func() ([]byte, int64, error) {
+		var iterator int64 = 0
+
+		iv := make([]byte, aes.BlockSize)
+		_, err := file.ReadAt(iv, iterator)
```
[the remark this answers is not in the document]
joshua-goldstein: we now return the number of bits written `n` and use this to increment the iterator

ee/audit/run_ee.go
```diff
@@ -107,37 +107,161 @@ func run() error {
...
 		}
-		x.Check2(file.ReadAt(iv[12:], iterator))
-		iterator = iterator + 4
+		iterator = iterator + aes.BlockSize
 
-		content := make([]byte, binary.BigEndian.Uint32(iv[12:]))
-		x.Check2(file.ReadAt(content, iterator))
-		iterator = iterator + int64(binary.BigEndian.Uint32(iv[12:]))
+		ct := make([]byte, len(x.VerificationTextDeprecated))
+		_, err = file.ReadAt(ct, iterator)
```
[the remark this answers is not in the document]
joshua-goldstein: we now return the number of bytes written `n` and use this to increment the iterator

joshua-goldstein:
> I didn't realize that byte array/slice can't be declared as a constant. This article has a few ideas https://blog.boot.dev/golang/golang-constant-maps-slices/. I don't mind if we declare a function that returns the constant value. I think it is useful to avoid a allocation that converts byte slice to string.

What is the purpose of this?  We are doing exactly one allocation per log file (default ~200 mb), so this is not a memory issue?

ee/audit/run_ee.go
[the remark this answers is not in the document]
joshua-goldstein: As discussed we will refactor to use io.Reader object

ee/audit/run_ee.go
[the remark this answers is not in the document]
joshua-goldstein: todo: `error.Wrap(err, "unable to read IV")` e.g.

ee/audit/run_ee.go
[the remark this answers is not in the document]
joshua-goldstein: Refactored to use `file io.ReaderAt, outfile io.Writer`

ee/audit/run_ee.go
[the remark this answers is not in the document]
joshua-goldstein: added error wrapper

ee/audit/run_ee.go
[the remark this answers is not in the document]
joshua-goldstein: added error wrapper

ee/audit/run_ee.go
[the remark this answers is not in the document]
joshua-goldstein: added error wrapper

testutil/testaudit/audit.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	for _, m := range cmds {
+		if !logMap[m] {
+			t.Fatalf("audit logs not present for command %s", m)
+		}
+	}
+}
+
+// run manually to generate the encrypted audit log used in TestDecrypt
+func TestGenerateAuditForTestDecrypt(t *testing.T) {
```
joshua-goldstein, line 42: This is here so that in the future it is easy to generate an encrypted audit log; of course we only want to run this manually, so we should skip it

joshua-goldstein:
The scope of this PR is now much larger than originally intended.  This PR now introduces:
- Four new integration tests for encrypted audit logging in `systest/audit_encrypted/audit_test.go`, two of which verify that encrypted audit logs can be encrypted, and two that verify that deprecated audit logs can be decrypted
- One new unit test in `ee/audit/run_ee_test.go` to verify that truncating an encrypted audit log does not cause the decrypt process to panic
- A modified unit test to decrypt using the new audit log format

This PR also does some minor refactoring to the code.

testutil/testaudit/audit.go
mangalaman93: can we mention this in the code itself.

testutil/testaudit/docker-compose.yml
```diff
@@ -0,0 +1,46 @@
...
+    working_dir: /data/alpha1
+    ports:
+      - "8080"
+      - "9080"
+    volumes:
+      - type: bind
+        source: $GOPATH/bin
+        target: /gobin
+        read_only: true
+      - type: bind
```
mangalaman93, line 14: I still don't remember why can't we use volume instead of this

x/log_writer.go
```diff
@@ -41,7 +39,9 @@ const (
 	backupTimeFormat = "2006-01-02T15-04-05.000"
 	bufferSize       = 256 * 1024
 	flushInterval    = 10 * time.Second
-	VerificationText = "Hello World"
+	//  old logs before https://github.com/dgraph-io/dgraph/pull/8323 contain deprecated verification text in header
+	VerificationTextDeprecated = "Hello World"
```
mangalaman93, line 43: would it be more useful to version them or something? What if we need to change this tomorrow again? We can also just do it the next time if that makes more sense.

mangalaman93 (changes requested):
Did a quick review for now.  I will review it again tomorrow.

testutil/testaudit/audit.go
joshua-goldstein: I already have a comment above the test - what else do you suggest?

ee/audit/run_ee.go
```diff
@@ -107,39 +108,175 @@ func run() error {
...
-	var iterator int64 = 0
 
-	iv := make([]byte, aes.BlockSize)
-	x.Check2(file.ReadAt(iv, iterator))
-	iterator = iterator + aes.BlockSize
+	if err := decrypt(file, outfile, block, stat.Size()); err != nil {
+		return errors.Wrap(err, "could not decrypt audit log")
+	}
+
+	glog.Infof("Decryption of Audit file %s is Done. Decrypted file is %s",
```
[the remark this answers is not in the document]
joshua-goldstein: made it lowercase and shorter

ee/audit/run_ee.go
```diff
@@ -107,39 +108,175 @@ func run() error {
...
+		decryptCmd.Conf.GetString("in"),
+		decryptCmd.Conf.GetString("out"))
+	return nil
+}
+
+func decrypt(file io.ReaderAt, outfile io.Writer, block cipher.Block, sz int64) error {
+
```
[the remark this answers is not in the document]
joshua-goldstein: done

ee/audit/run_ee.go
```diff
@@ -107,39 +108,175 @@ func run() error {
...
+	// #################################################################
+	decryptBody := func() {
+		for {
+			// if its the end of data. finish decrypting
+			if iterator >= sz {
+				break
+			}
+			n, err := file.ReadAt(iv, iterator)
+			if err != nil {
+				glog.Warningf("received %v while decrypting audit log\n", err)
```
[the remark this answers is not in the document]
joshua-goldstein: removed

ee/audit/run_ee.go
```diff
@@ -107,39 +108,175 @@ func run() error {
...
+				glog.Warningf("read %v bytes, expected %v\n", n, len(iv))
+				break
+			}
+			iterator = iterator + 16
+			length := make([]byte, 4)
+			n, err = file.ReadAt(length, iterator)
+			if err != nil {
```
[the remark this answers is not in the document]
joshua-goldstein: removed

ee/audit/run_ee_test.go
```diff
@@ -0,0 +1,117 @@
...
+		case 540 <= i && i <= 790:
+			// at this point the output file will be empty
+			check(t, file.Truncate(sz-int64(i)))
+			// verify that decrypt does not panic
+			check(t, decrypt(file, outfile, block, sz))
+			check(t, outfile.Truncate(0))
+			_, err := outfile.Seek(0, 0)
+			check(t, err)
+		}
+
```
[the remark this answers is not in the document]
joshua-goldstein: done

x/log_writer.go
joshua-goldstein: I hope we never have to look at this ever again :)  But in all seriousness I don't think we need to version it at this point.  If we do get in that position tomorrow we can just change these variable names.

testutil/testaudit/docker-compose.yml
joshua-goldstein: I tried using a volume and ran into issues.  Then we tried debugging the problem together and couldn't figure it out.  The problem resolved when we switched back to using a bind mount.  We use bind mounts in the other audit log tests (that existed in the repo previously in `systest/audit`).  I don't want to introduce that kind of change here in this PR, because if we change it here we should also change it elsewhere too.

ee/audit/run_ee_test.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+	}()
+	defer os.RemoveAll("testfiles/zero_audit_0_1.log")
+	block, err := aes.NewCipher(key)
+	check(t, err)
+
+	zeroCmd := []string{"/removeNode", "/assign", "/moveTablet"}
+
+	// size of audit log is 825 bytes
+	for i := 0; i <= int(sz); i = i + 5 {
```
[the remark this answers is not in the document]
joshua-goldstein: added a comment

ee/audit/run_ee_test.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+// we will truncate copy of encrypted audit log file
+func copy(t *testing.T, src string, dst string) {
+	// could also us io.CopyN but this is a small file
+	data, err := os.ReadFile(src)
+	check(t, err)
```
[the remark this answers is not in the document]
joshua-goldstein: thanks, I thought there should be a better way...  using testify now

ee/zero_audit_0_1.log.enc
mangalaman93, line 1: is this file needed?

mangalaman93 (approved):

ee/zero_audit_0_1.log.enc
joshua-goldstein: nope, good catch...  removing

mangalaman93 (approved):

skrdgraph (approved):

PR #8319, opened 2022-09-19
fix(query): Prevent multiple entries for same predicate in mutations.2022

dshekhar95:
## Problem
Multiple entries for the same predicate can lead to inconsistent state,
 

## Solution
 hence adding an error for it instead.

Fixes DGRAPH-3225

PR #8332, opened 2022-09-29
fix(query): Prevent multiple entries for same predicate in mutations.2022

skrdgraph:
we had NOT squashed & merged this https://github.com/dgraph-io/dgraph/pull/8319

So fixing it up here

## Problem
Multiple entries for the same predicate can lead to inconsistent state,
 

## Solution
 hence adding an error for it instead.

Fixes DGRAPH-3225


dshekhar95 (approved):

joshua-goldstein (approved):
👍

PR #8335, opened 2022-09-29
fix(GraphQL): Nested Auth Rules not working properly (#7915)

dshekhar95:
## Problem
Nested Auth Rules causing issues with "And" in queries

https://discuss.dgraph.io/t/auth-directives-dont-apply-to-nested-objects-when-using-interfaces/14157

### DIFF
--- a/graphql/e2e/auth/auth_test.go
+++ b/graphql/e2e/auth/auth_test.go
@@ -513,6 +513,44 @@ func TestAuthOnInterfaces(t *testing.T) {
 	}
 }
 
+func TestNestedAndAuthRulesWithMissingJWT(t *testing.T) {
+	addParams := &common.GraphQLParams{
+		Query: `
+		mutation($user1: String!, $user2: String!){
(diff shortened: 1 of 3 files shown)


matthewmcneely:
Using the example of the problem outlined at https://discuss.dgraph.io/t/auth-directives-dont-apply-to-nested-objects-when-using-interfaces/14157/9, my build from this branch does *NOT* seem to fix the reported issue.

For those looking at the issue on Discuss, the TL;DR of the issue is that this auth filter 
```
queryApprovableChange(filter: { x: true, and: { x: false } }) {
```

should always return false (true AND false => false).  My tests show that it _does_ evaluate that way (in other words, works correctly) when applied to an SDL type that is not derived from an interface, and not an edge. Complicated.

I'm going to add debug output in the graphql resolver to see if I can understand what the heck is going on.

skrdgraph:
@dshekhar95 @matthewmcneely @meghalims im changing this to draft

PR #8336, opened 2022-09-29
feat(testing): simplify package testing

joshua-goldstein:
## Problem

Previously calling `make test` ran a set of tests specified in t/Makefile.  This had to be modified every time we want to run a specific package test.

## Solution

We can now run `make test PKG=<package>` to run a specific package test.  E.g., to run the posting package test, run `make test PKG=posting`.  If the package is not specified, a standard suite of tests is run as before.

This PR also fixes the `make install` command (the SUBDIRS variable was removed in a previous PR but still referenced in the `make install` command).  

## Remark

The command `make test` is only used when testing locally.  This change will not affect CI/CD jobs.

## TODO

What should be the standard suite of tests to run when testing locally?  Some tests require high resources and often (or sporadically) fail when run locally.  We should figure out which tests can be run consistently locally and set `make test` to run these by default.

### DIFF
--- a/t/Makefile
+++ b/t/Makefile
@@ -18,7 +18,7 @@
 GOOS          ?= $(shell go env GOOS)
 GOPATH        ?= $(shell go env GOPATH)
 
-.PHONY: test
+.PHONY: all test
 
 all: test
 
@@ -27,7 +27,14 @@ test:
 	@go build .
 #	clean go testcache
 	go clean -testcache
-#	run the tests
-	@./t --skip systest/backup,systest/online-restore,systest/loader,tlstest
+#	run the tests specified in pkg; otherwise run standard suite
+	@if [ -n "$(PKG)" ]; then \
+		./t --pkg=$(PKG); \
+	else \
+		./t --skip systest/backup,systest/online-restore,systest/loader,tlstest; \
+	fi
 #	clean up docker containers after test execution
 	./t -r
+#	Exit message
+	@echo "Testing complete, goodbye!"
+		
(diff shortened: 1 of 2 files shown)


skrdgraph:
@matthewmcneely - this was what we were discussing recently. Just an FYI

skrdgraph:
@joshua-goldstein do you think it would be good to pass it like this?

`make test <ARGS>`

and the `ARGS` can be controlled at `t.go` level, this way people can run `make test --help` or add `--pkg` flag etc.

joshua-goldstein:
> @joshua-goldstein do you think it would be good to pass it like this?
> 
> `make test <ARGS>`
> 
> and the `ARGS` can be controlled at `t.go` level, this way people can run `make test --help` or add `--pkg` flag etc.

That is a nice idea, it would give us more flexibility when passing arguments to t.go.

AFAIK the only way to do this would be something like `make test ARGS="--pkg=posting --coverage=false"`.  I am not aware of a more elegant way to write it.

t/Makefile
```diff
@@ -27,7 +27,14 @@ test:
...
+	@if [ -n "$(PKG)" ]; then \
+		./t --pkg=$(PKG); \
+	else \
+		./t --skip systest/backup,systest/online-restore,systest/loader,tlstest; \
+	fi
 #	clean up docker containers after test execution
 	./t -r
+#	Exit message
+	@echo "Testing complete, goodbye!"
+
```
skrdgraph, line 40: can you fix the new line at the end? I think there are 2

joshua-goldstein:
Similar changes merged in https://github.com/dgraph-io/dgraph/pull/8304

PR #8337, opened 2022-09-30
fix(build): Update linux build steps

joshua-goldstein:
## Problem

1.)  When following install instructions for Linux, one will encounter the following error due to missing libraries:

```
/usr/local/go/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/usr/bin/ld: cannot find -lstdc++
collect2: error: ld returned 1 exit status
```

2.)  Makefile in root Dgraph directory included a deprecated SUBDIRS variables.  This was referenced when `make install` is called, giving an error.

## Solution

1.)  Instead of `apt-get install make gcc` instructions should be `apt-get install build-essential` which includes gcc, make, as well as the necessary libc packages.

2.)  Update Makefile to run `make install` correctly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ GOPATH         ?= $(shell go env GOPATH)
 ###############
 
 .PHONY: dgraph all oss version install install_oss oss_install uninstall test help image image-local local-image
-all: $(SUBDIRS)
+all: dgraph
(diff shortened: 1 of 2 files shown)


skrdgraph (approved):
looks good

PR #8341, opened 2022-09-30
feat(testing): introduce setup script

joshua-goldstein:
## Problem

Previously it was tedious setting up a fresh EC2 instance to run tests.

## Solution

This script automates the setup of a fresh EC2 instance.  It mirrors the gh-actions runner but with some small modifications to make it work (e.g. Go must be installed, path configured, etc.).  This is a WIP.


joshua-goldstein:
We use a similar script in https://github.com/dgraph-io/dgraph/blob/main/contrib/gh-runner/gh-runner.sh

PR #8369, opened 2022-10-18
fix(ci): increase file descriptor limit in gh-runners

kevinmingtarja:
## Problem
21million load test occasionally fails due to the file descriptor limit being too low

## Solution
Increase the file descriptor limit in our github runners machine to 2000000


skrdgraph (approved):
Thanks for making these changes! Looks good!

skrdgraph:
Merging this!

PR #8370, opened 2022-10-20
fix(cd): add cd step

skrdgraph:
## Problem
we don't have a release pipeline

## Solution
setup a release pipeline


skrdgraph:
@joshua-goldstein closing this because you have forked off this.

PR #8372, opened 2022-10-21
fix(CD): add release pipeline

skrdgraph:
## Problem
broken release process

## Solution
add release pipeline & make necessary changes to support it


joshua-goldstein (approved):
👍 We will finish cleanup and do more testing tomorrow

PR #8378, opened 2022-10-21
chore: CD cleanup

joshua-goldstein:
## Problem

We continue work on CD pipeline.

## Solution
- Artifacts are now saved and uploaded to Github (i.e. binary and sha)
- `make linux-dependency` include protobuf-compiler and other build/runtime dependencies
- Add `go mod tidy` check in CD steps
- Replace DGRAPH_RELEASE_VERSION with DGRAPH_VERSION
- Remove build parameters from root makefile, so they are now located only in dgraph/Makefile (they were duplicated in both Makefile and dgraph/Makefile)
- `make version` now uses build parameters located in dgraph/Makefile
- Replace $(BIN) with dgraph 

Feel free to make any other additions or subtractions...

## Remarks

We found interesting cross compilation issues in this process.  First, building a dgraph binary on an M1 Mac via 
```
GOOS=linux GOARCH=amd64 make dgraph
```
does NOT produce the same binary as on a linux/amd64 machine.  This can be seen in the following:

Artifact binary (made on linux/amd64 github runner):
```
> file dgraph
dgraph: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=62ab033f988e88edf53ec37e1d9363d34204890f, for GNU/Linux 3.2.0, not stripped
```
Compare with my locally made binary (M1 Mac):
```
> file dgraph/dgraph 
dgraph/dgraph: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=_SlrYO5M_ydJ_Wmfq2Qd/IxFWCCM1HR57u9G_UJxW/ojAnfYroSwaLqpInrpxt/w4QvLEaQMAzelTEmavC8, with debug_info, not stripped
```

If I download the artifact binary on my M1 machine and build a local Docker image,
 ```
docker build -f contrib/Dockerfile -t dgraph/dgraph:local .
```
(the Dockerfile file will pick up the binary), I will not be able to run the container:
```
> docker run dgraph/dgraph:local
qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
```

Furthermore, if I spin up a clean ec2 instance and scp the two dgraph binaries into the instance (i.e. the artifact binary, and my locally compiled binary), I will see that jemalloc is enabled on the artifact binary but is not enabled on my locally made binary, _even though at build time the jemalloc tag was specified in both cases._

Why is all of this happening?  The answer probably has to do with jemalloc and the fact that it is a C dependency.  Go code can be easily cross compiled which we take for granted,  whereas C code must be compiled appropriately for each environment.  This does not matter so much in production since everyone uses linux, but this is something to keep in mind as we run tests on our local machines (darwin/amd64, darwin/arm64, etc.).  It is also important that the binaries we produce are compiled in the appropriate environment.

### DIFF
--- a/.github/workflows/cd-dgraph.yml
+++ b/.github/workflows/cd-dgraph.yml
@@ -5,23 +5,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - name: Get Go Version
-        run: |
-          #!/bin/bash
-          GOVERSION=$({ [ -f .go-version ] && cat .go-version; })
-          echo "GOVERSION=$GOVERSION" >> $GITHUB_ENV
-      - name: Set up Go
-        uses: actions/setup-go@v3
-        with:
-          go-version: ${{ env.GOVERSION }}
-      - name: Install protobuf-compiler
-        run: sudo apt-get install -y protobuf-compiler
-      - name: Check protobuf
-        run: |
-          cd ./protos
-          go mod tidy
-          make regenerate
-          git diff --exit-code -- .
       - name: Set Dgraph Release Version
         run: |
           #!/bin/bash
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: .github/workflows/cd-dgraph.yml [hunk 1]
```diff
@@ -33,19 +45,29 @@ jobs:
...
+      - name: Get SHA
+        run: sha256sum dgraph/dgraph | cut -c-64 > dgraph/dgraph-linux-amd64.sha256
+      - name: Make Archive
+        run: tar -zcvf dgraph/dgraph-linux-amd64.tar.gz dgraph/dgraph
+      - uses: actions/upload-artifact@v3
+        with:
+          name: dgraph
+          path: |
+            dgraph/dgraph-linux-amd64.sha256
+            dgraph/dgraph-linux-amd64.tar.gz
```
skrdgraph, line 62: @joshua-goldstein - this is ported into the referenced PR

joshua-goldstein:
Closing because most changes here have already been merged into CD pipeline

Issue #8305, opened 2022-09-16
[FEATURE]: GraphQL Query Log.


MichelDiz:
### What you wanted to do.

Similar to this one https://github.com/dgraph-io/dgraph/issues/4944

I would like to get logs of queries executed at GraphQL application level.

### What you actually did.

N/A

### Why wasn't it great, with examples.

We have only logging for performed DQL. It would be nice to have GraphQL and mutations.

### Additional information.

https://dgraph.io/docs/deploy/log-format/#request-logging

PR #8388, opened 2022-10-25
refactor(GraphQL): Rename logRequest name and its resolver.

MichelDiz:
## Problem

In order to add GraphQL Query Log(#8305) we need to rename the current DQL query log resolver and config.

## Solution

> Rename: logRequest => logDQLRequest 

Important change:

```diff
--- glog.Infof("Got a query: %+v", req.req)
+++ glog.Infof("Got a DQL query: %+v", req.req)
```

### DIFF
--- a/edgraph/server.go
+++ b/edgraph/server.go
@@ -1188,8 +1188,8 @@ func (s *Server) doQuery(ctx context.Context, req *Request) (resp *api.Response,
 		return nil, serverOverloadErr
 	}
 
-	if bool(glog.V(3)) || worker.LogRequestEnabled() {
-		glog.Infof("Got a query: %+v", req.req)
+	if bool(glog.V(3)) || worker.LogDQLRequestEnabled() {
+		glog.Infof("Got a DQL query: %+v", req.req)
 	}
 
 	isGraphQL, _ := ctx.Value(IsGraphql).(bool)
--- a/graphql/admin/admin.go
+++ b/graphql/admin/admin.go
@@ -299,10 +299,10 @@ const (
 		cacheMb: Float
 
 		"""
-		True value of logRequest enables logging of all the requests coming to alphas.
-		False value of logRequest disables above.
+		True value of logDQLRequest enables logging of all the requests coming to alphas.
+		False value of logDQLRequest disables above.
 		"""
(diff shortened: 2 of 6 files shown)


skrdgraph:
I have reviewed this change and it looks okay. 

@matthewmcneely could you also review this please?

matthewmcneely:
@MichelDiz  So when logging is turned on via the /admin endpoint, how does the GraphQL query get logged? Is that already in the codebase?

MichelDiz:
@matthewmcneely Right now I'm first renaming this resolver, and then I'll implement the query log for GraphQL. There is no log for GraphQL yet. I'm thinking of putting it to Mutations too. To be able to view mutations in the logs optionally.

matthewmcneely:
@MichelDiz Ah OK, so this is just laying the foundation. One interesting aspect of when GraphQL logging is enabled is that you'll be able to view both the GraphQL and then the generated DQL query.

One thought related to that, you might consider tagging the log entry with a identifier that maps to the query context id, that way the GraphQL and generated DQL query could be compared.

matthewmcneely (approved):

skrdgraph:
I really like the above idea suggested by @matthewmcneely. The identifier facilitates poor many tracing in the logs. 

Thinking a little larger here (may be a dumb question), doesn't the tracing setup we have help us here? https://dgraph.io/docs/deploy/tracing/#setting-up-multiple-dgraph-clusters-with-jaeger ..

skrdgraph (approved):
This PR looks good. 👍

MichelDiz:
I will see if it will be necessary for an identifier. As I recall both queries (Graphql and DQL) appear at the same time. One pasted with the other at runtime. But I'll test it anyway.

@skrdgraph It can help the user to understand which query was, for example, the cause of a crash. But Jaeger can help identify at another atomic level. But this log will not appear in Jaeger. However, yes both together can be a powerful see the big picture. Jaeger does the predicate-level trace. The Query log would help you know what to look for in Jaeger for example.

BTW, Jaeger is a pretty cool tool. You can understand a lot on how Dgraph execute things. It shows each call in details. So you can have an idea of it.

PR #8391, opened 2022-10-25
fix(cd): Reduce frequency of test runs

adityasadalage:
## Problem
Our queue is increasing because the test run frequency is high (every 15 mins)

## Solution
Frequency changed from every 15 mins to every 1 hr


skrdgraph:
fixing https://github.com/dgraph-io/projects/issues/2

skrdgraph (approved):

PR #8365, opened 2022-10-16
fix(deps): update dependencies on badger + ristretto

joshua-goldstein:
## Problem

Dgraph currently uses Ristretto v0.1.0 and Badger v3.2103.1.  We encounter certain issues on arm64 machines that were patched in Ristretto v0.1.1 and Badger v3.2103.3.  Badger v3.2103.4 also fixes a potential manifest corruption issue.  

## Solution

We bump up the dependencies.  

```
> go get github.com/dgraph-io/ristretto@v0.1.1
go: upgraded github.com/dgraph-io/ristretto v0.1.0 => v0.1.1
go: upgraded golang.org/x/sys v0.0.0-20220728004956-3c1f35247d10 => v0.0.0-20221010170243-090e33056c14
> go get github.com/dgraph-io/badger/v3@v3.2103.4
go: downloading github.com/dgraph-io/badger/v3 v3.2103.4
go: module github.com/golang/protobuf is deprecated: Use the "google.golang.org/protobuf" module instead.
go: upgraded github.com/dgraph-io/badger/v3 v3.2103.1 => v3.2103.4
> go mod tidy                                    
```

## Remark

[Changelog of Badger v3.2103.1 to v3.2103.2](https://github.com/dgraph-io/badger/compare/v3.2103.1...v3.2103.2)
[Changelog of Badger v3.2103.2 to v3.2103.3](https://github.com/dgraph-io/badger/compare/v3.2103.2...v3.2103.3)
[Changelog of Badger v3.2103.2 to v3.2103.4](https://github.com/dgraph-io/badger/compare/v3.2103.1...v3.2103.2)
[Changelog of Ristretto v0.1.0 to v0.1.1](https://github.com/dgraph-io/ristretto/compare/v0.1.0...v0.1.1)

PR #8405, opened 2022-11-02
chore(ci): add ci workflow for arm64

joshua-goldstein:
## Problem

We would like to support arm64.  Toward that end we want to add a ci job that will run tests on an arm instance.  This is a WIP because some tests are failing (see remarks below)

## Solution
- Get the latest Ristretto and Badger versions that fixed certain arm64 issues.  This is also in a PR #8365 that has not been merged yet.
```
go get github.com/dgraph-io/ristretto@v0.1.1
go get github.com/dgraph-io/badger/v3@v3.2103.3
go mod tidy 
```
- Remove GOOS/GOARCH constraints in Makefile and CI workflow.

## Current test failures

These are all of the places we currently see failures.  Unfortunately the skip flag in t.go is unable to skip these particular tests in graphql.  All other tests are passing.

[graphql/e2e/normal](https://gist.github.com/joshua-goldstein/fcd4579ee7366a13b9e28eacad20a2b7)
[graphql/e2e/directives](https://gist.github.com/joshua-goldstein/a339d5a285c621aa3e4045f407a87804)
[graphql/e2e/custom_logic](https://gist.github.com/joshua-goldstein/48d91ac2969ddc29afd705d07186fb8b) 
[systest/export](https://gist.github.com/joshua-goldstein/39eaa83e277eb53101c7edc46aca0b83)

## Remarks

I was experimenting with a fresh arm64 instance and running tests on there (same or very similar environment to CI environment).  These look like nontrivial failures that need to be investigated.  

I have a script that does all the setup (Go, Docker, etc.) in a similar way to the CI environment but for arm (see [here](https://gist.github.com/joshua-goldstein/2c0a90192ed46a4145e3d2495e4e1c48)).

## Todo

Set up Github actions to run these jobs on an arm instance automatically, once we have a baseline.

### DIFF
--- a/.github/workflows/ci-dgraph-load-tests.yml
+++ b/.github/workflows/ci-dgraph-load-tests.yml
@@ -1,4 +1,4 @@
-name: ci-dgraph-load-tests
+name: ci-dgraph-load-tests-amd64
 on:
   push:
     branches:
--- a/.github/workflows/ci-dgraph-load-tests.yml
+++ b/.github/workflows/ci-dgraph-load-tests.yml
@@ -1,4 +1,4 @@
-name: ci-dgraph-load-tests-amd64
+name: ci-dgraph-load-tests
 on:
   push:
     branches:
--- a/.github/workflows/ci-dgraph-tests-arm64.yml
+++ b/.github/workflows/ci-dgraph-tests-arm64.yml
@@ -19,10 +19,15 @@ jobs:
     runs-on: [self-hosted, ARM64]
     steps:
       - uses: actions/checkout@v3
+      - name: Get Go Version
+        run: |
+          #!/bin/bash
+          GOVERSION=$({ [ -f .go-version ] && cat .go-version; })
+          echo "GOVERSION=$GOVERSION" >> $GITHUB_ENV
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.18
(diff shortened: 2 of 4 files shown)


joshua-goldstein:
Thanks @adityasadalage for taking a look at this, I was debating whether to reopen your PR and make these changes there but it seemed like a lot to push.  If you see anything here that can be modified please feel free to edit as necessary.

.github/workflows/ci-dgraph-load-tests.yml
```diff
@@ -16,7 +16,7 @@ on:
 jobs:
   dgraph-load-tests:
     if: github.event.pull_request.draft == false
-    runs-on: self-hosted
+    runs-on: [self-hosted, X64]
```
skrdgraph, line 19: small check is needed here - is this an `and` or `or`?

joshua-goldstein: AND, confirmed [here.]( https://docs.github.com/en/actions/hosting-your-own-runners/using-self-hosted-runners-in-a-workflow#using-default-labels-to-route-jobs).  However I need to change label to ARM64 which is the default label github actions uses.

skrdgraph: +1 ^ thanks!

PR #8406, opened 2022-11-02
Artifact fix to directly get dgraph binary from tar

dshekhar95:
## Problem
fix(release): Current dgraph tar from Artifact has a folder (dgraph) with binary inside. We want the tar to directly return the binary when opened
 

## Solution
Fix cd-dgraph.yml to point tar directly to dgraph binary

### DIFF
--- a/.github/workflows/cd-dgraph.yml
+++ b/.github/workflows/cd-dgraph.yml
@@ -39,14 +39,14 @@ jobs:
       - name: Make Linux Build
         run: make dgraph DGRAPH_VERSION=${{ env.DGRAPH_RELEASE_VERSION }}
       - name: Generate SHA for Linux Build
-        run: cd dgraph && sha256sum dgraph | cut -c-64 > dgraph-linux-amd64.sha256
+        run: cd dgraph && sha256sum dgraph | cut -c-64 > dgraph-checksum-linux-amd64.sha256
       - name: Tar Archive for Linux Build
         run: cd dgraph && tar -zcvf dgraph-linux-amd64.tar.gz dgraph
       - name: Upload Dgraph Binary Build Artifacts
         uses: actions/upload-artifact@v3
         with:
           path: |
-            dgraph/dgraph-linux-amd64.sha256
+            dgraph/dgraph-checksum-linux-amd64.sha256
             dgraph/dgraph-linux-amd64.tar.gz
       - name: Make Dgraph Docker Image
         run: |
--- a/.github/workflows/cd-dgraph.yml
+++ b/.github/workflows/cd-dgraph.yml
@@ -39,15 +39,15 @@ jobs:
       - name: Make Linux Build
         run: make dgraph DGRAPH_VERSION=${{ env.DGRAPH_RELEASE_VERSION }}
       - name: Generate SHA for Linux Build
-        run: sha256sum dgraph/dgraph | cut -c-64 > dgraph/dgraph-linux-amd64.sha256
+        run: cd dgraph && sha256sum dgraph | cut -c-64 > dgraph-linux-amd64.sha256 && mv dgraph-linux-amd64.sha256 ..
       - name: Tar Archive for Linux Build
-        run: tar -zcvf dgraph/dgraph-linux-amd64.tar.gz dgraph/dgraph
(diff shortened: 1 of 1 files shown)


dshekhar95:
https://github.com/dshekhar95/dgraph/actions/runs/3379275569

link to check artifact from my fork

MichelDiz (approved):

.github/workflows/cd-dgraph.yml
```diff
@@ -39,15 +39,15 @@ jobs:
       - name: Make Linux Build
         run: make dgraph DGRAPH_VERSION=${{ env.DGRAPH_RELEASE_VERSION }}
       - name: Generate SHA for Linux Build
-        run: sha256sum dgraph/dgraph | cut -c-64 > dgraph/dgraph-linux-amd64.sha256
+        run: cd dgraph && sha256sum dgraph | cut -c-64 > dgraph-linux-amd64.sha256 && mv dgraph-linux-amd64.sha256 ..
```
skrdgraph, line 42: I think you can remove the `&& mv dgraph-linux-amd64.sha256 ..`

reason: you can upload the created artifact from the original path `dgraph/dgraph-linux-amd64.sha256`

.github/workflows/cd-dgraph.yml
```diff
@@ -39,15 +39,15 @@ jobs:
...
       - name: Tar Archive for Linux Build
-        run: tar -zcvf dgraph/dgraph-linux-amd64.tar.gz dgraph/dgraph
+        run: cd dgraph && tar -zcvf dgraph-linux-amd64.tar.gz dgraph && mv dgraph-linux-amd64.tar.gz ..
```
skrdgraph, line 44: I think you can remove the `&& mv dgraph-linux-amd64.tar.gz ..`

reason: you can upload the created artifact from the original path `dgraph/dgraph-linux-amd64.tar.gz`

skrdgraph (commented):
@dshekhar95 I think we can minimize the `mv`'s as they will just cause confusion for the next engineer. We can keep it simple, and I think my comments do make it simple. Let me know your thoughts. If you agree, it will be good to have them fixed up in this PR.

.github/workflows/cd-dgraph.yml
```diff
@@ -39,15 +39,15 @@ jobs:
...
       - name: Upload Dgraph Binary Build Artifacts
         uses: actions/upload-artifact@v3
         with:
           path: |
-            dgraph/dgraph-linux-amd64.sha256
-            dgraph/dgraph-linux-amd64.tar.gz
+            dgraph-linux-amd64.sha256
+            dgraph-linux-amd64.tar.gz
```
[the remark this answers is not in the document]
dshekhar95: Hey Sudhish let me test that out on my fork. You are probably right but i mucked around with the bottom part first and ended up doing like this

.github/workflows/cd-dgraph.yml
```diff
@@ -39,9 +39,9 @@ jobs:
       - name: Make Linux Build
         run: make dgraph DGRAPH_VERSION=${{ env.DGRAPH_RELEASE_VERSION }}
       - name: Generate SHA for Linux Build
-        run: sha256sum dgraph/dgraph | cut -c-64 > dgraph/dgraph-linux-amd64.sha256
+        run: cd dgraph && sha256sum dgraph | cut -c-64 > dgraph-linux-amd64.sha256
```
skrdgraph, line 42: @dshekhar95 small request, I should have made this clear on the issue. I just realized this, so I am raising it on this PR. 

could you rename `dgraph-linux-amd64.sha256` to `dgraph-checksum-linux-amd64.sha256` please? This was a typo that had gotten in, from our release work. Apparently the old team was using `dgraph-checksum-*` and those references are made in other places which would break, if we don't make this fix. 

Could you squeeze that change in this PR itself? (again apologies for late feedback)

Also, you don't have to test that - I did look at the artifacts from your branch. It looks right. If you make a change on line#42, you will also have to change that in `upload artifact step`

skrdgraph (changes requested):
small request for renaming, left a comment (should have given that feedback earlier - apologies)

skrdgraph (approved):

skrdgraph:
Thanks @dshekhar95

PR #8413, opened 2022-11-04
chore(ci): fix breaking tests

joshua-goldstein:
## Problem

Since we added arm64 runners, they are being picked up by our workflows.  Tests are failing after about 1 minute with exec format error (due to documented Badger/Ristretto issue).  

For more info on runner labels [see here.](https://docs.github.com/en/actions/hosting-your-own-runners/using-self-hosted-runners-in-a-workflow#using-default-labels-to-route-jobs)

## Solution

We modify workflows to only pickup amd runners.  We also pin the Go version.

### DIFF
--- a/.github/workflows/ci-dgraph-load-tests.yml
+++ b/.github/workflows/ci-dgraph-load-tests.yml
@@ -1,4 +1,4 @@
-name: ci-dgraph-load-tests-amd64
+name: ci-dgraph-load-tests
 on:
   push:
     branches:
--- a/.github/workflows/ci-dgraph-load-tests.yml
+++ b/.github/workflows/ci-dgraph-load-tests.yml
@@ -1,4 +1,4 @@
-name: ci-dgraph-load-tests
+name: ci-dgraph-load-tests-amd64
 on:
   push:
     branches:
@@ -16,13 +16,18 @@ on:
 jobs:
   dgraph-load-tests:
     if: github.event.pull_request.draft == false
-    runs-on: self-hosted
+    runs-on: [self-hosted, x64]
     steps:
       - uses: actions/checkout@v3
+      - name: Get Go Version
+        run: |
+          #!/bin/bash
+          GOVERSION=$({ [ -f .go-version ] && cat .go-version; })
+          echo "GOVERSION=$GOVERSION" >> $GITHUB_ENV
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.18
+          go-version: ${{ env.GOVERSION }}
       - name: Set up Node
         uses: actions/setup-node@v3
         with:
--- a/.github/workflows/ci-golang-lint.yml
+++ b/.github/workflows/ci-golang-lint.yml
@@ -33,7 +33,7 @@ jobs:
         env:
           # prevent OOM
           GOGC: 10
-        uses: golangci/golangci-lint-action@latest
+        uses: golangci/golangci-lint-action@v3
         with:
           # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
           version: latest
--- a/.github/workflows/ci-golang-lint.yml
+++ b/.github/workflows/ci-golang-lint.yml
@@ -33,7 +33,7 @@ jobs:
         env:
           # prevent OOM
           GOGC: 10
-        uses: golangci/golangci-lint-action@v3
+        uses: golangci/golangci-lint-action
         with:
           # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
           version: latest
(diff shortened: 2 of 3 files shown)


joshua-goldstein:
@skrdgraph  This should fix the issue of failing workflows.

.github/workflows/ci-dgraph-load-tests.yml
```diff
@@ -16,13 +16,18 @@ on:
...
     if: github.event.pull_request.draft == false
-    runs-on: self-hosted
+    runs-on: [self-hosted, x64]
     steps:
       - uses: actions/checkout@v3
+      - name: Get Go Version
+        run: |
+          #!/bin/bash
+          GOVERSION=$({ [ -f .go-version ] && cat .go-version; })
+          echo "GOVERSION=$GOVERSION" >> $GITHUB_ENV
```
skrdgraph, line 26: +1 ^ thanks for adding this, could you also do this for `go-lint`?

.github/workflows/ci-dgraph-load-tests.yml
```diff
@@ -1,4 +1,4 @@
-name: ci-dgraph-load-tests
+name: ci-dgraph-load-tests-amd64
```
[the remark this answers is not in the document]
joshua-goldstein: @skrdgraph yeah I wasn't sure on this point - I just reverted the change

.github/workflows/ci-dgraph-load-tests.yml
joshua-goldstein: @skrdgraph Sure, previously the linter yaml was skipping the Go installation step entirely with `skip-go-installation: true` .  I just pushed changes that model the linter after the [Badger linter setup](https://github.com/dgraph-io/badger/blob/main/.github/workflows/ci-golang-lint.yml) (hence the version bumps).  Let me know if that looks okay.

.github/workflows/ci-golang-lint.yml
```diff
@@ -19,15 +19,23 @@ jobs:
...
+      - name: Setup Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GOVERSION }}
       - name: golang-lint
         env:
           # prevent OOM
           GOGC: 10
-        uses: golangci/golangci-lint-action@v2
+        uses: golangci/golangci-lint-action@v3.2.0
```
skrdgraph, line 36: should we just do @latest here?

.github/workflows/ci-golang-lint.yml
```diff
@@ -19,15 +19,23 @@ jobs:
...
         with:
           # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
-          version: v1.36
+          version: v1.48
```
skrdgraph, line 39: also should we just do latest here?

joshua-goldstein: latest can work here - currently latest is [v1.50.1](https://github.com/golangci/golangci-lint/releases) so not a big bump

.github/workflows/ci-dgraph-load-tests.yml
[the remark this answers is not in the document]
adityasadalage: Yeah ci-dgraph-tests-arm64.yml pipeline was given the new name. Would that need to be reverted to accommodate the matrix change later on?

.github/workflows/ci-golang-lint.yml
skrdgraph: it's easier if we pin to latest. We shouldn't be doing `latest` pins on core code though.

The other thing is there may be some `go version` <> `go-lang-ci version` compatibility issues.

.github/workflows/ci-golang-lint.yml
skrdgraph: if it does break, we would know something changed in the action side. Times when this would break is when the underlying schema would change things the way this action is invoked or some new fields coming in. 

Irrespective it's good to do latest here as well IMO

joshua-goldstein: Makes sense!   I pinned it to @ latest

.github/workflows/ci-golang-lint.yml
joshua-goldstein: Also pinned to latest.  👍

.github/workflows/ci-golang-lint.yml
joshua-goldstein: Interesting, @ latest is not possible ([see the failed run here](https://github.com/dgraph-io/dgraph/actions/runs/3396771636)).  But @ v3 works and will pull the latest minor versions

skrdgraph: they do have a `latest` tag https://github.com/golangci/golangci-lint-action/releases <- here

skrdgraph: you can probably just do `golangci/golangci-lint-action` without a version, this is still a guess

skrdgraph (approved):
👍

PR #8422, opened 2022-11-10
fix(cd) badger artifacts

skrdgraph:
## Problem
badger artifacts should be released with dgraph artifacts

## Solution
update the cd script to use the generated artifacts from badger project with correct version picks

### DIFF
--- a/.github/workflows/cd-dgraph.yml
+++ b/.github/workflows/cd-dgraph.yml
@@ -26,7 +26,7 @@ jobs:
         run: |
           #!/bin/bash
           BADGER_RELEASE_VERSION=$(cat go.mod | grep -i "github.com/dgraph-io/badger" | awk '{print $2}')
-          echo "making a new release for badger "$BADGER_RELEASE_VERSION
+          echo "setting badger version "$BADGER_RELEASE_VERSION
           echo "BADGER_RELEASE_VERSION=$BADGER_RELEASE_VERSION" >> $GITHUB_ENV
       - name: Download Badger Artifacts
         run: |
--- a/.github/workflows/cd-dgraph.yml
+++ b/.github/workflows/cd-dgraph.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Set Badger Release Version
         run: |
           #!/bin/bash
-          BADGER_RELEASE_VERSION=$(cat go.mod | grep -i badger | awk '{print $2}')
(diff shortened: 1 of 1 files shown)


skrdgraph:
validated it here: https://github.com/skrdgraph/dgraph/actions/runs/3435108039

.github/workflows/cd-dgraph.yml
```diff
@@ -22,6 +22,19 @@ jobs:
           go mod tidy
           make regenerate
           git diff --exit-code -- .
+      - name: Set Badger Release Version
+        run: |
+          #!/bin/bash
+          BADGER_RELEASE_VERSION=$(cat go.mod | grep -i badger | awk '{print $2}')
```
joshua-goldstein, line 28: Can use `go list -f '{{.Version}}' -m github.com/dgraph-io/badger/v3`

joshua-goldstein:
Quick comment before I forget (we will discuss more later): since this change exists only on the CD side, if someone is building Dgraph locally they would be required to pull the correct version of the Badger CLI.  E.g. after running `make install` to fetch Dgraph, one would have to run `go install github.com/dgraph-io/badger/v3@CORRECT_VERSION`.

.github/workflows/cd-dgraph.yml
skrdgraph: I was thinking about this ^, I think we really need to have a talk here ^. I did not like pinning to `github.com/dgraph-io/badger/v3` where we specifically choose v3. We wanted to make semver changes to badger repo, and possibly cut 4.0.0. This may complicate that process.  

Also, when I tried `github.com/dgraph-io/badger` <- it spat out a v 1.x* version.

joshua-goldstein: Addressing the latter point: Go considers `github.com/dgraph-io/badger` and `github.com/dgraph-io/badger/v3` to be two totally distinct modules (and hence they have two distinct import paths).  The expectation is that whenever a major version is released, dependent projects must modify the import path of the dependency to point to the new module (so it is not simply an upgrade of an existing module).

I do see your point - making a major version upgrade on Badger would require remembering to modify the Makefile, which is not great.

skrdgraph: yeah +1 ^ same issues noticed, lets talk and make a decision today. The nice part is we have 2 working approaches. A choice is to be made.

joshua-goldstein (approved):
Can change "making new release for Badger" to "fetching Badger version number..." or similar.  Otherwise looks good.

PR #8429, opened 2022-11-13
feat(DQL): Rename gql package to dql

MichelDiz:
## Problem

GraphQL Spec uses the acronym GQL to reference its language in its packages. Dgraph used this acronym in the past because the language was called "GraphQL+-". But today as the name has been updated and we have native support for GraphQL. This becomes a problem as it confuses package names.

## Solution

Rename GQL package to DQL


matthewmcneely (approved):
LGTM... A welcomed change!

.gitignore
```diff
@@ -9,6 +9,10 @@ gql/gql-fuzz.zip
 gql/fuzz-data/crashers
 gql/fuzz-data/suppressions
```
joshua-goldstein, line 10: here too?

matthewmcneely:
This will likely *demolish* many in-process PR's. I'd recommend this gets merged into main immediately after we cut the next release (when more PRs will be settled)

skrdgraph:
> This will likely _demolish_ many in-process PR's. I'd recommend this gets merged into main immediately after we cut the next release (when more PRs will be settled)

+1 ^ re-iterating Matthew's message

PR #8430, opened 2022-11-14
fix(chore): add the word "dev" to the version to avoid breaking the local test

MichelDiz:
## Problem

When you debug locally, without going through the build process. Dgraph doesn't set a default version. It returns empty. And that breaks Ratel and possibly some test that depends on the version information.

## Solution
 
 Check for missing version and put "dev" in place if it returns null.

### DIFF
--- a/edgraph/server.go
+++ b/edgraph/server.go
@@ -148,12 +148,12 @@ func GetGQLSchema(namespace uint64) (uid, graphQLSchema string, err error) {
 	resp, err := (&Server{}).Query(ctx,
 		&api.Request{
 			Query: `
-			query {
-			  ExistingGQLSchema(func: has(dgraph.graphql.schema)) {
-				uid
-				dgraph.graphql.schema
-			  }
-			}`})
+			 query {
+			   ExistingGQLSchema(func: has(dgraph.graphql.schema)) {
+				 uid
+				 dgraph.graphql.schema
+			   }
+			 }`})
 	if err != nil {
 		return "", "", err
 	}
@@ -719,7 +719,7 @@ func buildUpsertQuery(qc *queryContext) string {
 			//      * be empty if the condition is true
 			//      * have 1 UID (the 0 UID) if the condition is false
 			upsertQuery += qc.condVars[i] + ` as var(func: uid(0)) ` + cond + `
-			 `
+			  `
 		}
 	}
 	upsertQuery += `}`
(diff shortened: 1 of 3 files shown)


joshua-goldstein:
How are you building Dgraph locally?  It should be done using our Makefiles.  In Dgraph/Makefile, the version of Dgraph is passed as an LDFLAG to the `go build` command when making the binary.  By default (if no version is specified), the version is now "local".

MichelDiz:
When I run the debug it runs the code live. There's no building process there. I'm not building. It is similar to "go run main.go". And we need to simulate a real build with a live code.

nswamy (approved):
LGTM

x/init.go
```diff
@@ -65,6 +65,10 @@ func Init() {
 	// TODO: why is this here?
 	// Config.QueryEdgeLimit = 1e6
 
+	if dgraphVersion == "" {
+		dgraphVersion = "dev"
```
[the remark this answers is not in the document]
MichelDiz: I don't know if it would be a conflict. I think we're doing the same thing, just for different purposes.

If there is a value. My change won't do anything. So DevVersion() won't break. It will add "dev" only for unforeseen cases.

The DevVersion() function expects some value to exist. This is probably a specific scenario. For example, "rc" releases. It's in a "dev" version. But my goal is to run main.go directly(other scenario). This function does not take this into account. And I wouldn't want to modify DevVersion(). Would have to add switch case logic. Unless you really think I should do this.

Also, this function just verify if it is a dev version release. It doesn't inject values in dgraphVersion variable. It is used just for sentry contextual purposes.

Again, same idea but totally different purposes.

Also, there is an internal old discussion about remove sentry completely. I think it was generating panics.

Cheers.

edgraph/server.go
```diff
@@ -149,11 +149,11 @@ func GetGQLSchema(namespace uint64) (uid, graphQLSchema string, err error) {
...
-			  ExistingGQLSchema(func: has(dgraph.graphql.schema)) {
-				uid
-				dgraph.graphql.schema
-			  }
-			}`})
+				ExistingGQLSchema(func: has(dgraph.graphql.schema)) {
+					uid
+					dgraph.graphql.schema
+				  }
+				}`})
```
skrdgraph, line 156: I don't see any code change happening here, do you want to roll this back? I guess your IDE settings may be doing this. We will have a larger PR to fix these aesthetics - I don't think these changes should go in as part of this.

MichelDiz: This is the golang fmt. It is run automatically by the VsCode. I prefer not to fight it. After all, it follows the standards recommended by the Golang maintainers.

matthewmcneely (approved):

edgraph/server.go
MichelDiz: PS. Let me know your call.

PR #8432, opened 2022-11-14
Fix(lsbackup): Fix profiler in lsBackup (#7729)

dshekhar95:
cherry pick from #7729

## Problem
wrong config set for profiler (set to Restore instead of Backup)

## Solution
set profiler to Backup.conf

### DIFF
--- a/ee/backup/run.go
+++ b/ee/backup/run.go
@@ -1,4 +1,3 @@
-//go:build !oss
 // +build !oss
 
 /*
--- a/ee/backup/run.go
+++ b/ee/backup/run.go
@@ -41,9 +41,6 @@ import (
 	"google.golang.org/grpc"
 )
 
-// Restore is the sub-command used to restore a backup.
-var Restore x.SubCommand
-
 // LsBackup is the sub-command used to list the backups in a folder.
 var LsBackup x.SubCommand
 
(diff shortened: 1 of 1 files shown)


meghalims (dismissed):
Hi Dilip, as we have discussed, please add unit test for backup/run.go.

ee/backup/run.go
```diff
@@ -42,6 +43,7 @@ import (
 )
 
 // Restore is the sub-command used to restore a backup.
+
```
akon-dey, line 46: This is an unnecessary change.

ee/backup/run.go
```diff
@@ -162,7 +164,7 @@ func initBackupLs() {
 		Short: "List info on backups in a given location",
 		Args:  cobra.NoArgs,
 		Run: func(cmd *cobra.Command, args []string) {
-			defer x.StartProfile(Restore.Conf).Stop()
+			defer x.StartProfile(LsBackup.Conf).Stop()
```
akon-dey, line 167: LGTM

ee/backup/run.go
dshekhar95: @akon-dey the var Restore x.SubCommand on line 47 is used in other part of the code. So I had to re-introduce to fix the errors

dshekhar95: looks like github isnt showing the line 47 addition unless you click view changes, not sure if you missed that

dshekhar95:
> 

This will need an backup integration test

anurags92:
Please add some details about the change in the PR description.

matthewmcneely (approved):

dshekhar95 (commented):
fixed

meghalims (approved):
Looks good. Please merge

damonfeldman:
This fix allows profiling of the Backup operation to work properly if certain profilng flags are specified in the configuration. Previously, the system was profiling restore instead of backup.

PR #8437, opened 2022-11-16
fix(cmd/debug): Print banned namespaces correctly. (#7929)

SashaDgraph:
Earlier we were showing the opposite status: unbanned namespaces were shown as banned. This change fixes that.

(cherry picked from commit 43a1bd0b09f157144a075ab872c7faae4b28abb4)

Co-authored-by: Daniel Mai <daniel@dgraph.io>

### DIFF
--- a/dgraph/cmd/debug/run.go
+++ b/dgraph/cmd/debug/run.go
@@ -38,6 +38,7 @@ import (
 	"github.com/dgraph-io/badger/v3"
 	bpb "github.com/dgraph-io/badger/v3/pb"
 	"github.com/dgraph-io/ristretto/z"
+	humanize "github.com/dustin/go-humanize"
 
 	"github.com/dgraph-io/dgraph/codec"
 	"github.com/dgraph-io/dgraph/ee"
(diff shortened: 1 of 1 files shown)


matthewmcneely:
@SashaDgraph This PR differs from https://github.com/dgraph-io/dgraph/pull/7929/files. Can you elaborate?

SashaDgraph:
> @SashaDgraph This PR differs from https://github.com/dgraph-io/dgraph/pull/7929/files. Can you elaborate?

Let me look at it.

MichelDiz:
@SashaDgraph something is wrong here. I see no file changes. And the commit https://github.com/dgraph-io/dgraph/commit/43a1bd0b09f157144a075ab872c7faae4b28abb4 happens after `func printSummary` is added to the code. Currently in main there's no printSummary function at all. So, how you gonna change `!ok {` to `ok {`? if there is none?

This cherry pick is compromised.

SashaDgraph:
Screenshot 2022-11-21 at 5 12 31 PM
@MichelDiz @matthewmcneely attached is the screenshot of the conflict when cherry-picking the PR. Line 879-880 show that the entire function printSummary is currently not in the code. So if we choose this PR, we would have to choose the entire function.

MichelDiz:
@SashaDgraph exactly. We need the history of that line. Cuz if you add just that single func, it will be a "dangling" code. Used nowhere, right? We need to seed the history. Find its related code, where it's being used. And bring it all to a single PR. Gotta be careful with those Cherry-picks. As they can introduce null codes. With no effect.

matthewmcneely:
@SashaDgraph, @MichelDiz, @meghalims Actually, I think this cherry pick should not have been selected for this release. It's a change to a function that only exists downstream toward v21.12.

meghalims:
> @SashaDgraph, @MichelDiz, @meghalims Actually, I think this cherry pick should not have been selected for this release. It's a change to a function that only exists downstream toward v21.12.

I Agree Matthew. That was the agreement we had when we looked at the conflict on Monday.

meghalims:
After reviewing this cherry-pick, it turns out to be a change to a function that only exists in v21.12. So, scoping it out of 22.0.2 and closing this PR

PR #8441, opened 2022-11-19
fix(sort): Only filter out nodes with positive offsets. (#8077)

joshua-goldstein:
Cherry pick from #8077.

Steps to reproduce:
```
make image-local
cd contrib/local-test && make up
make schema-dql
make load-data-dql-rdf
make query-dql
curl: (52) Empty reply from server
make: *** [query-dql] Error 52
<alpha will crash and restart>
```

with the following minimal schema / data / query (should be placed in contrib/local-test):

schema.dql:
```
type Person {
    name
    age
    friend
}

name: string @index(term)  .
age: int @index(int) .
friend: [uid] @count .
```

dql-data.rdf
```
{
  set {
    _:michael <name> "Michael" .
    _:michael <dgraph.type> "Person" .
    _:michael <age> "314" .
    _:michael <friend> _:alice .
    _:michael <friend> _:bob .
    _:michael <friend> _:charlie .

    _:alice <name> "Alice" .
    _:alice <dgraph.type> "Person" .
    _:alice <age> "1" .

    _:bob <name> "Bob" .
    _:bob <dgraph.type> "Person" .
    _:bob <age> "2" .

    _:charlie <name> "Charlie" .
    _:charlie <dgraph.type> "Person" .
  }
}
```

query.dql
```
{
  michael_friends_first(func: allofterms(name, "Michael")) {
    name
    age
    friend (orderasc: age, offset: -1) {
      name
      age
    }
  }
}
```

### Remarks

- The cherry-pick does resolve the above crash
- This issue exists in main
- Why do we allow negative offsets?  If they are not useful, we should return an error when a user attempt to query using a negative offset.

### DIFF
--- a/query/query0_test.go
+++ b/query/query0_test.go
@@ -550,14 +550,14 @@ func TestCascadeWithSort(t *testing.T) {
 func TestNegativeOffset(t *testing.T) {
 	query := `
 	{
-		me(func: type(Person2), offset: -1, orderasc: name2) {
+		me(func: type(Person2), offset: -1, orderasc: age2) {
 			name2
 			age2
 		}
 	}
 	`
 	js := processQueryNoErr(t, query)
-	require.JSONEq(t, `{"data":{"me":[{"name2":"Alice","age2":10},{"age2":20}]}} `, js)
+	require.JSONEq(t, `{"data":{"me":[{"age2":20},{"name2":"Alice"}]}}`, js)
 }
 
 func TestLevelBasedFacetVarAggSum(t *testing.T) {
--- a/query/query0_test.go
+++ b/query/query0_test.go
@@ -543,7 +543,7 @@ func TestCascadeWithSort(t *testing.T) {
 	}
 	`
 	js := processQueryNoErr(t, query)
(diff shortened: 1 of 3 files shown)


query/query0_test.go
```diff
@@ -546,6 +546,20 @@ func TestCascadeWithSort(t *testing.T) {
 	require.JSONEq(t, `{"data":{"me":[{"name": "Daryl Dixon","alive": false},{"name": "Rick Grimes","alive": true}]}}`, js)
 }
 
+// Resolved in PR #8441
```
matthewmcneely, line 549: Other tests in the past have given a full link to the issue, i.e.

```
// Regression test for https://github.com/dgraph-io/dgraph/issues/3657.
```

meghalims (approved):
Approving this Joshua. Please go ahead and merge

MichelDiz (approved):

damonfeldman:
Note this fixes errors reported as ```panic: runtime error: slice bounds out of range [-4:]``` where a negative offset is used, a predicate is indexed, and nodes are included that do not have the predicate. It is questionable why or when negative offsets are useful, but this can happen.

PR #8446, opened 2022-11-21
chore(test)- Added tests for S3 backup and restore (full as well as i…

SiddeshUbale:
…ncremental)


systest/backup/s3-backup/backup_test.go
```diff
@@ -0,0 +1,496 @@
...
+	fmt.Fprintf(os.Stderr, msg+"\n", args...)
+	os.Exit(1)
+}
+
+func getEnvs() [4]string {
+
+	S3_ACCESS_KEY := os.Getenv("S3_ACCESS_KEY")
+	S3_SECRET_KEY := os.Getenv("S3_SECRET_KEY")
+	S3_BUCKET_NAME := os.Getenv("S3_BUCKET_NAME")
+	S3_REGION := os.Getenv("S3_REGION")
```
joshua-goldstein, line 485: Not sure if it is best practice to keep S3 secrets stored in environment variables (e.g. [ref](https://blog.diogomonica.com/2017/03/27/why-you-shouldnt-use-env-variables-for-secret-data/)).  This information can be easily leaked.  Also I don't see where this information is being added to the environment.  @darkn3rd has more expertise in this area.

PR #8447, opened 2022-11-21
chore(vlog): Fix outdated doc comments on RunVlogGC

kevinmingtarja:
## Problem
This [commit](https://github.com/dgraph-io/dgraph/commit/786ff4af79802b146e8d67ba8ed4c009dd0a7f16) changes the behavior of `x/x.go/RunVlogGC` to run **unconditionally every minute**. But the author forgot to update the comments on the function to reflect that change. 

The doc comments still says it `"runs GC unconditionally after every 10 minutes. Additionally it also runs GC if vLogSize has grown more than 1 GB in last minute"`, so it can be confusing for readers.

## Solution
 Fix the comments to reflect the latest behavior, which is that it runs **uncoditionally every minute**

### DIFF
--- a/x/x.go
+++ b/x/x.go
@@ -1135,12 +1135,11 @@ func IsGuardian(groups []string) bool {
 	return false
 }
 
-// RunVlogGC runs value log gc on store. It runs GC unconditionally after every 10 minutes.
-// Additionally it also runs GC if vLogSize has grown more than 1 GB in last minute.
+// RunVlogGC runs value log gc on store. It runs GC unconditionally after every 1 minute.
 func RunVlogGC(store *badger.DB, closer *z.Closer) {
 	defer closer.Done()
 
-	// Runs every 1m, checks size of vlog and runs GC conditionally.
+	// Runs every 1m
 	ticker := time.NewTicker(1 * time.Minute)
 	defer ticker.Stop()
 


x/x.go
```diff
@@ -1135,12 +1135,11 @@ func IsGuardian(groups []string) bool {
 	return false
 }
 
-// RunVlogGC runs value log gc on store. It runs GC unconditionally after every 10 minutes.
-// Additionally it also runs GC if vLogSize has grown more than 1 GB in last minute.
+// RunVlogGC runs value log gc on store. It runs GC unconditionally after every 1 minute.
 func RunVlogGC(store *badger.DB, closer *z.Closer) {
 	defer closer.Done()
 
-	// Runs every 1m, checks size of vlog and runs GC conditionally.
```
joshua-goldstein, line 1143: Conditionally on what?  Can be good to make that clear.  I believe is the discard ratio (looks like 0.7 in `err = store.RunValueLogGC(0.7)` below).  We can make a quick note that this is a number we chose based on performance considerations.

kevinmingtarja: It no longer runs conditionally now. And the discard ratio is well documented in the `badger` code. 

For the discard ratio number, it's quite well documented in the `badger` code, and we don't really know how the old team picked `0.7` per se. We only know that on paper, setting it to 0.7 means a file will be rewritten if at least 70% of the space can be discarded. This means fewer space reclaims will happen which means less background activity. So I think it wouldn't add much value as of now if we add it to the comments.

joshua-goldstein: I am sorry, I completely missed that your comment says "unconditionally" 😄 so this looks good.

joshua-goldstein (approved):
Looks good 👍

x/x.go
kevinmingtarja: haha no worries Josh, but I do think we need to benchmark that `0.7` discard ratio "magic number" some time in the future, see if it really is the optimal value for our ideal case workload.

x/x.go
```diff
@@ -1135,12 +1135,11 @@ func IsGuardian(groups []string) bool {
...
+	// Runs every 1m
 	ticker := time.NewTicker(1 * time.Minute)
 	defer ticker.Stop()
```
mangalaman93, line 1145: Do you see why this if is present in the code below?
```
if abs(lastSz, sz) > 512<<20 {
```

kevinmingtarja: That's a good question. Actually I'm not sure as well. It will certainly reduce the number of logs it outputs, but for what purpose and why 512MiB in particular I don't know. 

This was the commit that introduced that block of code https://github.com/dgraph-io/dgraph/pull/6889/commits/356f246422f261d3d3b7bda54a74ebd3adc83025.

mangalaman93: seems like it just could be for debugging.

matthewmcneely (approved):

PR #7962, opened 2021-07-28
fix(query): Do not execute filters if there are no source uids

ahsanbarkati:
No need to execute filter subgraph if there are no source UIDs.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7962)

PR #8452, opened 2022-11-23
fix(query): Do not execute filters if there are no source uids (#7962…

all-seeing-code:
No need to execute filter subgraph if there are no source UIDs.

(cherry picked from commit 849b587e101100059287e15fea0bf0be689904ea)

## Description
The original PR adds a fail fast methodology, where it doesn't expand on a filter subgraph if the srcUid is nil. The current cherry-pick PR brings that change in and adds a test case that scenario.

## Test
`TestFilterWithNoSrcUid`

### DIFF
--- a/query/query.go
+++ b/query/query.go
@@ -230,25 +230,27 @@ type Function struct {
 // SubGraph is the way to represent data. It contains both the request parameters and the response.
 // Once generated, this can then be encoded to other client convenient formats, like GraphQL / JSON.
 // SubGraphs are recursively nested. Each SubGraph contain the following:
-// * SrcUIDS: A list of UIDs that were sent to this query. If this subgraph is a child graph, then the
-//            DestUIDs of the parent must match the SrcUIDs of the children.
-// * DestUIDs: A list of UIDs for which there can be output found in the Children field
-// * Children: A list of child results for this query
-// * valueMatrix: A list of values, against a single attribute, such as name (for a scalar subgraph).
-//                This must be the same length as the SrcUIDs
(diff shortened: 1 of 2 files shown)


MichelDiz (approved):

query/query.go
```diff
@@ -230,25 +230,27 @@ type Function struct {
 // SubGraph is the way to represent data. It contains both the request parameters and the response.
 // Once generated, this can then be encoded to other client convenient formats, like GraphQL / JSON.
 // SubGraphs are recursively nested. Each SubGraph contain the following:
-// * SrcUIDS: A list of UIDs that were sent to this query. If this subgraph is a child graph, then the
```
mangalaman93, line 233: Are we changing anything significant in the comments?

query/query.go
```diff
@@ -2210,6 +2212,10 @@ func ProcessGraph(ctx context.Context, sg, parent *SubGraph, rch chan error) {
 			}
 
 			filter.SrcUIDs = sg.DestUIDs
+			if len(filter.SrcUIDs.Uids) == 0 {
```
mangalaman93, line 2215: Is there an existing test for this scenario?

query/query.go
anurags92: No, we are not. But these are go fmt changes which can be safely accepted.

query/query.go
anurags92: I added a test `TestFilterWithNoSrcUid` here.

nswamy (approved):
LGTM. Thanks for bringing in the change and adding the test.

PR #8454, opened 2022-11-24
fix(acl): filter out the results based on type Add a test case to ver…

anurags92:
Adds a test case to check fixes in #7978


nswamy (approved):
Thank you for adding the test
LGTM

matthewmcneely (approved):

PR #8458, opened 2022-11-27
chore(front): Add new message in front dashboard

MichelDiz:
## Problem

 The text was out of date. There's no dgraph-ratel
 
## Context

This text will appear at http://localhost:8080/ It is a text that helps to know if the Dgraph is running and healthy.

## Solution
 
Change the text and add some more contextual text.

```

Dgraph is up and running...

Now you can access your cluster via Ratel Dashboard. 
Remember, the PORT to use there to connect to Alpha is 8080 + offset if the case.
if you are running GraphQL you can use any GraphQL Client from the open source community.

Thank you for using Dgraph!

Important Links:

https://dgraph.io/docs
https://discuss.dgraph.io/
https://cloud.dgraph.io/


Copyright 2016-2022 Dgraph Labs, Inc. and Contributors.
```


dgraph/cmd/alpha/dashboard.go
```diff
@@ -39,8 +39,8 @@ func homeHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	dgraph := "\nDgraph is up and running...\n\n"
-	dgraph = dgraph + "Now you can access your cluster via Ratel Dashboard. Remember, the PORT to use there to connect to Alpha is 8080 + offset if the case.\n"
+	dgraph := "\nDgraph is up and running...\n\n" // swiftlint:disable:next line_length
+	dgraph = dgraph + `Now you can access your cluster via Ratel Dashboard.Remember, the PORT to use there to connect to Alpha is 8080 + offset if the case.\n`
```
matthewmcneely, line 43: This all should be refactored into one statement. For instance:

```golang
dgraph := `Dgraph is up and running...

Now you can access...
`

```

MichelDiz: Yeah, I know. But when printing it adds spaces. And it looks messy visually.

MichelDiz: I have to do something like this
```
	dgraph = dgraph + `Now you can access your cluster via Ratel Dashboard.
Remember, the PORT to use there to connect to Alpha is 8080 + offset if the case.\n`
```

matthewmcneely: Have a look at this: https://go.dev/play/p/vAqBiZZ0tYT

matthewmcneely: Also, I don't think you should call out what port to use. It might have been mapped to some other port. You could examine the http.Request.Host parameter and echo that back to the user. Or, I'd say leave it out entirely.

MichelDiz: I'm not in favor of using backtick here. Cuz if I need a concise spacing, distancing. I can't. If it would be HTML/CSS. Would be nice. But this is go printing.

I finally have found a nice way of doing it.

Thank you Matthew. Your comments helped me to find another way hehehe

About the port. It is just a tipo. Anyone who changes it, should be at a level that this tip is not necessary. Right? If you by conscience change the offset of the port. You know what you are doing.

matthewmcneely: Did you try running that go.dev link that I supplied? It formats everything as it is formatted between the ticks. I'm pretty sure it'll format the same way to the Writer and onto to the browser window. Are you seeing something different?

I can't get on board with adding extra machine instructions to simply concat a static string.

MichelDiz: Yes. It does. it is "the same". More clean in a way - that was my first attempt. But the lines need to be aligned to left in the code. And if you need to add more lines. You will place literal lines inside the block. With concat lines with quoted strings the fmt adds 2 tabs. It gets more "compact".

Both ways result in the same result. But I'd prefer without backticks.

matthewmcneely: So a good practice regarding "what style should I use?" when something like this comes up is to look at the go standard library and see what's done there. Here's one of many places there where long string literals are encoded as I recommended: https://github.com/golang/go/blob/bb917bd1b212dc8fff3852fa164667cd06b9f653/src/cmd/vendor/github.com/google/pprof/internal/driver/cli.go#L304

matthewmcneely: OK fine. But the "if" should be capitalized.

MichelDiz: Done.

matthewmcneely:
Added @rderbier as reviewer

dgraph/cmd/alpha/dashboard.go
```diff
@@ -38,8 +38,18 @@ func homeHandler(w http.ResponseWriter, r *http.Request) {
...
-	x.Check2(w.Write([]byte(
-		"Dgraph browser is available for running separately using the dgraph-ratel binary")))
+
+	dgraph := "\nDgraph is up and running...\n\n" +
+		"Now you can access your cluster via Ratel Dashboard.\n" +
+		"Remember, the PORT to use there to connect to Alpha is 8080 + offset if the case.\n" +
+		"If you are running GraphQL you can use any GraphQL Client from the open source community.\n" +
+		"\nThank you for using Dgraph!\n" +
+		"\nImportant Links:\n\nhttps://dgraph.io/docs\nhttps://discuss.dgraph.io/\nhttps://cloud.dgraph.io/ " +
+		"\n\n\nCopyright 2016-2022 Dgraph Labs, Inc. and Contributors.\n"
```
joshua-goldstein, line 48: Just a nitpick, is there a reason we don't use a raw string literal here?  i.e.
```
`
Dgraph is up and running...

Now you can access...
Remember....
`
```

matthewmcneely: Ha. See https://github.com/dgraph-io/dgraph/pull/8458#discussion_r1036323692

MichelDiz: Okay, I dicide to use HTML/CSS. I've share a printscreen in Slack. 
Basically it will use backtick and this won't have any side effects that I was complaning about Plain Text rendering.

```
dgraph := `<body>
	<BR><h2>Dgraph is up and running...
```

PR #8466, opened 2022-11-29
Micheldiz/fmt state

MichelDiz:


mangalaman93:
We have updated the predicate structure to use hyphen, we probably do not need this change any more. Feel free to reopen the PR if that is not true.

PR #8467, opened 2022-11-29
fix(JoinCluster): Avoid retrying JoinCluster indefinitely (#7961)

dshekhar95:
If a Zero follower receives a JoinCluster request, it won't process the proposal because it can't forward the request to leader. Raft doesn't send an error back in this case. In the current code, it would retry indefinitely. With this change, we now use the request context to give up at some point.

### DIFF
--- a/conn/node.go
+++ b/conn/node.go
@@ -573,7 +573,7 @@ var errInternalRetry = errors.New("Retry proposal again")
 func (n *Node) proposeConfChange(ctx context.Context, conf raftpb.ConfChange) error {
 	if ctx.Err() != nil {
 		// If ctx has already errored out, return without proposing.
-		return errors.Wrapf(ctx.Err(), "while proposeConfChange")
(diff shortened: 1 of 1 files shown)


dshekhar95:
PR to run tests, this other PR is relevant: https://github.com/dgraph-io/dgraph/pull/2467

conn/node.go
```diff
@@ -571,7 +571,14 @@ func (n *Node) DeletePeer(pid uint64) {
 var errInternalRetry = errors.New("Retry proposal again")
 
 func (n *Node) proposeConfChange(ctx context.Context, conf raftpb.ConfChange) error {
-	cctx, cancel := context.WithTimeout(ctx, 3*time.Second)
+	if ctx.Err() != nil {
+		// If ctx has already errored out, return without proposing.
+		return errors.Wrapf(ctx.Err(), "while proposeConfChange")
```
matthewmcneely, line 576: This can just be `.Wrap( ` as there's no formatting done.

PR #8470, opened 2022-12-01
chore(tests): Unit test for edgraph/validateToken ()

jbhamra1:
## Problem
Missing unit tests for validateToken (), getAcessJwt (), getRefreshJwt () of edgraph package

## Solution
TestValidateToken (), TestGetAccessJwt () and TestRefreshJwt () generates a signed JSON Web token using the static data defined in the test file, using the helper function generateJWT (). The generated token includes the token expiry time which is passed as an argument to the helper function.
TestValidateToken () calls validateToken and extracts data from the signed token which is then compared to validate with the original data.

TestGetAccessJwt() and TestRefreshJwt () are direct wrappers over jwt.NewWithClaims () but with their predefined expiry duration.

### DIFF
--- a/edgraph/access_ee_test.go
+++ b/edgraph/access_ee_test.go
@@ -149,7 +149,10 @@ func TestGetAccessJwt(t *testing.T) {
 		 * One of the factor [signature] would differ is based on the expiry time.
 		 * This will differ for tokstr and jwtstr and hence ignoring it.
 		 */
-		compareTokenParts (tokstr, jwtstr, 1, 2 ,0) // compare 1st and 2nd part, ignore the 3rd.
+		match := compareTokenParts (tokstr, jwtstr, 1, 2 ,0) // compare 1st and 2nd part, ignore the 3rd.
+		if match == false {
+			t.Errorf ("Actual output is not equal to the expected output")
+		}
 	}
 }
 
@@ -169,6 +172,9 @@ func TestGetRefreshJwt(t *testing.T) {
 		 * One of the factor [signature] would differ is based on the expiry time.
 		 * This will differ for tokstr and jwtstr and hence ignoring it.
 		 */
-		compareTokenParts (tokstr, jwtstr, 1, 2 ,0) // compare 1st and 2nd part, ignore the 3rd.
+		match := compareTokenParts (tokstr, jwtstr, 1, 2 ,0) // compare 1st and 2nd part, ignore the 3rd.
+		if match == false {
+			t.Errorf ("Actual output is not equal to the expected output")
+		}
 	}
 }
--- a/edgraph/access_ee_test.go
+++ b/edgraph/access_ee_test.go
@@ -13,36 +13,34 @@ import (
 
 var (
 	userDataList = []userData{
-		userData{1234567890, "user1", []string{"701", "702"}},
-		userData{2345678901, "user2", []string{"703", "701"}},
-		userData{3456789012, "user3", []string{"702", "703"}},
+		{1234567890, "user1", []string{"701", "702"}},
+		{2345678901, "user2", []string{"703", "701"}},
+		{3456789012, "user3", []string{"702", "703"}},
 	}
 )
 
 func generateJWT(namespace uint64, userId string, groupIds []string, exp int64) (string, error) {
-	var e1 int64
 	var token *jwt.Token
 
-	if exp == 0 {
+	e1 := exp
+	if e1 == 0 {
 		e1 = time.Now().Add(time.Minute * 30).Unix()
-	} else {
-		e1 = exp
 	}
 
+	claims := &jwt.MapClaims{
+		"namespace": namespace,
+		"userid":    userId,
+		"exp":       e1,
+	}
 	if groupIds != nil {
-		token = jwt.NewWithClaims(jwt.SigningMethodHS256, &jwt.MapClaims{
+		claims = &jwt.MapClaims{
 			"namespace": namespace,
 			"userid":    userId,
 			"groups":    groupIds,
 			"exp":       e1,
-		})
-	} else {
-		token = jwt.NewWithClaims(jwt.SigningMethodHS256, &jwt.MapClaims{
-			"namespace": namespace,
-			"userid":    userId,
-			"exp":       e1,
-		})
+		}
 	}
+	token = jwt.NewWithClaims(jwt.SigningMethodHS256, claims)
 
 	tokenString, err := token.SignedString([]byte(worker.Config.HmacSecret))
 	if err != nil {
@@ -66,20 +64,20 @@ func sliceCompare(x, y []string) bool {
 	return true
 }
 
-func compareTokenParts (tok1 string, tok2 string, part ...int) bool {
+func compareTokenParts(tok1 string, tok2 string, part ...int) bool {
 	var (
-		match bool         = true
+		match bool = true
 	)
 
-	t1 := strings.Split (tok1, ".")
-	t2 := strings.Split (tok2, ".")
+	t1 := strings.Split(tok1, ".")
+	t2 := strings.Split(tok2, ".")
 
 	for _, p := range part {
 		if p <= 0 {
 			continue
 		}
 
-		if t1 [p - 1] != t2 [p - 1] {
+		if t1[p-1] != t2[p-1] {
 			match = false
 			break
 		}
@@ -97,13 +95,11 @@ func TestValidateToken(t *testing.T) {
 	for _, userdata := range userDataList {
 		if tokenString, err = generateJWT(userdata.namespace, userdata.userId, userdata.groupIds, 0); err != nil {
 			t.Errorf("JWT token string generation error: %s", err)
-			return
 		}
 
 		ud, err := validateToken(tokenString)
 		if err != nil {
 			t.Errorf("validateToken() error: %s", err)
-			return
 		}
 		if ud.namespace != userdata.namespace || ud.userId != userdata.userId || !sliceCompare(ud.groupIds, userdata.groupIds) {
 			t.Errorf("Actual output %+v is not equal to the expected output %+v", userdata, ud)
@@ -149,9 +145,9 @@ func TestGetAccessJwt(t *testing.T) {
 		 * One of the factor [signature] would differ is based on the expiry time.
 		 * This will differ for tokstr and jwtstr and hence ignoring it.
 		 */
-		match := compareTokenParts (tokstr, jwtstr, 1, 2 ,0) // compare 1st and 2nd part, ignore the 3rd.
+		match := compareTokenParts(tokstr, jwtstr, 1, 2, 0) // compare 1st and 2nd part, ignore the 3rd.
 		if match == false {
-			t.Errorf ("Actual output is not equal to the expected output")
+			t.Errorf("Actual output is not equal to the expected output")
 		}
 	}
 }
@@ -172,9 +168,9 @@ func TestGetRefreshJwt(t *testing.T) {
 		 * One of the factor [signature] would differ is based on the expiry time.
 		 * This will differ for tokstr and jwtstr and hence ignoring it.
 		 */
-		match := compareTokenParts (tokstr, jwtstr, 1, 2 ,0) // compare 1st and 2nd part, ignore the 3rd.
+		match := compareTokenParts(tokstr, jwtstr, 1, 2, 0) // compare 1st and 2nd part, ignore the 3rd.
 		if match == false {
-			t.Errorf ("Actual output is not equal to the expected output")
+			t.Errorf("Actual output is not equal to the expected output")
 		}
 	}
 }
--- a/edgraph/access_ee_test.go
+++ b/edgraph/access_ee_test.go
@@ -1,48 +1,26 @@
 package edgraph
 
 import (
-	"fmt"
 	"strings"
 	"testing"
 	"time"
 
 	"github.com/dgraph-io/dgraph/ee/acl"
 	"github.com/dgraph-io/dgraph/worker"
+
 	jwt "github.com/dgrijalva/jwt-go"
 )
 
-var (
-	userDataList = []userData{
-		{1234567890, "user1", []string{"701", "702"}},
-		{2345678901, "user2", []string{"703", "701"}},
-		{3456789012, "user3", []string{"702", "703"}},
-	}
-)
-
-func generateJWT(namespace uint64, userId string, groupIds []string, exp int64) (string, error) {
-	var token *jwt.Token
-
-	e1 := exp
-	if e1 == 0 {
-		e1 = time.Now().Add(time.Minute * 30).Unix()
-	}
-
-	claims := jwt.MapClaims{
-		"namespace": namespace,
-		"userid":    userId,
-		"exp":       e1,
-	}
+func generateJWT(namespace uint64, userId string, groupIds []string, expiry int64) string {
+	claims := jwt.MapClaims{"namespace": namespace, "userid": userId, "exp": expiry}
 	if groupIds != nil {
 			claims["groups"] = groupIds
 	}
-	token = jwt.NewWithClaims(jwt.SigningMethodHS256, &claims)
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, &claims)
 
-	tokenString, err := token.SignedString([]byte(worker.Config.HmacSecret))
-	if err != nil {
-		_ = fmt.Errorf("Something Went Wrong: %s", err.Error())
-		return "", err
-	}
-	return tokenString, nil
+	tokenString, _ := token.SignedString([]byte(worker.Config.HmacSecret))
+
+	return tokenString
 }
 
 func sliceCompare(x, y []string) bool {
@@ -50,8 +28,8 @@ func sliceCompare(x, y []string) bool {
 		return false
 	}
 
-	for i, s := range x {
-		if s != y[i] {
+	for i := range x {
+		if x[i] != y[i] {
 			return false
 		}
 	}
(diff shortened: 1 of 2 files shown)


edgraph/access_ee_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+		jwtstr, err = getAccessJwt(userdata.userId, grpLst, userdata.namespace)
+		if err != nil {
+			t.Errorf("getAccessJwt () error: %s", err)
+			return
+		}
+		t2 := strings.Split(jwtstr, ".")
+
+		/*
+		 * JWT token format = [header].[payload].[signature].
```
anurags92, line 139: We can use a helper function to do this validation and use it in the two functions below.

edgraph/access_ee_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+	for _, userdata := range userDataList {
+		if tokstr, err = generateJWT(userdata.namespace, userdata.userId, nil, exp); err != nil {
+			t.Errorf("JWT token string generation error: %s", err)
+			return
+		}
+		t1 := strings.Split(tokstr, ".")
+
+		jwtstr, err = getRefreshJwt(userdata.userId, userdata.namespace)
+		if err != nil {
```
anurags92, line 165: We don't need to do error handling in tests, because if it errors out, the test will fail, which is good enough.

anurags92:
We can improve the description with details on the tests.

edgraph/access_ee_test.go
```diff
@@ -0,0 +1,180 @@
...
+	} else {
+		token = jwt.NewWithClaims(jwt.SigningMethodHS256, &jwt.MapClaims{
+			"namespace": namespace,
+			"userid":    userId,
+			"exp":       e1,
+		})
+	}
+
+	tokenString, err := token.SignedString([]byte(worker.Config.HmacSecret))
+	if err != nil {
```
anurags92, line 50: Remove error handling

edgraph/access_ee_test.go
```diff
@@ -0,0 +1,180 @@
...
+		userData{2345678901, "user2", []string{"703", "701"}},
+		userData{3456789012, "user3", []string{"702", "703"}},
+	}
+)
+
+func generateJWT(namespace uint64, userId string, groupIds []string, exp int64) (string, error) {
+	var e1 int64
+	var token *jwt.Token
+
+	if exp == 0 {
```
matthewmcneely, line 26: This could be cleaner, for instance:

```golang
e1 := exp
if e1 == 0 {
    e1 = time.Now()...
}
```

edgraph/access_ee_test.go
```diff
@@ -0,0 +1,180 @@
...
+		e1 = time.Now().Add(time.Minute * 30).Unix()
+	} else {
+		e1 = exp
+	}
+
+	if groupIds != nil {
```
matthewmcneely, line 32: This block could be easier to read by creating the token outside the if branch, then applying the "groups" member value iif `groupsIds != nil`

edgraph/access_ee_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/dgraph-io/dgraph/ee/acl"
+	"github.com/dgraph-io/dgraph/worker"
+	jwt "github.com/dgrijalva/jwt-go"
```
mangalaman93, line 11: separate out the dgraph and non-dgraph imports

edgraph/access_ee_test.go
```diff
@@ -0,0 +1,171 @@
...
+)
+
+var (
+	userDataList = []userData{
```
mangalaman93, line 15: move this inside the test where it is used

edgraph/access_ee_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+func generateJWT(namespace uint64, userId string, groupIds []string, exp int64) (string, error) {
+	var token *jwt.Token
+
+	e1 := exp
+	if e1 == 0 {
+		e1 = time.Now().Add(time.Minute * 30).Unix()
+	}
+
+	claims := jwt.MapClaims{
```
mangalaman93, line 30: you could put this whole thing in one line, should fit

edgraph/access_ee_test.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	return tokenString, nil
+}
+
+func sliceCompare(x, y []string) bool {
+	if len(x) != len(y) {
+		return false
+	}
+
+	for i, s := range x {
```
mangalaman93, line 53: just do something like:
```
for i := range x {
if x[i] != y[i]
```

edgraph/access_ee_test.go
```diff
@@ -0,0 +1,171 @@
...
+	t1 := strings.Split(tok1, ".")
+	t2 := strings.Split(tok2, ".")
+
+	for _, p := range part {
+		if p <= 0 {
+			continue
+		}
+
+		if t1[p-1] != t2[p-1] {
+			match = false
```
mangalaman93, line 76: you could just return here

edgraph/access_ee_test.go
```diff
@@ -0,0 +1,171 @@
...
+		if s != y[i] {
+			return false
+		}
+	}
+
+	return true
+}
+
+func compareTokenParts(tok1 string, tok2 string, part ...int) bool {
+	var (
```
mangalaman93, line 63: you don't need this variable at all

edgraph/access_ee_test.go
```diff
@@ -0,0 +1,171 @@
...
+			match = false
+			break
+		}
+	}
+
+	return match
+}
+
+func TestValidateToken(t *testing.T) {
+	var (
```
mangalaman93, line 85: you don't need to declare these variables at all

edgraph/access_ee_test.go
```diff
@@ -0,0 +1,171 @@
...
+		err         error
+		tokenString string
+	)
+
+	for _, userdata := range userDataList {
+		if tokenString, err = generateJWT(userdata.namespace, userdata.userId, userdata.groupIds, 0); err != nil {
```
mangalaman93, line 91: just use `:=` here

edgraph/access_ee_test.go
```diff
@@ -0,0 +1,171 @@
...
+			t.Errorf("validateToken() error: %s", err)
+		}
+		if ud.namespace != userdata.namespace || ud.userId != userdata.userId || !sliceCompare(ud.groupIds, userdata.groupIds) {
+			t.Errorf("Actual output %+v is not equal to the expected output %+v", userdata, ud)
+		}
+	}
+}
+
+func TestGetAccessJwt(t *testing.T) {
+	var (
```
mangalaman93, line 106: Let's not declare variables again, just use `:=` when needed

edgraph/access_ee_test.go
```diff
@@ -0,0 +1,171 @@
...
+		g := acl.GetGroupIDs(grpLst)
+		tokstr, _ = generateJWT(userdata.namespace, userdata.userId, g, exp)
+		jwtstr, _ = getAccessJwt(userdata.userId, grpLst, userdata.namespace)
+
+		/*
+		 * JWT token format = [header].[payload].[signature].
+		 * One of the factor [signature] would differ is based on the expiry time.
+		 * This will differ for tokstr and jwtstr and hence ignoring it.
+		 */
+		match := compareTokenParts(tokstr, jwtstr, 1, 2, 0) // compare 1st and 2nd part, ignore the 3rd.
```
mangalaman93, line 143: I would rather do some sort of validateToken call than actually checking what is inside the token

mangalaman93 (changes requested):

edgraph/access_ee_test.go
billprovince: +1. Here and elsewhere.
Variable declaration should be deferred as late as possible to the point where it is going to be used.
I think the only one I would pre-declare would be the expiry (sic) variable, since it requires a context switch to initialize this.

edgraph/access_ee_test.go
billprovince: Personally, I am ok with using a canonical set of tokens compared with the retrieved token.
The validation process for the other parts of the token would likely not look much different from "Hey, I expected it to match THIS string!", so the string comparison, I think is valid.

Having said that, our handling of making sure to ignore the timestamp seems a bit clunky. I agree that we should ignore the timestamp token (or at least have a different comparison mechanism), but having it controlled via a varargs type structure where the value either indicates an index (offset by 1) or a 0 seems odd.

A couple ideas here:
1.  Pass a slice of booleans for whether or not to compare instead of index offsets with 0 having special value.
2. A bit overboard, but if we want to apply different logic based on token types, we could pass in a slice of types where the validation procedure is decided based on the type.

edgraph/access_ee_test.go
```diff
@@ -0,0 +1,171 @@
...
+		if match == false {
```
billprovince, line 144: Prefer `if !match` over `if match == false`
Or, even better:
`if !compareTokenParts(...)` (since you don't really need the match variable).

Not a big deal, but I think this would be slightly preferable.

billprovince (commented):
See comments inline. Nothing that really prevents checkin in my opinion, but some suggestions for improvements.

edgraph/access_ee_test.go
```diff
@@ -1,113 +1,78 @@
...
 	"testing"
 	"time"
 
 	"github.com/dgraph-io/dgraph/ee/acl"
 	"github.com/dgraph-io/dgraph/worker"
-	jwt "github.com/dgrijalva/jwt-go"
-)
 
-var (
-	userDataList = []userData{
```
jbhamra1: I have moved this slice to TestValidateToken(), TestGetAccessJwt(), TestGetRefreshJwt().
Hope this was the expectation.

mangalaman93 (approved):

mangalaman93 (changes requested):
I do not think we should compare tokens first and the second part to check whether a token is valid. A token is valid if it has the right data and signature and that is why we should call a validataToken sort of a function in order to verify the token.

edgraph/access_ee_test.go
```diff
@@ -0,0 +1,118 @@
...
+func TestGetRefreshJwt(t *testing.T) {
+	expiry := time.Now().Add(worker.Config.RefreshJwtTtl).Unix()
+	userDataList := []userData{
+		{1234567890, "user1", []string{"701", "702"}},
+		{2345678901, "user2", []string{"703", "701"}},
+		{3456789012, "user3", []string{"702", "703"}},
+	}
+
+	for _, userdata := range userDataList {
+		tokstr := generateJWT(userdata.namespace, userdata.userId, nil, expiry)
```
mangalaman93, line 109:
```
jwtstr, _ := getRefreshJwt(userdata.userId, userdata.namespace)
ud2, _ := validateToken(jwtstr)
```

This is all you need and then you can check whether the values you passed to getRefreshJwt vs what was returned by validateToken.

edgraph/access_ee_test.go
```diff
@@ -0,0 +1,118 @@
...
+	g := acl.GetGroupIDs(grpLst)
+	expiry := time.Now().Add(worker.Config.AccessJwtTtl).Unix()
+	userDataList := []userData{
+		{1234567890, "user1", []string{"701", "702"}},
+		{2345678901, "user2", []string{"703", "701"}},
+		{3456789012, "user3", []string{"702", "703"}},
+	}
+
+	for _, userdata := range userDataList {
+		tokstr := generateJWT(userdata.namespace, userdata.userId, g, expiry)
```
mangalaman93, line 89: same here

edgraph/access_ee_test.go
```diff
@@ -0,0 +1,112 @@
...
+	g := acl.GetGroupIDs(grpLst)
+	userDataList := []userData{
+		{1234567890, "user1", []string{"701", "702"}},
+		{2345678901, "user2", []string{"703", "701"}},
+		{3456789012, "user3", []string{"702", "703"}},
+	}
+
+	for _, userdata := range userDataList {
+		jwtstr, _ := getAccessJwt(userdata.userId, grpLst, userdata.namespace)
+		ud, _ := validateToken (jwtstr)
```
billprovince, line 89: Similar issue with validateToken.
See definition here:
https://github.com/dgraph-io/dgraph/blob/bc71dc3297880e8c282be638506eddab9270b010/edgraph/access.go#L93

It doesn't do anything aside from return nil, nil.

edgraph/access_ee_test.go
```diff
@@ -0,0 +1,112 @@
...
+	userDataList := []userData{
+		{1234567890, "user1", []string{"701", "702"}},
+		{2345678901, "user2", []string{"703", "701"}},
+		{3456789012, "user3", []string{"702", "703"}},
+	}
+
+	for _, userdata := range userDataList {
+		tokenString := generateJWT(userdata.namespace, userdata.userId, userdata.groupIds, expiry)
+
+		ud, _ := validateToken(tokenString)
```
billprovince, line 50: I suspect that you have a local change, as otherwise, at line 51 below, we should expect an error since ud would be nil.
See: https://github.com/dgraph-io/dgraph/blob/bc71dc3297880e8c282be638506eddab9270b010/edgraph/access.go#L93

billprovince (commented):
There's an issue with our usage of validateToken. To put it bluntly, I don't see how the tests would pass with the change you specified unless you have additional changes not reflected in github.

edgraph/access_ee_test.go
mangalaman93: I think you want to look at the file access_ee.go instead.

mangalaman93: And therefore, we need to add the following to the test file
```
//go:build !oss
// +build !oss
```

edgraph/access_ee_test.go
```diff
@@ -0,0 +1,112 @@
+package edgraph
```
mangalaman93, line 1: add the copyright, this will be under community license (not apache v2) and also the right tags.

edgraph/access_ee_test.go
jbhamra1: edgraph/access.go and edgraph/access_ee.go declares "func validateToken(jwtStr string) ([]string, error)" and "func validateToken(jwtStr string) (*userData, error)" respectively.

The validateToken(tokenString) call from edgraph/access_ee_test.go resolves to "func validateToken(jwtStr string) (*userData, error)" declared in edgraph/access_ee.go but not to "func validateToken(jwtStr string) ([]string, error)" declared in edgraph/access.go, at least, in my linux workspace. The test output shown below. The text Italic bold is the result of an extra statement I have added to the test code, 'fmt.Println (ud)', for verification.

jassi Wed Dec 14 11:58:08 edgraph_validatetoken_getaccessjwt_getrefreshjwt edgraph > go test -v -count=1 ./...
=== RUN   TestValidateToken
**_&{1234567890 user1 [701 702]}
&{2345678901 user2 [703 701]}
&{3456789012 user3 [702 703]}_**
--- PASS: TestValidateToken (0.00s)
=== RUN   TestGetAccessJwt
--- PASS: TestGetAccessJwt (0.00s)
=== RUN   TestGetRefreshJwt
--- PASS: TestGetRefreshJwt (0.00s)
=== RUN   TestParseNQuads
--- PASS: TestParseNQuads (0.00s)
=== RUN   TestValNquads
. . .

When I write a sample program with an Add() function I see a different behaviour altogether, go interpreter throws a message- 'Add redeclared', whereas this is not seen for edgraph/ above.
This is the sample program code-

jassi Wed Dec 14 12:11:06 edgraph_validatetoken_getaccessjwt_getrefreshjwt math > ls -l
total 12
-rw-r--r-- 1 jassi users  65 Dec 14 07:56 m.go
-rw-r--r-- 1 jassi users 148 Dec 14 08:04 m_1.go
-rw-r--r-- 1 jassi users 515 Dec 14 10:45 m_1_test.go
jassi Wed Dec 14 12:11:08 edgraph_validatetoken_getaccessjwt_getrefreshjwt math > 
jassi Wed Dec 14 12:11:09 edgraph_validatetoken_getaccessjwt_getrefreshjwt math > cat m.go
package math

func Add (x int, y int) ([]string) {
	return nil
}
jassi Wed Dec 14 12:11:15 edgraph_validatetoken_getaccessjwt_getrefreshjwt math > cat m_1.go
package math

type userData struct {
	name string
	uid int64
}

func Add (x int, y int) (*userData) {
	return &userData{name: "Hello", uid: 2233}
}
jassi Wed Dec 14 12:11:25 edgraph_validatetoken_getaccessjwt_getrefreshjwt math >
jassi Wed Dec 14 12:11:33 edgraph_validatetoken_getaccessjwt_getrefreshjwt math > cat m_1_test.go
package math

import (
	"testing"
)

func Test_Add (t *testing.T) {
	Add (3, 5)
/*
	got := Add (3, 5)
	want := 8

	if got != want {
		t.Errorf ("got %v, wanted %v", got, want)
	}
*/
}
jassi Wed Dec 14 12:12:03 edgraph_validatetoken_getaccessjwt_getrefreshjwt math > go test -v -count=1 ./...
# github.com/dgraph-io/dgraph/gotest/math [github.com/dgraph-io/dgraph/gotest/math.test]
./m_1.go:8:6: Add redeclared in this block
	./m.go:3:6: other declaration of Add
FAIL	github.com/dgraph-io/dgraph/gotest/math [build failed]
FAIL
jassi Wed Dec 14 12:12:25 edgraph_validatetoken_getaccessjwt_getrefreshjwt math >

I don't know what is causing this discrepancy.

mangalaman93 (approved):

meghalims (approved):
Looks good

PR #8472, opened 2022-12-01
chore(ci): let pkg flag take multiple arguments

joshua-goldstein:
## Problem
Previously running

```
./t --pkg="systest/backup/common,systest/backup/encryption"
```

would not run any packages, because the pkg flag is currently parsed as a single string and matched against the list of packages in the repository.  Internally, a pkg.ID looks like 
```
github.com/dgraph-io/dgraph/systest/backup/common
```

so we are matching against this string.

## Solution

We split the string argument provided to --pkg by comma.  Now the above command will run two packages as expected.

### DIFF
--- a/t/t.go
+++ b/t/t.go
@@ -495,7 +495,7 @@ func getPackages() []task {
 
 	slowPkgs := []string{"systest", "ee/acl", "cmd/alpha", "worker", "e2e"}
 	skipPkgs := strings.Split(*skip, ",")
-	runPkgs := strings.Split(*runPkg, ",")
+	runPkgs  := strings.Split(*runPkg, ",")
 
 	moveSlowToFront := func(list []task) []task {
 		// These packages typically take over a minute to run.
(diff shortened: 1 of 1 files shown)


MichelDiz:
@joshua-goldstein that wouldn't be a typo? a single quote.

joshua-goldstein:
@anurags92 @MichelDiz This is a small 

> @joshua-goldstein that wouldn't be a typo? a single quote.

Sorry, I didn't follow?  We are splitting by comma `,`

joshua-goldstein:
Added a remark about what internal `pkg.ID` strings look like.

anurags92 (approved):

t/t.go
```diff
@@ -530,8 +531,18 @@ func getPackages() []task {
 
 	var valid []task
 	for _, pkg := range pkgs {
-		if len(*runPkg) > 0 && !strings.HasSuffix(pkg.ID, "/"+*runPkg) {
-			continue
+		if len(*runPkg) > 0 {
```
nswamy, line 534: nit: this is not necessary, the for loop would skip

nswamy (approved):

PR #8474, opened 2022-12-02
chore(ci): pin runners to ubuntu-20.04

joshua-goldstein:
## Problem

As per [Github](https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/) the `ubuntu-latest` label is transitioning to Ubuntu 22.04.  We would like to continue running on Ubuntu 20.04 for now.

## Solution

Pin runner to Ubuntu 20.04.

PR #8475, opened 2022-12-03
Update base ubuntu image

anurags92:
## Problem
Our base image for building dgraph docker images is `ubuntu:20.04`. This causes issue when we want to run tests on github provided machines. 
## Solution
PR updates base image to `ubuntu:22.04`.

### DIFF
--- a/contrib/Dockerfile
+++ b/contrib/Dockerfile
@@ -4,7 +4,7 @@
 # This gets built as part of release.sh. Must be run from /tmp/build, with the linux binaries
 # already built and placed there.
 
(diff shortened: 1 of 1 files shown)


joshua-goldstein:
@anurags92 I think we want to do the opposite - we can stick to 20.04 runners, since all of our CI/CD and development is happening on there.  We can move to 22.04 later (20.04 will not be deprecated anytime soon).

joshua-goldstein:
I opened #8474 to pin github runners to 20.04.

PR #8488, opened 2022-12-05
Add intersection tests for linear and binary methods

anurags92:
Add test cases for `IntersectWithLin`, `IntersectWithJump` and `IntersectWithBin`

### DIFF
--- a/algo/uidlist_test.go
+++ b/algo/uidlist_test.go
@@ -260,7 +260,6 @@ func TestUIDListIntersect5(t *testing.T) {
 
 func TestUIDListIntersect6(t *testing.T) {
 	common, other, _ := fillNums(10, 500)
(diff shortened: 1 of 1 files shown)


matthewmcneely (approved):

algo/uidlist_test.go
```diff
@@ -258,6 +258,30 @@ func TestUIDListIntersect5(t *testing.T) {
 	require.Equal(t, []uint64{3}, u.Uids)
 }
 
+func TestUIDListIntersect6(t *testing.T) {
```
nswamy, line 261: does go test framework provide a way to parameterize the values you want to test with (fllNums here). otherwise may be we can write a single test looping through the values ?

meghalims (approved):
LGTM.

PR #8490, opened 2022-12-06
chore(logging): Add more logging and trace tags for debugging

anurags92:
Adds logging and sets query `tag` for improved tracing in Jaeger.

### DIFF
--- a/edgraph/server.go
+++ b/edgraph/server.go
@@ -1398,8 +1398,11 @@ func processQuery(ctx context.Context, qc *queryContext) (*api.Response, error) 
 		glog.Infof("Finished a query that started at: %+v",
 			qr.Latency.Start.Format(time.RFC3339))
 	}
+
 	if err != nil {
-		glog.Infof("Error processing query: %+v\n", err.Error())
+		if bool(glog.V(3)) {
+			glog.Infof("Error processing query: %+v\n", err.Error())
+		}
 		return resp, errors.Wrap(err, "")
 	}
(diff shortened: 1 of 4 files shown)


edgraph/server.go
```diff
@@ -1394,7 +1394,12 @@ func processQuery(ctx context.Context, qc *queryContext) (*api.Response, error)
 	// Core processing happens here.
 	er, err := qr.Process(ctx)
 
+	if bool(glog.V(3)) || worker.LogDQLRequestEnabled() {
+		glog.Infof("Finished a query that started at: %+v",
+			qr.Latency.Start.Format(time.RFC3339))
+	}
 	if err != nil {
+		glog.Infof("Error processing query: %+v\n", err.Error())
```
matthewmcneely, line 1402: Could `err` here contain query processing errors? If so, maybe it should be restricted to the same v3 logging level?  People testing out with different syntax queries could make for some spammy logs.

anurags92: Good idea, gated the logging behind log level.

damonfeldman:
This enhancement ensures that the duration of all queries are logged if either the -v 3 logging level is set (level 3 is debug) or if "request logging" is enabled on the alpha. In addition, all query errors are logged if logging is set to -v 3. Previously, errors were not logged on the alpha (only reported to the client).

Also, for a particular error in zero logs reading "Error while writing: ..." the root cause of the error is now logged.

PR #8491, opened 2022-12-06
fix(ci/cd): pin linux deps

skrdgraph:
## Problem
The current problem with our Dgraph Docker environment is that the base linux packages are NOT pinned. We build the Docker container in our CI setup every time, and are pulling the `latest` tags. It's a good practice to make these dependencies static. This will avoid potential issues that can arise from environment changes.

## Solution
use `apt-cache policy <package name>` to identify the candidate version & old installed version. 
pin the packages correctly, and maintain them when CVE fixes are needed in the future.

### DIFF
--- a/contrib/Dockerfile
+++ b/contrib/Dockerfile
@@ -11,12 +11,12 @@ LABEL maintainer="Dgraph Labs <contact@dgraph.io>"
 RUN apt-get update
 # pin package versions always & bring in CVE fixes as needed
 RUN apt-get install -y --no-install-recommends ca-certificates
+RUN apt-get install -y --no-install-recommends sysstat
 RUN apt-get install -y --no-install-recommends curl="7.68.0-1ubuntu2.14"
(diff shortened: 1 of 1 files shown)


matthewmcneely (approved):

joshua-goldstein:
Looks like it is having trouble pulling sysstat?

skrdgraph:
@joshua-goldstein that's correct, I believe it had a n/w issue of some kind. The `make dgraph-image` is also invoked in trivy test - which we are passing atm. I am going to have to investigate this later in the evening.

skrdgraph:
Also another FYI - This PR also fixes about 60+ CVEs. But please remember, we will have to do a security exposure review monthly.

joshua-goldstein (approved):
Looks good now - I see it looks like a cache issue

meghalims (approved):
Looks good 👍

matthewmcneely (approved):

PR #8498, opened 2022-12-09
Lambda image fix for arm

adityasadalage:


adityasadalage:
Sanity check PR, closing

PR #7825, opened 2021-05-17
handle manifest version logic, update manifest version to 2105

NamanJain8:
The backward compatibility of the backup's manifest was broken by https://github.com/dgraph-io/dgraph/pull/7810, although the tool was added (https://github.com/dgraph-io/dgraph/pull/7815) that enables smooth migration of manifest.
This PR makes backup backward compatible, by updating the manifest(in-memory) after reading. 

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7825)

PR #7815, opened 2021-05-13
add update_manifest tool

NamanJain8:
This PR adds `dgraph update_manifest` tool. This can be used to migrate the manifest of the backups taken on release/v21.03 or their parallels to the state after this change https://github.com/dgraph-io/dgraph/pull/7810

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7815)

PR #8499, opened 2022-12-09
fix(version): bump up dgraph version in source

joshua-goldstein:
## Problem
 
Dgraph version is hard coded in source.  Currently it is still at 2103.  This should be bumped up to the latest version.  Thanks to @MichelDiz for noticing this issue.

## Solution

Added todo's because we probably should not be hard coding this into source.  We already pass the version in via ldflags, we could potentially use this instead.

### DIFF
--- a/x/x.go
+++ b/x/x.go
@@ -140,7 +140,7 @@ const (
 		"X-CSRF-Token, X-Auth-Token, X-Requested-With"
 	DgraphCostHeader = "Dgraph-TouchedUids"
 
-	// todo: we should use ldflag in init.go (currently not exported) to set this version
+	// TODO: we should use ldflag in init.go (currently not exported) to set this version
 	DgraphVersion = 2200
 )
 
--- a/x/x.go
+++ b/x/x.go
@@ -140,7 +140,8 @@ const (
 		"X-CSRF-Token, X-Auth-Token, X-Requested-With"
 	DgraphCostHeader = "Dgraph-TouchedUids"
 
-	DgraphVersion = 2103
+	// todo: we should use ldflag in init.go (currently not exported) to set this version
+	DgraphVersion = 2200
 )
 
 var (
--- a/worker/backup_ee.go
+++ b/worker/backup_ee.go
@@ -215,9 +215,7 @@ func ProcessBackupRequest(ctx context.Context, req *pb.BackupRequest) error {
 		}
 	}
 
-	// DgraphVersion hard coded in x.go
-	// todo: dgraph version probably should not be hard coded in source
-	// todo: we should use ldflag in init.go (currently not exported) to set this version
+	// TODO: we should use ldflag in init.go to set this version (currently DgraphVersion hard coded in x.go)
 	dir := fmt.Sprintf(backupPathFmt, req.UnixTs)
 	m := Manifest{
 		ReadTs:         req.ReadTs,
--- a/worker/backup_ee.go
+++ b/worker/backup_ee.go
@@ -217,7 +217,7 @@ func ProcessBackupRequest(ctx context.Context, req *pb.BackupRequest) error {
 
 	// DgraphVersion hard coded in x.go
 	// todo: dgraph version probably should not be hard coded in source
-	// todo: we should use ldflag in init.go (currently not exported)
+	// todo: we should use ldflag in init.go (currently not exported) to set this version
 	dir := fmt.Sprintf(backupPathFmt, req.UnixTs)
 	m := Manifest{
 		ReadTs:         req.ReadTs,
--- a/worker/backup_ee.go
+++ b/worker/backup_ee.go
@@ -1,3 +1,4 @@
+//go:build !oss
 // +build !oss
 
 /*
@@ -214,6 +215,9 @@ func ProcessBackupRequest(ctx context.Context, req *pb.BackupRequest) error {
 		}
 	}
 
+	// DgraphVersion hard coded in x.go
+	// todo: dgraph version probably should not be hard coded in source
+	// todo: we should use ldflag in init.go (currently not exported)
 	dir := fmt.Sprintf(backupPathFmt, req.UnixTs)
 	m := Manifest{
 		ReadTs:         req.ReadTs,


skrdgraph:
Looks like some our backup worker code uses `x.DgraphVersion` .. good catch!

x/x.go
```diff
@@ -140,7 +140,8 @@ const (
 		"X-CSRF-Token, X-Auth-Token, X-Requested-With"
 	DgraphCostHeader = "Dgraph-TouchedUids"
 
-	DgraphVersion = 2103
+	// TODO: we should use ldflag in init.go (currently not exported) to set this version
+	DgraphVersion = 2200
```
matthewmcneely, line 144: Maybe we should add this to the "Doing a Release" doc, basically just making sure this value is correct when the release is cut. But yeah, I agree in principle that it really should be pulling from the environment.

MichelDiz:
Do we already know how we use this variable? My theory is that we use this in the "upgrade" subprogram.

x/x.go
skrdgraph: +1 ^ would it be possible to pull this info from the environment at build time instead?

matthewmcneely:
Actually this PR is not needed. The -X build tag in the Makefile [here](https://github.com/dgraph-io/dgraph/blob/8c72479b33cfd2f803089de314220e3d8028d053/dgraph/Makefile#L39) sets this variable at build time. @skrdgraph, I'm assuming the release pipeline still sets env DGRAPH_VERSION as expected. 

Maybe a comment could be added just above to let future readers know that it's set at build time and/or set the value to sometime like 0? @skrdgraph, feel free to close this PR once you're satisfied the new release pipeline is setting up DGRAPH_VERSION correctly.

skrdgraph:
> Actually this PR is not needed. The -X build tag in the Makefile [here](https://github.com/dgraph-io/dgraph/blob/8c72479b33cfd2f803089de314220e3d8028d053/dgraph/Makefile#L39) sets this variable at build time. @skrdgraph, I'm assuming the release pipeline still sets env DGRAPH_VERSION as expected.
> 
> Maybe a comment could be added just above to let future readers know that it's set at build time and/or set the value to sometime like 0? @skrdgraph, feel free to close this PR once you're satisfied the new release pipeline is setting up DGRAPH_VERSION correctly.

yeah ^ that makes sense. 

In the release pipeline we obtain the version like [this](https://github.com/dgraph-io/dgraph/blob/main/.github/workflows/cd-dgraph.yml#L38-L51) and this gets passed to the build step like [this](https://github.com/dgraph-io/dgraph/blob/main/.github/workflows/cd-dgraph.yml#L52-L53) 

I know that the `dgraph version` on the generated binary shows what we pass in the release pipeline. But regarding this PR, there was something more to this - @MichelDiz & @joshua-goldstein - we are voting to close this ... unless you remember how this came up ....

MichelDiz:
Guys, I think this is necessary. See https://github.com/dgraph-io/dgraph/blob/6f7a6d2124052abcec25cb53f5a557c5973418dc/x/x.go#L143
It was renamed to Manifest version. So this is related to internal files.

See the PR https://github.com/dgraph-io/dgraph/pull/7825

MichelDiz:
@matthewmcneely as far I undestood in the PRs related to this var. The -X build tag is irrelevant to this. It is related to backups and upgrades. So I think this variable needs attentio everytime we have a big change in the code that migh be breaking. Manually attention I mean. As its use has not been documented, we need to understand and document it.

joshua-goldstein:
@matthewmcneely @skrdgraph Please take a look at my comment in the PR.  There is a file `x/init.go` that contains the strings that are set by ldflags.  See [here](https://github.com/dgraph-io/dgraph/blob/main/x/init.go#L37).  But this version is not being used by the backup code.  Backup code uses the hardcoded variable in x.go (see [here](https://github.com/dgraph-io/dgraph/blob/main/x/x.go#L143))

joshua-goldstein:
> Guys, I think this is necessary. See
> 
> https://github.com/dgraph-io/dgraph/blob/6f7a6d2124052abcec25cb53f5a557c5973418dc/x/x.go#L143
> 
> 
> It was renamed to Manifest version. So this is related to internal files. In the case to Backup files.
> > The backward compatibility of the backup's manifest was broken by #7810, although the tool was added (#7815) that enables smooth migration of manifest.
> > This PR makes backup backward compatible, by updating the manifest(in-memory) after reading.
> 
> See the PR #7825

This is a good catch, I wasn't aware this was renamed at some point...  Looks like that PR was merged into slash so presumably this will come up as we merge commits from slash into main.

matthewmcneely:
> It was renamed to Manifest version. So this is related to internal files. In the case to Backup files.

@MichelDiz I think this tool (ee/updatemanifest) was removed, right?

MichelDiz:
@matthewmcneely Not sure, I never saw that tool. The backup and upgrades feels like isolated codes(subcommands) that wasn't well documented for future maintenance.

matthewmcneely:
> Not sure, I never saw that tool. The backup and upgrades feels like isolated codes(subcommands) that wasn't well documented for future maintenance.

OK, but what I can verify is that the on the main branch with the current Makefile, that variable x.DgraphVersion is set by whatever is in the -X build flag.

MichelDiz:
@matthewmcneely correct, but it is a different variable. This one DgraphVersion is a exported variable, the one you're mentioning is in init.go which is private variable. That private one is set via -ldflags. If they are different, must be a reason for this.

matthewmcneely:
> @matthewmcneely correct, but it is a different variable. This one DgraphVersion is an exported variable, the one you're mentioning is in init.go which is private variable. That private one is set via -ldflags. If they are different, must be a reason for this.

Ah right, sorry. So I guess the question remains, should this variable be set by an ldflag and not hardcoded? From what I can see, it's only used by worker/backup_ee.go::ProcessBackupRequest. Which I'm assuming we'd want that function to actually have a correct (current) version when it creates the backup `Manifest`.

skrdgraph:
This was changed to ManifestVersion, this solves the current naming issues. 

Please leave a comment while closing existing PRs / Issues (just for clarifications).

PR #8500, opened 2022-12-09
fix(ci): reduce triggers by disabling on draft runs

skrdgraph:
## Problem
ci needs draft run trigger to be disabled, cost

## Solution
 disabled in ci yawls


meghalims (approved):
LGTM

dshekhar95 (approved):

PR #8503, opened 2022-12-11
fix(ci): resolve systest failures on arm

joshua-goldstein:
## Problem

Systest/export backup to minio was failing on arm64.

## Solution

The minio image that we require was published before multi-architecture docker images were introduced.  As such, there are two tags, namely `RELEASE.2020-11-13T20-10-18Z` (which we currently use for systest/export) and `RELEASE.2020-11-13T20-10-18Z-arm64` (which is required for arm64).  See the Dockerhub page [here](https://hub.docker.com/r/minio/minio/tags?page=1&name=RELEASE.2020-11-13T20-10-18Z).

Unfortunately docker-compose files do not offer some sort of conditional logic.  As such the easiest way to use this image is to create a new docker-compose-arm64.yml file and a flag that specifies when it should be uses at the t.go level.  This is not so elegant but is the easiest solution currently since we lack a way to dynamically generate docker-compose files.

This PR also improves some testing functionality.  In brief, all that is required to run tests now is to run `make test` from the root dgraph directory.  If one wants to run specific package tests (e.g. systest/export) one can use `make test args="--pkg=systest/export"`.   What this does is pass the `--pkg=systest/export` flag to the t.go binary when it is run.  This has the benefit that any number of arguments can be passed to t.go.  E.g., in the process of testing this PR, I ran
```
make test args="--pkg=systest/export --arch=arm64"
```

which (after building the dgraph binary and making a dgraph docker image with local tag) builds t.go and executes `./t --pkg=systest/export --arch-arm64`

Finally, we fix a logic error that was introduced recently in t.go.  Coverage files should only be copied over when t.go is run in coverage mode.


skrdgraph (approved):

skrdgraph:
This looks good 👍 - ill merge it tomorrow! I want a review by @anurags92 here as well ...

systest/backup/minio/docker-compose-arm64.yml
```diff
@@ -0,0 +1,97 @@
+# Auto-generated with: [./compose -a 3 -z 1 -r 1 --minio --minio_port=9001 --minio_env_file=../../backup.env -w --port_offset=0 --expose_ports=false --mem= --names=false -O ../systest/backup/minio/docker-compose.yml]
```
anurags92, line 1: Same comment as above

systest/backup/minio/docker-compose-arm64.yml
```diff
@@ -0,0 +1,97 @@
...
+      target: /gobin
+      read_only: true
+    - type: bind
+      source: ../../../tlstest/mtls_internal/tls/zero1
+      target: /dgraph-tls
+      read_only: true
+    command: /gobin/dgraph  ${COVERAGE_OUTPUT} zero --raft "idx=1;" --my=zero1:5080 --replicas=1 --logtostderr -v=2 --bindall
+      --tls "ca-cert=/dgraph-tls/ca.crt; server-cert=/dgraph-tls/node.crt; server-key=/dgraph-tls/node.key; internal-port=true; client-cert=/dgraph-tls/client.zero1.crt; client-key=/dgraph-tls/client.zero1.key;"
+  minio:
+    image: minio/minio:RELEASE.2020-11-13T20-10-18Z-arm64
```
anurags92, line 91: Same comment as above

systest/export/docker-compose-arm64.yml
```diff
@@ -0,0 +1,94 @@
+# Auto-generated with: [./compose -l -a 3 -z 1 -r 1 --alpha_volume data:/data --minio --minio_port=9001 --minio_env_file=export.env --port_offset=0 --expose_ports=false --alpha_env_file=export.env --mem= --names=false -O ../systest/export/docker-compose.yml]
```
anurags92, line 1: Same as above

systest/export/docker-compose-arm64.yml
```diff
@@ -0,0 +1,94 @@
...
+      read_only: true
+    - type: volume
+      source: data
+      target: /data
+      read_only: false
+    command: /gobin/dgraph ${COVERAGE_OUTPUT} alpha --my=alpha3:7080 --zero=zero1:5080 --logtostderr
+      -v=2 
+      --security "whitelist=10.0.0.0/8,172.16.0.0/12,192.168.0.0/16;"
+  minio:
+    image: minio/minio:RELEASE.2020-11-13T20-10-18Z-arm64
```
anurags92, line 72: Same as above

t/Makefile
```diff
@@ -26,8 +26,12 @@ test:
...
+	@go clean -testcache
+#	run t.go with specified arguments; otherwise run standard suite
+	@if [ -n "$(args)" ]; then \
+		./t $(args); \
+	else \
+		./t; \
+	fi
 #	clean up docker containers after test execution
-	./t -r
+	@./t -r
```
anurags92, line 37: Why do we use `@` ?

t/t.go
```diff
@@ -95,6 +95,7 @@ var (
 		"comma separated list of packages that needs to be skipped. "+
 			"Package Check uses string.Contains(). Please check the flag carefully")
 	runCoverage = pflag.Bool("coverage", false, "Set true to calculate test coverage")
+	arch        = pflag.String("arch", "x86", "Machine architecture (required for systest/export test)")
```
anurags92, line 98: Add comment on which values it accepts. eg. Valid values: [arm, x86 etc.]

t/t.go
```diff
@@ -204,26 +205,28 @@ func stopCluster(composeFile, prefix string, wg *sync.WaitGroup, err error) {
 			fmt.Printf("CLUSTER STOPPED: %s\n", prefix)
 		}
 
-		// get all matching containers, copy /usr/local/bin/coverage.out
-		containers := testutil.AllContainers(prefix)
-		for _, c := range containers {
-			tmp := fmt.Sprintf("%s.%s", tmpCoverageFile, c.ID)
+		if *runCoverage == true {
```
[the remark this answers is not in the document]
anurags92: Good catch!

t/t.go
```diff
@@ -515,8 +519,16 @@ type task struct {
 	isCommon bool
 }
 
+// for custom cluster tests (i.e. those not using default docker-compose.yml)
 func composeFileFor(pkg string) string {
 	dir := strings.Replace(pkg, "github.com/dgraph-io/dgraph/", "", 1)
+	if *arch == "arm64" {
```
anurags92, line 525: This is not required if you decide to update the PR based on environment variables,

anurags92:
Wanted to test if the changes I proposed would work: https://github.com/dgraph-io/dgraph/pull/8508. cc: @joshua-goldstein @skrdgraph

t/Makefile
joshua-goldstein: Without the @ the makefile will print out the command and then execute it; with the @ the command is just executed.

systest/backup/minio-large/docker-compose-arm64.yml
```diff
@@ -0,0 +1,97 @@
+# Auto-generated with: [./compose -a 3 -z 1 -r 1 --minio --minio_port=9001 --minio_env_file=../../backup.env -w --port_offset=0 --expose_ports=false --mem= --names=false -O ../systest/backup/minio-large/docker-compose.yml]
```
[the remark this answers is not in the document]
joshua-goldstein: Agreed - it would be nice to eventually use compose.go but for now I modify this line

systest/backup/minio/docker-compose-arm64.yml
joshua-goldstein: Done

systest/backup/minio/docker-compose-arm64.yml
joshua-goldstein: Done

systest/export/docker-compose-arm64.yml
joshua-goldstein: Done

t/t.go
joshua-goldstein: done

systest/backup/minio-large/docker-compose-arm64.yml
```diff
@@ -0,0 +1,97 @@
...
+      read_only: true
+    - type: bind
+      source: ../../../tlstest/mtls_internal/tls/alpha3
+      target: /dgraph-tls
+      read_only: true
+    command: /gobin/dgraph  ${COVERAGE_OUTPUT} alpha --my=alpha3:7080 --zero=zero1:5080 --logtostderr -v=2 
+      --security "whitelist=10.0.0.0/8,172.16.0.0/12,192.168.0.0/16;"
+      --tls "ca-cert=/dgraph-tls/ca.crt; server-cert=/dgraph-tls/node.crt; server-key=/dgraph-tls/node.key; internal-port=true; client-cert=/dgraph-tls/client.alpha3.crt; client-key=/dgraph-tls/client.alpha3.key;"
+  minio:
+    image: minio/minio:RELEASE.2020-11-13T20-10-18Z-arm64
```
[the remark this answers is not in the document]
joshua-goldstein: This is a very nice solution...  I wasn't sure if this was possible, so thank you for confirming it in your PR!  I just updated this PR.

systest/backup/minio/docker-compose-arm64.yml
joshua-goldstein: Done

systest/export/docker-compose-arm64.yml
joshua-goldstein: Done

t/t.go
joshua-goldstein: removed

joshua-goldstein:
> Wanted to test if the changes I proposed would work: #8508. cc: @joshua-goldstein @skrdgraph

I was also wondering if the environment variable would get substituted at the appropriate time...  Looks like it works fine.  Thanks for the idea and testing it!  I've updated this PR.

anurags92:
@joshua-goldstein we can remove architecture flag and let the `t.go` figure it out itself. Much cleaner. Also can we log somewhere which architecture we are running tests on, so that we are sure that our changes flow through?

joshua-goldstein:
> @joshua-goldstein we can remove architecture flag and let the `t.go` figure it out itself. Much cleaner. Also can we log somewhere which architecture we are running tests on, so that we are sure that our changes flow through?

So sometime like a `runtime.GOARCH` check?  We can also do that easily.

anurags92:
Another small comment which came up during the code review call, we can define default values for docker variables used in yml files. See: https://stackoverflow.com/questions/70772517/docker-compose-env-default-with-fallback 

Should we use that and insert a different value only for ARM? 

(Apologies for a separate comment, should have added this earlier but forgot).

t/Makefile
```diff
@@ -26,8 +26,12 @@ test:
 # 	build the t.go binary
 	@go build .
 #	clean go testcache
-	go clean -testcache
-#	run the tests
-	@./t --skip systest/backup,systest/online-restore,systest/loader,tlstest
+	@go clean -testcache
+#	run t.go with specified arguments; otherwise run standard suite
+	@if [ -n "$(args)" ]; then \
```
anurags92, line 31: If `$arags` is empty, wouldn't that default to `./t`. Do we need an `if-else`?

joshua-goldstein:
> Another small comment which came up during the code review call, we can define default values for docker variables used in yml files. See: https://stackoverflow.com/questions/70772517/docker-compose-env-default-with-fallback
> 
> Should we use that and insert a different value only for ARM?
> 
> (Apologies for a separate comment, should have added this earlier but forgot).

Also a good catch - what I did was add a default value now, so if the environment variable is not set it will use the previously used tag.

anurags92 (approved):
Thanks Joshua! Looks good to me.

PR #8492, opened 2022-12-06
fix(ci/cd): Fix failing test packages for ARM CI runs

adityasadalage:
## Problem
We are currently skipping some test packages in ARM CI to get passing test runs. This prevents the ARM CI from having true parity with amd64 CI runs and we need a baseline for ARM tests as well as amd64.

## Solution
The fix for this is to update the lambda image to have an ARM binary as part of the "latest" tag. There is also an old dockerfile in e2e/custom_logic/cmd that needs to be updated to work with arm machines.

PR #8505, opened 2022-12-11
fix(ci): unskip directives and normal in graphql/e2e for arm

joshua-goldstein:
## Problem

We were skipping these tests due to lambda containers not being compatible with arm.

## Solution

We pushed a new multi-architecture lambda image and use this when running arm tests.


joshua-goldstein:
Closing in favor of #8492

PR #8506, opened 2022-12-12
fix(ACL): Prevents permissions overrride and merges acl cache to persist permissions across different namespaces (#8418)

anurags92:
In the current implementation, when you update ACL rules across
different namespaces, the previous rules are overwritten. This results
in incorrect results for other namespaces (the ones which were created
earlier).

To reproduce:
1. Create a `namespace-1` with acl rules configured
2. Create a `namespace-2` with a different set of acl rules configured
3. Login to `namespace-1` and do a query. It will not respect the rules
configured in step-1.

1. `TestTwoPermissionSetsInNameSpacesWithAcl`

(cherry picked from commit de758acec0e046b78703e8c6ff3593c99cf71029)

### DIFF
--- a/edgraph/access_ee.go
+++ b/edgraph/access_ee.go
@@ -1,3 +1,4 @@
+//go:build !oss
 // +build !oss
 
 /*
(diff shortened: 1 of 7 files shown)


edgraph/access_ee.go
```diff
@@ -1274,17 +1274,17 @@ func groupFilter(groupIds []string) *gql.FilterTree {
...
+	 addUserFilterToFilter makes sure that user can't misue filters to access other user's info.
+	 If the *filter* have type(dgraph.type.Group) or type(dgraph.type.User) functions,
+	 it generate a *newFilter* with function like eq(dgraph.xid, userId) or eq(dgraph.xid,groupId...)
+	 and return a filter of the form
+
+			&gql.FilterTree{
+				Op: "AND",
+				Child: []gql.FilterTree{
+					{filter, newFilter}
+				}
```
skrdgraph, line 1286: [nitpick ~ no need to fix] I notice the `tabbing` & related aesthetic changes coming in from your PRs - possibly from your IDE formatter. I personally think we should avoid introducing these changes, when the PR is targeting an older branch.

skrdgraph (commented):
[nitpick] we should try to keep the diff's tightened ONLY to the issue at hand. It would be nice to raise a separate PR for these aesthetic changes. I am guessing this happened because of your IDE, and you should probably minimize changes for easier review in these side-releases. (no action is needed)

edgraph/access_ee.go
anurags92: So the original change went into main and it had the formatting changes like you mentioned. Since this PR is only a cherry-pick, I brought in all the changes from the other merge. Fixing (removing) these would require changes on top of cherry-pick PR.

meghalims (approved):
LGTM

mangalaman93 (approved):
We should also merge this change to main, I do have comments but I do not want to block this change. Will review it again when we merge to main.

anurags92:
> We should also merge this change to main, I do have comments but I do not want to block this change. Will review it again when we merge to main.

This is merged in main. This is only a cherry-pick to slash branch for one customer who is facing issues. If you have any urgent comments, can you file it on the original PR or as a separate issue. I'll fix them in main.

PR #8544, opened 2022-12-16
fix(cd): Update dgraph binary builds' filenames

adityasadalage:
## Problem
cd-dgraph.yml builds dgraph binaries of the same name for both **ARM** and **amd** architectures. This causes confusion when downloading artifacts as the filename and file extensions are the same when opening in MacOS Finder. 

## Solution
Specify the right name for the kind of architecture we are building the architecture for.


joshua-goldstein (approved):
Looks good now 👍

skrdgraph:
@adityasadalage I have closed this pull request, lets wait for @MichelDiz 's thoughts here.

MichelDiz:
@skrdgraph as long you guys keep the same name of the file. Keep the binaries at first level instead of nested folder. It will be fine. For ARM, I have to add a new entry in the bash script to identify the arm machine.

PR #8546, opened 2022-12-16
chore(ci): remove arch from ldflag

joshua-goldstein:
Dgraph version ldflag should not have arch (only version number).  We update the Makefile to not add this in the ldflags build arguments.


skrdgraph (approved):
looks good!

meghalims (approved):
Looks good

PR #8549, opened 2022-12-16
fix(ci): reduce cron runs for self-hosted runners

skrdgraph:
## Problem
we are quite stable now, and we don't want to burn $'s

## Solution
reduce cron runs


adityasadalage (approved):
LGTM

PR #8556, opened 2023-01-02
add update_manifest tool (#7815)

mangalaman93:
This PR adds `dgraph updatemanifest` tool. This can be used to migrate the manifest of the backups taken on release/v21.03 or their parallels to the state after this change #7810

### DIFF
--- a/worker/backup_ee.go
+++ b/worker/backup_ee.go
@@ -560,7 +560,7 @@ func (pr *BackupProcessor) CompleteBackup(ctx context.Context, m *Manifest) erro
 	}
 	manifest.Manifests = append(manifest.Manifests, m)
 
-	if err := createManifest(handler, uri, manifest); err != nil {
+	if err := CreateManifest(handler, uri, manifest); err != nil {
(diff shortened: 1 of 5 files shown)


matthewmcneely:
@mangalaman93 Is this just for v21.03 upgrades? Will there be an option to move from say, v22.0 to the next major version (via a sub-flag?)? If it's just going to be for v21.03, maybe that should be reflected in the name of the subcommand.

worker/backup_ee.go
```diff
@@ -674,7 +674,7 @@ func checkAndGetDropOp(key []byte, l *posting.List, readTs uint64) (*pb.DropOper
 		// * DROP_NS;ns
 		// So, accordingly construct the *pb.DropOperation.
 		dropOp := &pb.DropOperation{}
-		dropInfo := strings.Split(string(val), ";")
+		dropInfo := strings.SplitN(string(val), ";", 2)
```
matthewmcneely, line 677: Is this change to enable processing of namespaces or attributes that might contain semicolons?

PR #8559, opened 2023-01-04
fix(restore): consider the banned namespaces while bumping (#7839)

mangalaman93:
We were not considering the banned namespaces while bumping the lease. So, if we restore a backup with namespace 0,1,2 not-banned and 3,4 banned. We will bump only up to 2. This would error out when creating a namespace

### DIFF
--- a/worker/restore_map.go
+++ b/worker/restore_map.go
@@ -620,6 +620,7 @@ func RunMapper(req *pb.RestoreRequest, mapDir string) (*mapResult, error) {
 	dropAll := false
 	dropAttr := make(map[string]struct{})
 	dropNs := make(map[uint64]struct{})
+	var maxBannedNs uint64
 
 	// manifests are ordered as: latest..full
 	for i, manifest := range manifests {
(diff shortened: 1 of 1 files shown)


matthewmcneely (approved):

MichelDiz:
These cherry-picks for branches confuse me a bit. But maybe it's the safest way to avoid breaking something. Only I wonder if the CI is correct. If it is running the right context. It feels like it wasn't triggered here.

mangalaman93:
Don't worry about that Michel, when we merge, we will merge to main. Feel free to review the change.

skrdgraph (dismissed):

anurags92:
> We were not considering the banned namespaces while bumping the lease. So, if we restore a backup with namespace 0,1,2 not-banned and 3,4 banned. We will bump only up to 2. This would error out when creating a namespace

- Can you explain a little bit more? 
- What if we have a scenario where `0,1` -- allowed `2,3` -- banned and `4,5,6` -- allowed ?
- Probably needs a test.

mangalaman93:
@anurags92 in your scenario, when we do restore, we will observe the namespace 6 and we will bump up to the correct value for maxNamespace. Siddhesh is working on a test already.

harshil-goel (approved):

anurags92 (approved):

PR #8560, opened 2023-01-04
fix(debug): Check length of wal entry before parsing

mangalaman93:
Check for the length of wal entry's data before parsing it for key


joshua-goldstein (approved):
Looks like this is a debug tool and instead of panicking we want to print the key.

skrdgraph (approved):
looks good 👍

PR #8561, opened 2023-01-04
fix(groot): do not upsert groot for all namespaces on restart

mangalaman93:
Earlier, whenever the alpha starts(or restarts), we were upserting guardian and groot for all the namespaces. This is not actually needed. The change was made in the PR #7759 to fix a bulk loader edge case. This PR fixes that by generating the required RDFs in the bulk loader itself. Essentially, it inserts the ACL RDFs when force loading into non-Galaxy namespace.

(cherry picked from commit 6730f10b62da7c7dc3bf5985ef957f08d3c733b4)

### DIFF
--- a/dgraph/cmd/bulk/run.go
+++ b/dgraph/cmd/bulk/run.go
@@ -118,7 +118,9 @@ func init() {
 	flag.Bool("new_uids", false,
 		"Ignore UIDs in load files and assign new ones.")
 	flag.Uint64("force-namespace", math.MaxUint64,
-		"Namespace onto which to load the data. If not set, will preserve the namespace.")
+		"Namespace onto which to load the data. If not set, will preserve the namespace."+
+			" When using this flag to load data into specific namespace, make sure that the "+
+			"load data do not have ACL data.")
 
 	flag.String("badger", BulkBadgerDefaults, z.NewSuperFlagHelp(BulkBadgerDefaults).
 		Head("Badger options (Refer to badger documentation for all possible options)").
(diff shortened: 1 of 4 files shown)


joshua-goldstein:
Adding this for future reference: https://github.com/dgraph-io/dgraph/pull/7917 
This PR reverts #7759

MichelDiz (approved):

matthewmcneely (dismissed):

harshil-goel (approved):

dgraph/cmd/bulk/run.go
```diff
@@ -118,7 +118,9 @@ func init() {
 	flag.Bool("new_uids", false,
 		"Ignore UIDs in load files and assign new ones.")
 	flag.Uint64("force-namespace", math.MaxUint64,
-		"Namespace onto which to load the data. If not set, will preserve the namespace.")
+		"Namespace onto which to load the data. If not set, will preserve the namespace."+
+			" When using this flag to load data into specific namespace, make sure that the "+
```
anurags92, line 122: What's the behaviour in case it does have the ACL data?

mangalaman93: I think it will just sail through into the p directory though it is probably not recommended. The user should rather add rules using dgraph APIs. Maybe, we could check for this during bulk loader.

harshil-goel (approved):

anurags92 (approved):

matthewmcneely (approved):

PR #8562, opened 2023-01-04
opt(schema): Load schema and types using Stream framework 

mangalaman93:
For big datasets, we're seeing a big slowdown due to loading schema and types serially using a single iterator. Using the Stream framework, makes this metadata loading step much faster, resulting in a much faster Alpha initialization.

(cherry picked from commit d03d5ad14e8ec5044d1e7f16c8be271d890346f1)

### DIFF
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -17,23 +17,23 @@
 package schema
 
 import (
-	"bytes"
 	"context"
 	"encoding/hex"
 	"fmt"
 	"math"
 	"sync"
 
-	"github.com/golang/glog"
-	"github.com/golang/protobuf/proto"
-	"github.com/pkg/errors"
-	"golang.org/x/net/trace"
-
 	"github.com/dgraph-io/badger/v3"
+	badgerpb "github.com/dgraph-io/badger/v3/pb"
 	"github.com/dgraph-io/dgraph/protos/pb"
 	"github.com/dgraph-io/dgraph/tok"
 	"github.com/dgraph-io/dgraph/types"
 	"github.com/dgraph-io/dgraph/x"
+
+	"github.com/golang/glog"
+	"github.com/golang/protobuf/proto"
+	"github.com/pkg/errors"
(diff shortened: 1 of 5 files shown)


anurags92 (approved):

schema/schema.go
```diff
@@ -499,82 +499,72 @@ func Load(predicate string) error {
...
+				State().SetType(pk.Attr, t)
+				return nil
+			})
+			return nil, err
+		}
+		glog.Fatalf("Invalid load type")
+		return nil, errors.New("shouldn't reach here")
 	}
-	return nil
+	return stream.Orchestrate(context.Background())
```
matthewmcneely, line 567: Shouldn't the stream use the context of the group's closer?

mangalaman93: what do you mean by group's closer?

mangalaman93: improved the code a bit

schema/schema.go
```diff
@@ -498,83 +498,73 @@ func Load(predicate string) error {
...
+
+		switch loadType {
+		case loadSchema:
+			var s pb.SchemaUpdate
+			err := item.Value(func(val []byte) error {
+				if len(val) == 0 {
+					s = pb.SchemaUpdate{Predicate: pk.Attr, ValueType: pb.Posting_DEFAULT}
+				}
+				x.Checkf(s.Unmarshal(val), "Error while loading schema from db")
+				State().Set(pk.Attr, &s)
```
harshil-goel, line 548: One place it's &s, other place it is just t. Any reason for this?

harshil-goel (commented):
LGTM. Left a question. 
I feel like the stream framework api is a bit weird. Like we should be just passing the KeyToList function in the object creation itself? What do you think, maybe we should take a look at all the apis that badger has exposed.

schema/schema.go
mangalaman93: Done.

harshil-goel (approved):

PR #8564, opened 2023-01-04
feat(cdc): Add superflag to enable TLS without CA or certs (#7946)

mangalaman93:
Does the following changes:
 * will attempt to connect to Kafka over TLS using the system certs.
 * add helper function x.TLSBaseConfig.
 * sets the min TLS version to v1.2 along with the minimum cipher suites.

### DIFF
--- a/worker/server_state.go
+++ b/worker/server_state.go
@@ -45,7 +45,7 @@ const (
 	SecurityDefaults  = `token=; whitelist=;`
 	LudicrousDefaults = `enabled=false; concurrency=2000;`
 	CDCDefaults       = `file=; kafka=; sasl_user=; sasl_password=; ca_cert=; client_cert=; ` +
-		`client_key=; sasl-mechanism=PLAIN;`
+		`client_key=; sasl-mechanism=PLAIN; tls=false;`
 	LimitDefaults = `mutations=allow; query-edge=1000000; normalize-node=10000; ` +
 		`mutations-nquad=1000000; disallow-drop=false; query-timeout=0ms; txn-abort-after=5m; ` +
 		` max-retries=-1;max-pending-queries=10000`
--- a/worker/sink_handler.go
+++ b/worker/sink_handler.go
@@ -86,10 +86,18 @@ func newKafkaSink(config *z.SuperFlag) (Sink, error) {
 	saramaConf.Producer.Return.Successes = true
 	saramaConf.Producer.Return.Errors = true
 
-	if config.GetPath("ca-cert") != "" {
-		tlsCfg := &tls.Config{
-			MinVersion: tls.VersionTLS12,
+	if config.GetBool("tls") && config.GetPath("ca-cert") == "" {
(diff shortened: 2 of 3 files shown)


anurags92:
@mangalaman93 @Rajakavitha1 when we update main with new flags, I think we should flag it for doc update etc.? What do you think about that? Should we create a new issue/tag someone here/add a label?

mangalaman93:
we will have to do it for all the changes. We can do it once slash and main are more aligned. We will just go through the list once again I guess.

worker/sink_handler.go
```diff
@@ -86,10 +86,18 @@ func newKafkaSink(config *z.SuperFlag) (Sink, error) {
 	saramaConf.Producer.Return.Successes = true
 	saramaConf.Producer.Return.Errors = true
 
-	if config.GetPath("ca-cert") != "" {
-		tlsCfg := &tls.Config{
-			MinVersion: tls.VersionTLS12,
+	if config.GetBool("tls") && config.GetPath("ca-cert") == "" {
```
matthewmcneely, line 89: What's the action if one sets `tls=true` but doesn't specify `ca-cert`?

worker/sink_handler.go
```diff
@@ -86,10 +86,18 @@ func newKafkaSink(config *z.SuperFlag) (Sink, error) {
...
+		tlsCfg := x.TLSBaseConfig()
```
matthewmcneely, line 90: There's quite a lot of this duplicated in the `elif` test below. Not a deal-breaker, but definitely could be cleaned up.

matthewmcneely:
@meghalims Kafka CDC (and by extension tls with Kafka) is completely uncovered by tests. Maybe a good task for a junior?

worker/sink_handler.go
mangalaman93: we just use the system certs

worker/sink_handler.go
mangalaman93: a few lines yeah. I tried putting this code inside the TLSBaseConfig function but realized that it can error out while reading system certs. We may not want that in certain cases.

matthewmcneely (approved):

worker/server_state.go
```diff
@@ -45,7 +45,7 @@ const (
 	SecurityDefaults  = `token=; whitelist=;`
 	LudicrousDefaults = `enabled=false; concurrency=2000;`
 	CDCDefaults       = `file=; kafka=; sasl_user=; sasl_password=; ca_cert=; client_cert=; ` +
-		`client_key=; sasl-mechanism=PLAIN;`
+		`client_key=; sasl-mechanism=PLAIN; tls=false;`
```
harshil-goel, line 48: Why default False?

mangalaman93: we can change the default later on. This could cause problems to our current customer.

harshil-goel (approved):

harshil-goel (approved):

PR #8510, opened 2022-12-13
fix(backup): use StreamWriter instead of KVLoader during backup restore

mangalaman93:
cherry-pick PR https://github.com/dgraph-io/dgraph/pull/7753
    
This commit is a major rewrite of online restore code. It used to use KVLoader in badger. Now it instead uses StreamWriter that is much faster for writes in the case of restore.

PR #8566, opened 2023-01-04
fix(probe): do not contend for lock in lazy load (#8037) (#8041)

mangalaman93:
Earlier the admin server mutex lock was used to protect the graphql schema map. But now we store that in schema store that internally handles the concurrency. Hence, we don't need to take the admin server's read lock to access schema.

/probe/graphql is used as health check and is called very frequently. This rlock on adminserver mutex makes the /probe/graphql requests block while lazy loading when restore operation gets triggered at the startup. That leads to so many go routines being spun up.

(cherry picked from commit 5ad40d8479163697049612f28ba40429652b765a)

### DIFF
--- a/graphql/admin/admin.go
+++ b/graphql/admin/admin.go
@@ -1026,12 +1026,9 @@ func (as *adminServer) resetSchema(ns uint64, gqlSchema schema.Schema) {
 
 func (as *adminServer) lazyLoadSchema(namespace uint64) error {
 	// if the schema is already in memory, no need to fetch it from disk
-	as.mux.RLock()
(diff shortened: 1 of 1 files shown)


anurags92:
@mangalaman93 It would be helpful to add a link to the PR which changes the graphql schema store in the PR description to verify this.

mangalaman93:
This change is there in part of the #8510

mangalaman93:
The old PR is #7925

matthewmcneely (approved):

skrdgraph (approved):

harshil-goel (approved):

anurags92 (approved):

PR #8573, opened 2023-01-04
fix(zero): fix update membership to make bulk tablet proposal instead of multiple small (#8090)

anurags92:
Converting the smaller tablet proposals into a single bulk call makes the execution faster. Else, this update becomes of the order of O(n^2), where n is the number of tablet updates.

(cherry picked from commit 695c6d763b9e75edb0e47dd1f32a0af98222511a)


skrdgraph (approved):
Looks good!

mangalaman93 (approved):

PR #8574, opened 2023-01-04
fix(GraphQL): Pass on HTTP request headers for subscriptions (#7806) (#7807)

anurags92:
Include headers from subscribe request and pass it along the downstream for processing.

Co-authored-by: aman-bansal <bansalaman2905@gmail.com>

Co-authored-by: Abhimanyu Singh Gaur <12651351+abhimanyusinghgaur@users.noreply.github.com>
(cherry picked from commit dbd57443992dd9f208f334095911069b3b3efa3f)

### DIFF
--- a/graphql/admin/http.go
+++ b/graphql/admin/http.go
@@ -158,10 +158,10 @@ func (gs *graphqlSubscription) Subscribe(
 	variableValues map[string]interface{}) (payloads <-chan interface{},
 	err error) {
 
+	reqHeader := http.Header{}
 	// library (graphql-transport-ws) passes the headers which are part of the INIT payload to us
 	// in the context. We are extracting those headers and passing them along.
 	headerPayload, _ := ctx.Value("Header").(json.RawMessage)
-	reqHeader := http.Header{}
 	if len(headerPayload) > 0 {
 		headers := make(map[string]interface{})
 		if err = json.Unmarshal(headerPayload, &headers); err != nil {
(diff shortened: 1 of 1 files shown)


MichelDiz (approved):

graphql/admin/http.go
```diff
@@ -175,14 +175,31 @@ func (gs *graphqlSubscription) Subscribe(
...
 		OperationName: operationName,
 		Query:         document,
 		Variables:     variableValues,
 		Header:        reqHeader,
 	}
 	namespace := x.ExtractNamespaceHTTP(&http.Request{Header: reqHeader})
-	LazyLoadSchema(namespace) // first load the schema, then do anything else
+	glog.Infof("namespace: %d. Got GraphQL request over websocket.", namespace)
+	// first load the schema, then do anything else
+	if err = LazyLoadSchema(namespace); err != nil {
```
mangalaman93, line 200: :=

mangalaman93:
the description of the commit is not accurate, please fix it.

mangalaman93 (changes requested):

graphql/admin/http.go
anurags92: Do you want to re-initialize the named return parameter `err`? I don't think that's idiomatic go. See a [function](https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/io/io.go;drc=bae7d772e800207ed764b70a9e5963f0c1e8f0aa;l=326-340) which I could find in the core lib which has similar named parameter. Inadvertently `:=` leads to variable shadowing which can cause issues I guess. I don't foresee what could fail in this case, but I remember reading that we should avoid variable shadowing with named return parameters.

mangalaman93: I didn't realize that. Now I think we should remove the named parameter completely. I don't think named parameters are a good idea.

anurags92: That sounds better. I have removed named parameters from the function. Apart from not needing named params here, we weren't using `payload` anywhere in the function. So it didn't really need a named param.

mangalaman93 (approved):
please squash & merge

PR #8592, opened 2023-01-06
fix(ci): adjust cron schedules

skrdgraph:
## Problem
Our scheduled runs happen too often. We now have a stable baseline - and we can limit these runs. We are also adjusting triggering evented runs when a PR is opened, and when a PR is merged*. Reducing the scheduled run makes more sense now.

## Solution
changing these runs to happen twice daily


kevinmingtarja (approved):
LGTM!

adityasadalage (approved):
LGTM

PR #8599, opened 2023-01-10
BREAKING: fix json marshal unmarshal for namespace > 127 (#7810)

harshil-goel:

Live loader
Backup and List Backup
Http clients and Ratel
Fix:
New Format: <anmespace>-<attribute> (- is the hyphen literal)


x/keys.go
```diff
@@ -84,21 +86,25 @@ func GalaxyAttr(attr string) string {
 
 // ParseNamespaceAttr returns the namespace and attr from the given value.
 func ParseNamespaceAttr(attr string) (uint64, string) {
-	return binary.BigEndian.Uint64([]byte(attr[:8])), attr[8:]
+	splits := strings.SplitN(attr, NsSeparator, 2)
+	return strToUint(splits[0]), splits[1]
 }
 
 func ParseNamespaceBytes(attr string) ([]byte, string) {
```
mangalaman93, line 93: can we live without this function?

mangalaman93 (changes requested):
Almost LGTM

PR #8600, opened 2023-01-10
add update_manifest tool (#7815)

harshil-goel:
This PR adds `dgraph updatemanifest` tool. This can be used to migrate the manifest of the backups taken on release/v21.03 or their parallels to the state after this change #7810


mangalaman93:
I think, we should remove this tool. We should make sure that later versions of Dgraph are backward compatible.

PR #8601, opened 2023-01-10
Store namespace in predicate as a hex separated by a hyphen to prevent json marshal issues

harshil-goel:

- Live loader using guardian of galaxy
- Backup and List Backup
- Http clients and Ratel
- Schema and predicate

Fix:
New Format: <namespace>-<attribute> (- is the hyphen literal) <namespace> is a string "81" in hex

We also update the manifest version after update. This diff takes care that older backups are still compatible and can be used to restore.


mangalaman93 (changes requested):

worker/backup_ee.go
```diff
@@ -554,13 +554,13 @@ func (pr *BackupProcessor) CompleteBackup(ctx context.Context, m *Manifest) erro
...
-	manifest, err := GetManifest(handler, uri)
+	manifest, err := GetManifestNoUpgrade(handler, uri)
 	if err != nil {
 		return err
 	}
 	manifest.Manifests = append(manifest.Manifests, m)
 
-	if err := createManifest(handler, uri, manifest); err != nil {
+	if err := CreateManifest(handler, uri, manifest); err != nil {
 		return errors.Wrap(err, "Complete backup failed")
```
matthewmcneely, line 564: Realize this is not your doing, but error text should not be capitalized

worker/restore_map.go
```diff
@@ -721,15 +721,14 @@ func RunMapper(req *pb.RestoreRequest, mapDir string) (*mapResult, error) {
...
-						return nil, errors.Wrap(err, "Map phase failed to parse namespace")
-					}
+				if op.DropValue == "" {
+					// In 2103, we do not support namespace level drop data.
+					dropAll = true
+					continue
+				}
+				ns, err := strconv.ParseUint(op.DropValue, 0, 64)
+				if err != nil {
+					return nil, errors.Wrap(err, "Map phase failed to parse namespace")
```
matthewmcneely, line 731: Error text should not be capitalized

matthewmcneely (changes requested):

ee/updatemanifest/run.go
```diff
@@ -0,0 +1,137 @@
+// +build !oss
+
+/*
+ * Copyright 2021 Dgraph Labs, Inc. and Contributors
```
mangalaman93, line 4: fix the license

x/keys.go
```diff
@@ -57,8 +57,23 @@ const (
...
 	NamespaceOffset = 1
+	// NsSeparator is the separator between between the namespace and attribute.
+	NsSeparator = "-"
 )
 
+// Invalid bytes are replaced with the Unicode replacement rune.
+// See https://golang.org/pkg/encoding/json/#Marshal
+const replacementRune = rune('\ufffd')
+
+func AttrFrom2103(attr string) (string, error) {
```
mangalaman93, line 68: is this function defined again here?

mangalaman93: And can we explain what is going on here? maybe an example in the comments.

mangalaman93 (changes requested):

mangalaman93 (dismissed):

mangalaman93 (approved):

matthewmcneely (approved):

PR #8602, opened 2023-01-10
BREAKING: fix json marshal unmarshal for namespace > 127 (#7810,#7838,#8091) #8599

harshil-goel:
We used to store predicate as | (pipe | signifies concatenation). We store this as a string. is 8 bytes uint64, which when marshaled to JSON bytes mess up the predicate. This is because for the namespace greater than 127, the UTF-8 encoding might take up several bytes (also if the mapping does not exist, then it replaces it with some other rune). This affects three identified places in Dgraph:

Live loader
Backup and List Backup
Http clients and Ratel
Fix:
New Format: - (- is the hyphen literal)

### DIFF
--- a/dgraph/cmd/alpha/http_test.go
+++ b/dgraph/cmd/alpha/http_test.go
@@ -428,7 +428,7 @@ func TestTransactionBasic(t *testing.T) {
 	require.Equal(t, 2, len(mr.preds))
 	var parsedPreds []string
 	for _, pred := range mr.preds {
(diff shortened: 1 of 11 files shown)


mangalaman93 (dismissed):

matthewmcneely:
> A better idea is to use UTF-8 string even for internal operations.

@harshil-goel Was this accomplished in this PR? Have all use-cases been addressed with this namespace formatting change (migration, etc)?

mangalaman93 (approved):

PR #8607, opened 2023-01-16
chore(CI): Update Labeler

MichelDiz:
## Problem

Labeler needs update and cover more places.

## Solution

Edit the CI configs.


.github/workflows/labeler.yml
```diff
@@ -3,12 +3,15 @@ on:
...
   label:
+    permissions:
+      contents: read
+      pull-requests: write
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/labeler@v2
-      with:
-        repo-token: "${{ secrets.GITHUB_TOKEN }}"
+      - uses: actions/labeler@v4
```
skrdgraph, line 15: this tabbing doesn't look correct

skrdgraph: yeah that looks right, resolving

.github/workflows/labeler.yml
```diff
@@ -3,12 +3,15 @@ on:
   pull_request:
     branches:
       - main
-      - 'release/**'
+      - "release/**"
```
skrdgraph, line 6: could you label to `release/v**`?

skrdgraph (dismissed):
looks okay to me. **Please do not merge without @rderbier's review on this PR**.

@rderbier your review is needed here. can you review `label` names? and see if we want to create something new instead of using the old ones?

anurags92 (approved):

.github/labeler.yml
```diff
@@ -14,11 +15,37 @@ area/integrations:
   - .github/**/*
   - .travis/**/*
 area/testing/jepsen:
-  - contrib/jepsen/**/* 
+  - contrib/jepsen/**/*
 area/commercial:
   - ee/**/*
-  - wiki/content/enterprise-features/**/*
+area/enterprise:
```
mangalaman93, line 21: why do we have two tags for this code, area/enterprise and area/commercial?

mangalaman93 (changes requested):

.github/labeler.yml
MichelDiz: The two were already there when I arrived. So I used both. Which should we use? At the time, no one clarified their use, so I adopted both.

MichelDiz: See 
https://github.com/dgraph-io/dgraph/labels/area%2Fcommercial
https://github.com/dgraph-io/dgraph/labels/area%2Fenterprise

Enterprise looks more used.

mangalaman93: let's use that then.

mangalaman93 (approved):

rderbier (approved):

PR #8614, opened 2023-01-18
bug(core): Fixed infinite loop in CommitToDisk

harshil-goel:
Whenever an error happens in CommitToDisk, we retry the function according to x.config.MaxRetries. However, the value was set to -1, which causes the function to get stuck in infinite loop. This leads to dgraph not being able to commit the transaction, nor move forward to newer queries. 

CommitToDisk function is required to be pushed to disk. In case of an error, different alphas in a group can end up having different data leading to data loss. To avoid such issues, we panic in case we are not able to CommitToDisk after 10 retries. Once the alpha is restarted, if the issue is fixed, the alpha would start to work again. This way, Alphas wont fail silently and we would know if there was any issue occuring.

Fixes: https://github.com/dgraph-io/projects/issues/85

### DIFF
--- a/worker/draft.go
+++ b/worker/draft.go
@@ -819,8 +819,10 @@ func (n *node) commitOrAbort(pkey uint64, delta *pb.OracleDelta) error {
 			return
 		}
 		txn.Update()
-		err := x.RetryUntilSuccess(int(x.Config.MaxRetries),
-			10*time.Millisecond, func() error {
+		// We start with 20 ms, so that we end up waiting 5 mins by the end.
+		// If there is any transient issue, it should get fixed within that timeframe.
+		err := x.ExponentialRetry(int(x.Config.MaxRetries),
+			20*time.Millisecond, func() error {
 				err := txn.CommitToDisk(writer, commit)
 				if err == badger.ErrBannedKey {
 					glog.Errorf("Error while writing to banned namespace.")
@@ -832,6 +834,7 @@ func (n *node) commitOrAbort(pkey uint64, delta *pb.OracleDelta) error {
 		if err != nil {
 			glog.Errorf("Error while applying txn status to disk (%d -> %d): %v",
 				start, commit, err)
+			panic(err)
 		}
 	}
 
--- a/worker/server_state.go
+++ b/worker/server_state.go
@@ -48,7 +48,7 @@ const (
 		`client_key=; sasl-mechanism=PLAIN; tls=false;`
 	LimitDefaults = `mutations=allow; query-edge=1000000; normalize-node=10000; ` +
 		`mutations-nquad=1000000; disallow-drop=false; query-timeout=0ms; txn-abort-after=5m; ` +
-		` max-retries=-1;max-pending-queries=10000`
+		` max-retries=10;max-pending-queries=10000`
 	ZeroLimitsDefaults = `uid-lease=0; refill-interval=30s; disable-admin-http=false;`
 	GraphQLDefaults    = `introspection=true; debug=false; extensions=true; poll-interval=1s; ` +
 		`lambda-url=;`
--- a/x/x.go
+++ b/x/x.go
@@ -613,11 +613,27 @@ func Max(a, b uint64) uint64 {
 	return b
 }
 
+// ExponentialRetry runs the given function until it succeeds or can no longer be retried.
+func ExponentialRetry(maxRetries int, waitAfterFailure time.Duration,
+	f func() error) error {
+	var err error
+	for retry := maxRetries; retry > 0; retry-- {
+		if err = f(); err == nil {
+			return nil
+		}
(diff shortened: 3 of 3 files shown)


skrdgraph (approved):

skrdgraph (approved):

x/x.go
```diff
@@ -619,7 +619,7 @@ func Max(a, b uint64) uint64 {
 func RetryUntilSuccess(maxRetries int, waitAfterFailure time.Duration,
 	f func() error) error {
 	var err error
-	for retry := maxRetries; retry != 0; retry-- {
+	for retry := maxRetries; retry > 0; retry-- {
```
nswamy, line 622: so this fails transactions with large keys ?

worker/server_state.go
```diff
@@ -48,7 +48,7 @@ const (
 		`client_key=; sasl-mechanism=PLAIN;`
 	LimitDefaults = `mutations=allow; query-edge=1000000; normalize-node=10000; ` +
 		`mutations-nquad=1000000; disallow-drop=false; query-timeout=0ms; txn-abort-after=5m; ` +
-		` max-retries=-1;max-pending-queries=10000`
+		` max-retries=3;max-pending-queries=10000`
```
damonfeldman, line 51: It seems like a larger impact that we give up on retries after 3 tries. Looking at x.go::commitOrAbort() and x.go::RetryUntilSuccess() it looks like we will try 3 times 10ms apart with this setting.

Will this have a big impact if there is a short-term issue like a pod going down, some kind of stuck transaction? E.g. I notice that CommmitToDisk() waits for the cache.Lock() mutex. I'm worried that some contention will cause transactions to be aborted.

Perhaps more retries with exponential backoff would be safer, but I am not sure.

matthewmcneely: +1 on the backoff approach. 3x10ms might not be enough time for network hiccups.

There doesn't seem to be a backoff func in the x package, but easily added.

damonfeldman: I happened to just see this comment in oracle.go, which suggests we need to retry aggressively in some cases:

	// NOTE: It is important that we continue retrying proposeTxn until we succeed. This should
	// happen, irrespective of what the user context timeout might be. We check for it before
	// reaching this stage, but now that we're here, we have to ensure that the commit proposal goes
	// through. Otherwise, we should block here forever. If we don't do this, we'll see txn
	// violations in Jepsen, because we'll send out a MaxAssigned higher than a commit, which would
	// cause newer txns to see older data.

x/x.go
harshil-goel: No, Sorry, this issue doesn't fix the transaction with large keys. Currently if we have large keys, the write to badger api call fails. Once that request fails, the alpha gets stuck in a infinite loop, which causes it to not process anything further. This diff fixes the infinite loop. Aman has a diff, where we parse keys and fail the request before hand itself.

worker/draft.go
```diff
@@ -819,8 +819,8 @@ func (n *node) commitOrAbort(pkey uint64, delta *pb.OracleDelta) error {
 			return
 		}
 		txn.Update()
-		err := x.RetryUntilSuccess(int(x.Config.MaxRetries),
-			10*time.Millisecond, func() error {
+		err := x.ExponentialRetryUntilSuccess(int(x.Config.MaxRetries),
```
mangalaman93, line 822: ExponentialRetry?

worker/draft.go
```diff
@@ -819,8 +819,8 @@ func (n *node) commitOrAbort(pkey uint64, delta *pb.OracleDelta) error {
...
+			20*time.Millisecond, func() error {
```
mangalaman93, line 823: why 20ms?

worker/draft.go
```diff
@@ -832,6 +832,7 @@ func (n *node) commitOrAbort(pkey uint64, delta *pb.OracleDelta) error {
 		if err != nil {
 			glog.Errorf("Error while applying txn status to disk (%d -> %d): %v",
 				start, commit, err)
+			panic(err)
```
mangalaman93, line 835: can we add more logs here, enough to figure out the issue?

x/x.go
```diff
@@ -613,11 +613,27 @@ func Max(a, b uint64) uint64 {
 	return b
 }
 
+// RetryUntilSuccess runs the given function until it succeeds or can no longer be retried.
```
mangalaman93, line 616: change the comment

mangalaman93 (changes requested):

worker/draft.go
mangalaman93: I think panic is the most graceful way to die! The other things that we could do is retry but we already doing that.

mangalaman93 (dismissed):

mangalaman93 (approved):

anurags92 (approved):

PR #8615, opened 2023-01-18
Change intersect heuristic from 500 to 100

harshil-goel:
Currently for intersecting two different arrays, we either use linear or binary jump. This is decided on the basis of the ratio of the two arrays being intersected. Currently, the ratio is set to be 500. We have had certain customer issues where queries too long. After debugging, the issue seems to be the ratio. Changing the ratio back to 100 fixes the customer issues. Query time went from 9 sec to 0.3sec.


anurags92 (approved):

mangalaman93 (changes requested):
Can we write a benchmark for it showing that it does what you are saying and later on we can track if it goes the other way round due to a change?

harshil-goel:
We have better algorithm now

PR #8622, opened 2023-01-20
fix(proposalKey): Proposal key initialisation with correct byte ordering

anurags92:
Proposal key initialisation has a bug where we want to reserve first 2 bytes for node id but do not do because  we read random bytes in all 8 bytes and do a logical OR. This PR adds a test case and fixes the logic of initialising the proposal key.

### DIFF
--- a/dgraph/cmd/zero/raft.go
+++ b/dgraph/cmd/zero/raft.go
@@ -93,8 +93,10 @@ func (n *node) initProposalKey(id uint64) error {
 		return err
 	}
 	proposalKey = n.Id<<48 | uint64(binary.BigEndian.Uint32(random4Bytes))<<16
-	// We want to avoid spillage to node id in case of overflow. So by setting 48th bit to
-	// 0 we ensure that we never overflow even if our random bytes are all 1s
+	// We want to avoid spillage to node id in case of overflow. For instance, if the
+	// random bytes end up being [xx,xx, 255, 255, 255, 255, 0 , 0] (xx, xx being the node id)
+	// we would spill to node id after 65535 calls to unique key.
+	// So by setting 48th bit to 0 we ensure that we never spill out to node ids.
 	proposalKey &= ^(uint64(1) << 47)
 	return nil
 }
--- a/dgraph/cmd/zero/raft.go
+++ b/dgraph/cmd/zero/raft.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"log"
 	"math"
+	"math/rand"
 	"sort"
 	"strings"
 	"sync"
@@ -88,6 +89,9 @@ func (n *node) AmLeader() bool {
 func (n *node) initProposalKey(id uint64) error {
 	x.AssertTrue(id != 0)
 	random4Bytes := make([]byte, 4)
+	if _, err := rand.Read(random4Bytes); err != nil {
+		return err
+	}
 	proposalKey = n.Id<<48 | uint64(binary.BigEndian.Uint32(random4Bytes))<<16
 	return nil
 }
--- a/dgraph/cmd/zero/raft.go
+++ b/dgraph/cmd/zero/raft.go
@@ -18,11 +18,11 @@ package zero
 
 import (
 	"context"
+	"crypto/rand"
 	"encoding/binary"
 	"fmt"
 	"log"
 	"math"
-	"math/rand"
 	"sort"
 	"strings"
 	"sync"
--- a/dgraph/cmd/zero/raft.go
+++ b/dgraph/cmd/zero/raft.go
@@ -96,10 +96,6 @@ func (n *node) initProposalKey(id uint64) error {
 	return nil
 }
 
-func (n *node) proposalKey() uint64 {
-	return proposalKey
-}
-
 func (n *node) uniqueKey() uint64 {
 	return atomic.AddUint64(&proposalKey, 1)
 }
--- a/dgraph/cmd/zero/raft.go
+++ b/dgraph/cmd/zero/raft.go
@@ -18,7 +18,6 @@ package zero
 
 import (
 	"context"
-	"crypto/rand"
 	"encoding/binary"
 	"fmt"
 	"log"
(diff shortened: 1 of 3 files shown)


dgraph/cmd/zero/zero_test.go
```diff
@@ -83,3 +89,30 @@ func TestIdBump(t *testing.T) {
 	res, err = zc.AssignIds(ctx, &pb.Num{Val: 10, Type: pb.Num_UID, Bump: true})
 	require.Contains(t, err.Error(), "Nothing to be leased")
 }
+
+func extractNodeIdFrom(proposalKey uint64) uint64 {
```
mangalaman93, line 93: can't we just shift the uint64 number by 6*8=48 and get the node ID?

dgraph/cmd/zero/zero_test.go
```diff
@@ -83,3 +89,30 @@ func TestIdBump(t *testing.T) {
...
+
+	return binary.BigEndian.Uint64(numSlice)
+}
+
+func TestProposalKey(t *testing.T) {
+	dir, err := ioutil.TempDir("", "test_pk")
+	require.NoError(t, err)
+	defer os.RemoveAll(dir)
+
+	store := raftwal.Init(dir)
```
mangalaman93, line 109: do you need to call this? even when store is nil inside the node, the call to uniqueKey and initProposalKey should work

dgraph/cmd/zero/zero_test.go
```diff
@@ -83,3 +89,30 @@ func TestIdBump(t *testing.T) {
...
+
+	id := uint64(2)
+	rc := &pb.RaftContext{Id: id}
+	n := conn.NewNode(rc, store, nil)
+	node := &node{Node: n, ctx: context.Background(), closer: z.NewCloser(1)}
+	node.initProposalKey(node.Id)
+
+	require.Equal(t, id, extractNodeIdFrom(node.proposalKey()), "id extracted from proposal key is not equal to initial value")
```
mangalaman93, line 117: can you call it a few times in the test and make sure that all the numbers are unique?

mangalaman93 (changes requested):

dgraph/cmd/zero/raft.go
```diff
@@ -88,13 +88,20 @@ func (n *node) AmLeader() bool {
...
+	random4Bytes := make([]byte, 4)
+	if _, err := rand.Read(random4Bytes); err != nil {
 		return err
 	}
+	copy(b[4:8], random4Bytes)
 	proposalKey = n.Id<<48 | binary.BigEndian.Uint64(b)<<16
 	return nil
 }
 
+func (n *node) proposalKey() uint64 {
```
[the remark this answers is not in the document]
anurags92: Its not a field but a global variable.

dgraph/cmd/zero/zero_test.go
anurags92: Yes, this is required for correct initialization. The code crashes with nil store.

dgraph/cmd/zero/raft.go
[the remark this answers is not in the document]
mangalaman93: we can access the global variable in the test too.

dgraph/cmd/zero/zero_test.go
mangalaman93: You only need a node object, what code/function causes crash?

dgraph/cmd/zero/raft.go
```diff
@@ -18,11 +18,11 @@ package zero
 
 import (
 	"context"
-	"crypto/rand"
 	"encoding/binary"
 	"fmt"
 	"log"
 	"math"
+	"math/rand"
```
mangalaman93, line 25: this should be crypto/rand I believe.

dgraph/cmd/zero/zero_test.go
anurags92: https://github.com/dgraph-io/dgraph/blob/c04ce807f9399cc9d00159d54333d7a6a1de18fe/conn/node.go#L90

dgraph/cmd/zero/zero_test.go
```diff
@@ -83,3 +88,34 @@ func TestIdBump(t *testing.T) {
 	res, err = zc.AssignIds(ctx, &pb.Num{Val: 10, Type: pb.Num_UID, Bump: true})
 	require.Contains(t, err.Error(), "Nothing to be leased")
 }
+
+func extractNodeIdFrom(proposalKey uint64) uint64 {
```
mangalaman93, line 92: we probably don't need this one liner now

dgraph/cmd/zero/zero_test.go
mangalaman93: Can we not live without calling newNode? We can just create the object ourselves and call these functions.

anurags92: Removed.

mangalaman93 (approved):
Looks good. Can you update the description to maybe talk about the bug instead of the solution? or maybe both?

dgraph/cmd/zero/raft.go
```diff
@@ -80,8 +84,19 @@ func (n *node) AmLeader() bool {
 	return time.Since(n.lastQuorum) <= 5*time.Second
 }
 
+// {2 bytes Node ID} {4 bytes for random} {2 bytes zero}
```
matthewmcneely, line 87: NIT: this is duplicated in the worker package (below).  The proposalKey concept could be implemented in the x package or elsewhere... not a deal-breaker...

matthewmcneely (approved):

dgraph/cmd/zero/raft.go
```diff
@@ -88,11 +88,11 @@ func (n *node) AmLeader() bool {
 // {2 bytes Node ID} {4 bytes for random} {2 bytes zero}
 func (n *node) initProposalKey(id uint64) error {
 	x.AssertTrue(id != 0)
-	b := make([]byte, 8)
-	if _, err := rand.Read(b); err != nil {
+	random4Bytes := make([]byte, 4)
```
mangalaman93, line 91: let's keep the 17th bit from the left as 0 when initializing the proposal key. That will ensure that the node ID bytes won't be modified during calls to uniqueKey.

dgraph/cmd/zero/raft.go
```diff
@@ -88,11 +88,14 @@ func (n *node) AmLeader() bool {
...
+	if _, err := rand.Read(random4Bytes); err != nil {
 		return err
 	}
-	proposalKey = n.Id<<48 | binary.BigEndian.Uint64(b)<<16
+	proposalKey = n.Id<<48 | uint64(binary.BigEndian.Uint32(random4Bytes))<<16
+	// We want to avoid spillage to node id in case of overflow. So by setting 48th bit to
```
mangalaman93, line 96: Maybe add more details of why there could be spillage and why setting 48th bit to 0 would solve it. It may not be immediately obvious to the person reading the code.

mangalaman93 (approved):

PR #8623, opened 2023-01-20
fix(mutation): validate mutation before applying it

mangalaman93:
This change ensures that the value in the mutation is not too big. The challenge here is that the keys in badger have a limitation on their size (< 2<<16). We need to ensure that no key, either primary or secondary index key is bigger than that.

Fixes https://github.com/dgraph-io/projects/issues/73

### DIFF
--- a/edgraph/server.go
+++ b/edgraph/server.go
@@ -54,6 +54,7 @@ import (
 	"github.com/dgraph-io/dgraph/query"
 	"github.com/dgraph-io/dgraph/schema"
 	"github.com/dgraph-io/dgraph/telemetry"
+	"github.com/dgraph-io/dgraph/tok"
 	"github.com/dgraph-io/dgraph/types"
 	"github.com/dgraph-io/dgraph/types/facets"
 	"github.com/dgraph-io/dgraph/worker"
@@ -608,6 +609,11 @@ func (s *Server) doMutate(ctx context.Context, qc *queryContext, resp *api.Respo
 		},
 	}
 
+	// ensure that we do not insert very large (> 64 KB) value
+	if err := validateMutation(ctx, edges); err != nil {
+		return err
+	}
+
 	qc.span.Annotatef(nil, "Applying mutations: %+v", m)
 	resp.Txn, err = query.ApplyMutations(ctx, m)
 	qc.span.Annotatef(nil, "Txn Context: %+v. Err=%v", resp.Txn, err)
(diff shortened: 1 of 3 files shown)


edgraph/server.go
```diff
@@ -675,6 +681,52 @@ func (s *Server) doMutate(ctx context.Context, qc *queryContext, resp *api.Respo
 	return nil
 }
 
+// validateMutation ensures that the value in the edge is not too big.
+// The challange here is that the keys in badger have a limitation on their size (< 2<<16).
+// We need to ensure that no key, either primary or secondary index key is bigger than that.
+// See here for more details: https://github.com/dgraph-io/projects/issues/73
+func validateMutation(ctx context.Context, edges []*pb.DirectedEdge) error {
+	const maxKeySize = 63 << 10
```
anurags92, line 689: Add some comment on this.

mangalaman93: Done.

edgraph/server.go
```diff
@@ -668,6 +674,56 @@ func (s *Server) doMutate(ctx context.Context, qc *queryContext, resp *api.Respo
...
+		}
+		// only string type can have large values that could cause us issues later
+		if update.GetValueType() != pb.Posting_STRING {
+			continue
+		}
+
+		storageVal := types.Val{Tid: types.TypeID(e.GetValueType()), Value: e.GetValue()}
+		schemaVal, err := types.Convert(storageVal, types.TypeID(update.GetValueType()))
+		if err != nil {
+			return fmt.Errorf("too large value to store: %w", err)
```
matthewmcneely, line 707: I think this error could be misleading, there are many reasons that types.Convert might fail. Maybe better to errors.Wrap with a more generic message?

mangalaman93: Done

worker/graphql_schema.go
```diff
@@ -258,6 +258,7 @@ func (w *grpcWorker) UpdateGraphQLSchema(ctx context.Context,
 // WaitForIndexing does a busy wait for indexing to finish or the context to error out,
 // if the input flag shouldWait is true. Otherwise, it just returns nil straight away.
 // If the context errors, it returns that error.
+// TODO(aman): we should return an error if the indexing fails
```
harshil-goel, line 261: Since we now panic in case of badger write fails, we should look into whats the behaviour currently.

harshil-goel (commented):
We need to look at what happens when indexing fails, as we can't just skip the transcation anymore.

mangalaman93:
@harshil-goel  indexing request comes as part of the Alter request and it is not a transaction. Today, it is acceptable that the indexing could fail. We don't have a good way of letting the user know yet and we should build one. The nice thing here is that the schema remains consistent either way.

worker/graphql_schema.go
mangalaman93: what do you mean? could you elaborate?

anurags92 (dismissed):

harshil-goel (approved):

anurags92 (approved):

harshil-goel (approved):

PR #7770, opened 2021-04-30
feat(cloud): add shared-instance flag in limit superflag in alpha

NamanJain8:
This PR adds a `shared-instance` flag to `--limit` superflag. 
When set to true (false by default), it will:
- Restrict access to any of the ACL operations like `Login`, add/remove/update user from non-galaxy namespaces.
- Prevent the leaking of environment variables for minio and aws.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7770)

PR #8625, opened 2023-01-23
feat(cloud): add shared-instance flag in limit superflag in alpha (#7770)

anurags92:
This PR adds a shared-instance flag to --limit superflag.
When set to true (false by default), it will:

- Restrict access to any of the ACL operations like Login, add/remove/update user from non-galaxy namespaces.
- Prevent the leaking of environment variables for minio and aws.

(cherry picked from commit 5f3cece75da375077c45ab64228cb7053cf03d02)

### DIFF
--- a/edgraph/access_ee.go
+++ b/edgraph/access_ee.go
@@ -49,6 +49,10 @@ type predsAndvars struct {
 func (s *Server) Login(ctx context.Context,
 	request *api.LoginRequest) (*api.Response, error) {
 
+	if !shouldAllowAcls(request.GetNamespace()) {
+		return nil, errors.New("operation is not allowed in shared cloud mode")
+	}
+
 	if err := x.HealthCheck(); err != nil {
 		return nil, err
 	}
@@ -628,18 +632,15 @@ type authPredResult struct {
 }
 
 func authorizePreds(ctx context.Context, userData *userData, preds []string,
-	aclOp *acl.Operation) (*authPredResult, error) {
+	aclOp *acl.Operation) *authPredResult {
 
-	ns, err := x.ExtractNamespace(ctx)
-	if err != nil {
-		return nil, errors.Wrapf(err, "While authorizing preds")
-	}
 	if !worker.AclCachePtr.Loaded() {
 		RefreshACLs(ctx)
 	}
 
 	userId := userData.userId
 	groupIds := userData.groupIds
+	ns := userData.namespace
 	blockedPreds := make(map[string]struct{})
 	for _, pred := range preds {
 		nsPred := x.NamespaceAttr(ns, pred)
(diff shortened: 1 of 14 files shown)


edgraph/access_ee.go
```diff
@@ -49,6 +49,10 @@ type predsAndvars struct {
 func (s *Server) Login(ctx context.Context,
 	request *api.LoginRequest) (*api.Response, error) {
 
+	if !shouldAllowAcls(request.GetNamespace()) {
+		return nil, errors.New("operation is not allowed in cloud mode")
```
mangalaman93, line 53: shared cloud mode

systest/cloud/cloud_test.go
```diff
@@ -0,0 +1,184 @@
...
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/dgraph-io/dgo/v210/protos/api"
```
mangalaman93, line 28: fix imports

systest/cloud/cloud_test.go
```diff
@@ -0,0 +1,184 @@
...
+	resp = testutil.Export(t, nsToken, minioDest, "accesskey", "secretkey")
+	require.Zero(t, len(resp.Errors))
+
+	// Galaxy guardian should provide the crednetials as well.
+	resp = testutil.Export(t, galaxyToken, minioDest, "accesskey", "secretkey")
+	require.Zero(t, len(resp.Errors))
+
+}
+
+func TestMain(m *testing.M) {
```
mangalaman93, line 181: maybe we can live without it

mangalaman93 (changes requested):
Why are we doing this in cloud? Can we live without it? And what is the comment about that says that this PR avoids leaking secrets.

anurags92:
> Why are we doing this in cloud? Can we live without it? And what is the comment about that says that this PR avoids leaking secrets.

I am not very sure about why do we need this in cloud. @rarvikar  could you put some thoughts on this? 

> PR avoids leaking secrets.
I couldn't find a place that pertained to this. I do see a test in `cloud_test.go` which tests that `access creds` are required when exporting.

edgraph/access_ee.go
```diff
@@ -979,14 +986,24 @@ func authorizeQuery(ctx context.Context, parsedReq *dql.Result, graphql bool) er
...
+		namespace = userData.namespace
 
 		if x.IsGuardian(groupIds) {
-			// Members of guardian groups are allowed to query anything.
-			return nil, nil, nil
+			if shouldAllowAcls(userData.namespace) {
+				// Members of guardian groups are allowed to query anything.
+				return nil, nil, nil
+			}
+			blocked := make(map[string]struct{})
```
mangalaman93, line 996: this code seems to be duplicated. Does it make sense to create a function instead?

systest/cloud/cloud_test.go
mangalaman93: The order needs to be opposite, first non-dgraph imports and then dgraph imports.

testutil/utils.go
```diff
@@ -24,10 +24,11 @@ import (
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/require"
-
 	"github.com/dgraph-io/dgo/v210"
 	"github.com/dgraph-io/dgraph/x"
+	"github.com/dgrijalva/jwt-go"
```
mangalaman93, line 29: same here, import order needs to be changed

mangalaman93 (changes requested):

testutil/utils.go
anurags92: Addressed.

rarvikar:
Hi @anurags92 @mangalaman93 , we use this flag to bypass ACL logins for users of shared backends. This flag IS in-use on all of our regional shared Dgraph clusters hosting shared and free customer backends.

mangalaman93 (approved):

mangalaman93 (approved):

harshil-goel (approved):

PR #8638, opened 2023-01-30
fix(backup): create directory before writing backup

mangalaman93:
We used to create the directory while creating a backup if one didn't exist. Slash branch doesn't create one but I think we should continue with the behaviour of creating a directory as was there in main branch.

### DIFF
--- a/worker/backup_handler.go
+++ b/worker/backup_handler.go
@@ -62,11 +62,6 @@ const (
 )
 
 func createBackupFile(h UriHandler, uri *url.URL, req *pb.BackupRequest) (io.WriteCloser, error) {
-	if !h.DirExists("./") {
-		if err := h.CreateDir("./"); err != nil {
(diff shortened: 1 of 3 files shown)


harshil-goel (approved):

worker/backup_handler.go
```diff
@@ -62,11 +62,6 @@ const (
 )
 
 func createBackupFile(h UriHandler, uri *url.URL, req *pb.BackupRequest) (io.WriteCloser, error) {
-	if !h.DirExists("./") {
-		if err := h.CreateDir("./"); err != nil {
-			return nil, errors.Wrap(err, "while creating backup file")
-		}
-	}
```
anurags92, line 69: This piece of code was no-op because it would have failed already if the `backup` dir was not present when checking for manifest file. Is that correct?

mangalaman93: exactly .

PR #8649, opened 2023-02-03
chore(other): Rename Master's references remaining

MichelDiz:
## Problem

We moved from master to main repo naming.

## Solution
 
 Rename remaining references.
 
 Replace: #8448


worker/backup_manifest.go
```diff
@@ -123,18 +122,18 @@ func getFilteredManifests(h UriHandler, manifests []*Manifest,
 	return manifests, nil
 }
 
-// getConsolidatedManifest walks over all the backup directories and generates a master manifest.
```
mangalaman93, line 126: We should undo the changes in this file.

worker/multi_tenancy_ee.go
```diff
@@ -1,5 +1,4 @@
 //go:build !oss
-// +build !oss
```
mangalaman93, line 2: I am not sure we should get rid of these yet. I think Go version on your laptop is probably newer.

mangalaman93 (changes requested):
Thanks for the PR @MichelDiz. Please address the comments and this should be good to go.

worker/multi_tenancy_ee.go
MichelDiz: What is the difference? and yes, this was automatic from fmt.

mangalaman93: I am thinking that we should do this uniformly in all files when we do it. Dgraph can still compile with older versions of Go and removing this can cause backward incompatibility.

MichelDiz: Hum, weird. Why the maintainers(go devs basically) of fmt would do such move without taking this into account?

The big problem is that it's a war to be fought. Every time I save the file it automatically does this. You would have to create a configuration and share it with everyone and tell everyone to use it. But it is necessary to understand what is the real effect that this causes. We can't just stay in theory.

I didn't want to change that line. And every time I save the file it is unavoidable. If I remove autoformatting I lose other correction advantages.

Actually, I wanted to understand more what this line does. Is there any document?

mangalaman93: do you know what tool is doing this on your laptop? Maybe that will help us figure out why it is being done. I am on go1.20 and saving a file doesn't do this for me.

MichelDiz: It is a PC. And I'm using VSCode with latest go updates recomended by vscode. If I deactivate the fmt this stops.

mangalaman93: I think for now then, let's not do this. At least not until we understand this better.

MichelDiz: lol, i thought you do otherwise why would you care about something that you don't know what results?

Anyway. The information about this little tag is quite rare. 
What I have found (In this discussion https://github.com/dgraph-io/dgraph/pull/8448#discussion_r1030030428):

This is some auto fmt from latest goimports. Not sure what it does. If you have a go server with goimports installed. Everytime you save the file it will update that comment.

There are few expressions with this
```
!dev
!oss (which means "open source software" as far as I know. Also it could be "cross platform OS")
!debug
!windows
```

What this fmt does is just add `go:` in whatever it finds.

I've just chose not fight against it. Cuz it is something that will update no matter what I do against it.

matthewmcneely: The Go authors recommend keeping both: https://go.dev/doc/go1.17#build-lines

mangalaman93 (approved):
Thank you for doing this Michel. LGTM

MaxxReis (approved):

matthewmcneely (approved):

MaxxReis (approved):

PR #8657, opened 2023-02-08
fix(query): handle bad timezone correctly

mangalaman93:
in Go 1.20, time.MarshalJSON function checks for time zone strictly wrt RFC 3339 i.e. zone ∈ [0, 23]. In Dgraph, we have been simply following what time package has been doing and we need to handle the bad time zones gracefully now.

The way we handle it is that, we do not allow ingesting bad time zones going forward. Such a mutation will return an error. On the other hand, if data is already ingested, we return the result as is and ensure that such a query does not fail.


harshil-goel (dismissed):

query/outputnode.go
```diff
@@ -484,7 +484,7 @@ func (enc *encoder) AddValue(fj fastJsonNode, attr uint16, v types.Val) error {
 func (enc *encoder) AddListValue(fj fastJsonNode, attr uint16, v types.Val, list bool) error {
 	bs, err := valToBytes(v)
 	if err != nil {
-		return nil // Ignore this.
+		return fmt.Errorf("error while converting badger data to json: %v", err)
```
mangalaman93, line 487: badger data -> stored data

query/outputnode_test.go
```diff
@@ -262,3 +263,43 @@ func TestChildrenOrder(t *testing.T) {
...
+			out: "\"2018-05-30T09:30:10+30:00\""},
+		{in: time.Date(2018, 5, 30, 9, 30, 10, 0, time.FixedZone("", -99*60*60)),
+			out: "\"2018-05-30T09:30:10-99:00\""},
+		{in: time.Date(2018, 5, 30, 9, 30, 10, 0, time.FixedZone("", 99*60*60)),
+			out: "\"2018-05-30T09:30:10+99:00\""},
+	}
+
+	for _, tc := range timesToMarshal {
+		out, err := marshalTimeJson(tc.in)
+		if tc.err == "" {
```
matthewmcneely, line 297: `tc.err` will always be "", ya? did you mean to add some failing fixtures in the timesToMarshall slice?

query/outputnode.go
```diff
@@ -677,7 +677,7 @@ func valToBytes(v types.Val) ([]byte, error) {
 		return boolFalse, nil
 	case types.DateTimeID:
 		t := v.Value.(time.Time)
-		return t.MarshalJSON()
+		return marshalTimeJson(t)
```
anurags92, line 680: add comment at the top why we need this function, maybe link golang discussion from the go repo.

query/outputnode.go
mangalaman93: removed the error, we are keeping the behaviour same

query/outputnode_test.go
mangalaman93: Done

query/outputnode.go
```diff
@@ -689,6 +689,32 @@ func valToBytes(v types.Val) ([]byte, error) {
...
+	// If zone >23 or <-23, we need to handle this case ourselves.
+	// This is because, in go1.20, MarshalJSON fails for invalid zones.
+	// We, for now, call MarshalJSON for timestamp without the zone (or making it UTC zone).
+	b, err := t.Add(time.Duration(offset) * time.Second).UTC().MarshalJSON()
+	if err != nil {
+		return nil, err
+	}
+
+	// we will get a byte slice that has Z appended at the end
+	zone := offset / 60
```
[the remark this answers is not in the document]
mangalaman93: Done

query/outputnode.go
mangalaman93: Done

matthewmcneely (dismissed):

types/scalar_types.go
```diff
@@ -196,6 +197,21 @@ func ValueForType(id TypeID) Val {
...
 // By default, Go parses time in UTC unless specified in the data itself.
 func ParseTime(val string) (time.Time, error) {
+	t, err := parseTimeNonStrict(val)
+	if err != nil {
+		return t, err
+	}
+
+	_, offset := t.Zone()
+	boundary := 23*60*60 + 59*60
+	if offset > boundary || offset < -1*boundary {
```
anurags92, line 207: nit: This condition is checked in another place as well. Do we want to export this in a function and call that?

anurags92 (dismissed):

harshil-goel (dismissed):

types/scalar_types.go
mangalaman93: Done

harshil-goel (approved):

skrdgraph (approved):
cc: @rarvikar 

This looks okay. 

The plan is as follows: 
- merge this
- this PR will not break reads
- this PR will need a communication in the next release, indicating writes would error out

If customers need a fix for the `write`, we will make a patch release (as this is an anti-pattern for go-standards).

PR #8665, opened 2023-02-13
Added build tags for each test file

shivaji-dgraph:
Added build tags for each of the test file and changed t.go accordingly

### DIFF
--- a/algo/heap_test.go
+++ b/algo/heap_test.go
@@ -1,5 +1,3 @@
-//go:build unit_test
-
 /*
  * Copyright 2016-2023 Dgraph Labs, Inc. and Contributors
  *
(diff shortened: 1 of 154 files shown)


algo/heap_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build unit_test
```
mangalaman93, line 1: you don't need a tag for unit test. No tag means it is a unit test

mangalaman93 (changes requested):

algo/heap_test.go
shivaji-dgraph: done

mangalaman93 (changes requested):

mangalaman93 (dismissed):

mangalaman93 (approved):

MichelDiz (approved):

PR #8684, opened 2023-02-18
feat(upgrade): introduce basic upgrade test

joshua-goldstein:
Use existing LDBC test suite to run a basic upgrade test.  E.g.

```
./t --suite=ldbc --ldbcLoadOnly=true
./t --suite=ldbc --ldbcQueryOnly=true --download=false
```

will essentially split LDBC test suite into two steps.  Firstly we run Bulkloader on LDBC data.  Secondly we run queries against this data.  

This allows us to test upgrade paths.  E.g. use bulkloader with Dgraph v21.03, then query data using Dgraph v22.00.  To do this we have added a new Makefile target `make test-upgrade`.  This target looks for the directory `t/bin` and expects to see two binaries, `dgraph.old` and `dgraph.new`.  The bulkloader will run using the old binary, and the queries will be run using the new binary.  This provides us one check of data incompatibility.

Note that this target does not compile any binaries.  We expect this target to be incorporated as part of a larger upgrade script, which will take responsibility for compiling the base dgraph binary and target dgraph binary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 /build
 /installs
+/t/bin
+/linux
 /coverage.out
 /dgraph-bulk-loader
 /osx-docker-gopath
--- a/Makefile
+++ b/Makefile
@@ -65,6 +65,14 @@ test: docker-image
 	@mv dgraph/dgraph ${GOPATH}/bin/dgraph
 	@$(MAKE) -C t test
 
+test-upgrade: docker-image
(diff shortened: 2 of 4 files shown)


mangalaman93:
I think this is great. Our plan for testing upgrade is slightly different though, we plan to use build tags instead. I think keep this PR into draft and once we have the upgrade library, we will take this and align it with our existing work. Thanks for doing this.

mangalaman93:
closing this in lieu of #8933

PR #8702, opened 2023-02-24
Add flags to statically link dgraph binary

### LINKED ISSUE #8701: [BUG]: Docker images don't work when built locally on a 22.04 ubuntu machine

Dgraph version   : local
Dgraph codename  : dgraph
Dgraph SHA-256   : 57783495102de2b9461ed0f79e9984994e448c5d3b37b2906642d892a38ee0c0
Commit SHA-1     : d981392d8
Commit timestamp : 2023-02-24 11:20:24 +0530
Branch           : main
Go version       : go1.20
jemalloc enabled : true

### Tell us a little more about your go-environment?

go version go1.20 linux/amd64

Yes

### What is the hardware spec (RAM, CPU, OS)?

RAM
```
               total        used        free      shared  buff/cache   available
Mem:            31Gi        10Gi       6.3Gi       2.1Gi        13Gi        17Gi
Swap:          2.0Gi        12Mi       2.0Gi

```
CPU
```
Architecture:            x86_64
  CPU op-mode(s):        32-bit, 64-bit
  Address sizes:         39 bits physical, 48 bits virtual
  Byte Order:            Little Endian
CPU(s):                  20
  On-line CPU(s) list:   0-19
Vendor ID:               GenuineIntel
  Model name:            12th Gen Intel(R) Core(TM) i7-12700H
    CPU family:          6
    Model:               154

```
OS
```
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.1 LTS
Release:	22.04
Codename:	jammy

```

### What steps will reproduce the bug?

`make docker-image`

`docker run dgraph/dgraph:local dgraph version`

### Expected behavior and actual result.

Expected: 

It should output the version of dgraph inside the docker image

Actual:
```
dgraph: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by dgraph)
dgraph: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by dgraph)
```

### Additional information

This happens when I try to build a docker image on Ubuntu22.04 machine.

anurags92:
Fixes #8701

### DIFF
--- a/dgraph/Makefile
+++ b/dgraph/Makefile
@@ -42,7 +42,7 @@ gitBranch       = github.com/dgraph-io/dgraph/x.gitBranch
 lastCommitSHA   = github.com/dgraph-io/dgraph/x.lastCommitSHA
 lastCommitTime  = github.com/dgraph-io/dgraph/x.lastCommitTime
 
-BUILD_FLAGS   ?= -ldflags '-X ${lastCommitSHA}=${BUILD} -X "${lastCommitTime}=${BUILD_DATE}" -X "${dgraphVersion}=${BUILD_VERSION}" -X "${dgraphCodename}=${BUILD_CODENAME}" -X ${gitBranch}=${BUILD_BRANCH}'
+BUILD_FLAGS   ?= -ldflags '-extldflags "-static" -X ${lastCommitSHA}=${BUILD} -X "${lastCommitTime}=${BUILD_DATE}" -X "${dgraphVersion}=${BUILD_VERSION}" -X "${dgraphCodename}=${BUILD_CODENAME}" -X ${gitBranch}=${BUILD_BRANCH}'
 
 # Insert build tags if specified
 ifneq ($(strip $(BUILD_TAGS)),)


anurags92:
Extra warnings:
```
# github.com/dgraph-io/dgraph/dgraph
/usr/bin/ld: /tmp/go-link-1471705237/000028.o: in function `pluginOpen':
/_/plugin/plugin_dlopen.go:19: warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /tmp/go-link-1471705237/000005.o: in function `mygetgrouplist':
/_/os/user/getgrouplist_unix.go:15: warning: Using 'getgrouplist' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /tmp/go-link-1471705237/000004.o: in function `mygetgrgid_r':
/_/os/user/cgo_lookup_cgo.go:45: warning: Using 'getgrgid_r' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /tmp/go-link-1471705237/000004.o: in function `mygetgrnam_r':
/_/os/user/cgo_lookup_cgo.go:54: warning: Using 'getgrnam_r' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /tmp/go-link-1471705237/000004.o: in function `mygetpwnam_r':
/_/os/user/cgo_lookup_cgo.go:36: warning: Using 'getpwnam_r' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /tmp/go-link-1471705237/000004.o: in function `mygetpwuid_r':
/_/os/user/cgo_lookup_cgo.go:27: warning: Using 'getpwuid_r' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /tmp/go-link-1471705237/000025.o: in function `_cgo_cbcce81e6342_C2func_getaddrinfo':
/tmp/go-build/cgo-gcc-prolog:58: warning: Using 'getaddrinfo' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
make[1]: Leaving directory '/home/alvis/go/src/github.com/dgraph-io/dgraph/dgraph'

```

joshua-goldstein:
So I was playing around with this and I think I figured out (roughly) what is happening...  The `ldd` output tells us what shared libraries are being dyamically linked into the dgraph binary (from dgraph v22.0.2 but version shouldn't matter);
```
ldd dgraph/dgraph 
	linux-vdso.so.1 (0x00007fff86df2000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f252d24c000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f252d229000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f252d037000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f252d267000)

```

This makes sense because in the ristretto/z package, we have the following C preamble in [ristretto/calloc_jemalloc.go](https://github.com/dgraph-io/ristretto/blob/f4fa6df711eec3a61b8ccbec6bbb76e95ae4ecce/z/calloc_jemalloc.go#L7-L14)
```
package z


/*
#cgo LDFLAGS: /usr/local/lib/libjemalloc.a -L/usr/local/lib -Wl,-rpath,/usr/local/lib -ljemalloc -lm -lstdc++ -pthread -ldl
#include <stdlib.h>
#include <jemalloc/jemalloc.h>
*/
import "C"
```

My understanding is that the `-rpath` flag specifies the libraries that need to be linked at runtime, which look like the same libraries in the `ldd` output.  My guess is that this is required for jemalloc to function correctly, but I am not familiar with the technical details of jemalloc enough to say why those particular libraries are being dynamically linked.  

Incidentally, if we want a statically linked binary, all you have to do is `CGO_ENABLED=0 make dgraph`, which will produce a static binary.  This will of course not use jemalloc.
```
CGO_ENABLED=0 make dgraph
file dgraph/dgraph
dgraph: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=CbWTp2eD9xUtr1MvY2_k/YatVwXL7yCppYGXEHrdp/xpzaqBAuS3nmtWlJdx6w/RQEFQpuZugErT1WYYtwE, with debug_info, not stripped
```

Note I am not using the flag in this PR; this can be done from any version of Dgraph currently.

Going back to the original issue...  Let's assume my local system is on 22.04.  Since we do our dgraph builds locally (i.e. not inside a container), the dgraph binary will dynamically link (?) the shared libraries from ubuntu 22.04.  However, when Dgraph is actually running inside a container that is pinned at ubuntu 20.04, it will see the wrong shared libraries inside the container and hence give this error.

One solution might be to move the build process to inside the container, but that would have other effects and we would have to think about the implications.  We will have to think about this issue more when we start deploying our dgraph binaries in different environments.

Just wanted to put down my observations here in case we discuss this issue again in the future @anurags92 @mangalaman93 .  I think for now we can close this PR since the warnings are telling us that we need these libraries to be dynamically linked if we want jemalloc to work correctly.

mangalaman93:
> Incidentally, if we want a statically linked binary, all you have to do is `CGO_ENABLED=0 make dgraph`, which will produce a static binary. This will of course not use jemalloc.

Why would CGO_ENABLED=0 case not use jemalloc?

joshua-goldstein:
> > Incidentally, if we want a statically linked binary, all you have to do is `CGO_ENABLED=0 make dgraph`, which will produce a static binary. This will of course not use jemalloc.
> 
> Why would CGO_ENABLED=0 case not use jemalloc?

Because jemalloc is a [C library](https://github.com/jemalloc/jemalloc).  It is a custom implementation of malloc to reduce fragmentation.  There was a nice blog post on why we use jemalloc (see [here](https://dgraph.io/blog/post/manual-memory-management-golang-jemalloc/)).

joshua-goldstein:
@mangalaman93 Just wanted to follow up here so all our discussion is in one place...  that is not accurate actually.  Every platform does not need a separate build.  The issue is very specific:
- Developer is on Ubuntu 22.04. and builds Dgraph locally.  The Dgraph binary is therefore dynamically linked to shared libraries in Ubuntu 22.04.  However, when building a Docker image **locally**, the image is pinned to Ubuntu 20.04.  The shared libraries that exist inside the image are not *forward compatible* (i.e. the 20.04 libraries will not support a binary linked to 22.04 libraries
- **This is *only an issue for developers, on 22.04, who build the docker image locally*!**  The fix is extremely straightforward - in `contrib/Dockerfile` change the base image from ubuntu-20.04 to ubuntu-22.04 and dgraph will build the image locally without a problem.

This actually ***resolves*** my concerns (that I previously discussed with @skrdgraph), namely:

- Q: Is jemalloc statically or dynamically linked?  What happens when a customer doesn't have jemalloc locally and `wgets` our binary, for example?
- Answer: *jemalloc is statically linked* and therefore there is no problem, jemalloc does not need to be installed locally
- Q: What about those dynamically linked libaries?
- Answer: They look like core linux libraries that every linux distribution has (e.g. `/lib/x86_64-linux-gnu/libdl.so.2` is in every ubuntu image).
- Q: What about compatibility between 22.04 and 20.04?
- Answer: Because these libraries are **careful about maintaining backwards compatibility**, and because we build Dgraph on 20.04, the Dgraph binary (and Dgraph images) will run without a problem on both Ubuntu 20.04 and 22.04.  Ubuntu 20.04 is very far from end of life and I think what we have is currently optimal.

tl;dr: we do not need to do anything now.  Our binaries and images work fine on both 20.04 and 22.04, which is exactly the use case we should be aiming for.  We should not complicate our build process without an extremely good reason.  @skrdgraph

mangalaman93:
This makes a lot of sense, thank you for looking into it. One question, how do we know jemalloc is statically linked? The Go toolchain does, that's amazing.

anurags92:
Thanks @joshua-goldstein! That makes sense.

PR #8703, opened 2023-02-24
fix(ci) increase frequency

skrdgraph:
## Problem
Need more test runs because we are closing in on our next release

## Solution
Increase frequency for critical CI runs to happen every 2hrs
This will increase our CI spends, but it's better we keep this running for a few days to detect any failures.

**note** 
- we are doing this because we noticed some flaky failures in our tests on `main` (*`scheduled runs`*)
- the last change was on jan-6  https://github.com/dgraph-io/dgraph/pull/8592, where we reduced the frequency because of a stable baseline

### DIFF
--- a/.github/workflows/ci-dgraph-code-coverage.yml
+++ b/.github/workflows/ci-dgraph-code-coverage.yml
@@ -14,7 +14,7 @@ on:
       - main
       - 'release/**'
   schedule:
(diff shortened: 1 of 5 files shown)


akon-dey (approved):
LGTM

adityasadalage (approved):
Looks good

PR #8707, opened 2023-02-25
chore(test): add layout for dgraphtest package

mangalaman93:
This package will provide functions to setup cluster for testing and benchmarking going forward


dgraphtest/config.go
```diff
@@ -0,0 +1,78 @@
...
+	cc.numZeros = n
+	return cc
+}
+
+func (cc ClusterConfig) WithReplicas(n int) ClusterConfig {
+	cc.replicas = n
+	return cc
+}
+
+func (cc ClusterConfig) WithACL() ClusterConfig {
```
mangalaman93, line 65: we need to take TTL as an argument here and configure it correctly while running alpha

mangalaman93 (commented):
one comment, good to go otherwise

contrib/Dockerfile
```diff
@@ -29,8 +29,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 ADD linux /usr/local/bin
 
-EXPOSE 8080
-EXPOSE 9080
```
skrdgraph, line 33: This is the same Dockerfile that is being used for build, test & release. This could break things for folks who choose to simply do `docker run` directly on the image without port-mapping. https://stackoverflow.com/a/22150099 #2 is important.

mangalaman93: If someone is running a Dgraph Cluster, I am wondering how only exposing alpha ports will make things work. We are not exposing zero ports here.

skrdgraph:
@mangalaman93 this may be redundant, I understand the purpose of this. But have you looked at https://github.com/dgraph-io/dgraph/blob/main/compose/compose.go ... 

The old team (from what I can tell) was planning to transition `t.go` dgraph-cluster CRUDs to `compose.go`. This wrapper tool is actually quite handy. If it can be repurposed it will serve your purpose IMO.

mangalaman93:
The compose tool gives you a very static cluster. For our testing going forward, we would like a more dynamic control on the cluster. Think about upgrades for example, we want to setup a cluster with a dgraph version for which we don't even have a docker image. This could be easily done if we have the binary and we can mount that binary inside the container. We can also do the same when we want to upgrade the cluster and so on. This is just the base of the work, we are going to build more interesting tools on top of this work. I am also thinking that using the same library, we could setup cluster on AWS, ECS if needed and even a cluster on Dgraph cloud. The interface to the test would be consistent so that the same test could run in different environment. Maybe, we could integrate the compose tool as one of the ways of setting up the cluster in the library.

dgraphtest/cluster.go
```diff
@@ -0,0 +1,301 @@
...
+	conf ClusterConfig
+
+	// resources
+	dcli   *docker.Client
+	net    cnet
+	zeros  []dnode
+	alphas []dnode
+}
+
+func NewCluster(conf ClusterConfig) (*Cluster, error) {
```
anurags92, line 53: If you return a pointer, I have seen it returned as a named param. But I don't think it has any operational difference.

Found this while reviewing -- https://medium.com/@philpearl/bad-go-pointer-returns-340f2da8289
(just instructive, do not have an issue with the signature)

anurags92 (changes requested):

dgraphtest/cluster.go
```diff
@@ -0,0 +1,301 @@
...
+		return errors.Wrap(err, "unable to talk to docker daemon")
+	}
+
+	if err := c.createNetwork(); err != nil {
+		return errors.Wrap(err, "error creating network")
+	}
+
+	for i := 0; i < c.conf.numZeros; i++ {
+		zo := &zero{id: i}
+		zo.coname = fmt.Sprintf(zeroNameFmt, c.conf.prefix, zo.id)
```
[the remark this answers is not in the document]
mangalaman93: coname -> container name, didn't want to use a longer name
loname -> local name 

I have explained them in struct where I have defined them. If there is a suggestion for a better name and happy to replace.

dgraphtest/dgraph.go
```diff
@@ -0,0 +1,238 @@
...
+	"path/filepath"
+
+	"github.com/docker/docker/api/types/mount"
+	docker "github.com/docker/docker/client"
+	"github.com/docker/go-connections/nat"
+	"github.com/pkg/errors"
+)
+
+const (
+	zeroNameFmt   = "%v_zero%d"
```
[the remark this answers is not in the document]
mangalaman93: same here, zeroNameFmt is the container name format, zeroLNameFmt is the zero's containers local name. Once you look at the struct, it will make more sense.

dgraphtest/cluster.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+
+	return c, nil
+}
+
+func (c *Cluster) log(format string, args ...any) {
+	if c == nil || c.conf.logr == nil {
+		return
+	}
+	c.conf.logr.Logf(format, args...)
```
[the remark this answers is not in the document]
mangalaman93: I was thinking about it, but I thought an explicit configuration reference will be more readable. If I embed it, reading the code could be confusing.

dgraphtest/cluster.go
mangalaman93: This is a good idea. We already have lots of pointers inside the struct. We don't need the Cluster to be a pointer then.

contrib/Dockerfile
```diff
@@ -29,8 +29,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 ADD linux /usr/local/bin
 
-EXPOSE 8080
```
harshil-goel, line 32: Why are we removing this?

dgraphtest/dgraph.go
```diff
@@ -0,0 +1,238 @@
...
+	ports() nat.PortSet
+	cmd(*Cluster) []string
+	workingDir() string
+	mounts(ClusterConfig) ([]mount.Mount, error)
+	healthURL(c *Cluster) (string, error)
+}
+
+type zero struct {
+	id     int    // 0, 1, 2
+	coid   string // container ID in docker world
```
harshil-goel, line 65: Coid, and coname confused me at first until I saw the struct. What is loname? 
Would it be possible to have friendlier names?

dgraphtest/cluster.go
```diff
@@ -0,0 +1,301 @@
...
+	docker "github.com/docker/docker/client"
+	"github.com/pkg/errors"
+)
+
+var (
+	requestTimeout = 30 * time.Second
+	stopTimeout    = time.Minute
+)
+
+type cnet struct {
```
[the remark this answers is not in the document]
mangalaman93: Done

dgraphtest/cluster.go
```diff
@@ -0,0 +1,301 @@
...
+
+func (c *Cluster) stopContainer(dc dnode) error {
+	ctx, cancel := context.WithTimeout(context.Background(), requestTimeout)
+	defer cancel()
+	if err := c.dcli.ContainerStop(ctx, dc.cid(), &stopTimeout); err != nil {
+		return errors.Wrapf(err, "error stopping container [%v]", dc.cname())
+	}
+	return nil
+}
+
```
harshil-goel, line 209: Probably not a part of this diff, but can we also include options for collecting logs here?

dgraphtest/cluster.go
```diff
@@ -0,0 +1,301 @@
...
+	defer cancel()
+	resp, err := c.dcli.ContainerCreate(ctx, cconf, hconf, networkConfig, dc.cname())
+	if err != nil {
+		return "", errors.Wrapf(err, "error creating container %v", dc.cname())
+	}
+
+	return resp.ID, nil
+}
+
+func (c *Cluster) healthCheck() error {
```
harshil-goel, line 261: Is checking for health the same as checking if license is applied? Today we got an issue where we were checking for  health, but still license failed. Can we add a check for is license applied? Okay to be done in a separate diff.

dgraphtest/cluster.go
```diff
@@ -0,0 +1,301 @@
...
+
+func (c *Cluster) createContainer(dc dnode) (string, error) {
+	cmd := dc.cmd(c)
+	image := c.dgraphImage()
+	mts, err := dc.mounts(c.conf)
+	if err != nil {
+		return "", err
+	}
+
+	cconf := &container.Config{Cmd: cmd, Image: image, WorkingDir: dc.workingDir(), ExposedPorts: dc.ports()}
```
harshil-goel, line 240: Can we log the final conf? Might be helpful while debugging.

dgraphtest/image.go
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package dgraphtest
+
+func (c *Cluster) dgraphImage() string {
+	return "dgraph/dgraph:local"
```
harshil-goel, line 20: Is this intended to be updated in future to contain various different functions? If not, what is it's function?

dgraphtest/config.go
```diff
@@ -0,0 +1,78 @@
...
+	cc.acl = true
+	return cc
+}
+
+func (cc ClusterConfig) WithEncryption() ClusterConfig {
+	cc.encryption = true
```
harshil-goel, line 73: Can we let the user give an encryption key? Is that something for later?

dgraphtest/config.go
```diff
@@ -0,0 +1,78 @@
...
+type Logger interface {
+	Logf(format string, args ...any)
+}
+
+type ClusterConfig struct {
+	prefix     string
+	numAlphas  int
+	numZeros   int
+	replicas   int
+	verbosity  int
```
harshil-goel, line 34: WithVerbosity() function is missing

dgraphtest/config.go
mangalaman93: something for later

dgraphtest/image.go
mangalaman93: yeah, that is possible.

dgraphtest/dgraph.go
mangalaman93: done

contrib/Dockerfile
mangalaman93: because it only works for alpha, it doesn't work for Zero

dgraphtest/dgraph.go
```diff
@@ -0,0 +1,238 @@
...
+		},
+	}, nil
+}
+
+func (z *zero) healthURL(c *Cluster) (string, error) {
+	publicPort, err := publicPort(c.dcli, z, zeroHttpPort)
+	if err != nil {
+		return "", err
+	}
+	return "http://localhost:" + publicPort + "/health", nil
```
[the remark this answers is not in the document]
mangalaman93: we haven't allowed TLS yet. This would be for later.

dgraphtest/dgraph.go
```diff
@@ -0,0 +1,238 @@
...
+	return nat.PortSet{
+		alphaGrpcPort: {},
+		alphaHttpPort: {},
+	}
+}
+
+func (a *alpha) cmd(c *Cluster) []string {
+	acmd := []string{"/gobin/dgraph", "alpha", fmt.Sprintf("--my=%s:%v", a.lname(), alphaInterPort),
+		"--bindall", "--logtostderr", fmt.Sprintf("-v=%d", c.conf.verbosity),
+		`--security=whitelist=10.0.0.0/8,172.16.0.0/12,192.168.0.0/16`}
```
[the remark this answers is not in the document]
mangalaman93: what will be required in zero also?

dgraphtest/dgraph.go
```diff
@@ -0,0 +1,238 @@
...
+	defer cancel()
+
+	info, err := dcli.ContainerInspect(ctx, dc.cid())
+	if err != nil {
+		return "", errors.Wrap(err, "error inspecting container")
+	}
+
+	for port, bindings := range info.NetworkSettings.Ports {
+		if port.Port() == privatePort {
+			return bindings[0].HostPort, nil
```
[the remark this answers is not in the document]
mangalaman93: done

dgraphtest/cluster.go
mangalaman93: yes, for sure

dgraphtest/cluster.go
mangalaman93: could you elaborate?

dgraphtest/config.go
mangalaman93: Done

dgraphtest/cluster.go
mangalaman93: you could just do docker inspect or something. Not sure whether we can easily log all the config.

contrib/Dockerfile
anurags92, line 32: I think it should be a separate PR.

dgraphtest/local_cluster.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+func (c LocalCluster) healthCheck() error {
+	c.log("checking health of containers")
+	for i := 0; i < c.conf.numZeros; i++ {
+		url, err := c.zeros[i].healthURL(c)
+		if err != nil {
+			return errors.Wrap(err, "error getting health URL")
+		}
+		if err := c.containerHealthCheck(url); err != nil {
```
anurags92, line 268: Why not re-use: https://github.com/dgraph-io/dgraph/blob/ec9d53543b5627b3aefd5bb0c8eea0994d5521eb/x/health.go#L51

mangalaman93: That function is different than what we are doing here.

contrib/Dockerfile
mangalaman93: Good idea. https://github.com/dgraph-io/dgraph/pull/8733

harshil-goel (approved):

anurags92 (approved):

PR #8709, opened 2023-02-28
chore(deps): bump badger up to v4

joshua-goldstein:
We bump up the dependency on Badger to v4.  This requires regenerating protobuf files.  Do not merge yet.


skrdgraph (approved):
👍

PR #8713, opened 2023-03-01
fix(changelog): update release version

skrdgraph:
## Problem
Follow Strict SerVer for Beta / RC & GA

## Solution
https://commons.apache.org/releases/versioning.html

Beta Release Numbers
Beta releases are denoted by adding "beta<beta version number>" after the release number. For example, if the current release version is 2.0.4, and a developer wished to preview the next major release, the release would be labeled 3.0-beta1.


meghalims (approved):
LGTM

joshua-goldstein (approved):

PR #8714, opened 2023-03-02
chore(logging): add jaeger tag handler for queries and mutations

### LINKED ISSUE #8677: [FEATURE]: Add custom tags for Jaeger Tracing.
### Use case

Sometimes we need to compare two specific queries stats in a cluster with numerous namespaces and many clients making requests at the same time. Adding a tag will allow us to isolate our query and perform our analysis without too much noise.

### Links to Discuss, RFC or previous Issues and PRs

_No response_

### Links to examples and research

_No response_

### Current state

We are only able to filter out namespaces. Which helps, but not ideal.

### Solution proposal

We could have a function that reads a comment in the query body and adds the tag for us.
e.g.
```
{
 #jaeger.query=myQueryUsers033
   q(func: ...) @filter(...) {
    ...
   }
}
```

```
func foo(ctx context.Context) {
    ctx, span := trace.StartSpan(ctx, "foo")
    defer span.End()

    span.AddAttributes(trace.StringAttribute("my-custom-tag", "my-custom-value"))

    // ...
}
```

### Additional Information

_No response_

MichelDiz:
## Problem

It's hard to find a specific query. So having a tag in the query body as a comment helps filter through the sea of logs.

 Fixes #8677
 
## Solution

Usage is very simple. Add a comment #tag: + your tag name. It works for queries and mutations.

```gql
{
 set {
  #tag:mutation2334
   _:New <test> "test" .
 }
}
```

```gql
#tag:TESTING4
#xxxxxxxxxx only the tag TESTING4 will be captured

schema{}
```

PS. In Upsert Block the tag works only in the query block, not the mutation.

How you can use in Jaeger

Captura de Tela (16)

For mutations you have to use 
```gql
mutation.tag=
```

### DIFF
--- a/edgraph/server.go
+++ b/edgraph/server.go
@@ -1313,7 +1313,7 @@ func (s *Server) doQuery(ctx context.Context, req *Request) (resp *api.Response,
 		if !isQuery {
 			m = req.req.String()
 		}
-		if strings.Contains(m, "#") {
+		if strings.Contains(m, "#tag") {
 			q := isQuery
 			validateComment(span, m, q)
 		}
@@ -1328,6 +1328,9 @@ func (s *Server) doQuery(ctx context.Context, req *Request) (resp *api.Response,
 				"Non guardian of galaxy user cannot bypass namespaces. "+s.Message())
 		}
 	}
+
+	// extras tags for tracing
+	span.AddAttributes(otrace.BoolAttribute("isGraphQL", isGraphQL))
 
 	qc := &queryContext{
 		req:      req.req,
--- a/edgraph/server.go
+++ b/edgraph/server.go
@@ -25,6 +25,7 @@ import (
 	"fmt"
 	"math"
 	"net"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -148,12 +149,12 @@ func GetGQLSchema(namespace uint64) (uid, graphQLSchema string, err error) {
 	resp, err := (&Server{}).QueryNoGrpc(ctx,
 		&api.Request{
 			Query: `
-			query {
-				ExistingGQLSchema(func: has(dgraph.graphql.schema)) {
-					uid
-					dgraph.graphql.schema
-				  }
-				}`})
+			 query {
+				 ExistingGQLSchema(func: has(dgraph.graphql.schema)) {
+					 uid
+					 dgraph.graphql.schema
+				   }
+				 }`})
 	if err != nil {
 		return "", "", err
 	}
@@ -546,6 +547,19 @@ func (s *Server) Alter(ctx context.Context, op *api.Operation) (*api.Payload, er
 func annotateNamespace(span *otrace.Span, ns uint64) {
 	span.AddAttributes(otrace.Int64Attribute("ns", int64(ns)))
 }
+func validateComment(span *otrace.Span, Query string, q bool) {
+	// grab with regex the comment from the query. That starts with # and ends with dot
+	re := regexp.MustCompile(`(?m)#tag:(\w+)`)
+	querytag := re.FindStringSubmatch(Query)
+	if len(querytag) > 1 {
+		switch {
+		case !q:
+			span.AddAttributes(otrace.StringAttribute("mutation.tag", querytag[1]))
+		case q:
+			span.AddAttributes(otrace.StringAttribute("query.tag", querytag[1]))
+		}
+	}
+}
 
 func annotateStartTs(span *otrace.Span, ts uint64) {
 	span.AddAttributes(otrace.Int64Attribute("startTs", int64(ts)))
@@ -760,7 +774,7 @@ func buildUpsertQuery(qc *queryContext) string {
 			//      * be empty if the condition is true
 			//      * have 1 UID (the 0 UID) if the condition is false
 			upsertQuery += qc.condVars[i] + ` as var(func: uid(0)) ` + cond + `
-			 `
+			  `
 		}
 	}
 	upsertQuery += `}`
@@ -1292,6 +1306,18 @@ func (s *Server) doQuery(ctx context.Context, req *Request) (resp *api.Response,
 	}
 	if isMutation {
 		ostats.Record(ctx, x.NumMutations.M(1))
+	}
+
+	if isQuery || isMutation {
+		m := req.req.Query
+		if !isQuery {
+			m = req.req.String()
+		}
+		print(strings.Contains(m, "#"))
+		if strings.Contains(m, "#") {
+			q := isQuery
+			validateComment(span, m, q)
+		}
 	}
 
 	if req.doAuth == NeedAuthorize && x.IsGalaxyOperation(ctx) {
--- a/edgraph/server.go
+++ b/edgraph/server.go
@@ -548,7 +548,7 @@ func annotateNamespace(span *otrace.Span, ns uint64) {
 	span.AddAttributes(otrace.Int64Attribute("ns", int64(ns)))
 }
 func validateComment(span *otrace.Span, Query string, q bool) {
-	// grab with regex the comment from the query. That starts with # and ends with dot
+	// grab with regex the comment from the query.
 	re := regexp.MustCompile(`(?m)#tag:(\w+)`)
 	querytag := re.FindStringSubmatch(Query)
 	if len(querytag) > 1 {
@@ -1313,7 +1313,6 @@ func (s *Server) doQuery(ctx context.Context, req *Request) (resp *api.Response,
 		if !isQuery {
 			m = req.req.String()
 		}
-		print(strings.Contains(m, "#"))
 		if strings.Contains(m, "#") {
 			q := isQuery
 			validateComment(span, m, q)
--- a/edgraph/server.go
+++ b/edgraph/server.go
@@ -552,10 +552,10 @@ func validateComment(span *otrace.Span, Query string, q bool) {
 	re := regexp.MustCompile(`(?m)#tag:(\w+)`)
 	querytag := re.FindStringSubmatch(Query)
 	if len(querytag) > 1 {
-		switch {
-		case !q:
+		switch q {
+		case false:
 			span.AddAttributes(otrace.StringAttribute("mutation.tag", querytag[1]))
-		case q:
+		case true:
 			span.AddAttributes(otrace.StringAttribute("query.tag", querytag[1]))
 		}
 	}
--- a/query/query4_test.go
+++ b/query/query4_test.go
@@ -1564,3 +1564,51 @@ func TestNumUids(t *testing.T) {
 	require.Equal(t, metrics.NumUids["name"], uint64(16))
 	require.Equal(t, metrics.NumUids["_total"], uint64(26))
 }
+
+func TestJaegerTag(t *testing.T) {
+	query := `
+		 {
+			 #tag:test
+			 me(func: uid(0x01)) {
+				 uid
+			 }
+		 }
+	 `
+	js := processQueryNoErr(t, query)
+	require.JSONEq(t,
+		`{"data": {"me":[ {"uid": "0x1"}]}}`,
+		js)
+}
+
+func TestJaegerTag1(t *testing.T) {
+	query := `
+		 {
+			 me(func: uid(0x01)) {
+				 uid
+				 #tag:test
+			 }
+		 }
+	 `
+	js := processQueryNoErr(t, query)
+	require.JSONEq(t,
+		`{"data": {"me":[ {"uid": "0x1"}]}}`,
+		js)
+}
+
+func TestJaegerTag2(t *testing.T) {
+	query := `
+		 {
+			 me(func: has(<#tag:test>)) {
+				 uid
+				 <#tag:test>
+			 }
+		 }
+	 `
+	js := processQueryNoErr(t, query)
+	require.JSONEq(t,
+		`{"data": {"me":[ {
+			"uid": "0x12c",
+			"#tag:test": "Bob"
+		  }]}}`,
+		js)
+}
--- a/query/common_test.go
+++ b/query/common_test.go
@@ -274,7 +274,7 @@ type Speaker {
 }
 
 type JaegerTest {
-	name
+	_name
 	<#tag:test>
 }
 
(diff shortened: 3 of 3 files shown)


skrdgraph:
@MichelDiz is there a way you can add a test for this?

skrdgraph (changes requested):
@MichelDiz is it possible to add a test-case for your `validateComment`?

MichelDiz:
> @MichelDiz is it possible to add a test-case for your `validateComment`?

what would be the best approach?
I have to.
1. Spin up a cluster with Jaeger.
2. Create a script that use the filter tag in Jaeger. Which I have no idea how without some hacking.

Give me some ideas. I'm not good in creating tests.

edgraph/server.go
```diff
@@ -148,12 +149,12 @@ func GetGQLSchema(namespace uint64) (uid, graphQLSchema string, err error) {
 	resp, err := (&Server{}).QueryNoGrpc(ctx,
 		&api.Request{
 			Query: `
-			query {
```
mangalaman93, line 151: have we made a change here? If not, let's undo this.

edgraph/server.go
```diff
@@ -1294,6 +1308,17 @@ func (s *Server) doQuery(ctx context.Context, req *Request) (resp *api.Response,
 		ostats.Record(ctx, x.NumMutations.M(1))
 	}
 
+	if isQuery || isMutation {
+		m := req.req.Query
+		if !isQuery {
+			m = req.req.String()
+		}
+		if strings.Contains(m, "#") {
```
mangalaman93, line 1316: This is not a very robust logic and this should be implemented at the time of parsing the query instead. I am also curious how one would specify this tag in a client.

mangalaman93 (changes requested):
We can easily add some unit tests but I think we need to think about the design a little more.

edgraph/server.go
MichelDiz: That is go fmt, I won't pick a fight with it. Either I disable it (and lose any benefit from using it) or I just let it do what it wants to do. Is there any default config to enforce go fmt not doing it?

edgraph/server.go
MichelDiz: We have a very narrow window(to have it in beta) and we need this. Any improvement can come later. We can open a Jira Ticket to follow up.

I gave just the idea and I couldn't wait anyone to pick it up(feels like nobody would pick it). if you have a better way you can add your changes now or later. But this will help us a LOT during debugging in the cloud.

MichelDiz: Any client will work the same. It is just a comment, which is "native" for the queries and mutations.

edgraph/server.go
mangalaman93: I am not sure why it is doing it for you. My vscode also runs go fmt and it is not updating this file for me. In any case, you still should be able to undo this change. (run `git checkout origin/main edgraph/server.go`

edgraph/server.go
mangalaman93: Let's talk about this. I understand the urgency of it but I would really like to make sure the quality of release to the best that we can.

edgraph/server.go
MichelDiz: This command undo everything(all commits, not just that part) and also bring new stuff. It doesn't work as you expected.

I have nothing fancy in my Go env. Just started it and done. Maybe it is the version?

```
GO111MODULE="on"
GOARCH="amd64"
GOBIN=""
GOCACHE="/root/.cache/go-build"
GOENV="/root/.config/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/root/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/root/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GOVCS=""
GOVERSION="go1.20.1"
GCCGO="gccgo"
GOAMD64="v1"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/root/dev/dgraph/go.mod"
GOWORK=""
CGO_CFLAGS="-O2 -g"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-O2 -g"
CGO_FFLAGS="-O2 -g"
CGO_LDFLAGS="-O2 -g"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -Wl,--no-gc-sections -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build463741918=/tmp/go-build -gno-record-gcc-switches"
```

edgraph/server.go
damonfeldman: To me the two important things are:

 1) Is this a good, long-term way to tag queries that will work for the users?
 2) is this a safe change?

How it is coded is less critical, because if the user behavior is desirable, and it works, we can easily move it to a parsing layer (or somewhere else) later.

Similarly for testing - for now we can test manually, and next release we can add those same checks as automated unit test.

So the key as I see it is the behavior that is exposed to the users:  the first ```#tag:<sometext>``` comment in a query will cause a new attribute tag=<sometext> to be added to the otrace span. This seems like a good behavior to me, but this is the key item I think we should discuss.

Is it safe? Since the new code paths are "protected" first by a check if '#' is in the query, and then a regex match, it seems that this is very unlikely to break existing queries. It's also a little safer because the only behavior is adding an otrace attribute, vs changing any query behavior.

mangalaman93: I agree with most of what you said, Damon. These are my concerns.

I can't see right now that this would be a robust change. I am wondering what if my query itself contains the expected tag. It could show up as part of a string in the query. Will our code still work?

Another question I have is, will this be extensible to our dgraph clients, what about the JSON API for querying data? What about mutations and upserts later? The language change has to be well defined too and that would be true when lexer and parser can make sense of it correctly. 

We would need tests to ensure that we don't break any existing behaviour, I can't think of any other way to ensure that.

Why don't we use our TXNs API instead of modifying the query language? We could add another parameter to the API and let the user pass the tag there instead.

joshua-goldstein: @damonfeldman I'd like to echo Aman's concerns above.  The way this change is written right now is not very robust.  Some fundamental issues are:
- If the query itself contains the tag (not in a comment, but inside the query) then this will probably get picked up by the regex, and this is certainly undesirable behavior and may cause panics / errors since the user may not have tracing set up
- Furthermore this mixes up the logic of parsing the query (which belongs in a different package) into the `edgraph` package here which should not contain this kind of logic.  We have a lexer, parser, etc. for this purpose and should not be matching regexes due to the potential issues pointed out above.

I agree with @mangalaman93 above that we could modify the transaction API to let the user pass the tags there instead.  There are some subtleties there to think about but I think that is a much cleaner way to make such a change.

joshua-goldstein (changes requested):
See discussion above

edgraph/server.go
MichelDiz: I hope you read this. I'm addressing every point mentioned above. I agree that the code could be improved and put in a place (if there is a place for it) to look neat and tidy. However, all concerns raised are invalid. Which leads me to believe that you didn't test your theories before raising them. You didn't tested out the code. Mentally or in practice.

----------------

The point about being "robust" can be assessed later if your concern is beauty or performance - We need as it is now to be able to obtain data in the surgical search for problems with a specific query. 

The code only checks if it has a comment which starts with #. If there is no hashtag first it will be ignored. Yes, if there is another field with # it will be positive. But there's another check, with the regex that looks for a more complex pattern. If the pattern does not match. It doesn't matter if there are other hashtags in other corners of the query it will be ignored. That logic is simple to read from the code.

>I am wondering what if my query itself contains the expected tag. It could show up as part of a string in the query. Will our code still work?

I'm not sure I understood the question well. Because the answer is obvious, see, If your query has a comment (the comment is native to Dgraph, we support comments in queries and mutations, you can have 90% of the query only comments and it won't be a problem) with the expected tag. it will just append the tag to the span trace that was already created during the start of the transaction. It's just an append.

The part "It could show up as part of a string in the query" I don't understand. Do you mean the response or the body of the trace? Yes, all comments whether tagged or not - with this PR or not - will be present in the trace. This is the default behavior, I didn't define it with this code. But it would never be part of the response. That's impossible. 

>Another question I have is, will this be extensible to our dgraph clients, what about the JSON API for querying data? What about mutations and upserts later?

For those who understand the flow of the query execution, the answer to this question would be simple. The parsing of the query happens way before this step in which I put this check. The query transformation has already been done. It doesn't matter if the query will come in JSON or DQL or RDF. It was already parsed well before reaching this step of the code. 

The only thing missing is mutation in JSON. Since JSON does not support comments. Mutations in JSON will be stopped long before the transaction even starts. Because JSON does not support comments.

Yes. This code works for RDF and Upserts (Read the comment of the PR - I have put an example for RDF). In the case of Upserts they only work if the comment is in the query block, if you put it in the mutation block it will be ignored.

> The language change has to be well defined too and that would be true when lexer and parser can make sense of it correctly.

There's NO language change here. I'm using an existing feature which is comments.

> Why don't we use our TXNs API instead of modifying the query language?

Again, there's NO language change... How can't you see that? oO

> We could add another parameter to the API and let the user pass the tag there instead.

That would be nice, if not take months or years to do. I mean, pretty sure it would be fast to code. But the problem is "willing" and collaboration.

MichelDiz: @joshua-goldstein 

>If the query itself contains the tag (not in a comment, but inside the query) then this will probably get picked up by the regex

Never. We can bet. :P I can bet 5 months salary on this.

First: The query has to have the hashtag. Which is possible. You can have a predicate like <#.hashtag> or a value that has hashtags. That is fine. But only the first check it will grab it.

Second: The regex expression  `(?m)#tag:(\w+)` is used to match a pattern that begins with the string "#tag:" followed by one or more word characters without space. If your query doesn't start with # and doesn't have the word "tag" it will be ignored. You can test out in https://regex101.com/r/SB97n7/1 if you wanna confirm. In this example we have two valig tags that the regex will capture. BUT the lines 557 and 559 in the code I'm selecting only the first found with `querytag[1]`.

So the Regex will trap only one match and that match has to start with #tag: and have a word after the colon. If you don't do like this, it will be ignored.

>Furthermore this mixes up the logic of parsing the query

Not really. I'm not parsing anything. Just checking if it exist and picking. Not transforming anything. It is similar to parsing, but I'm not transforming the query into an object that will be used in another part of the code. That's the meaning of parse. Transform something meaningful for your architecture.


>I agree with @mangalaman93 above that we could modify the transaction API to let the user pass the tags there instead. There are some subtleties there to think about but I think that is a much cleaner way to make such a change.

I also agree, but I'm pretty sure this code is fine. And you can do whatever you want later. All concerns above are not valid. And we need this feature for customer success and other debug activities. We don't have time to wait months for someone to make the decision to do it. I commented on this feature in Slack weeks ago. And it was ignored, nobody wanted to work or discuss it. I went there and did it. Well, cuz it's simple and easy. Now, if you want to do something more bureaucratic and supposedly better and safer. I support it and I'm glad this caught your attention and you're interested in improving. That is good.

MichelDiz: By the way, I would be very happy to review any designs you guys come up with. Truly glad you guys invested your time developing something on this idea.

The idea of adding a new entry in the API is good, but it seems to me that you want to use a Bazooka against an ant. Create an entire design for a single tag. Like, spending a fortune on time for such small thing.

Like, there's a (fake) joke about Russians and Americans in space. Where the US spent billions of dollars to develop a pen that could write in space. While the Russians used a simple pencil.

damonfeldman: These are all good points. I do think the change is pretty safe, since annotations are lower impact than query or process changes, but even so, testing is always important.

Let's do what I think both Aman and Michel are suggesting and set up a design meeting on this once the v23 crunch is over. We'll end up with an ideal solution in the end, I am sure.

query/query4_test.go
```diff
@@ -1564,3 +1564,51 @@ func TestNumUids(t *testing.T) {
...
+		`{"data": {"me":[ {"uid": "0x1"}]}}`,
+		js)
+}
+
+func TestJaegerTag1(t *testing.T) {
+	query := `
+		 {
+			 me(func: uid(0x01)) {
+				 uid
+				 #tag:test
```
mangalaman93, line 1588: @MichelDiz will this tag the query in jager? Is it possible to check for that?

query/query4_test.go
```diff
@@ -1564,3 +1564,51 @@ func TestNumUids(t *testing.T) {
...
+		`{"data": {"me":[ {"uid": "0x1"}]}}`,
+		js)
+}
+
+func TestJaegerTag2(t *testing.T) {
+	query := `
+		 {
+			 me(func: has(<#tag:test>)) {
+				 uid
+				 <#tag:test>
```
mangalaman93, line 1603: same question here.

mangalaman93 (changes requested):

query/query4_test.go
MichelDiz: Yes, I carried out all these tests while running Jaeger concurrently. They all succeeded.

It would be conceivable to think of a process that could somehow infer this from Jaeger's API. However, I did not delve into this due to its complexity. Therefore, I focused on the requests, aiming to reproduce panics with the tags, which in this case, occurred without any panics.

mangalaman93: But that is not right. Isn't this a `#tag:test` predicate in this case? It shouldn't be interpreted as a jaeger tag from what I understand.

MichelDiz: Nope, it is ignored by the parse as it is a comment. You can comment anywhere in the query. As long it doesn't break the syntax.

If you check the code. What I'm doing is just inferring comments. With that I can capture the tag. Because Dgraph passes the complete query at this stage of execution. This step is before parsing. Parsing happens after that stage.

query/query4_test.go
MichelDiz: For this one. You can see in `query/common_test.go` that I added the predicate `<#tag:test>`. This is captured by the logic as the function uses `"strings.Contains(m, "#tag")" `, but it doesn't cause any side effects. I consider this query to be okay. It's rare for someone to use this pattern in predicates, and even if it does happen, it's not an issue, its effect is innocuous.

But yes. This will be captured. As the hashtag along with the name "tag" and the colon is tha pattern that I'm looking for.

PR #8718, opened 2023-03-03
chore(linter): fix unconvert linter issues on linux

mangalaman93:


skrdgraph (approved):
Looks good. (@mangalaman93 add a description, as that translates to commit messages - we are trying to maintain a consistent approach on detailing the #problem & #solution in the templated description).

anurags92 (approved):

PR #8719, opened 2023-03-03
fix: check index for next word and replace global index with local one

hezhizhen:
## Problem
 

`nextWord()` used a global index and simply got the word from the list based on the index without checking if it's in the range.

## Solution
 

Pass the index as an argument to the function and let the callers maintain the index themselves (increase it here), and check if the index is valid.

I also fixed some typo and comments by the way.

### DIFF
--- a/contrib/integration/swap/main.go
+++ b/contrib/integration/swap/main.go
@@ -140,8 +140,7 @@ func main() {
 func getNextWord(index int) string {
 	// check if index is in the range of words
 	if index < 0 || index >= len(words) {
-		x.Log(errors.Errorf("invalid index: %d", index), "get next word")
-		return ""
+		x.Fatalf("invalid index for getting next word: %d", index)
 	}
 	return words[index]
 }
--- a/contrib/integration/swap/main.go
+++ b/contrib/integration/swap/main.go
@@ -137,10 +137,21 @@ func main() {
 
 }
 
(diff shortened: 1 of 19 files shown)


hezhizhen:
@skrdgraph

contrib/integration/swap/main.go
```diff
@@ -137,10 +137,21 @@ func main() {
 
 }
 
+func getNextWord(index int) string {
+	// check if index is in the range of words
+	if index < 0 || index >= len(words) {
+		x.Log(errors.Errorf("invalid index: %d", index), "get next word")
```
mangalaman93, line 143: we should crash or something here instead of silently failing

mangalaman93 (changes requested):
one comment, LG otherwise. Thanks for the PR

contrib/integration/swap/main.go
hezhizhen: Done. PTAL

mangalaman93:
rerunning the failed test. It is failing due to a known issue, we are looking into it.

mangalaman93 (approved):

hezhizhen:
> rerunning the failed test. It is failing due to a known issue, we are looking into it.

@mangalaman93 How can I re-run failed tests?

MichelDiz (approved):

PR #8723, opened 2023-03-03
Add Upgrade and setup binaries function

shivaji-dgraph:


dgraphtest/cluster.go
```diff
@@ -207,6 +224,21 @@ func (c *Cluster) stopContainer(dc dnode) error {
 	return nil
 }
 
+func (c *Cluster) Upgrade(versions versions) error {
```
mangalaman93, line 227: You can just take a `version string` as an argument

dgraphtest/cluster.go
```diff
@@ -207,6 +224,21 @@ func (c *Cluster) stopContainer(dc dnode) error {
...
+	//stop all containers
+	c.Stop()
+	// setup function
+
+	if err := flipVersion(); err != nil {
```
mangalaman93, line 232: you should pass this version here so that we can upgrade to the right version

dgraphtest/cluster.go
```diff
@@ -21,29 +21,46 @@ import (
...
+const (
+	dgraphRepoUrl = "https://github.com/dgraph-io/dgraph.git"
 )
 
 // cluster's network struct
 type cnet struct {
 	id   string
 	name string
 }
+type versions struct {
```
mangalaman93, line 56: we don't need this type

dgraphtest/config.go
```diff
@@ -35,6 +35,7 @@ type ClusterConfig struct {
 	acl        bool
 	encryption bool
 	logr       Logger
+	isUpgrade  bool
```
mangalaman93, line 38: we don't need it

dgraphtest/config.go
```diff
@@ -78,3 +79,8 @@ func (cc ClusterConfig) WithLogger(logr Logger) ClusterConfig {
 	cc.logr = logr
 	return cc
 }
+
+func (cc ClusterConfig) WithUpgrade() ClusterConfig {
```
mangalaman93, line 83: don't need it

dgraphtest/dgraph.go
```diff
@@ -102,14 +102,31 @@ func (z *zero) workingDir() string {
...
-	return []mount.Mount{
-		{
-			Type:     mount.TypeBind,
-			Source:   os.Getenv("GOPATH") + "/bin",
-			Target:   "/gobin",
-			ReadOnly: true,
-		},
-	}, nil
+
+	if conf.isUpgrade {
```
mangalaman93, line 106: we should decide this based on the current version of dgraph that the cluster is running

dgraphtest/dgraph.go
```diff
@@ -173,13 +190,25 @@ func (a *alpha) workingDir() string {
 }
 
 func (a *alpha) mounts(conf ClusterConfig) ([]mount.Mount, error) {
-	mounts := []mount.Mount{
-		{
+	var mounts []mount.Mount
+	if conf.isUpgrade {
```
mangalaman93, line 194: same

dgraphtest/cluster.go
```diff
@@ -300,3 +332,104 @@ func (c *Cluster) containerHealthCheck(url string) error {
 
 	return fmt.Errorf("failed health check on [%v]", url)
 }
+
+func SetupBinaries(versions versions) error {
```
mangalaman93, line 336: you can move all the binary related code to `image.go`

dgraphtest/cluster.go
```diff
@@ -21,29 +21,46 @@ import (
...
+	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/pkg/errors"
 )
 
 var (
-	requestTimeout = 30 * time.Second
-	stopTimeout    = time.Minute
+	requestTimeout    = 30 * time.Second
+	stopTimeout       = time.Minute
+	homePath          = os.Getenv("HOME") + "/data/dgraph"
```
mangalaman93, line 41: I think what we should do here is that, we should have a relative folder where we keep all the binaries. The name of the binaries could be something like `dgraph_{version}`. Then, every time we need to mount the binary inside the dir, we copy the binary into a temp folder (we can get it using os.TempDir) and then mount it inside the container. Doing this will ensure that we can run multiple clusters in parallel.

PR #8727, opened 2023-03-04
chore(debug): Add VSCode config file for debugging.

### LINKED ISSUE #8752: [ENHANCEMENT]: Add Vscode debug config.
### Use case and current behavior

N/A

### Enhancement

N/A

### Solution proposal

Add .vscode/launch.json at the root directory.

### Links to Discuss, RFC or previous Issues and PRs

https://github.com/dgraph-io/dgraph/pull/8727

### Links to examples and research

_No response_

### Additional Information

This is just for tracking as Akon requested.

MichelDiz:
Just sharing the configuration so that others can also debug via VS code. The extra addition is the possibility to use Jaeger.

Fix #8752


mangalaman93:
@MichelDiz Thanks for the PR. I am not sure if we want to keep this in the Dgraph repo. Everyone has their own preferences for using vscode and for configuring it. For example, I use delve. I know other engineers who don't even use vscode. I would prefer to keep things that are really needed so that we reduce the our maintenance work.

MichelDiz:
@mangalaman93  we share a bunch of things at https://github.com/dgraph-io/dgraph/tree/main/contrib and we say
>Please note that they could be out of date, since they do not receive the same attention as the rest of the repository.

I believe that the more we facilitates for a possible contributors, the better. That's why I'm sharing this. And I believe that maintenance to this config will only be done in decades.

I've been sharing this config internally since 2020 and it has proven to be useful even for those who have other methods. Why not share here? instead of copy paste in the chat.

Cheers.

mangalaman93:
I would still be in favour of a public gist instead of this repo. We can also add it to the documentation, that could still be okay.

mangalaman93:
> Think of it like this, argument in favor of not adding in the core repo - mainly aesthetics, argument in favor of adding - useful for devs using vscode. They can get started without going into docs etc.

I don't think aesthetics is the reason I do not prefer to add it in the repo. It is because of the maintenance work that follows. It is easy to add things to repo but something always changes somewhere, either in Dgraph or in vscode (in this case). And then someone needs to fix this part of the code. Or it is gonna just lie there until it fails for someone.

Not just that, this PR only adds config for vscode. Tomorrow, someone else may want to add config for vim and then IntelliJ too. And now it is a whole folder full of config that either is not maintained well, or requires some maintenance work to ensure that it still works.

I personally would like to minimize the maintenance work that has to be done.

anurags92:
> It is because of the maintenance work that follows. It is easy to add things to repo but something always changes somewhere, either in Dgraph or in vscode (in this case). And then someone needs to fix this part of the code. Or it is gonna just lie there until it fails for someone.

Maintenance work would be same if we add it in docs. And I feel it is okay for it lie there until it fails for someone, they will open an issue and we will try to fix it.

> Not just that, this PR only adds config for vscode. Tomorrow, someone else may want to add config for vim and then IntelliJ too.

There's a difference, as I see it between vscode and other editors. `vim` is pure editor in a sense that configurations are not project specific. Here the `vscode` `launch.json` is specific to Dgraph to be able to run Dgraph in vscode environment. For Goland (intellij) I don't think we export similar config file to run projects. If we do, it can as well stay in .intellij folder or something. 

Marginal cost of maintenance (which would be there in docs as well) is less than the benefit of finding the supported config in the repo.

MichelDiz:
> It is because of the maintenance work that follows.

As I mentioned before. Any maintenance for that config will take decades. The first version of this from 2.3 years ago still works. And this one just adds Jaeger and ACL.

> Tomorrow, someone else may want to add config for vim and then IntelliJ too. And now it is a whole folder full of config that either is not maintained well, or requires some maintenance work to ensure that it still works.

Easy, say that we "support" only one. I think vscode debugging is a killer thing. Even customers can play around with it as it is no brainer to use it. Just click a button. That would bring inclusion as anyone(not just Go Wizards) can go there add some breakpoints and test anything they want. Including understanding the code by following the "step in" steps.

>We can also add it to the documentation, that could still be okay.

I think let it in docs is bad cuz it is hidden. And in Docs we don't teach the user on how to contribute in the core code. The docs are meant for DQL/GraphQL usage. Not core code. And we don't have a documentation specific for developers. Things get unclear even reading comments in the code. Nothing is really documented. This will change when we bring out the RFCs as an obligation.

But for this case, I think it is just a piece of config file that isn't really a problem to keep there.

akon-dey:
I think it makes sense to have this configuration. Only VSCode users will benefit but non-VSCode users will have no impact.

MaxxReis (approved):

PR #7870, opened 2021-05-28
opt(snapshot): use full table copy when streaming the entire data

NamanJain8:
When streaming the entire data in the snapshot (`snap.SinceTs=0`), we can do an entire table copy instead of iterating over the KVs. This brings about 3x performance improvement, as well as the CPU of the sender, is idle. Refer https://github.com/dgraph-io/badger/pull/1700 for more details.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7870)

PR #8736, opened 2023-03-07
opt(snapshot): use full table copy when streaming the entire data (#7870)

mangalaman93:
When streaming the entire data in the snapshot (snap.SinceTs=0), we can do an entire table copy instead of iterating over the KVs. This brings about 3x performance improvement, as well as the CPU of the sender, is idle. Refer dgraph-io/badger#1700 for more details.


anurags92 (approved):

ryanfoxtyler:
closing PR for now, keeping branch active

PR #8740, opened 2023-03-08
chore(worker): unify mapper receiver names

hezhizhen:
## Problem
 

`mapper` methods have different receiver names: `m` and `mw`.

## Solution
 

Unify receiver names for `mapper`.

Simplify some code btw.

### DIFF
--- a/worker/backup_handler.go
+++ b/worker/backup_handler.go
@@ -276,10 +276,9 @@ func (h *s3Handler) FileExists(path string) bool {
 		return true
 	}
 	errResponse := minio.ToErrorResponse(err)
-	if errResponse.Code == "NoSuchKey" {
-		return false
+	if errResponse.Code != "NoSuchKey" {
+		glog.Errorf("Failed to verify object existence: %v", err)
 	}
-	glog.Errorf("Failed to verify object existence: %v", err)
 	return false
 }
 
--- a/worker/backup_handler.go
+++ b/worker/backup_handler.go
@@ -272,16 +272,15 @@ func (h *s3Handler) DirExists(path string) bool  { return true }
 func (h *s3Handler) FileExists(path string) bool {
 	objectPath := h.getObjectPath(path)
 	_, err := h.mc.StatObject(h.bucketName, objectPath, minio.StatObjectOptions{})
-	if err != nil {
-		errResponse := minio.ToErrorResponse(err)
-		if errResponse.Code == "NoSuchKey" {
-			return false
-		} else {
-			glog.Errorf("Failed to verify object existence: %v", err)
-			return false
-		}
+	if err == nil {
+		return true
 	}
-	return true
+	errResponse := minio.ToErrorResponse(err)
+	if errResponse.Code == "NoSuchKey" {
+		return false
+	}
+	glog.Errorf("Failed to verify object existence: %v", err)
+	return false
 }
 
 func (h *s3Handler) JoinPath(path string) string {
--- a/worker/draft.go
+++ b/worker/draft.go
@@ -119,7 +119,7 @@ func (n *node) startTask(id op) (*z.Closer, error) {
 
 // startTaskAtTs is used to check whether an op is already running. If a rollup is running,
 // it is canceled and startTask will wait until it completes before returning.
-// If the same task is already running, this method returns an errror.
+// If the same task is already running, this method returns an error.
 // Restore operations have preference and cancel all other operations, not just rollups.
 // You should only call Done() on the returned closer. Calling other functions (such as
 // SignalAndWait) for closer could result in panics. For more details, see GitHub issue #5034.
@@ -179,9 +179,8 @@ func (n *node) startTaskAtTs(id op, ts uint64) (*z.Closer, error) {
 				if otherOp.ts < ts {
 					// If backup is running at higher timestamp, then indexing can't be executed.
 					continue
-				} else {
-					return nil, errors.Errorf("operation %s is already running", otherId)
 				}
+				return nil, errors.Errorf("operation %s is already running", otherId)
 			case opRollup:
 				// Remove from map and signal the closer to cancel the operation.
 				delete(n.ops, otherId)
@@ -192,13 +191,12 @@ func (n *node) startTaskAtTs(id op, ts uint64) (*z.Closer, error) {
 		}
 	case opSnapshot, opPredMove:
 		for otherId, otherOp := range n.ops {
-			if otherId == opRollup {
-				// Remove from map and signal the closer to cancel the operation.
-				delete(n.ops, otherId)
-				otherOp.SignalAndWait()
-			} else {
+			if otherId != opRollup {
 				return nil, errors.Errorf("operation %s is already running", otherId)
 			}
+			// Remove from map and signal the closer to cancel the operation.
+			delete(n.ops, otherId)
+			otherOp.SignalAndWait()
 		}
 	default:
 		glog.Errorf("Got an unhandled operation %s. Ignoring...", id)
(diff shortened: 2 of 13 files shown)


worker/backup_handler.go
```diff
@@ -272,16 +272,15 @@ func (h *s3Handler) DirExists(path string) bool  { return true }
...
-		if errResponse.Code == "NoSuchKey" {
-			return false
-		} else {
-			glog.Errorf("Failed to verify object existence: %v", err)
-			return false
-		}
+	if err == nil {
+		return true
+	}
+	errResponse := minio.ToErrorResponse(err)
```
mangalaman93, line 278: can we do something like this instead
```
if resp := minio.ToErrorResponse(err); resp.Code != "NoSuchKey" {
    glog.Errorf("Failed to verify object existence: %v", err)
}
return false
```

worker/draft.go
```diff
@@ -192,13 +191,12 @@ func (n *node) startTaskAtTs(id op, ts uint64) (*z.Closer, error) {
 		}
 	case opSnapshot, opPredMove:
 		for otherId, otherOp := range n.ops {
-			if otherId == opRollup {
```
mangalaman93, line 195: I understand that the modified code is more Go like but the original code seems more readable to me.

mangalaman93 (changes requested):

worker/backup_handler.go
hezhizhen: Sure.

worker/draft.go
hezhizhen: `if` and `else` are rather straightforward, which makes the code seem more readable. Go's style is a few small tweaks without loosing the readability.

In this case, I first change the order of `if` and `else`:

```go
if otherId != opRollup { return nil, errors.Errorf("operation %s is already running", otherId) }
else {
    // Remove from map and signal the closer to cancel the operation.
    delete(n.ops, otherId)
    otherOp.SignalAndWait()
}
```

Then I remove the unnecessary `else`:

```go
if otherId != opRollup { return nil, errors.Errorf("operation %s is already running", otherId) }

// Remove from map and signal the closer to cancel the operation.
delete(n.ops, otherId)
otherOp.SignalAndWait()
```

mangalaman93: yeah, I get that the code is equivalent. What I don't like about the modified code is that it does `delete(n.ops, otherId)` and I don't know what otherId is until I see the if condition. This seems a bit convoluted to read to me.

hezhizhen: I see... Since you read the function in a linear way (top down), you see the `if` condition first, and then you see `delete(n.ops, otherId)`. Although it is not wrapped in a `else` block, the `return` in the `if` block indicates it. 

I suppose that you are not accustomed to this style because you are more familiar with other languages. It's okay to revert it if it really bothers you.

hezhizhen:
@mangalaman93 The code coverage test failed. Could you please re-run it?

worker/draft.go
mangalaman93: maybe we could rewrite the code like this. This would feel more explicit and clear.
```
if otherId != opRollup { return nil, errors.Errorf("operation %s is already running", otherId) }

// Remove from map and signal the closer to cancel the operation.
delete(n.ops, opRollup)
otherOp.SignalAndWait()
```

hezhizhen: Sure.

hezhizhen: But why changing `otherId` to `opRollup`?

hezhizhen: `otherId` is the same as `opRollup`, but the former is a key of `n.ops`, and the latter is something used to compare with the key. So the semantical meaning is  little wierd. 

Maybe I can update comments like `otherId is opRollup, remove from map and signal the closer to cancel the operation.`?

mangalaman93: Could you elaborate on this?
```
otherId is the same as opRollup, but the former is a key of n.ops, and the latter is something used to compare with the key. So the semantical meaning is little wierd.
```

hezhizhen: > Could you elaborate on this?
> 
> ```
> otherId is the same as opRollup, but the former is a key of n.ops, and the latter is something used to compare with the key. So the semantical meaning is little wierd.
> ```

`ns.ops` is a map whose keys are op integers and values are operation structs. `otherId` is its key when iterating it, so `delete(ns.ops, otherId)` means deleting the `otherId` key from the map. For the case above, it means deleting a key of the map when it is `opRollup` (otherwise don't).

`opRollup` is constant, indicating rolling up operation (I'm not quite sure of it, and correct me if I'm wrong). `delete(ns.ops, opRollup)` means deleting the certain key `opRollup` from the map, no matter if it exists. For the case above, it means deleting the `opRollup` itself when a key of the map is `opRollup`.

In my opinion, the slight difference lies in what you delete: a key of the map or the certain key itself.

hezhizhen: I'm not sure if I elaborated it well, and the difference is rather slight.

mangalaman93: well, that is exactly my concern. When we do the check `== opRollup` it is guaranteed that we delete the `opRollup` key from the map. When we do the check `!= opRollup`, we don't know what we are deleting. (well, the code would only delete from the map if it is opRollup but,). Usually what happens is that a nearby change in code could lead to a case where we end up deleting other keys here in the code because the check is different. And this code won't complain that we are deleting other key than opRollup.

hezhizhen: > When we do the check == opRollup it is guaranteed that we delete the opRollup key from the map. When we do the check != opRollup, we don't know what we are deleting.

In this case, we only delete the key when the check `== opRollup` is passed, so we don't need to worry about *When we do the check != opRollup, we don't know what we are deleting*.

It seems that you don't enough confidence in the correctness when `delete(n.ops, otherId)` is not in a `if` block.

As for further modification like deleting other keys:

- in the old style: expand the `if` condition like `if otherId == opRollup || otherId == opBackup`
- in the new style: expand the `if` condition like `if otherId != opRollup && otherId != opBackup`

What may be a concern is the expansion itself: will `||` (or) be easier to be understood than `&&` (and), or vice versa?

mangalaman93: I have full confidence in the code, and you are right that it would work correctly. My challenge is with the situation when we make changes to this code in the future and we may miss out on this fact that the delete happens only for a particular case because that is not immediately obvious. I hope that clears things up.

hezhizhen: The initial version is: 

```go
if otherId == opRollup {
    // Remove from map and signal the closer to cancel the operation. 
    delete(n.ops, otherId) 
    otherOp.SignalAndWait()
} else {
    return nil, errors.Errorf("operation %s is already running", otherId) 
}
```

It is rather straightfoward.

Go's version is:

```go
if otherId != opRollup {
    return nil, errors.Errorf("operation %s is already running", otherId) 
}

// Remove from map and signal the closer to cancel the operation. 
delete(n.ops, otherId) 
otherOp.SignalAndWait()
```

Here is your suggestion:

```go
if otherId != opRollup {
    return nil, errors.Errorf("operation %s is already running", otherId) 
}

// Remove from map and signal the closer to cancel the operation. 
delete(n.ops, opRollup) 
otherOp.SignalAndWait()
```

It is to make sure that the deleted key is `opRollup` (correct me if I'm wrong).

Let's think about adding more handling logic for key `opBackup`.

- initial version: add an `else if` block
- Go's version: use `switch` and handle `opRollup` and `opBackup` cases respectively (my suggestion)
- your suggestion: just like Go's version

Actually I'm not sure if I'm simplifying it or decreasing its readability. I'm accustomed to Go's style and it does reduces complexity. Last but not the least, we can add unit tests to ensure its correctness.

mangalaman93: Let's do that. If you write unit test for it, then I don't care what design you choose. Because if I make a change later on that is messing up something, my unit tests will fail. I would suggest that get this PR merged without any change here and raise a separate PR with this change and tests for it.

hezhizhen: > Let's do that. If you write unit test for it, then I don't care what design you choose. Because if I make a change later on that is messing up something, my unit tests will fail. I would suggest that get this PR merged without any change here and raise a separate PR with this change and tests for it.

Okay. I'll do it later.

hezhizhen: Done. PTAL @mangalaman93

mangalaman93 (approved):

anurags92 (approved):

PR #8750, opened 2023-03-13
chore(tests): add upgrade tests in query package

mangalaman93:


query/upgrade_test.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		panic(err)
+	}
+	client = dg
+
+	// do mutations
+	populateCluster()
+
+	// upgrade
+	x.Panic(c.Upgrade("v23.0.0-beta1"))
```
joshua-goldstein, line 51: how will this work in the future?  we have hard-coded the next version into testmain, what comes next?  Should we replace this with the next version?  This will be difficult to version and keep track of

query/upgrade_test.go
```diff
@@ -0,0 +1,62 @@
...
+	"testing"
+	"time"
+
+	"github.com/dgraph-io/dgraph/dgraphtest"
+	"github.com/dgraph-io/dgraph/x"
+)
+
+func TestMain(m *testing.M) {
+	conf := dgraphtest.NewClusterConfig().WithNumAlphas(3).WithNumZeros(3).WithReplicas(3).
+		WithACL(20 * time.Second).WithEncryption().WithVersion("0c9f60156")
```
joshua-goldstein, line 31: what SHA does this refer to?

query/rdf_result_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build integration || cloud
+//go:build integration || cloud || upgrade
```
joshua-goldstein, line 1: keeping track of these tags will become difficult, is there another way to initialize the test spin up?

query/upgrade_test.go
mangalaman93: Probably one before the `v23.0.0-beta1` but this will get replaced in the next PR. This PR ensures that we can run upgrade tests in query package, any upgrade tests.

query/rdf_result_test.go
mangalaman93: I am hoping we won't have any more tags. I am also hoping that eventually, integration2 will also be renamed to integration.

query/common_test.go
```diff
@@ -352,7 +352,7 @@ func populateCluster() {
 		panic(err)
 	}
 
-	if err := dc.AssignUids(21000); err != nil {
+	if err := dc.AssignUids(65536); err != nil {
```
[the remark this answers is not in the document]
mangalaman93: doing this in the next PR.

harshil-goel (approved):

PR #8758, opened 2023-03-16
fix(testing info) : Updated info

SashaDgraph:
## Problem
 

The testing information in CONTRIBUTING.md was outdated.

## Solution
 
Updated the testing information in CONTRIBUTING.md


CONTRIBUTING.md
```diff
@@ -138,22 +138,11 @@ Docker images that contains `dgraph` and `badger` commands.
...
-    INFO: Running test for github.com/dgraph-io/dgraph/codec/benchmark
-    ?   	github.com/dgraph-io/dgraph/codec/benchmark	[no test files]
-    …
-
-Run `test.sh --help` for more info.
+1. Change directory to t directory. 
+2. If all packages need to be tested, run
+      make test
+   If only a specific package needs to be tested, run
+      make test args="--pkg=desired_package_name"
```
mangalaman93, line 145: could we add an example here? It may not be clear as too what -pkg should be set to. And I am wondering whether we should also specify how to use the `t` binary?

mangalaman93 (changes requested):

CONTRIBUTING.md
skrdgraph: can we clean up the makefile and reference the `make test` here?

PR #271, opened 2016-10-06
Making all but one tests pass

pawanrawal:
All except the JoinCluster test pass, have skipped JoinCluster test for now. I will probably fix sometime later once the group related logic stabilizes.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/271)

PR #8766, opened 2023-03-19
Fix zero's health endpoint to return a json response

jbhamra1:
Zero's /health endpoint returns just "OK" where the same Alpha's endpoint returns a JSON. Now, zero's '/health' endpoint also returns a JSON response if the "Accept" HTTP header in the request it set to "application/json". It continues to return "OK" by default.

```
$ curl -H "Accept: application/json" http://localhost:61983/health
{"instance":"zero","address":"zero3:5080","status":"healthy","version":"v23.0.0-beta1-28-g05b90a580","uptime":9223372036,"lastEcho":1679305435}
```

Earlier:
```
$ curl http://localhost:61983/health
OK
```

### DIFF
--- a/dgraph/cmd/zero/http.go
+++ b/dgraph/cmd/zero/http.go
@@ -239,14 +239,14 @@ func (s *Server) zeroHealth(ctx context.Context) (*api.Response, error) {
 		return nil, ctx.Err()
 	}
 
-	healthAll := []pb.HealthInfo{pb.HealthInfo{
+	healthAll := pb.HealthInfo{
 		Instance:    "zero",
 		Address:     x.WorkerConfig.MyAddr,
 		Status:      "healthy",
 		Version:     x.Version(),
 		Uptime:      int64(time.Since(x.WorkerConfig.StartTime) / time.Second),
 		LastEcho:    time.Now().Unix(),
-	}}
+	}
 
 	jsonOut, err := json.Marshal(healthAll)
 	if err != nil {
@@ -256,22 +256,29 @@ func (s *Server) zeroHealth(ctx context.Context) (*api.Response, error) {
 }
 
 func (st *state) pingResponse(w http.ResponseWriter, r *http.Request) {
-	var err error
 	x.AddCorsHeaders(w)
 
 	switch r.Header.Get("Accept") {
 		case "application/json":
-			var resp *api.Response
-			if resp, err = (st.zero).zeroHealth(r.Context()); err != nil {
+			resp, err := (st.zero).zeroHealth(r.Context())
+			if err != nil {
 				x.SetStatus(w, x.Error, err.Error())
 				return
 			}       
 			w.Header().Set("Content-Type", "application/json; charset=utf-8")
 			w.WriteHeader(http.StatusOK)
-			_, _ = w.Write(resp.Json)
+			_, err = w.Write(resp.Json)
+			if err != nil {
+				x.SetStatus(w, x.Error, err.Error())
+				return
+			}       
 		default:
 			w.Header().Set("Content-Type", "text/plain; charset=utf-8")
 			w.WriteHeader(http.StatusOK)
-			_, _ = w.Write([]byte("OK"))
+			_, err := w.Write([]byte("OK"))
+			if err != nil {
+				x.SetStatus(w, x.Error, err.Error())
+				return
+			}       
 	}
 }
--- a/dgraph/cmd/zero/http.go
+++ b/dgraph/cmd/zero/http.go
@@ -267,17 +267,15 @@ func (st *state) pingResponse(w http.ResponseWriter, r *http.Request) {
 			}       
 			w.Header().Set("Content-Type", "application/json; charset=utf-8")
 			w.WriteHeader(http.StatusOK)
-			_, err = w.Write(resp.Json)
-			if err != nil {
-				x.SetStatus(w, x.Error, err.Error())
+			if _, err := w.Write(resp.Json); err != nil {
+				glog.Warningf("Could not send error msg=%+v code=%+v due to http error %+v", err.Error(), x.Error, err)
 				return
 			}       
 		default:
 			w.Header().Set("Content-Type", "text/plain; charset=utf-8")
 			w.WriteHeader(http.StatusOK)
-			_, err := w.Write([]byte("OK"))
-			if err != nil {
-				x.SetStatus(w, x.Error, err.Error())
+			if _, err := w.Write([]byte("OK")); err != nil {
+				glog.Warningf("Could not send error msg=%+v code=%+v due to http error %+v", err.Error(), x.Error, err)
 				return
 			}       
 	}
--- a/dgraph/cmd/zero/http.go
+++ b/dgraph/cmd/zero/http.go
@@ -23,12 +23,16 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"encoding/json"
 
+	"github.com/pkg/errors"
 	"github.com/gogo/protobuf/jsonpb"
 	"github.com/golang/glog"
 
 	"github.com/dgraph-io/dgraph/protos/pb"
+	"github.com/dgraph-io/dgo/v210/protos/api"
 	"github.com/dgraph-io/dgraph/x"
+	"github.com/dgraph-io/dgraph/worker"
 )
 
 // intFromQueryParam checks for name as a query param, converts it to uint64 and returns it.
@@ -231,8 +235,59 @@ func (st *state) getState(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (s *Server) zeroHealth(ctx context.Context, all bool) (*api.Response, error) {
+	if ctx.Err() != nil {
+		return nil, ctx.Err()
+	}
+
+	var healthAll []pb.HealthInfo
+
+	// Append self.
+	healthAll = append(healthAll, pb.HealthInfo{
+		Instance:    "zero",
+		Address:     x.WorkerConfig.MyAddr,
+		Status:      "healthy",
+		Group:       strconv.Itoa(int(worker.GroupId())),
+		Version:     x.Version(),
+		Uptime:      int64(time.Since(x.WorkerConfig.StartTime) / time.Second),
+		LastEcho:    time.Now().Unix(),
+	})
+
+	var err error
+	var jsonOut []byte
+	if jsonOut, err = json.Marshal(healthAll); err != nil {
+		return nil, errors.Errorf("Unable to Marshal. Err %v", err)
+	}
+	return &api.Response{Json: jsonOut}, nil
+}
+
 func (st *state) pingResponse(w http.ResponseWriter, r *http.Request) {
+	var err error
 	x.AddCorsHeaders(w)
+
+	var ok = false
+	for _, s := range r.Header["Content-Type"] {
+		if s == "application/json" {
+			ok = true
+			break
+		}
+	}
+	if ok {
+		var resp *api.Response
+		if resp, err = (st.zero).zeroHealth(context.Background(), false); err != nil {
+			x.SetStatus(w, x.Error, err.Error())
+			return
+		}       
+		if resp == nil {
+			x.SetStatus(w, x.ErrorNoData, "No health information available.")
+			return
+		} 
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write(resp.Json)
+
+		return
+	}
 
 	w.WriteHeader(http.StatusOK)
 	_, _ = w.Write([]byte("OK"))
--- a/dgraph/cmd/zero/http.go
+++ b/dgraph/cmd/zero/http.go
@@ -25,7 +25,6 @@ import (
 	"time"
 	"encoding/json"
 
-	"github.com/pkg/errors"
 	"github.com/gogo/protobuf/jsonpb"
 	"github.com/golang/glog"
 
@@ -235,15 +234,12 @@ func (st *state) getState(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-func (s *Server) zeroHealth(ctx context.Context, all bool) (*api.Response, error) {
+func (s *Server) zeroHealth(ctx context.Context) (*api.Response, error) {
 	if ctx.Err() != nil {
 		return nil, ctx.Err()
 	}
 
-	var healthAll []pb.HealthInfo
-
-	// Append self.
-	healthAll = append(healthAll, pb.HealthInfo{
+	healthAll := []pb.HealthInfo{pb.HealthInfo{
 		Instance:    "zero",
 		Address:     x.WorkerConfig.MyAddr,
 		Status:      "healthy",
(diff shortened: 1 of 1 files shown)


dgraph/cmd/zero/http.go
```diff
@@ -231,9 +235,60 @@ func (st *state) getState(w http.ResponseWriter, r *http.Request) {
...
 }
 
+func (s *Server) zeroHealth(ctx context.Context, all bool) (*api.Response, error) {
+	if ctx.Err() != nil {
+		return nil, ctx.Err()
+	}
+
+	var healthAll []pb.HealthInfo
+
+	// Append self.
```
mangalaman93, line 245: don't need this comment.

dgraph/cmd/zero/http.go
```diff
@@ -231,9 +235,60 @@ func (st *state) getState(w http.ResponseWriter, r *http.Request) {
...
+		Instance:    "zero",
+		Address:     x.WorkerConfig.MyAddr,
+		Status:      "healthy",
+		Group:       strconv.Itoa(int(worker.GroupId())),
+		Version:     x.Version(),
+		Uptime:      int64(time.Since(x.WorkerConfig.StartTime) / time.Second),
+		LastEcho:    time.Now().Unix(),
+	})
+
+	var err error
```
mangalaman93, line 256: declaration is probably not needed. And can move out the json.Marshal line out of the if

dgraph/cmd/zero/http.go
```diff
@@ -231,9 +235,60 @@ func (st *state) getState(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (s *Server) zeroHealth(ctx context.Context, all bool) (*api.Response, error) {
```
mangalaman93, line 238: do we need the all arg here?

dgraph/cmd/zero/http.go
```diff
@@ -231,9 +235,60 @@ func (st *state) getState(w http.ResponseWriter, r *http.Request) {
...
+	var ok = false
+	for _, s := range r.Header["Content-Type"] {
+		if s == "application/json" {
+			ok = true
+			break
+		}
+	}
+	if ok {
+		var resp *api.Response
+		if resp, err = (st.zero).zeroHealth(context.Background(), false); err != nil {
```
mangalaman93, line 277: we should get the context from the request.

mangalaman93 (changes requested):

mangalaman93 (commented):
We should also fix the commit message to fit the limit and add more details as part of the description after a new line. And is it going to be a breaking change? If yes, we should probably specify everywhere that it is a breaking change.

dgraph/cmd/zero/http.go
```diff
@@ -231,9 +235,60 @@ func (st *state) getState(w http.ResponseWriter, r *http.Request) {
...
 
+func (s *Server) zeroHealth(ctx context.Context, all bool) (*api.Response, error) {
+	if ctx.Err() != nil {
+		return nil, ctx.Err()
+	}
+
+	var healthAll []pb.HealthInfo
+
+	// Append self.
+	healthAll = append(healthAll, pb.HealthInfo{
```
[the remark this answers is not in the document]
jbhamra1: bash jassi Mon Mar 20 15:13:28 sprint1_issue_5517 dgraphtest > dock ps
CONTAINER ID   IMAGE                 COMMAND                  CREATED      STATUS              PORTS                                                                  NAMES
9218571d27df   dgraph/dgraph:local   "/gobin/dgraph alpha…"   6 days ago   Up About a minute   0.0.0.0:61990->8080/tcp, 0.0.0.0:61991->9080/tcp                       dgraph-alpha6-1
0d7b0aa67cd1   dgraph/dgraph:local   "/gobin/dgraph zero …"   6 days ago   Up About a minute   8080/tcp, 9080/tcp, 0.0.0.0:61982->5080/tcp, 0.0.0.0:61983->6080/tcp   dgraph-zero3-1
137866cb4a23   dgraph/dgraph:local   "/gobin/dgraph zero …"   6 days ago   Up About a minute   8080/tcp, 9080/tcp, 0.0.0.0:61978->5080/tcp, 0.0.0.0:61979->6080/tcp   dgraph-zero2-1
46000a547cad   dgraph/dgraph:local   "/gobin/dgraph alpha…"   6 days ago   Up About a minute   0.0.0.0:61988->8080/tcp, 0.0.0.0:61989->9080/tcp                       dgraph-alpha5-1
98cbe9b2542d   dgraph/dgraph:local   "/gobin/dgraph alpha…"   6 days ago   Up About a minute   0.0.0.0:61986->8080/tcp, 0.0.0.0:61987->9080/tcp                       dgraph-alpha4-1
1e46030bc95a   dgraph/dgraph:local   "/gobin/dgraph zero …"   6 days ago   Up About a minute   8080/tcp, 9080/tcp, 0.0.0.0:61973->5080/tcp, 0.0.0.0:61974->6080/tcp   dgraph-zero1-1
83114283343e   dgraph/dgraph:local   "/gobin/dgraph alpha…"   6 days ago   Up About a minute   0.0.0.0:61984->8080/tcp, 0.0.0.0:61985->9080/tcp                       dgraph-alpha3-1
76cb634b311d   dgraph/dgraph:local   "/gobin/dgraph alpha…"   6 days ago   Up About a minute   0.0.0.0:61980->8080/tcp, 0.0.0.0:61981->9080/tcp                       dgraph-alpha2-1
90f85ec636fd   dgraph/dgraph:local   "/gobin/dgraph alpha…"   6 days ago   Up About a minute   0.0.0.0:61976->8080/tcp, 0.0.0.0:61975->9080/tcp                       dgraph-alpha1-1
e54bbc98b883   minio/minio:latest    "/usr/bin/docker-ent…"   6 days ago   Up About a minute   9000/tcp, 0.0.0.0:61972->9001/tcp                                      dgraph-minio-1
bash jassi Mon Mar 20 15:13:31 sprint1_issue_5517 dgraphtest > 
bash jassi Mon Mar 20 15:13:34 sprint1_issue_5517 dgraphtest > 
bash jassi Mon Mar 20 15:13:34 sprint1_issue_5517 dgraphtest > echo ""; curl -H "Content-Type: application/json" http://localhost:61983/health; echo -e "\n"

[{"instance":"zero","address":"zero3:5080","status":"healthy","group":"0","version":"v23.0.0-beta1-28-g05b90a580","uptime":9223372036,"lastEcho":1679305435}]

bash jassi Mon Mar 20 15:13:55 sprint1_issue_5517 dgraphtest > 
bash jassi Mon Mar 20 15:14:04 sprint1_issue_5517 dgraphtest > echo ""; curl http://localhost:61983/health; echo -e "\n"

OK

bash jassi Mon Mar 20 15:14:22 sprint1_issue_5517 dgraphtest >

dgraph/cmd/zero/http.go
```diff
@@ -231,9 +234,49 @@ func (st *state) getState(w http.ResponseWriter, r *http.Request) {
...
+func (s *Server) zeroHealth(ctx context.Context) (*api.Response, error) {
+	if ctx.Err() != nil {
+		return nil, ctx.Err()
+	}
+
+	healthAll := []pb.HealthInfo{pb.HealthInfo{
+		Instance:    "zero",
+		Address:     x.WorkerConfig.MyAddr,
+		Status:      "healthy",
+		Group:       strconv.Itoa(int(worker.GroupId())),
```
mangalaman93, line 246: Group is probably an Alpha concept. We can remove it.

dgraph/cmd/zero/http.go
```diff
@@ -231,9 +234,49 @@ func (st *state) getState(w http.ResponseWriter, r *http.Request) {
...
+		Version:     x.Version(),
+		Uptime:      int64(time.Since(x.WorkerConfig.StartTime) / time.Second),
+		LastEcho:    time.Now().Unix(),
+	}}
+
+	jsonOut, _ := json.Marshal(healthAll)
```
mangalaman93, line 252: we should check the error here, than ignoring it.

mangalaman93 (changes requested):

dgraph/cmd/zero/http.go
jbhamra1: I misunderstood your earlier comment. This is corrected in the recent upload.

dgraph/cmd/zero/http.go
```diff
@@ -231,9 +234,44 @@ func (st *state) getState(w http.ResponseWriter, r *http.Request) {
...
+
+	jsonOut, err := json.Marshal(healthAll)
+	if err != nil {
+		return nil, errors.Errorf("Unable to Marshal. Err %v", err)
+	}
+	return &api.Response{Json: jsonOut}, nil
+}
+
 func (st *state) pingResponse(w http.ResponseWriter, r *http.Request) {
+	var err error
```
mangalaman93, line 259: we shouldn't declare variables unless we have to

dgraph/cmd/zero/http.go
```diff
@@ -231,9 +234,44 @@ func (st *state) getState(w http.ResponseWriter, r *http.Request) {
...
 	x.AddCorsHeaders(w)
 
-	w.WriteHeader(http.StatusOK)
-	_, _ = w.Write([]byte("OK"))
+	switch r.Header.Get("Accept") {
+		case "application/json":
+			var resp *api.Response
```
mangalaman93, line 264: same, no need for declaration

dgraph/cmd/zero/http.go
```diff
@@ -231,9 +234,44 @@ func (st *state) getState(w http.ResponseWriter, r *http.Request) {
...
+			if resp, err = (st.zero).zeroHealth(r.Context()); err != nil {
+				x.SetStatus(w, x.Error, err.Error())
+				return
+			}       
+			w.Header().Set("Content-Type", "application/json; charset=utf-8")
+			w.WriteHeader(http.StatusOK)
+			_, _ = w.Write(resp.Json)
```
mangalaman93, line 271: we shouldn't ignore the error here.

dgraph/cmd/zero/http.go
```diff
@@ -231,9 +234,44 @@ func (st *state) getState(w http.ResponseWriter, r *http.Request) {
...
 
+func (s *Server) zeroHealth(ctx context.Context) (*api.Response, error) {
+	if ctx.Err() != nil {
+		return nil, ctx.Err()
+	}
+
+	healthAll := []pb.HealthInfo{pb.HealthInfo{
+		Instance:    "zero",
+		Address:     x.WorkerConfig.MyAddr,
+		Status:      "healthy",
```
mangalaman93, line 245: I am thinking whether there is a situation when zero can be unhealthy?

mangalaman93 (changes requested):

dgraph/cmd/zero/http.go
```diff
@@ -231,9 +234,44 @@ func (st *state) getState(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (s *Server) zeroHealth(ctx context.Context) (*api.Response, error) {
+	if ctx.Err() != nil {
+		return nil, ctx.Err()
+	}
+
+	healthAll := []pb.HealthInfo{pb.HealthInfo{
```
mangalaman93, line 242: why is this a list?

mangalaman93 (commented):
we should fix the title and description of the PR. That is what goes into the final commit message. The title is longer than Git's limit.

dgraph/cmd/zero/http.go
jbhamra1: Actually, resp is used at line #271, if I restrict its declaration to 'if' block, the compiler would throw error, and since zeroHealth()'s return is multi-valued, I think, we need to have both, resp and err as they are. But, this is what I can do- move statement (var err error) at line# 259 to line# 264 just to reduce it's scope.

dgraph/cmd/zero/http.go
jbhamra1: No need for this to be a list, corrected.

dgraph/cmd/zero/http.go
mangalaman93: then you can just move the statement outside the if block

dgraph/cmd/zero/http.go
mangalaman93: yeah, later is fine.

dgraph/cmd/zero/http.go
```diff
@@ -256,22 +256,29 @@ func (s *Server) zeroHealth(ctx context.Context) (*api.Response, error) {
...
-			if resp, err = (st.zero).zeroHealth(r.Context()); err != nil {
+			resp, err := (st.zero).zeroHealth(r.Context())
+			if err != nil {
 				x.SetStatus(w, x.Error, err.Error())
 				return
 			}       
 			w.Header().Set("Content-Type", "application/json; charset=utf-8")
 			w.WriteHeader(http.StatusOK)
-			_, _ = w.Write(resp.Json)
+			_, err = w.Write(resp.Json)
```
mangalaman93, line 270: I think we should just log (warning) the error. Good chance, if write has failed, something is really wrong with the request and SetStatus may not work.

dgraph/cmd/zero/http.go
```diff
@@ -256,22 +256,29 @@ func (s *Server) zeroHealth(ctx context.Context) (*api.Response, error) {
...
+			if err != nil {
+				x.SetStatus(w, x.Error, err.Error())
+				return
+			}       
 		default:
 			w.Header().Set("Content-Type", "text/plain; charset=utf-8")
 			w.WriteHeader(http.StatusOK)
-			_, _ = w.Write([]byte("OK"))
+			_, err := w.Write([]byte("OK"))
```
mangalaman93, line 278: same

dgraph/cmd/zero/http.go
```diff
@@ -231,9 +234,51 @@ func (st *state) getState(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (s *Server) zeroHealth(ctx context.Context) (*api.Response, error) {
```
mangalaman93, line 237: I was looking at how alpha's health endpoint implemented and now I understood why there was an `all` arg and we were returning a list. We could do something similar here too, it'd be nice to have it behave similar to how alpha's endpoint is returning. It would also be useful to know what this node thinks of the health of other zero nodes. We will have to look into if we have access to health of other zero nodes.

mangalaman93 (changes requested):

dgraph/cmd/zero/http.go
```diff
@@ -231,9 +234,51 @@ func (st *state) getState(w http.ResponseWriter, r *http.Request) {
...
+	return &api.Response{Json: jsonOut}, nil
+}
+
 func (st *state) pingResponse(w http.ResponseWriter, r *http.Request) {
 	x.AddCorsHeaders(w)
 
-	w.WriteHeader(http.StatusOK)
-	_, _ = w.Write([]byte("OK"))
+	switch r.Header.Get("Accept") {
+		case "application/json":
```
mangalaman93, line 262: By looking at the reference here, it seems like Accept header could be set to more than one values including */*. We should handle that case too. On the other hand, I am thinking this header may get removed in the CORS case. Should we just consider using a query parameter instead?

jbhamra1: I was going through the text on removal of Accept header in case of CORS and have hit this- https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header

So, if 'Accept' is safe then why it can be removed in case of CORS. I mean, I am not an expert in this area and so, is thinking if someone could explain even why this would happen and under what situation. I tried searching the web on this and could not find any relevant writings on this (removal of Accept for CORS).

jbhamra1: "By looking at the reference here, it seems like Accept header could be set to more than one values including */*. We should handle that case too."
>>> Right, so, we handle switch case "application/json" to send json and this corresponds to the '<MIME_type>/<MIME_subtype> and rest all other cases under the 'default' to send "OK", which makes it backward compatible. The 'default' would cover "<MIME_type>/*" and "*/*". 
Other than these 2 switch cases which other cases we need to cover specifically. Aren't these 2 cases sufficient for now?

dgraph/cmd/zero/http.go
jbhamra1: I have uploaded code changes for these comments.

PR #8767, opened 2023-03-20
chore(test): skip a test when dgraph's version does not support the feature

shivaji-dgraph:
When we do upgrade testing, it is possible that some of the tests that were added later fails for an older version of Dgraph. We would like to skip those tests when dgraph's version does not support that feature.

### DIFF
--- a/dgraphtest/cluster.go
+++ b/dgraphtest/cluster.go
@@ -17,10 +17,13 @@
 package dgraphtest
 
 import (
+	"testing"
+
 	"github.com/dgraph-io/dgo/v210"
 )
 
 type Cluster interface {
 	AssignUids(num uint64) error
 	Client() (*dgo.Dgraph, error)
+	SkipTest(t *testing.T, commit string) error
 }
--- a/dgraphtest/cluster.go
+++ b/dgraphtest/cluster.go
@@ -17,13 +17,11 @@
 package dgraphtest
 
 import (
-	"testing"
-
 	"github.com/dgraph-io/dgo/v210"
 )
 
 type Cluster interface {
 	AssignUids(num uint64) error
 	Client() (*dgo.Dgraph, error)
-	SkipTest(t *testing.T, commit string) error
+	GetVersion() string
 }
--- a/dgraphtest/local_cluster.go
+++ b/dgraphtest/local_cluster.go
@@ -24,12 +24,15 @@ import (
 	"net/http"
 	"os"
 	"strings"
+	"testing"
 	"time"
 
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/network"
 	docker "github.com/docker/docker/client"
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
 	"google.golang.org/grpc"
@@ -65,6 +68,7 @@ type LocalCluster struct {
 func NewLocalCluster(conf ClusterConfig) (*LocalCluster, error) {
 	c := &LocalCluster{conf: conf}
 	if err := c.init(); err != nil {
+		glog.Infof(err.Error())
 		c.Cleanup()
 		return nil, err
 	}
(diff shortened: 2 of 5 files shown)


dgraphtest/local_cluster.go
```diff
@@ -433,3 +437,44 @@ func (c *LocalCluster) AssignUids(num uint64) error {
 	}
 	return nil
 }
+
+// CompareCommits compare given commit with current cluster's commit and return
```
mangalaman93, line 441: This statement and the function name could be simpler. Not sure which commit is the "cluster's commit", the names have to be clear.

CompareCommits -> IsParent(ancestor, descendant string) (bool, error)


// IsParent checks whether `ancestor` is the ancestor commit of the given descendant commit.

dgraphtest/local_cluster.go
```diff
@@ -433,3 +437,44 @@ func (c *LocalCluster) AssignUids(num uint64) error {
...
+// true if cluster's commit is ancestor of given commit else return false
+func CompareCommits(shaToBeCompare, baseSha string) (bool, error) {
+	repo, err := git.PlainOpen(repoDir)
+	if err != nil {
+		return false, errors.Wrap(err, "error while opening git repo")
+	}
+	hash, err := repo.ResolveRevision(plumbing.Revision(shaToBeCompare))
+	if err != nil {
+		return false, errors.Wrap(err, "error while getting commit hash")
```
mangalaman93, line 450: mention which hash, ancestor or descendant

dgraphtest/local_cluster.go
```diff
@@ -433,3 +437,44 @@ func (c *LocalCluster) AssignUids(num uint64) error {
...
+	}
+	commitToBeCompare, err := repo.CommitObject(*hash)
+	if err != nil {
+		return false, errors.Wrap(err, fmt.Sprintf("error while getting commit object of hash [%v]", hash))
```
mangalaman93, line 454: use errors.Wrapf instead. Use small but clear error such as "error finding commit object [%v]"

avoid terms such as "while", "of has" (in this case). It doesn't be hash if you think about it, it could be a tag too

dgraphtest/local_cluster.go
```diff
@@ -433,3 +437,44 @@ func (c *LocalCluster) AssignUids(num uint64) error {
 	}
 	return nil
 }
+
+// CompareCommits compare given commit with current cluster's commit and return
+// true if cluster's commit is ancestor of given commit else return false
+func CompareCommits(shaToBeCompare, baseSha string) (bool, error) {
```
mangalaman93, line 443: do we need to export this function?

dgraphtest/local_cluster.go
```diff
@@ -433,3 +437,44 @@ func (c *LocalCluster) AssignUids(num uint64) error {
...
+		return false, errors.Wrap(err, fmt.Sprintf("error while getting commit object of hash [%v]", hash))
+	}
+	isParentCommit, err := baseCommit.IsAncestor(commitToBeCompare)
+	if err != nil {
+		return false, errors.Wrap(err, "error while comparing commits")
+	}
+	return isParentCommit, nil
+}
+
+func (c *LocalCluster) SkipTest(t *testing.T, commit string) error {
```
mangalaman93, line 471: we probably don't need to return an error. This would be a one liner at the caller then.

mangalaman93: we would fail the test if errors occurs

dgraphtest/local_cluster.go
```diff
@@ -433,3 +437,44 @@ func (c *LocalCluster) AssignUids(num uint64) error {
...
+	isParentCommit, err := CompareCommits(commit, c.conf.version)
+	if err != nil {
+		return err
+	}
+	if isParentCommit {
```
mangalaman93, line 476: can you provide the reason with enough information, for example, say that "test is valid for {hash/version} > {version}".

mangalaman93 (changes requested):

dgraphtest/cluster.go
```diff
@@ -17,10 +17,13 @@
...
 import (
+	"testing"
+
 	"github.com/dgraph-io/dgo/v210"
 )
 
 type Cluster interface {
 	AssignUids(num uint64) error
 	Client() (*dgo.Dgraph, error)
+	SkipTest(t *testing.T, commit string) error
```
mangalaman93, line 28: This function can be common across all clusters. It shouldn't be part of the interface. Instead a funtion `GetVersion` should be that the SkipTest can use

dgraphtest/local_cluster.go
```diff
@@ -433,3 +436,44 @@ func (c *LocalCluster) AssignUids(num uint64) error {
...
+		return false, err
+	}
+	descendantCommit, err := repo.CommitObject(*descendantHash)
+	if err != nil {
+		return false, errors.Wrapf(err, "error finding commit object [%v]", descendant)
+	}
+	isParentCommit, err := descendantCommit.IsAncestor(ancestorCommit)
+	if err != nil {
+		return false, errors.Wrapf(err, "unable to compare commit [%v] to commit [%v]",
+			descendantCommit.Hash, ancestorCommit.Hash)
```
mangalaman93, line 465: We should include what was passed in the function arg

query/query0_test.go
```diff
@@ -3515,6 +3515,7 @@ func TestMatchingWithPagination(t *testing.T) {
 }
 
 func TestInvalidRegex(t *testing.T) {
+	dc.SkipTest(t, "e0cc0450b88593b7496c0947aea016fc6457cb61")
```
mangalaman93, line 3518: SkipTest ~> ShouldSkipTest

mangalaman93 (changes requested):

dgraphtest/local_cluster.go
mangalaman93: Not the hashes

mangalaman93 (commented):
Also improve the title of the PR and description if needed. This will go into the final commit message

dgraphtest/cluster.go
shivaji-dgraph: Could you please elaborate more on this

mangalaman93: Think about this, if I have a cloud cluster and also a cluster running on AWS and cluster running on my own machine. Now, in all 3 cases, I would like to skip the test. Would I be implementing the whole SkipTest function again for 3 different clusters?

dgraphtest/local_cluster.go
```diff
@@ -433,3 +436,48 @@ func (c *LocalCluster) AssignUids(num uint64) error {
...
+		return false, errors.Wrapf(err, "unable to compare commit [%v] to commit [%v]",
+			descendant, ancestor)
+	}
+	return isParentCommit, nil
+}
+
+func ShouldSkipTest(t *testing.T, testCommit, clusterVersion string) error {
+	isParentCommit, err := isParent(testCommit, clusterVersion)
+	if err != nil {
+		t.Fatalf(err.Error())
```
mangalaman93, line 473: t.Fatalf(err) is enough

PR #8768, opened 2023-03-23
add upgrade tests in query package for all version combinations

mangalaman93:
This PR allows us to run upgrade tests for various combinations of before and after dgraph versions in the query package.

### DIFF
--- a/query/upgrade_test.go
+++ b/query/upgrade_test.go
@@ -20,6 +20,7 @@ package query
 
 import (
 	"context"
+	"log"
 	"testing"
 	"time"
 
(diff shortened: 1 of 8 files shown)


query/upgrade_test.go
```diff
@@ -27,36 +28,70 @@ import (
...
+		{"v21.03.0-92-g0c9f60156", "v23.0.0-beta1"},
+		// {"v21.03.0-96-g65fff46c4-slash", "v23.0.0-beta1"},
+		// {"v21.03.0-98-g19f71a78a-slash", "v23.0.0-beta1"},
+		// {"v21.03.0-99-g4a03c144a-slash", "v23.0.0-beta1"},
+		// {"v21.03.1", "v23.0.0-beta1"},
+		// {"v21.03.2", "v23.0.0-beta1"},
+		// {"v21.12.0", "v23.0.0-beta1"},
+		// {"v22.0.0", "v23.0.0-beta1"},
+		{"v22.0.1", "v23.0.0-beta1"},
+		// {"v22.0.2", "v23.0.0-beta1"},
```
joshua-goldstein, line 78: we might want to paralellize this in the future, perhaps by making this configurable via a command line argument to `go test`

query/upgrade_test.go
```diff
@@ -27,36 +28,70 @@ import (
...
+	}
+	for _, cv := range comboVersions {
+		log.Printf("running: backup in [%v], restore in [%v]", cv.before, cv.after)
```
joshua-goldstein, line 81: can add if/else logic here based on arguments provided

query/upgrade_test.go
harshil-goel: We can just run the official go test using m.Run("name"). If we use proper names, go test will run a specific test only.

query/upgrade_test.go
```diff
@@ -27,36 +28,70 @@ import (
...
+		runTest(cv.before, cv.after)
+	}
 }
+
+// userClient, err := testutil.DgraphClient("0.0.0.0:33196")
```
harshil-goel, line 86: Does this need to be removed?

harshil-goel (approved):
This is really great to see. The test could be written in minutes after the new framework. Lets write more tests now!!

query/upgrade_test.go
mangalaman93: no, that works when we are inside a Test not in TestMain

query/upgrade_test.go
mangalaman93: wip, I will fix it.

harshil-goel (approved):

anurags92 (approved):

PR #8769, opened 2023-03-25
fix(Query): Do an error check before bubbling up nil error

anurags92:
## Problem
Without the check, the code erraneously bubbles up nil error even when there's been an error in one of the child subgraph. This later causes a panic because the downstream code expects no error.

Reproduce this:

DQL Query:
```
{
    var(func: eq(name, "Alicey")){
        friend(orderdesc: id) {
            f as count(uid)
        }
    }
    q(func: uid(f)){
        name
    } 
}
```

Fixes: https://github.com/dgraph-io/projects-internal/issues/89

## Solution
Add an error check.

### DIFF
--- a/query/query.go
+++ b/query/query.go
@@ -2843,10 +2843,6 @@ func (req *Request) ProcessQuery(ctx context.Context) (err error) {
 				continue
 			}
 
-			fmt.Println("##################################################")
-			fmt.Printf("Executing: %d\n", idx)
-			fmt.Println("##################################################")
-
 			switch {
 			case sg.Params.Alias == "shortest":
 				// We allow only one shortest path block per query.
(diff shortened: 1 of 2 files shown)


mangalaman93 (approved):

skrdgraph (approved):
Looks good 👍

query/query.go
```diff
@@ -2363,7 +2363,7 @@ func ProcessGraph(ctx context.Context, sg, parent *SubGraph, rch chan error) {
 		}
 	}
 
-	if sg.DestUIDs == nil || len(sg.DestUIDs.Uids) == 0 {
+	if (sg.DestUIDs == nil || len(sg.DestUIDs.Uids) == 0) && childErr == nil {
```
harshil-goel, line 2366: Add a comment explaining the isseu?

PR #8780, opened 2023-04-03
make dgraphtest multitenancy aware

mangalaman93:


dgraphtest/cluster.go
```diff
@@ -263,11 +326,84 @@ loop:
...
+	}
+	if _, err := hc.RunAdminQuery(params); err != nil {
+		return err
+	}
+	return nil
+}
+
+// == GRPC ==
+// SetupSchema sets up DQL schema
+func SetupSchema(client *dgo.Dgraph, dbSchema string) error {
```
harshil-goel, line 377: GRPC Client just like HTTP client?

harshil-goel (approved):

harshil-goel (dismissed):

dgraphtest/cluster.go
mangalaman93: Done

anurags92 (changes requested):
Add more details in the PR description. 

Once we have both clients (http and grpc) in the Client object, would there be a recommendation on which one to use? Typically we have used gRPC for everything accept graphql.

mangalaman93:
> 

I think both clients have different purposes. Grpc client is used to make grpc requests using dql whereas HTTP client uses graphql over HTTP. There could be some overlap on the operations that one could do using both clients but that is on the discretion of the user of the library.

anurags92 (approved):

harshil-goel (approved):

PR #8788, opened 2023-04-05
chore(CI): Add spell checking with Vale.sh

MichelDiz:
## Problem

The current project repository does not have spell checking enabled, which can lead to inconsistencies and errors in the written content.

## Solution

Added spell checking functionality using the Vale linting tool, which can check for spelling and grammar errors in Markdown files and code comments. The Vale configuration file was added to the repository to allow for customized spell checking rules.

Changes made to the project files to enable spell checking.


.github/workflows/ci-spell-checking.yml
```diff
@@ -0,0 +1,44 @@
...
+    - synchronize
+    - ready_for_review
+    branches: 
+    - main
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
skrdgraph, line 20: why is this needed for spell check?

skrdgraph:
Screen Shot 2023-04-05 at 16 01 55

Adding this ^ to show that files are not audited for spell checks.

.github/workflows/ci-spell-checking.yml
```diff
@@ -0,0 +1,44 @@
...
+    types:
+    - opened
+    - reopened
+    - synchronize
+    - ready_for_review
+    branches: 
+    - main
+
+jobs:
+  lint:
```
skrdgraph, line 17: this should be renamed to something meaningful ... and should not be `lint`.

.github/workflows/ci-spell-checking.yml
MichelDiz: I was using for other thing. I will remove it.

.github/workflows/ci-spell-checking.yml
MichelDiz: Do you have any suggestion?

.github/workflows/ci-spell-checking.yml
```diff
@@ -0,0 +1,44 @@
...
+        id: changed-files-specific
+        uses: tj-actions/changed-files@v35
+        with:
+          files: ./
+          files_ignore: './.vscode/'
+      - name: Install Vale
+        if: steps.changed-files-specific.outputs.any_changed == 'true'
+        run: |
+          sudo apt-get update
+          sudo apt-get install curl
```
mangalaman93, line 33: we should avoid sudo commands. And it seems reasonable to assume curl is present.

.github/workflows/ci-spell-checking.yml
```diff
@@ -0,0 +1,44 @@
...
+          curl -Ls https://api.github.com/repos/errata-ai/vale/releases/latest \
+            | grep "browser_download_url.*vale_[0-9.]*_Linux_64-bit.tar.gz" \
+            | cut -d : -f 2,3 \
+            | tr -d \" \
+            | xargs curl -Ls \
+            | tar xz
+            sudo mv ./vale /usr/bin/vale
```
mangalaman93, line 40: we don't need to do this. We can run the command from a local folder.

mangalaman93 (changes requested):
I think it'd be useful to have a spell checker in the repo. It is not uncommon to make spelling mistakes in the comments etc. I am hoping and it looks like Michel has done lot of the work to avoid false positives.

.github/workflows/ci-spell-checking.yml
MichelDiz: I did this before, I ran into some issues about setting up Vale. So doing it that way solved the problems I had at the time. I don't remember anymore, but I remember that I did it to solve a specific problem in the Vale.

PR #8791, opened 2023-04-06
chore(CI): add paths to ignore

MichelDiz:
## Problem

Previously, our GitHub Actions CI workflows were processing unnecessary files, which was consuming extra compute resources and increasing our CI costs. The included files are such as documentation files, image files, and configuration files. This slows down the CI process and increases resource usage / costs.

## Solution

To reduce CI costs and improve efficiency, I have added a set of paths to be ignored in our CI workflows using paths-ignore.

By ignoring these files and directories, our CI workflows will only process the necessary files, which should reduce the overall CI execution time and cost.


mangalaman93:
@MichelDiz how much does it save for us on CI?

MichelDiz:
@mangalaman93 Kind of hard to say. But I was talking with @skrdgraph about this. It is a waster of resources doing CI with all those tests at the same time all the time someone adds a markdown, gif, png or unrelated code. Makes no sense run CI in such simple cases right? We also run the test in CRON. No need to run when someone changes a md file.

mangalaman93:
Very few of the PRs in this repo are non code related. only 2 out of last 35 PRs merged are actually non-code related. If cost is a problem, we can figure something out there. I don't think this is really needed.

MichelDiz:
@mangalaman93 this is needed, not based on the past only. But in the future. We will have RFC(documentation). The compose path is code, but not related with any test we have. Dgraphtest can have it's own CI. And any image and other stuff we don't want to run it.

Also, this is just the beginning. We gonna fine tune it. I have plan to make the CI as "fail fast" instead of waiting the whole test to run. Just hold-on.

.github/workflows/ci-aqua-security-trivy-tests.yml
```diff
@@ -1,6 +1,26 @@
 name: ci-aqua-security-trivy-tests
 on:
   push:
+    paths-ignore:
+      - '.vscode/CODEOWNERS'
+      - '.vscode/*.json'
+      - 'compose/**'
+      - 'contrib/**'
```
joshua-goldstein, line 8: our canonical Dockerfile is located in `contrib`, this should be removed

.github/workflows/ci-aqua-security-trivy-tests.yml
```diff
@@ -1,6 +1,26 @@
...
+      - 'dgraphtest/**'
```
joshua-goldstein, line 9: this should be removed - if we make a change to our test suite, we want to see if the tests pass or fail

.github/workflows/ci-aqua-security-trivy-tests.yml
```diff
@@ -1,6 +1,26 @@
...
+      - 'licenses/**'
+      - 'paper/**'
+      - 'RFC/**'
+      - 'static/**'
+      - 't/**'
```
joshua-goldstein, line 14: this should be removed - again if we make a change to our test executable, we would like to see tests run against this change

.github/workflows/ci-aqua-security-trivy-tests.yml
joshua-goldstein, line 9: this should be removed - if we make a change to our test suite, we want to see if the tests pass or fail

.github/workflows/ci-aqua-security-trivy-tests.yml
joshua-goldstein, line 8: our canonical Dockerfile is located in `contrib`, this should be removed

.github/workflows/ci-aqua-security-trivy-tests.yml
```diff
@@ -1,6 +1,26 @@
...
+      - 'telemetry/**'
```
joshua-goldstein, line 15: why should `telemetry` not be in our CI?

.github/workflows/ci-aqua-security-trivy-tests.yml
```diff
@@ -1,6 +1,26 @@
...
+      - 'wiki/**'
+      - '**.dockerignore'
+      - '**.gitignore'
+      - '**/**.md'
+      - '**/**.png'
+      - '**/**.sh'
```
joshua-goldstein, line 21: this is overly broad and should be removed

joshua-goldstein (changes requested):
Echoing @mangalaman93 concerns, why is this necessary right now?  We are not making frequent changes to these files, and when we do so we can always keep the PR as a draft and then only open when ready to merge.  This means that CI runs at most two times when such a change is proposed and merged.  That is not a huge overhead.

My comments on the PR are the more serious concerns I have.

.github/workflows/ci-aqua-security-trivy-tests.yml
MichelDiz: Cuz there's no test for it at all.

.github/workflows/ci-aqua-security-trivy-tests.yml
MichelDiz: This what? When you don't point out exactly what you mean. I can't tell if it is about .sh or all lines from 16 to 21.
If it is about .sh, yeah. I put it there to discuss. As I see we are more and more going away from .sh. And all .sh there is there are deprecated or just examples.

.github/workflows/ci-aqua-security-trivy-tests.yml
MichelDiz: The description of Contrib says:
>The contrib directory contains scripts, images, and other helpful things which are not part of the core dgraph distribution. Please note that they could be out of date, since they do not receive the same attention as the rest of the repository.

The objective of the tests are related to the core code or helpful things we use time to time?

The thing is. If your PR change some code AND change something in contrib. The CI will be triggered.

Now, let's say I wanna update something in /contrib/systemd/ or  /contrib/jepsen. Are the CI covering those? So why run the CI at all?

If you need to update some dockerfile or docker compose. We will have an option to trigger manually if needed. But it is something we can think more about. Maybe have a granular control here instead ignoring the whole contrib file.

.github/workflows/ci-aqua-security-trivy-tests.yml
MichelDiz: Okay.

.github/workflows/ci-aqua-security-trivy-tests.yml
MichelDiz: Okay. BTW, are we moving to /dgraphtest ? at some point this /t will be removed?

MichelDiz:
> Echoing @mangalaman93 concerns, why is this necessary right now? We are not making frequent changes to these files, and when we do so we can always keep the PR as a draft and then only open when ready to merge. This means that CI runs at most two times when such a change is proposed and merged. That is not a huge overhead.
> 
> My comments on the PR are the more serious concerns I have.

>why is this necessary right now?

Costs is always necessary thing to do. We are not philanthropy who receives infinite money from the government.

>We are not making frequent changes to these files,

Which means? It's not that you're not seeing it now that it's not going to happen.

>and when we do so we can always keep the PR as a draft and then only open when ready to merge.

The point is. Not only us open PRs. We can guide the team with "good practice". But we don't say anything to community contributors. This PR is a way to mitigate that. Plus, not every new engineer is going to know everything they need to know beforehand. That's why we automate things. Cuz we never know what the others will do.

> That is not a huge overhead.

That can be true now. But with time every system tends to increase its complexity. Imagine running CI to handle just a PNG? Imagine in the future having a complex CI sequence just to change a Markdown? The CI have a cost. It's like the cost of an appliance doing unnecessary work. In the month you do not see the costs of this. But if you add every year or a decade. You can buy a private jet.

Another problem we solve with this is the following. Imagine the situation, someone introduces a bug and accidentally goes through the CI. Only then in "my" turn I'm just changing a text in a README. The CI blocks it and I'll have all the work to find out what's going on and call someone. And the person fixing the readme might not have the right skill. But usually we see a PR with a failing CI and we don't help to solve it. "Let the dev take care of it alone".

So, if we have a Markdown/PNG/JPG/others change. We should fore SURE skip the whole CI. If the CI will do any good for that PR at all.

.github/workflows/ci-aqua-security-trivy-tests.yml
joshua-goldstein: Sorry if that wasn't clear; I was referring to the last line `**/**.sh`.  There are still a few places where we are using shell scripts (e.g. protobuf generation).  Also in general it is overly broad not to run CI against shell scripts.

.github/workflows/ci-aqua-security-trivy-tests.yml
joshua-goldstein: There is no unit test in the telemetry package.  It is however imported in two [different](https://github.com/dgraph-io/dgraph/blob/f085c32bae6f7e27475503f64684164f9e3bc085/dgraph/cmd/zero/zero.go#L36) [places](https://github.com/dgraph-io/dgraph/blob/f085c32bae6f7e27475503f64684164f9e3bc085/edgraph/server.go#L56).

.github/workflows/ci-aqua-security-trivy-tests.yml
joshua-goldstein: We are a long way off from removing `t.go`.  The idea behind introducing `dgraphtest` is to make writing new integration tests easier for developers.  But our existing test suite will remain on t.go, at least for some time.

.github/workflows/ci-aqua-security-trivy-tests.yml
MichelDiz: Yeah. It is imported. But if there's no test for it. I see no reason to run CI. The other packages have their own testing packages. But none of them take telemetry into consideration. It is not a tied relation. But yeah, I think I will remove it. As I have plans to change the telemetry and add tests for that.

This package was added in hurry back them. We were discussing how to get some important data from how ppl use Dgraph. But this was done in a hurry.

.github/workflows/ci-aqua-security-trivy-tests.yml
MichelDiz: Nice. I was thinking that would be just an append. Something to use sporadically and manually.

joshua-goldstein:
> > Echoing @mangalaman93 concerns, why is this necessary right now? We are not making frequent changes to these files, and when we do so we can always keep the PR as a draft and then only open when ready to merge. This means that CI runs at most two times when such a change is proposed and merged. That is not a huge overhead.
> > My comments on the PR are the more serious concerns I have.
> 
> > why is this necessary right now?
> 
> Costs is always necessary thing to do. We are not philanthropy who receives infinite money from the government.
> 
> > We are not making frequent changes to these files,
> 
> Which means? It's not that you're not seeing it now that it's not going to happen.
> 
> > and when we do so we can always keep the PR as a draft and then only open when ready to merge.
> 
> The point is. Not only us open PRs. We can guide the team with "good practice". But we don't say anything to community contributors. This PR is a way to mitigate that. Plus, not every new engineer is going to know everything they need to know beforehand. That's why we automate things. Cuz we never know what the others will do.
> 
> > That is not a huge overhead.
> 
> That can be true now. But with time every system tends to increase its complexity. Imagine running CI to handle just a PNG? Imagine in the future having a complex CI sequence just to change a Markdown? The CI have a cost. It's like the cost of an appliance doing unnecessary work. In the month you do not see the costs of this. But if you add every year or a decade. You can buy a private jet.
> 
> Another problem we solve with this is the following. Imagine the situation, someone introduces a bug and accidentally goes through the CI. Only then in "my" turn I'm just changing a text in a README. The CI blocks it and I'll have all the work to find out what's going on and call someone. And the person fixing the readme might not have the right skill. But usually we see a PR with a failing CI and we don't help to solve it. "Let the dev take care of it alone".
> 
> So, if we have a Markdown/PNG/JPG/others change. We should fore SURE skip the whole CI. If the CI will do any good for that PR at all.

All reasonable points -- I think we should have a discussion about file reorganization in the repo before making these changes.  There are already some strange things like the `static` directory that should probably either be removed or moved to a special directory.  That way we don't have a huge list of exceptions and can keep this no-run list as small as possible.

joshua-goldstein (approved):

skrdgraph (approved):
Thanks for working on this. This is a good first pass IMO.

skrdgraph:
@MichelDiz there may be a way to not do DRY here. Can you check on that before merging?

MichelDiz:
@aman-bansal let me do it

       - '.github/CODEOWNERS'
  This will ignore any changes to the text file that defines code owners.

       - '.vscode/**'

This will ignore any changes made to the Vscode settings. We don't do any tests with it.

       - 'compose/**'

We don't do any testing in compose. It's a tool that creates yml for docker and the like. Again, there's no testing with this or something strictly connected to that in the CI.

       - 'contrib/systemd/**'

There are no tests here. It's just a repository of sample configurations.

       - 'licenses/**'
Only license here. We don't do CI on licenses.

       - 'paper/**'
It makes no sense to do CI in text and PDF.

       - 'present/**'

This is an old folder that contained Pitch. It should be deleted. But who knows if anyone wants to use it yet.

       - 'RFC/**'
RFC is just text. There are no tests. It doesn't make sense to run CI when we insert an RFC.

       - 'static/**'
These are artifacts that should be deleted. Because it was used in the Readme. Today we no longer use it. But as this PR is not intended to delete anything. So let someone else decide what to do with this path. Until there we are ignoring it.

       - 'wiki/**'
That's a dead end, as the documentation has been moved to another repository.

       - '**/**.dockerignore'
       - '**/**.gitignore'
This is ignoring dockerignore and gitignore files. There are no tests for this.

       - '**/**.md'
This is just text. It makes no sense to run CI for that.

       - '**/**.png'
This is just image. It makes no sense to run CI for that.

       - '**/**.jpg'
This is just image. It makes no sense to run CI for that.

       - '**/**.gif'
This is just image. It makes no sense to run CI for that.

       - '**/**.ini'
This is just a config file, which maybe should be deleted. It makes no sense to run CI for that.

Let me know if some doubt remains.

MichelDiz:
This setting does NOT mean that if it find a markdown for example. Which will block the CI. If you edit anything else. It will naturally trigger the CI. This logic applied in this PR ignores only if you are editing something only like texts and the others quoted. There is no risk of not triggering the CI. Otherwise Github would not have created this feature. Or would make it very clear if it would work other way.

joshua-goldstein:
> Again, I'm not happy with this PR. There is so much duplicate things here. Please think about it and do better. @joshua-goldstein @skrdgraph

"Do better" is not actionable feedback.  What specifically do you want to see in this PR?  I agree that the list is long but for a first pass it looks reasonable now, since @MichelDiz removed the lines I had trouble with.  It would be great to have this list in only one place but AFAIK that is not possible.

mangalaman93:
Seems like there is no better way doing this. Merging this PR, thanks for doing this @MichelDiz

PR #8796, opened 2023-04-13
`fix`: Made a change for the ldbc test which will fail the test if file download fails.

SiddeshUbale:
`fix`: Made a change for the ldbc test which will fail the test if file download fails.

### DIFF
--- a/t/t.go
+++ b/t/t.go
@@ -806,7 +806,9 @@ func downloadDataFiles() {
 	}
 }
 
-func downloadLDBCFiles() {
+func downloadLDBCFiles() (out [][]byte, err []error) {
+	var outInfo [][]byte
+	var errInfo []error
 	if !*downloadResources {
 		fmt.Print("Skipping downloading of resources\n")
 		return
@@ -834,12 +836,15 @@ func downloadLDBCFiles() {
 			if out, err := cmd.CombinedOutput(); err != nil {
 				fmt.Printf("Error %v", err)
 				fmt.Printf("Output %v", out)
+				outInfo = append(outInfo, out)
+				errInfo = append(errInfo, err)
 			}
 			fmt.Printf("Downloaded %s to %s in %s \n", fname, *tmp, time.Since(start))
 		}(fname, link, &wg)
 	}
 	wg.Wait()
 	fmt.Printf("Downloaded %d files in %s \n", len(ldbcDataFiles), time.Since(start))
+	return outInfo, errInfo
 }
 
 func createTestCoverageFile(path string) error {
@@ -1006,9 +1011,6 @@ func run() error {
 		if testSuiteContains("load") || testSuiteContains("all") {
 			downloadDataFiles()
 		}
-		if testSuiteContains("ldbc") || testSuiteContains("all") {
-			downloadLDBCFiles()
-		}
 		for i, task := range valid {
 			select {
 			case testCh <- task:
@@ -1050,10 +1052,31 @@ func validateAllowed(testSuite []string) {
 	}
 }
 
+func validateDownload() {
+	// 	if len(*download) == 0 {
+	// 		return
+	// 	}
+	// 	if *download != "ldbc" {
+	// 		log.Fatalf("Only ldbc is supported for download")
+	// 	}
+	if testSuiteContains("ldbc") || testSuiteContains("all") {
+		out, err := downloadLDBCFiles()
+		length := len(out)
+		for i := 0; i < length; i++ {
+			str := string(out[i])
+			log.Println("Error Downloading a file: ", str)
+		}
+		if (len(out) != 0) || (len(err) != 0) {
+			log.Panic("Error downloading LDBC data files")
+		}
+	}
+}
+
 func main() {
 	pflag.Parse()
 	testsuite = strings.Split(*suite, ",")
 	validateAllowed(testsuite)
+	validateDownload()
 
 	rand.Seed(time.Now().UnixNano())
 	procId = rand.Intn(1000)
--- a/t/t.go
+++ b/t/t.go
@@ -994,14 +994,19 @@ func run() error {
 	go func() {
 		defer close(testCh)
 		valid := getPackages()
-		if *tmp == "" {
-			*tmp = os.TempDir()
-		}
-		x.Check(testutil.MakeDirEmpty([]string{*tmp}))
+
 		if testSuiteContains("load") || testSuiteContains("all") {
+			if *tmp == "" {
+				*tmp = os.TempDir()
+			}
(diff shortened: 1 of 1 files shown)


t/t.go
```diff
@@ -1006,9 +1011,6 @@ func run() error {
 		if testSuiteContains("load") || testSuiteContains("all") {
 			downloadDataFiles()
```
mangalaman93, line 1012: will we have the same problem if this function fails too?

t/t.go
```diff
@@ -1050,10 +1052,31 @@ func validateAllowed(testSuite []string) {
 	}
 }
 
+func validateDownload() {
+	// 	if len(*download) == 0 {
```
mangalaman93, line 1056: still need these?

t/t.go
```diff
@@ -1050,10 +1052,31 @@ func validateAllowed(testSuite []string) {
...
+	// 	}
+	// 	if *download != "ldbc" {
+	// 		log.Fatalf("Only ldbc is supported for download")
+	// 	}
+	if testSuiteContains("ldbc") || testSuiteContains("all") {
+		out, err := downloadLDBCFiles()
+		length := len(out)
+		for i := 0; i < length; i++ {
+			str := string(out[i])
+			log.Println("Error Downloading a file: ", str)
```
mangalaman93, line 1067: just use small letters in panic comment, same below

t/t.go
SiddeshUbale: Resolved: Applied the same fix for the load test.

t/t.go
SiddeshUbale: Resolved

t/t.go
SiddeshUbale: Resolved

t/t.go
```diff
@@ -802,11 +807,16 @@ func downloadDataFiles() {
 		if out, err := cmd.CombinedOutput(); err != nil {
 			fmt.Printf("Error %v", err)
 			fmt.Printf("Output %v", out)
+			outInfo = append(outInfo, out)
+			errInfo = append(errInfo, err)
 		}
 	}
+	return outInfo, errInfo
```
mangalaman93, line 814: can we check for error right here instead of returning it?

mangalaman93 (changes requested):

t/t.go
```diff
@@ -786,7 +786,12 @@ var ldbcDataFiles = map[string]string{
 	"ldbcTypes.schema": "https://github.com/dgraph-io/benchmarks/blob/master/ldbc/sf0.3/ldbcTypes.schema?raw=true",
 }
 
-func downloadDataFiles() {
+var outInfo [][]byte
```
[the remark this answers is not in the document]
SiddeshUbale: All the variables related to ldbc and load datafiles are declared here, that's why I appended both of these variables right below them, We can declare them at the top of the file.

t/t.go
SiddeshUbale: Resolved

t/t.go
```diff
@@ -800,8 +800,10 @@ func downloadDataFiles() {
 		cmd.Dir = *tmp
 
 		if out, err := cmd.CombinedOutput(); err != nil {
-			fmt.Printf("Error %v", err)
-			fmt.Printf("Output %v", out)
+			fmt.Printf("Error %v /n", err)
+			str := string(out)
```
mangalaman93, line 804: we don't need to do this assignment here, we can directly print it

t/t.go
```diff
@@ -800,8 +800,10 @@ func downloadDataFiles() {
...
+			log.Println("error downloading a file: ", str)
+			log.Panic("error downloading required load test files")
```
mangalaman93, line 806: why don't we just do `panic("error downloading required load test files")` here instead?

t/t.go
```diff
@@ -800,8 +800,10 @@ func downloadDataFiles() {
 		cmd.Dir = *tmp
 
 		if out, err := cmd.CombinedOutput(); err != nil {
-			fmt.Printf("Error %v", err)
-			fmt.Printf("Output %v", out)
+			fmt.Printf("Error %v /n", err)
+			str := string(out)
+			log.Println("error downloading a file: ", str)
```
mangalaman93, line 805: we should just use fmt everywhere.

mangalaman93 (changes requested):
A few more comments, this looks better. Could you show an example output too?

t/t.go
SiddeshUbale: Resolved

t/t.go
SiddeshUbale: Resolved.

mangalaman93 (changes requested):

t/t.go
```diff
@@ -791,17 +791,13 @@ func downloadDataFiles() {
...
-	}
-	x.Check(testutil.MakeDirEmpty([]string{*tmp}))
 	for fname, link := range datafiles {
 		cmd := exec.Command("wget", "-O", fname, link)
 		cmd.Dir = *tmp
 
 		if out, err := cmd.CombinedOutput(); err != nil {
-			fmt.Printf("Error %v", err)
-			fmt.Printf("Output %v", out)
+			fmt.Printf("Error %v /n", err)
```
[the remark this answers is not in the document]
SiddeshUbale: Resolved

t/t.go
```diff
@@ -833,7 +824,7 @@ func downloadLDBCFiles() {
 			cmd.Dir = *tmp
 			if out, err := cmd.CombinedOutput(); err != nil {
 				fmt.Printf("Error %v", err)
```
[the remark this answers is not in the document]
SiddeshUbale: Resolved

t/t.go
```diff
@@ -1003,11 +994,21 @@ func run() error {
 	go func() {
 		defer close(testCh)
 		valid := getPackages()
-		if testSuiteContains("load") || testSuiteContains("all") {
-			downloadDataFiles()
-		}
 		if testSuiteContains("ldbc") || testSuiteContains("all") {
+			if *tmp == "" {
```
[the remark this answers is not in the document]
SiddeshUbale: Resolved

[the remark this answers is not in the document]
mangalaman93: We can move the following code outside the ifs and before them:
```
if *tmp == "" {
    *tmp = filepath.Join(os.TempDir(), "/ldbcdata")
}
x.Check(testutil.MakeDirEmpty([]string{*tmp}))
```

mangalaman93 (changes requested):

mangalaman93 (approved):

all-seeing-code (approved):

PR #8798, opened 2023-04-19
fix(ci): frequency fix

skrdgraph:
## Problem
higher frequency was enabled to solve flaky tests .. this is more $$'s

## Solution
reduce frequency because we are done with RC testing

### DIFF
--- a/.github/workflows/ci-dgraph-code-coverage.yml
+++ b/.github/workflows/ci-dgraph-code-coverage.yml
@@ -14,7 +14,7 @@ on:
       - main
       - 'release/**'
   schedule:
-    - cron: "0 */2 * * *" # every 2hrs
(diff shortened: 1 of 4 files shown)


mangalaman93 (approved):

adityasadalage (approved):
LGTM

.github/workflows/ci-dgraph-code-coverage.yml
```diff
@@ -14,7 +14,7 @@ on:
       - main
       - 'release/**'
   schedule:
-    - cron: "0 */2 * * *" # every 2hrs
+    - cron: "0 0,6,12,18 * * *"
```
nswamy, line 17: whats the difference in the syntax ? can you just say `0 */2` if you are changing it to every 6 hours.

PR #8802, opened 2023-04-20
use env var for Mac for supporting dev workflow

mangalaman93:
Because MAC runs a linux VM for running docker container, we can't follow the normal development workflow on Mac. we need to make two changes in the workflow:
 1. Additionally run GOOS=linux make install whenever running make install
 2. export the following variable in your rc file
       export DGRAPH_BIN_PATH=$GOPATH/bin/linux_$GOARCH

### DIFF
--- a/compose/compose.go
+++ b/compose/compose.go
@@ -176,7 +176,7 @@ func initService(basename string, idx, grpcPort int) service {
 	if opts.LocalBin {
 		svc.Volumes = append(svc.Volumes, volume{
 			Type:     "bind",
-			Source:   "$GOPATH/bin",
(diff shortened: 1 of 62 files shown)


compose/compose.go
```diff
@@ -176,7 +176,7 @@ func initService(basename string, idx, grpcPort int) service {
 	if opts.LocalBin {
 		svc.Volumes = append(svc.Volumes, volume{
 			Type:     "bind",
-			Source:   "$GOPATH/bin",
+			Source:   "${DGRAPH_BIN_PATH:-$GOPATH/bin}",
```
sanjayk-github-dev, line 179: nit: can you put this on the same line?

sanjayk-github-dev (approved):

PR #8803, opened 2023-04-21
chore(test): improve health check for NFS backup tests

shivaji-dgraph:
NFS backup tests fail frequently in CI. The test installs the NFS client before starting the alpha service because of this the Alpha container takes more time to reach a healthy state and may not pass the health check. To solve this, we check the health of Alpha within the tests itself (essentially giving Alpha 100 seconds to reach a healthy state). Example failed run is available here https://github.com/dgraph-io/dgraph/actions/runs/4744244095/jobs/8424889366

### DIFF
--- a/systest/backup/nfs-backup/backup_test.go
+++ b/systest/backup/nfs-backup/backup_test.go
@@ -28,7 +28,6 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
-	"time"
 
 	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc"
@@ -51,11 +50,8 @@ var (
 
 func TestBackupHAClust(t *testing.T) {
 	//TODO: Polling for health-checks with a 1sec wait for 100 times
-	for i := 1; i <= 100; i++ {
-		if err := testutil.CheckHealthContainer(testutil.ContainerAddr("alpha1", 8080)); err == nil {
-			break
-		}
-		time.Sleep(1 * time.Second)
+	if err := testutil.CheckHealthContainer(testutil.ContainerAddr("alpha1", 8080)); err != nil {
+		require.NoError(t, err, "alpha1 did not pass health check")
 	}
 
 	backupRestoreTest(t, testutil.SockAddr, testutil.SockAddrAlpha4Http,
--- a/systest/backup/nfs-backup/backup_test.go
+++ b/systest/backup/nfs-backup/backup_test.go
@@ -50,12 +50,12 @@ var (
 )
 
 func TestBackupHAClust(t *testing.T) {
-	// check health of alpha1
-	for i := 1; i <= 50; i++ {
+	//TODO: Polling for health-checks with a 1sec wait for 100 times
+	for i := 1; i <= 100; i++ {
 		if err := testutil.CheckHealthContainer(testutil.ContainerAddr("alpha1", 8080)); err == nil {
 			break
 		}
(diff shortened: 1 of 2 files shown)


systest/backup/nfs-backup/backup_test.go
```diff
@@ -49,6 +50,13 @@ var (
 )
 
 func TestBackupHAClust(t *testing.T) {
+	// check health of alpha1
```
skrdgraph, line 53: add a comment here explaining the logic, with a `TODO` flag. For example: 
```
//TODO: Polling for health-checks with a 1sec wait for 100 times ... this can be changed to a push event action instead of polling this way. 

We can come back and fix it later with this TODO flag.

File: systest/backup/nfs-backup/backup_test.go [hunk 2]
```diff
@@ -49,6 +50,13 @@ var (
...
+	for i := 1; i <= 50; i++ {
+		if err := testutil.CheckHealthContainer(testutil.ContainerAddr("alpha1", 8080)); err == nil {
+			break
+		}
+		time.Sleep(2 * time.Second)
```
skrdgraph, line 58: nitpick - change from `2*time.second` to `1*time.second` && increase the for loop from `i<=50` to `I<=100`. 

this would be more readable.

skrdgraph:
@shivaji-dgraph pls add a description to the PR. Preferably in the format `Problem/Solution`. In your case you are raising this PR because of flaky test issue, and your temporary solution is to increase the health check window. You can also provide reference failures link from the CI jobs to justify the need for this PR.

Raising PRs with description is not a good practice. When we trace things back to this PR in the future, it will be difficult to understand the context.

skrdgraph (changes requested):
@shivaji-dgraph pls add a description to the PR. Preferably in the format `Problem/Solution`. In your case you are raising this PR because of flaky test issue, and your temporary solution is to increase the health check window. You can also provide reference failures link from the CI jobs to justify the need for this PR.

Raising PRs with description is not a good practice. When we trace things back to this PR in the future, it will be difficult to understand the context.

systest/backup/nfs-backup/backup_test.go
shivaji-dgraph: done @skrdgraph

systest/backup/nfs-backup/backup_test.go
shivaji-dgraph: done

skrdgraph:
@shivaji-dgraph I have a question ... it looks like you are controlling this outside of the `CheckHealthContainer` function. If you look at this function https://github.com/dgraph-io/dgraph/blob/47c6b68d35e92b20cec99ad35994e5527f2dba40/testutil/docker.go#L292-L314 .. it's already having a looped wait using polling mechanisms. I am wondering if this can enhanced inside the CheckHealthContainer itself. 

Any reason you chose to do this outside?

systest/backup/nfs-backup/backup_test.go
```diff
@@ -49,6 +49,10 @@ var (
 )
 
 func TestBackupHAClust(t *testing.T) {
+	//TODO: Polling for health-checks with a 1sec wait for 100 times
```
mangalaman93, line 52: why is there a TODO here?

mangalaman93 (changes requested):
minor comment.

PR #8808, opened 2023-04-26
chore(core): Update comment about how drop op behaves with rollups

harshil-goel:
## Problem
Dgraph.drop.op doesn't conflict out other transactions, unlike what we have written in code.

## Solution
We have to update the explanation about how dgraph.drop.op would behave under rollups.

### DIFF
--- a/posting/list.go
+++ b/posting/list.go
@@ -837,12 +837,14 @@ func (l *List) Length(readTs, afterUid uint64) int {
 //     read a given backup or not. A backup, which has a drop record, would render older backups
 //     unnecessary.
 //
-//     If we rollup the dgraph.drop.op, and store result on ts + 1, which moves the original record.
-//     We want to see if there can be any issues in backup/restore due to this. To ensure that there
-//     is no issue in writing on ts + 1, we do the following analysis. Analysis is done for
-//     drop op, but it would be the same for drop predicate and namespace. Assume that there were
-//     two backups, at b1 and b2. We move rollup ts around to see if it can cause any issues. There
-//     can be 3 cases:
+//     If we rollup the dgraph.drop.op, and store result on ts + 1, it effectively copies the
+//     original record into a new location. We want to see if there can be any issues in
+//     backup/restore due to this. To ensure that there is no issue in writing on ts + 1,
+//     we do the following analysis.
+//
+//     Analysis is done for drop op, but it would be the same for drop predicate and namespace.
+//     Assume that there were two backups, at b1 and b2. We move rollup ts around to see if it
+//     can cause any issues. There can be 3 cases:
 //
 //     1. b1 < ts < b2. In this case, we would have a drop record in b2. This is the same behaviour
 //     as we would have writen on ts.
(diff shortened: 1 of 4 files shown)


mangalaman93 (dismissed):

MichelDiz (dismissed):

posting/list.go
```diff
@@ -831,8 +831,30 @@ func (l *List) Length(readTs, afterUid uint64) int {
...
 //   - Drop operation can cause issues if they are rolled up. Since we are storing results at ts + 1,
-//     in dgraph.drop.op. If some operations were done then, they would be overwriten.
-//     This won't happen as the transactions should conflict out.
+//     in dgraph.drop.op. When we do drop op, we delete the relevant data first using a mutation.
+//     Then we write a record into the dgraph.drop.op. We use this record to figure out if a
+//     drop was performed. This helps us during backup, when we want to know if we need to
+//     read a given backup or not. A backup, which has a drop record, would render older backups
+//     unnecessary.
+//
+//     If we rollup the dgraph.drop.op, and store result on ts + 1, we would move the record. We
```
anurags92, line 840: >we would move the record.
Move what and where? Need more clarification.

posting/list.go
```diff
@@ -831,8 +831,30 @@ func (l *List) Length(readTs, afterUid uint64) int {
...
+//
+//     2. b1 = ts < b2. In this case, we would have a drop record in b1, and in b2. In this case
+//     originally we would have read both b2 and b1. However, last item in b1 is drop, so
+//     it wouldn't have any data anyways. So we don't need to read b1.
+//
+//     3. b1 < ts < ts + 1 = b2. In this case, we would have both drop drop records in b2. No issues
+//     in this case.
+//
+//     The only issue would come if a backup happened in between ts, and ts + 1. As our timestamps
+//     are integers, it's not possible. However, a bug could happen if we store the rollup at ts + k
```
anurags92, line 857: So what's the point? If a bug can happen, are we addressing it, acknowledging it?

I don't see the point of this example. It needs more explanation as to what are we trying to achieve here.

posting/list.go
anurags92: which moves the drop record that we originally created.

posting/list.go
```diff
@@ -831,8 +831,30 @@ func (l *List) Length(readTs, afterUid uint64) int {
...
-//     in dgraph.drop.op. If some operations were done then, they would be overwriten.
-//     This won't happen as the transactions should conflict out.
+//     in dgraph.drop.op. When we do drop op, we delete the relevant data first using a mutation.
+//     Then we write a record into the dgraph.drop.op. We use this record to figure out if a
+//     drop was performed. This helps us during backup, when we want to know if we need to
+//     read a given backup or not. A backup, which has a drop record, would render older backups
+//     unnecessary.
+//
+//     If we rollup the dgraph.drop.op, and store result on ts + 1, we would move the record. We
+//     want to see if there can be any issues in backup/restore due to this. Analysis is done for
```
anurags92, line 841: To ensure that there is not issue on writing at ts+1, we do the following analysis.

posting/list.go
```diff
@@ -831,8 +831,32 @@ func (l *List) Length(readTs, afterUid uint64) int {
...
+//     2. b1 = ts < b2. In this case, we would have a drop record in b1, and in b2. Originally, only
+//     b1 would have a drop record. With this new approach, b2 would also have a drop record. This
+//     is okay because last entry in b1 is drop, so it wouldn't have any data to be applied.
+//
+//     3. b1 < ts < ts + 1 = b2. In this case, we would have both drop drop records in b2. No issues
+//     in this case.
+//
+//     This proves that writing rollups at ts + 1 would not cause any issues with dgraph.drop.op.
+//     The only issue would come if a rollup happens at ts + k. If a backup happens in between
+//     ts and ts + k, it would lead to some data being dropped during restore.
```
anurags92, line 859: I would put it like --- `It could lead to some data`... Other than that, this looks great!

mangalaman93 (approved):

anurags92 (approved):

PR #8811, opened 2023-05-01
build: build dgraph inside docker

joshua-goldstein:
## Problem

Currently we build dgraph locally, then build our docker image with this locally built binary.  We also inject this binary into our docker-compose files.  However, since we use images with base image ubuntu-20, building dgraph on ubuntu-22 presents a problem since libc libraries on ubuntu 22 (which are linked into the dgraph binary) will mean that this binary will not work on ubuntu 20.  See e.g. [here](https://github.com/dgraph-io/dgraph/pull/8702).

We also have a slightly strange workflow on MacOS, in which we need to compile a Dgraph binary using the `GOOS=linux GOARCH=amd64` flags (to build dgraph for this platform), then move this binary into the GOPATH before running tests.  This build process would mean calling `make docker-image` would "just work" on any platform.  (We would still have to stop injecting the binary in the test docker-compose yamls for the workflow to feel natural).

## Solution

Use multistep build process to build dgraph and dgraph docker images.  This is currently a draft and can probably be optimized; e.g. build times are quite long and might not be practical.  However, this still could be useful for CD pipelines.


contrib/Dockerfile
```diff
@@ -1,10 +1,48 @@
-# This file is used to add the nightly Dgraph binaries and assets to Dgraph base
-# image.
+# This file builds the dgraph binary and dgraph docker image.
 
-# This gets built as part of release.sh. Must be run from /tmp/build, with the linux binaries
-# already built and placed there.
+# This gets built as part of our CD pipeline. Must be run through Makefile.
 
-FROM ubuntu:20.04
+FROM --platform=linux/amd64 ubuntu:20.04 AS build
```
joshua-goldstein, line 5: this flag is needed to make this work on apple m1; otherwise it will default to linux/arm64

contrib/Dockerfile
```diff
@@ -1,10 +1,48 @@
...
+    BUILD_DATE=$BUILD_DATE \
+    BUILD_BRANCH=$BUILD_BRANCH \
+    BUILD_VERSION=$BUILD_VERSION \
+    make dgraph
+
+RUN mkdir bin && mv dgraph/dgraph bin
+
+# -------------------------------------------------------------------------------
+
+FROM --platform=linux/amd64 ubuntu:20.04
```
joshua-goldstein, line 45: this flag is needed to make this work on apple m1; otherwise it will default to linux/arm64

mangalaman93: we should just add this comment to Dockerfile itself

contrib/Dockerfile
mangalaman93, line 5: maybe we could use the golang images instead of downloading Go ourselves

contrib/Dockerfile
```diff
@@ -1,10 +1,48 @@
...
+ARG BUILD_VERSION
+
+RUN BUILD=$BUILD \
+    BUILD_CODENAME=$BUILD_CODENAME \
+    BUILD_DATE=$BUILD_DATE \
+    BUILD_BRANCH=$BUILD_BRANCH \
+    BUILD_VERSION=$BUILD_VERSION \
+    make dgraph
+
+RUN mkdir bin && mv dgraph/dgraph bin
```
mangalaman93, line 41: why are we doing this?

contrib/Dockerfile
```diff
@@ -1,10 +1,48 @@
...
+    && tar -C /usr/local/ -xzf go${GO_VERSION}.linux-amd64.tar.gz \
+    && rm -f go${GO_VERSION}.linux-amd64.tar.gz
+
+ENV PATH=$PATH:/usr/local/go/bin:/root/go/bin \
+    GOPATH=/root/go
+
+# save time by caching this layer
+RUN cd dgraph && make jemalloc
+
+ARG BUILD
```
mangalaman93, line 28: where are these args used?

mangalaman93 (changes requested):

contrib/Dockerfile
joshua-goldstein: I don't follow?  this flag changes the base image that is used.

contrib/Dockerfile
joshua-goldstein: I considered that - but the issue is that the golang images use alpine linux, not ubuntu, which would not necessarily be a problem in itself.  The main issue is the alpine linux uses the musl c library instead of glibc, and they are not compatible.  See https://stackoverflow.com/questions/33382707/alpine-musl-vs-glibc-are-they-supposed-to-be-compatible

contrib/Dockerfile
joshua-goldstein: we cannot execute git commands inside of the container (since it is a very lightweight image is does not include `git`).  However part of our build process includes invoking `git` to fetch info about what branch we are on, commit SHA, etc (this is in `Makefile` and `dgraph/Makefile`).  So it is not possible to do this inside the container.  This workaround means that these parameters are passed inside the container as docker arguments.

contrib/Dockerfile
joshua-goldstein: because we eventually do `COPY --from=build /dgraph/bin /usr/local/bin`
I think it is cleaner and less error-prone to create a new directory and move just the dgraph executable into that `bin` directory.  The other option is to copy just the dgraph executable from the dgraph directory, but then we have to be careful that we are not copying the full contents of the directory.  This is also more explicit IMO

contrib/Dockerfile
joshua-goldstein: I should mention that this is an issue since we will eventually want to give first class support to linux/arm64 and these kinds of explicit parameters preclude that.

mangalaman93: the comment that you made here on GitHub, I am suggesting that the same comment should be present in line 45 of the Dockerfile.

contrib/Dockerfile
mangalaman93: I see debian based images, but I understand if you want to do it yourself. We should add a note to the Dockerfile about this too that we can't use golang docker images as our base image

contrib/Dockerfile
mangalaman93: makes sense. Let's add a comment in the Dockerfile about this too.

contrib/Dockerfile
mangalaman93: we can install git, that is a throw away container anyway instead of figuring out new workarounds.

contrib/Dockerfile
```diff
@@ -1,10 +1,48 @@
...
-# already built and placed there.
+# This gets built as part of our CD pipeline. Must be run through Makefile.
 
-FROM ubuntu:20.04
+FROM --platform=linux/amd64 ubuntu:20.04 AS build
+
+RUN mkdir /dgraph
+WORKDIR /dgraph
+
+COPY . ./
```
mangalaman93, line 10: we should delay this in steps

skrdgraph:
Going to block this because it's a good to have - but not helping us drastically. We can circle back on this later .. as we have a ton of other things.

billprovince (commented):
FYI: This failed on my ubuntu 22.04 laptop.

Bringing up cluster test-166-1 for package: ../dgraph/docker-compose.yml ...
While running command: "docker-compose --compatibility -f ../dgraph/docker-compose.yml -p test-166-1 up --force-recreate --build --remove-orphans --detach" Error: exit status 1
Running: /usr/local/go/bin/go test -failfast -v -tags=integration -timeout 30m github.com/dgraph-io/dgraph/ee/acl with test-166-1
Docker logs for 373a800ce770c28c051d83b7dbe12bd9ed7f0fcd92cc0ccc656fd407dc59cef8 is  with error <nil> Docker logs for 518a907c0316a29cbbf280183a463fbfbc28f1c5a4496f647f12f44881f718b5 is  with error <nil> ---> LOGS for test-166-1 written to ./test-166-11415513190.log .

billprovince (commented):
FYI: This failed on my ubuntu 22.04 laptop.

Bringing up cluster test-166-1 for package: ../dgraph/docker-compose.yml ...
While running command: "docker-compose --compatibility -f ../dgraph/docker-compose.yml -p test-166-1 up --force-recreate --build --remove-orphans --detach" Error: exit status 1
Running: /usr/local/go/bin/go test -failfast -v -tags=integration -timeout 30m github.com/dgraph-io/dgraph/ee/acl with test-166-1
Docker logs for 373a800ce770c28c051d83b7dbe12bd9ed7f0fcd92cc0ccc656fd407dc59cef8 is  with error <nil> Docker logs for 518a907c0316a29cbbf280183a463fbfbc28f1c5a4496f647f12f44881f718b5 is  with error <nil> ---> LOGS for test-166-1 written to ./test-166-11415513190.log .

I verified that GOPATH is properly set (export GOPATH=`env go GOPATH`)

billprovince (commented):
when testing on a Mac M1 (and after commenting out the line locally in t.go inside of startCluster that specifies cmd.stdErr = nil), I get the error info:

PR #8816, opened 2023-05-03
chore: wait longer for login to work

mangalaman93:
Seems like it takes longer for a 6 alpha cluster to be available for running queries.

### DIFF
--- a/dgraphtest/cluster.go
+++ b/dgraphtest/cluster.go
@@ -297,7 +297,9 @@ loop:
 		time.Sleep(waitDurBeforeRetry)
 
 		resp, err := c.AlphasHealth()
-		if err != nil {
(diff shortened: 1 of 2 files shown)


anurags92 (approved):

dgraphtest/cluster.go
```diff
@@ -297,7 +297,9 @@ loop:
 		time.Sleep(waitDurBeforeRetry)
 
 		resp, err := c.AlphasHealth()
-		if err != nil {
+		if err != nil && strings.Contains(err.Error(), "the server is in draining mode") {
```
harshil-goel, line 300: Can you make the error string a constant. Both here, and where it is emitted.

mangalaman93: The error string is actually much longer. I don't know if we want to compare such a long string here.

PR #8818, opened 2023-05-06
chore: fix flaky test TestGetAccessJwt

mangalaman93:


edgraph/access_ee_test.go
```diff
@@ -100,8 +99,10 @@ func TestGetAccessJwt(t *testing.T) {
 		{3456789012, "user3", []string{"702", "703"}},
 	}
 
+	worker.Config.AccessJwtTtl = 20 * time.Second
```
anurags92, line 102: Do you think its a good idea to override it here instead of in config?

mangalaman93: This is a unit test. There is not config.

harshil-goel (approved):

anurags92 (approved):

PR #8827, opened 2023-05-16
add @unique constraint support in schema for new predicates

shivaji-dgraph:
Partially Fixes #8827
Closes: DGRAPHCORE-206
Docs PR: https://github.com/dgraph-io/dgraph-docs/pull/638

This PR adds support for uniqueness constraint using @unique directive in DQL schema. This unique directive ensures that all values of the predicate are different in a Dgraph Cluster. This completes phase 1, and enables adding a new predicate with unique directive. As part of the phase 2, we will work on adding support for unique directive for existing predicates.

## Performance
Live Loader before this change on 21 million dataset took 10m54s whereas after this change took 11m02s. It did not make any significant different to non-unique predicates.

## How to Use
You can now specify unique in schema as follows: `email: string @unique @index(hash) @upsert .`. Now, Dgraph will ensure that no mutation adds a duplicate for the predicate email.

## Phase 2 [TODO]
- [ ] check if @unique can be added to schema depending upon whether existing data has any duplicates. If the existing data has any duplicates, we do not allow adding the @unique directive and return a query that allows user to identify these UIDs.
- [ ] If index computation is in progress, we should not allow mutations with predicates for which @unique is set
- [ ] Fix ACL to ensure that we do not end up adding duplicate users
- [ ] Ensure that unique constraint is not violated during Bulk loader


worker/mutation.go
```diff
@@ -424,6 +424,32 @@ func checkSchema(s *pb.SchemaUpdate) error {
 			x.ParseAttr(s.Predicate))
 	}
 
+	if s.Unique {
+		if s.ValueType == pb.Posting_STRING {
```
mangalaman93, line 428: unique directive does not mean that the predicate cannot have term index. It just means that we should also have the hash or exact index. We should prefer the hash index.

worker/schema.go
```diff
@@ -110,6 +110,8 @@ func populateSchema(attr string, fields []string) *pb.SchemaNode {
 			schemaNode.List = pred.GetList()
 		case "upsert":
```
mangalaman93, line 111: you could combine the upsert and the unique cases

worker/mutation.go
```diff
@@ -424,6 +424,32 @@ func checkSchema(s *pb.SchemaUpdate) error {
 			x.ParseAttr(s.Predicate))
 	}
 
+	if s.Unique {
```
mangalaman93, line 427: a switch statement would look better here. and add a todo for the types that are added here.

protos/pb/pb.pb.go
```diff
@@ -3577,6 +3577,7 @@ type SchemaUpdate struct {
 	// custom name. This field stores said name.
 	ObjectTypeName string `protobuf:"bytes,12,opt,name=object_type_name,json=objectTypeName,proto3" json:"object_type_name,omitempty"`
 	NoConflict     bool   `protobuf:"varint,13,opt,name=no_conflict,json=noConflict,proto3" json:"no_conflict,omitempty"`
+	Unique    bool                   `protobuf:"varint,14,opt,name=unique,proto3" json:"unique,omitempty"`
```
mangalaman93, line 3580: you will have to modify the api.proto file in dgo repo

mangalaman93 (changes requested):
A few comments, this looks pretty good to start with. We need to add test for parsing the schema.

edgraph/server.go
```diff
@@ -1544,8 +1544,30 @@ func parseRequest(qc *queryContext) error {
 			if err != nil {
 				return err
 			}
+			newGmu := dql.Mutation{Set: make([]*api.NQuad, 0)}
```
mangalaman93, line 1547: we should modify the request while executing it instead of parsing it.

edgraph/server.go
```diff
@@ -1544,8 +1544,30 @@ func parseRequest(qc *queryContext) error {
...
+			if isUnique(gmu, ctx) {
+				for _, singlePred := range gmu.Set {
+
+					predSchema, _ := schema.State().Get(ctx, fmt.Sprint(singlePred.GetNamespace())+"-"+singlePred.GetPredicate())
```
mangalaman93, line 1551: why are we not using singlePred directly and using fmt.Sprint?

edgraph/server.go
```diff
@@ -1544,8 +1544,30 @@ func parseRequest(qc *queryContext) error {
...
+					if predSchema.Unique {
```
mangalaman93, line 1552: we already checked in isUnique whether the mutation has a predicate with unique schema. Why are we checking here again?

mangalaman93 (changes requested):

edgraph/server.go
```diff
@@ -1573,6 +1609,64 @@ func parseRequest(qc *queryContext) error {
...
+// addQueryIfUniqueTrue modifies the request if @unique defined on Predicate
+func addQueryIfUniqueTrue(ctx context.Context, qc *queryContext) {
+	if len(qc.gmuList) == 0 {
+		return
+	}
+
+	qc.uniqueVars = map[string]uniqueVarsUids{}
+	for _, gmu := range qc.gmuList {
+
+		if !isUnique(gmu, ctx) {
```
mangalaman93, line 1621: We don't need to do this work twice.

edgraph/server.go
```diff
@@ -1573,6 +1609,64 @@ func parseRequest(qc *queryContext) error {
...
+			continue
+		}
+
+		buildQuery := ""
+		for i, singlePred := range gmu.Set {
+			predSchema, _ := schema.State().Get(ctx, x.NamespaceAttr(singlePred.Namespace, singlePred.Predicate))
+			if predSchema.Unique {
+				//use fmt.SprintF for buils queryma
+				buildQuery = buildQuery + fmt.Sprintf(`UID%v as v%v(func: eq(%v,"%v"))
```
mangalaman93, line 1630: variable name could be something like __UID_%v__ to avoid conflict

edgraph/server.go
```diff
@@ -1573,6 +1609,64 @@ func parseRequest(qc *queryContext) error {
...
+				{
+					uid
+				}`, strconv.Itoa(i), strconv.Itoa(i), singlePred.Predicate, singlePred.ObjectValue.GetDefaultVal())
+
+				// So here we are adding query for predicate if unique is defined
+				// for example  Schema => email: string @unique @index(exact) .
+				// To ensure uniqueness of email predicate we will query for value of predicate
+				// for example mutation =>  _:a <email> "example@email.com"  .
+				// then query will be
+				// UID1 as v1(func: eq(email,"example@email.com"))
```
mangalaman93, line 1640: we don't need `v1`

edgraph/server.go
```diff
@@ -1573,6 +1609,64 @@ func parseRequest(qc *queryContext) error {
...
+				//{
```
mangalaman93, line 1641: you don't need this part either.

edgraph/server.go
```diff
@@ -1573,6 +1609,64 @@ func parseRequest(qc *queryContext) error {
...
+				//	uid
+				//}
+				qc.uniqueVars["UID"+strconv.Itoa(i)] = uniqueVarsUids{singlePred.Subject, singlePred.ObjectValue.GetDefaultVal(), singlePred.Predicate}
```
mangalaman93, line 1644: we should construct the variable name once

edgraph/server.go
```diff
@@ -1473,6 +1494,19 @@ func processQuery(ctx context.Context, qc *queryContext) (*api.Response, error)
...
 
+	for name, value := range qc.uniqueVars {
+		if checkForDuplicateBlankNode(value, qc.uniqueVars) {
+			return resp, errors.New("duplicate values present in the mutation")
+		}
+
+		v := qr.Vars[name]
+		if v.Uids != nil {
+			if qc.uniqueVars[name].subjectUid != "0x"+strconv.FormatInt(int64(v.Uids.Uids[0]), 16) {
+				return resp, errors.New("abort mutation")
```
mangalaman93, line 1505: need to improve the error here as well.

mangalaman93 (changes requested):

edgraph/server.go
```diff
@@ -1311,7 +1333,7 @@ func (s *Server) doQuery(ctx context.Context, req *Request) (resp *api.Response,
 		graphql:  isGraphQL,
 		gqlField: req.gqlField,
 	}
-	if rerr = parseRequest(qc); rerr != nil {
+	if rerr = parseRequest(ctx, qc); rerr != nil {
```
mangalaman93, line 1336: we don't need to pass ctx here

edgraph/server.go
```diff
@@ -1473,6 +1495,22 @@ func processQuery(ctx context.Context, qc *queryContext) (*api.Response, error)
 	}
 	qc.span.Annotatef(nil, "Response = %s", resp.Json)
 
+	for name, value := range qc.uniqueVars {
+
```
mangalaman93, line 1499: nit: extra new line

edgraph/server.go
```diff
@@ -1473,6 +1495,22 @@ func processQuery(ctx context.Context, qc *queryContext) (*api.Response, error)
...
+		if checkForDuplicateBlankNode(value, qc.uniqueVars) {
+			return resp, errors.Errorf("Can not insert duplicate value [%v] for predicate [%v]",
```
mangalaman93, line 1501: Can not --> could not

edgraph/server.go
```diff
@@ -1573,6 +1613,62 @@ func parseRequest(qc *queryContext) error {
 	return validateQuery(qc.dqlRes.Query)
 }
 
+// addQueryIfUniqueTrue modifies the request if @unique defined on Predicate
```
mangalaman93, line 1616: addQueryIfUniqueTrue -> addQueryIfUnique

addQueryIfUniqueTrue adds dummy queries in the request for checking whether predicate is unique in the db

edgraph/server.go
```diff
@@ -1573,6 +1613,62 @@ func parseRequest(qc *queryContext) error {
...
+func addQueryIfUniqueTrue(ctx context.Context, qc *queryContext) {
+	if len(qc.gmuList) == 0 {
+		return
+	}
+
+	qc.uniqueVars = map[string]uniqueVarsUids{}
+	for _, gmu := range qc.gmuList {
+
```
mangalaman93, line 1624: nit: don't need this newline

edgraph/server.go
```diff
@@ -1573,6 +1613,62 @@ func parseRequest(qc *queryContext) error {
...
+		if !isUnique(gmu, ctx) {
+			continue
+		} else {
+			buildQuery := ""
+			for i, singlePred := range gmu.Set {
```
mangalaman93, line 1629: singlePred -> pred

edgraph/server.go
```diff
@@ -1573,6 +1613,62 @@ func parseRequest(qc *queryContext) error {
...
+				predSchema, _ := schema.State().Get(ctx, x.NamespaceAttr(singlePred.Namespace, singlePred.Predicate))
```
mangalaman93, line 1630: set write context here

edgraph/server.go
```diff
@@ -1573,6 +1613,62 @@ func parseRequest(qc *queryContext) error {
...
+	if len(qc.gmuList) == 0 {
+		return
+	}
+
+	qc.uniqueVars = map[string]uniqueVarsUids{}
+	for _, gmu := range qc.gmuList {
+
+		if !isUnique(gmu, ctx) {
+			continue
+		} else {
```
mangalaman93, line 1627: don't need else

edgraph/server.go
```diff
@@ -1573,6 +1613,62 @@ func parseRequest(qc *queryContext) error {
...
+			buildQuery := ""
+			for i, singlePred := range gmu.Set {
+				predSchema, _ := schema.State().Get(ctx, x.NamespaceAttr(singlePred.Namespace, singlePred.Predicate))
+				if predSchema.Unique {
+					queryVar := fmt.Sprintf("UID%v_", strconv.Itoa(i))
+					//use fmt.SprintF for buils queryma
```
mangalaman93, line 1633: remove this

edgraph/server.go
```diff
@@ -1573,6 +1613,62 @@ func parseRequest(qc *queryContext) error {
...
+	for _, gmu := range qc.gmuList {
+
+		if !isUnique(gmu, ctx) {
+			continue
+		} else {
+			buildQuery := ""
+			for i, singlePred := range gmu.Set {
+				predSchema, _ := schema.State().Get(ctx, x.NamespaceAttr(singlePred.Namespace, singlePred.Predicate))
+				if predSchema.Unique {
+					queryVar := fmt.Sprintf("UID%v_", strconv.Itoa(i))
```
mangalaman93, line 1632: `__dgraph_uniquecheck_%v__`
change the upsert variable name too

mangalaman93: i can be repeated, add a test case for this too

edgraph/server.go
```diff
@@ -1573,6 +1613,62 @@ func parseRequest(qc *queryContext) error {
...
+					//use fmt.SprintF for buils queryma
+					buildQuery = buildQuery + fmt.Sprintf(`%v as var(func: eq(%v,"%v"))
+				{
```
mangalaman93, line 1635: don't need this

edgraph/server.go
```diff
@@ -1573,6 +1613,62 @@ func parseRequest(qc *queryContext) error {
...
+					uid
+				}`, queryVar, singlePred.Predicate, singlePred.ObjectValue.GetDefaultVal())
```
mangalaman93, line 1637: there should be a better way to do this

edgraph/server.go
```diff
@@ -1573,6 +1613,62 @@ func parseRequest(qc *queryContext) error {
...
+
+					// So here we are adding query for predicate if unique is defined
```
mangalaman93, line 1639: Now, we add a query for a predicate if it is unique.

edgraph/server.go
```diff
@@ -1573,6 +1613,62 @@ func parseRequest(qc *queryContext) error {
...
+					// for example  Schema => email: string @unique @index(exact) .
```
mangalaman93, line 1640: For example, schema =>

edgraph/server.go
```diff
@@ -1573,6 +1613,62 @@ func parseRequest(qc *queryContext) error {
...
+					// To ensure uniqueness of email predicate we will query for value of predicate
```
mangalaman93, line 1641: To ensure uniqueness of email predicate, we will query for the value of predicate to ensure that we are not adding a duplicate value.

edgraph/server.go
```diff
@@ -1573,6 +1613,62 @@ func parseRequest(qc *queryContext) error {
...
+					// for example mutation =>  _:a <email> "example@email.com"  .
+					// then query will be
+					// UID1 as v(func: eq(email,"example@email.com"))
```
mangalaman93, line 1644: `//   __dgraph_uniquecheck_1__ as v(func: eq(email, "example@email.com"))`

mangalaman93: This query will tell us all the UIDs for which store the email value as "example@email.com". We can be sure that we will get at most one UID in return. If the returned UID, in case one is returned, is different than the UID we have in the mutation, then we abort the mutation.

edgraph/server.go
```diff
@@ -1573,6 +1613,62 @@ func parseRequest(qc *queryContext) error {
...
+					qc.uniqueVars[queryVar] = uniqueVarsUids{singlePred.Subject, singlePred.ObjectValue.GetDefaultVal(), singlePred.Predicate}
+				}
+			}
+			if qc.req.Query == "" {
+				buildQuery = "{" + buildQuery + "}"
+				qc.req.Query = buildQuery
+			} else {
+
```
mangalaman93, line 1652: nit: remove line

edgraph/server.go
```diff
@@ -1573,6 +1613,62 @@ func parseRequest(qc *queryContext) error {
...
+					// So here we are adding query for predicate if unique is defined
+					// for example  Schema => email: string @unique @index(exact) .
+					// To ensure uniqueness of email predicate we will query for value of predicate
+					// for example mutation =>  _:a <email> "example@email.com"  .
+					// then query will be
+					// UID1 as v(func: eq(email,"example@email.com"))
+					qc.uniqueVars[queryVar] = uniqueVarsUids{singlePred.Subject, singlePred.ObjectValue.GetDefaultVal(), singlePred.Predicate}
+				}
+			}
+			if qc.req.Query == "" {
```
mangalaman93, line 1648:
```
if len(buildQuery) == 0 {
  return
}
```

edgraph/server.go
```diff
@@ -1573,6 +1613,62 @@ func parseRequest(qc *queryContext) error {
...
+				qc.req.Query = strings.TrimRight(qc.req.Query, "}")
+				buildQuery = qc.req.Query + buildQuery + "}"
+				qc.req.Query = buildQuery
+			}
+		}
+	}
+
+}
+
+func isUnique(mu *dql.Mutation, ctx context.Context) bool {
```
mangalaman93, line 1662: we probably don't need this function

edgraph/server.go
```diff
@@ -109,6 +109,12 @@ type existingGQLSchemaQryResp struct {
 	ExistingGQLSchema []graphQLSchemaNode `json:"ExistingGQLSchema"`
 }
 
+type uniqueVarsUids struct {
```
mangalaman93, line 112: we don't need this type

mangalaman93 (changes requested):

edgraph/server.go
```diff
@@ -1570,6 +1648,73 @@ func parseRequest(qc *queryContext) error {
 	return validateQuery(qc.dqlRes.Query)
 }
 
+// addQueryIfUnique adds dummy queries in the request for checking whether predicate is unique in the db
```
wisdomatom, line 1651: This may works fine when no concurrent mutations or unique value already exist into db. But think about this, there may have two concurrent mutations which insert the same value which not exist yet into db, and below query "__dgraph_uniquecheck_%v__" can't find duplicate value, eventually there would be two same value inserted into db.

This may called MVCC duplicate key problem.

As i think, there may be two resolutions:
1. We first write duplicated values into db, but not commit, and then find way to read uncommited values, then find the unique value conflict and return unique key conflict error.
2. We set up an unique key buffer predicate into db, write and commit duplicate values into buffer predicate before execute mutations, and do unique check before commit and delete unique value from buffer after commit.

edgraph/server.go
```diff
@@ -1570,6 +1648,73 @@ func parseRequest(qc *queryContext) error {
...
+			// 2.  _:a <email> val(queryVariable)  .
+			// In the code below, we construct respective query for the above use-cases viz.
+			//  __dgraph_uniquecheck_1__ as var(func: eq(email,"example@email.com"))
+			//  __dgraph_uniquecheck_1__ as var(func: eq(email,val(queryVariable)))
+			// Each of the above query will check if there is already an existing value.
+			// We can be sure that we may get at most one UID in return.
+			// If the returned UID is different than the UID we have
+			// in the mutation, then we abort the mutation.
+
+			if !strings.HasPrefix(pred.ObjectId, "val(") {
```
mangalaman93, line 1686: if and else condition statements are exactly the same except the value in the query. We should just have an if..else on that.

edgraph/server.go
```diff
@@ -1570,6 +1648,73 @@ func parseRequest(qc *queryContext) error {
...
+	count := 1
+	buildQuery := ""
+	for gmuIndex, gmu := range qc.gmuList {
+		for rdfIndex, pred := range gmu.Set {
+			predSchema, _ := schema.State().Get(ctx, x.NamespaceAttr(pred.Namespace, pred.Predicate))
+			if !predSchema.Unique {
+				continue
+			}
+			queryVar := fmt.Sprintf("__dgraph_uniquecheck_%v__", strconv.Itoa(count))
+			uniqueVarMapKey := strconv.Itoa(gmuIndex) + strconv.Itoa(rdfIndex)
```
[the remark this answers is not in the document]
mangalaman93: One suggestion you could try is fmt.Sprintf("%v_%v", gmuIndex, rdfIndex). But this will have string keys for the map. Ideal would be ingeger keys. Having uint32/uint64 keys are really efficient, and you could achieve that using a pairing function. Read more here https://en.wikipedia.org/wiki/Pairing_function

edgraph/server.go
```diff
@@ -1570,6 +1648,73 @@ func parseRequest(qc *queryContext) error {
...
 
+// addQueryIfUnique adds dummy queries in the request for checking whether predicate is unique in the db
+func addQueryIfUnique(qc *queryContext) {
+	ctx := context.WithValue(context.Background(), schema.IsWrite, false)
+	if len(qc.gmuList) == 0 {
+		return
+	}
+
+	qc.uniqueVars = map[string]string{}
+	count := 1
```
mangalaman93, line 1659: can we use the same thing as we are using for map keys and skip using the count variable?

edgraph/server.go
```diff
@@ -1570,6 +1648,73 @@ func parseRequest(qc *queryContext) error {
...
+				//    __dgraph_uniquecheck_1__ as var(func: eq(email,val(queryVariable)))
+				buildQuery = buildQuery + fmt.Sprintf(`%v as var(func: eq(%v,%v))`,
+					queryVar, pred.Predicate, pred.ObjectId)
+				qc.uniqueVars[uniqueVarMapKey] = queryVar
+
+			}
+			count = count + 1
+
+		}
+		if qc.req.Query == "" && buildQuery != "" {
```
mangalaman93, line 1706: I am wondering if we should start using strings.Builder

edgraph/server.go
```diff
@@ -1907,3 +2052,27 @@ func isDropAll(op *api.Operation) bool {
 	}
 	return false
 }
+
+func parsePredSub(str string) (uint64, error) {
+	switch {
+	case strings.Contains(str, "_:"):
+		return 0, nil
+	case strings.Contains(str, "0x"):
```
mangalaman93, line 2060: isn't there a function in stdlib that could also parse the 0b case etc?

edgraph/server.go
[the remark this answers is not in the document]
mangalaman93: we could also do what we do for matrixes either row wise counting or column wise counting.

edgraph/server.go
```diff
@@ -1512,6 +1519,75 @@ func processQuery(ctx context.Context, qc *queryContext) (*api.Response, error)
...
 
+	for gmuIndex, gmu := range qc.gmuList {
+		for rdfIndex, pred := range gmu.Set {
+			condVar := qc.condVars[gmuIndex]
+			uids, isThere := qc.uidRes[condVar]
+
+			uniqueQueryVar := qc.uniqueVars[strconv.Itoa(gmuIndex)+strconv.Itoa(rdfIndex)]
+			uniqueQueryResult, ok := qr.Vars[uniqueQueryVar]
+
+			if !ok || !(isThere && len(uids) == 1) {
```
mangalaman93, line 1530: I think do it in small steps, instead of a big condition. It would be difficult to read this way.

edgraph/server.go
```diff
@@ -1512,6 +1519,75 @@ func processQuery(ctx context.Context, qc *queryContext) (*api.Response, error)
...
+				continue
+			}
+			isEmpty := func(l *pb.List) bool {
+				return l == nil || len(l.Uids) == 0
+
+			}
+			var subjectUid uint64
+			var predValue interface{}
+
+			// here we are setting subjectUid variable with actual value of uid(v)
```
mangalaman93, line 1540: we need to simplify these conditions, there are too many right now.

mangalaman93 (changes requested):

edgraph/server.go
wisdomatom: Very pleasure to hear that. Infect, we are using @upsert to achieve the unique constrain, it works well except a lot of "transaction abort" error, and we have to keep retrying until success.

Thank you!

systest/unique_test.go
wisdomatom, line 1: I wrote a test case for unique constrain in concurrent condition, but it can't pass. We may need a little more thinking on this.

func ConcurrencyDupValueForDiffBlankNodeTest(t *testing.T, c *dgo.Dgraph) {
	ctx := context.Background()

	op := &api.Operation{Schema: `email: string @unique  @index(exact)  .`}
	require.NoError(t, c.Alter(ctx, op))
	var rdf []string
	concurrency := 100
	for i := 0; i < concurrency; i++ {
		rdf = append(rdf, fmt.Sprintf(`_:%v <email> "example@email.com" .`, i))
	}
	errCh := make(chan error, concurrency)
	wg := &sync.WaitGroup{}
	for _, r := range rdf {
		wg.Add(1)
		go func(r string) {
			defer wg.Done()
			_, err := c.NewTxn().Mutate(ctx, &api.Mutation{
				SetNquads: []byte(r),
				CommitNow: true,
			})
			if err != nil {
				errCh <- err
			}
		}(r)
	}
	wg.Wait()
	close(errCh)
	for err := range errCh {
		// todo: unique constrain was broken, we have to find a way to fix this.
		// unexpected error here, this indicate that multiple values were inserted into db.
		// rpc error: code = Unknown desc = inconsistent data, found multiple nodes with same value forpredicate [email]. uids: [...]
		require.ErrorContains(t, err, "could not insert duplicate value [example@email.com] for predicate [email]")
	}
	const emailQuery = `
	{
		q(func: eq(email, "example@email.com")) {
			email
			}
	}`
	resp, err := c.NewTxn().Query(ctx, emailQuery)
	require.NoError(t, err)
	// there should be only one email data as expected.
	testutil.CompareJSON(t, `{"q":[{"email":"example@email.com"}]}`, string(resp.Json))
}

shivaji-dgraph: Hey @wisdomatom, now this test won't fail. I explicitly added the @upsert directive in the schema if @unique is defined. So, if we try to insert the same values concurrently, then Dgraph will return error  "Transaction has been aborted. Please retry."

systest/unique_test.go
```diff
@@ -0,0 +1,583 @@
...
+)
+
+const emailQuery = `
+{
+	q(func: eq(email, "example@email.com")) {
+		email
+		}
+}`
+
+type dqlSchema struct {
```
mangalaman93, line 45: remove these

systest/unique_test.go
```diff
@@ -0,0 +1,583 @@
+//go:build integration
+
+/*
+ * Copyright 2017-2023 Dgraph Labs, Inc. and Contributors
+ *
```
mangalaman93, line 5: format the queries, name the test a bit better, improve the concurrency test, use dgraphtest

mangalaman93 (changes requested):
We should for now, do not allow adding unique when the predicate already exists. Later on, we can support unique for existing predicate as well.

systest/unique_test.go
wisdomatom: Hey @shivaji-dgraph , this is a good solution for implementing uniqueness constraint, but it also has some issues. For example, when trying to create 100 non-duplicate users simultaneously, some of the user creations may fail and prompt "Transaction Aborted," which is not what we want to see.

As I mentioned earlier, I am using @upsert directive with conditional upsert block to implement uniqueness constraint. However, in order to avoid Transaction Abort, I have to consider other solutions.

mangalaman93: @wisdomatom In this case, I think it makes sense to abort one txn when both are trying to insert the same data. Is there a scenario where this may not be necessary? I am trying to figure out what you'd like to have here ideally speaking.

wisdomatom: @mangalaman93 
When inserting data concurrently, I expect the following:

When concurrently inserting 100 identical email addresses, only 1 or 0 successful insertions should occur, and the rest should fail.
When concurrently inserting 100 different email addresses, all 100 should be successfully inserted without any errors.

Later, I will write several test cases to illustrate the current issue and some of my personal solutions.

mangalaman93: Awesome, that seems like a reasonable expectations to me. Looking forward to the test cases.

wisdomatom: > Awesome, that seems like a reasonable expectations to me. Looking forward to the test cases.

Well, here is my test cases, i test it hundred times, it works fine for now.
Please mind that below test case is running using commit `e35112aafd9124425e7884baacf82c9e26fe0bec`.
I can't set up the test environment using the latest commit version.

```golang
type ConflictTS string

type uniqueDetect struct {
	Uid   string `json:"uid"`
	Key   string `json:"key"`
	Value string `json:"value"`
}

func uniqueConflictDetectInit(t *testing.T, c *dgo.Dgraph) {
	ctx := context.Background()
	op := &api.Operation{Schema: `
_Unique.key: string  @index(hash)  . 
_Unique.value: string @index(hash) . 
`}
	require.NoError(t, c.Alter(ctx, op))
}

func hasUniqueConstrain(p string) bool {
	// todo
	// here we should check if predicate has @unique directive
	// we only have one "email" predicate in test case, so we make it simple
	if p == "email" {
		return true
	}
	return false
}

// uniqueConflictDetect
// this func is used to do conflict detect for unique predicate
// first we write unique predicate values into db with an identify key(timestamp + uuid).
// and then we query all related unique predicate values from db.
// then we judge if there is duplicate values inserted by other concurrency request, and return conflict error or nil.
// returned ConflictTS is used to clear the cache inserted into db, so other request would work well.
func uniqueConflictDetect(nQuadsUnique []string, c *dgo.Dgraph) (ConflictTS, error) {
	var err error
	var ts ConflictTS
	var nQuadsDetect []string
	var conflictKeys []string
	ctx := context.Background()
	ts = ConflictTS(fmt.Sprintf("%v-%v", time.Now().UnixNano(), uuid.New().String()))
	for idx, n := range nQuadsUnique {
		spo := strings.Fields(n)
		p := strings.TrimRight(strings.TrimLeft(spo[1], "<"), ">")
		if !hasUniqueConstrain(p) {
			continue
		}
		o := strings.Trim(spo[2], `"`)
		conflictKey := fmt.Sprintf(`%v-%v`, p, o)
		conflictKeys = append(conflictKeys, conflictKey)
		nQuadsDetect = append(nQuadsDetect,
			fmt.Sprintf(`_:id%v <_Unique.key> "%v" .`, idx, conflictKey),
			fmt.Sprintf(`_:id%v <_Unique.value> "%v" .`, idx, ts),
		)
	}
	if len(nQuadsDetect) == 0 {
		// no unique predicate, we don't need the conflict check
		return ConflictTS(""), nil
	}
	txn := c.NewTxn()
	_, err = txn.Do(ctx, &api.Request{
		Mutations: []*api.Mutation{
			&api.Mutation{
				SetNquads: []byte(strings.Join(nQuadsDetect, "\n")),
				// CommitNow: true,
			},
		},
		// CommitNow: true,
	})
	if err != nil {
		return ts, err
	}
	err = txn.Commit(ctx)
	if err != nil {
		return ts, err
	}
	q := `
{
	q(func: eq(_Unique.key, %v)) {
		uid
		key:_Unique.key
		value:_Unique.value
	}
}
`
	bts, _ := json.Marshal(conflictKeys)
	q = fmt.Sprintf(q, string(bts))
	resp, err := c.NewReadOnlyTxn().Query(ctx, q)
	if err != nil {
		return ts, err
	}
	response := &struct {
		Q []uniqueDetect `json:"q"`
	}{}
	err = json.Unmarshal(resp.Json, response)
	if err != nil {
		return ts, err
	}
	arr := response.Q
	if len(arr) == 0 {
		return ts, fmt.Errorf("unique key empty")
	}
	uniqueMap := map[string][]string{}
	for i, a := range arr {
		_, ok := uniqueMap[a.Key]
		if !ok {
			uniqueMap[a.Key] = []string{a.Value}
			continue
		}
		uniqueMap[a.Key] = append(uniqueMap[a.Key], arr[i].Value)
	}
	for _, k := range conflictKeys {
		v, ok := uniqueMap[k]
		// In normal mode, Dgraph provides strong consistency
		// so len(v) should not be zero, because we just insert is above and committed
		// but in my test cluster, it sometimes returns empty result. this is very rare.
		// we return an error here to avoid that the unique constraints was broken.
		if !ok || len(v) == 0 {
			return ts, fmt.Errorf("unique key empty(%v)", k)
		}
		// there is the conflict detect judge condition
		// if len(v) > 1, this indicates that some other concurrency request inserted dup value, so there should be a conflict.
		// if v[0] != ts, this is to say we find another concurrency request inserted dup value, but not this ts, so we still return a conflict.
		if len(v) > 1 || v[0] != string(ts) {
			return ts, fmt.Errorf("unique key conflict(%v)", k)
		}
	}
	// fmt.Println(">>>>>>>>>", ts, uniqueMap, time.Now())
	return ts, nil
}

// uniqueConflictTSRemove
// this func is used to remove ConflictTS cache data.
// todo: we should setup a cronjob to detect if there has expired ConflictTS data, and if we find that, just delete it.
func uniqueConflictTSRemove(ts ConflictTS, c *dgo.Dgraph) error {
	var err error
	ctx := context.Background()
	_, err = c.NewTxn().Do(ctx, &api.Request{
		CommitNow: true,
		Query:     fmt.Sprintf(`query{var1 as var(func:eq(_Unique.value, "%v"))}`, ts),
		Mutations: []*api.Mutation{
			&api.Mutation{
				CommitNow: true,
				DelNquads: []byte(`
uid(var1) <_Unique.key> * .
uid(var1) <_Unique.value> * .
`),
			},
		},
	})
	return err
}

func ConcurrentDiffBlankNodeInsertTest(t *testing.T, c *dgo.Dgraph) {
	var err error
	uniqueConflictDetectInit(t, c)
	ctx := context.Background()

	op := &api.Operation{Schema: `email: string @unique  @index(exact)  .`}
	require.NoError(t, c.Alter(ctx, op))
	var rdf []string
	concurrency := 200
	for i := 0; i < concurrency; i++ {
		rdf = append(rdf, fmt.Sprintf(`_:id%v <email> "example%03d@email.com" .`, i, i))
	}
	errCh := make(chan error, concurrency)
	wg := &sync.WaitGroup{}
	for _, r := range rdf {
		wg.Add(1)
		go func(r string) {
			defer wg.Done()

			// conflict detect
			ts, err := uniqueConflictDetect([]string{r}, c)
			defer uniqueConflictTSRemove(ts, c)
			if err != nil {
				errCh <- err
				return
			}

			t := c.NewTxn()
			defer t.Discard(ctx)
			_, err = t.Mutate(ctx, &api.Mutation{
				SetNquads: []byte(r),
			})
			if err != nil {
				errCh <- err
				return
			}

			err = t.Commit(ctx)
			if err != nil {
				errCh <- err
				return
			}
		}(r)
	}
	wg.Wait()
	close(errCh)
	for err := range errCh {
		require.NoError(t, err)
	}
	const emailQuery = `
	{
		q(func: has(email), orderasc: email) {
			email
			}
	}`
	resp, err := c.NewTxn().Query(ctx, emailQuery)
	require.NoError(t, err)
	response := &struct {
		Q []struct {
			Email string `json:"email"`
		} `json:"q"`
	}{}
	err = json.Unmarshal(resp.Json, response)
	require.NoError(t, err)
	var values []string
	for i := 0; i < concurrency; i++ {
		values = append(values, fmt.Sprintf(`{"email":"example%03d@email.com"}`, i))
	}
	fmt.Println("created:", len(response.Q))
	testutil.CompareJSON(t, fmt.Sprintf(`{"q":[%v]}`, strings.Join(values, ",")), string(resp.Json))
}

func ConcurrentSameBlankNodeInsertTest(t *testing.T, c *dgo.Dgraph) {
	var err error
	uniqueConflictDetectInit(t, c)
	ctx := context.Background()

	op := &api.Operation{Schema: `email: string @unique  @index(exact)  .`}
	require.NoError(t, c.Alter(ctx, op))
	var rdf []string
	concurrency := 400
	for i := 0; i < concurrency; i++ {
		rdf = append(rdf, fmt.Sprintf(`_:id%v <email> "example@email.com" .`, i))
	}
	errCh := make(chan error, concurrency)
	wg := &sync.WaitGroup{}
	for _, r := range rdf {
		wg.Add(1)
		go func(r string) {
			defer wg.Done()
			// below random sleep is to increse the chances that we insert 1 data successfully into db.
			// without sleep all insert request would be failed.
			time.Sleep(time.Millisecond * time.Duration(rand.Int()%1000))
			// conflict detect
			ts, err := uniqueConflictDetect([]string{r}, c)
			defer func() {
				uniqueConflictTSRemove(ts, c)
				if err == nil {
					fmt.Println("delete cache:", ts, time.Now())
				}
			}()
			if err != nil {
				errCh <- err
				return
			}

			t := c.NewTxn()
			defer t.Discard(ctx)
			_, err = t.Mutate(ctx, &api.Mutation{
				SetNquads: []byte(r),
			})
			if err != nil {
				errCh <- err
				return
			}

			err = t.Commit(ctx)
			if err != nil {
				errCh <- err
				return
			}
		}(r)
	}
	wg.Wait()
	close(errCh)
	for err := range errCh {
		if strings.Contains(err.Error(), "unique key conflict") || 
			strings.Contains(err.Error(), "could not insert duplicate value [example@email.com] for predicate [email]") {
			continue
		}
		t.Error(err)
	}
	const emailQuery = `
	{
		q(func: has(email), orderasc: email) {
			email
			}
	}`
	resp, err := c.NewTxn().Query(ctx, emailQuery)
	require.NoError(t, err)
	if string(resp.Json) == `{"q":[{"email":"example@email.com"}]}` ||
		string(resp.Json) == `{"q":[]}` {
		// created data would be 1 or 0
		fmt.Println("created:", string(resp.Json))
		return
	}
	require.Error(t, fmt.Errorf("duplicate value inserted"), string(resp.Json))
}
```

systest/unique_test.go
```diff
@@ -0,0 +1,551 @@
...
+	_, err := txn1.Mutate(ctx, &api.Mutation{
+		SetNquads: []byte(rdf),
+	})
+	require.NoError(t, err)
+
+	txn2 := dg.NewTxn()
+	_, err = txn2.Mutate(ctx, &api.Mutation{
+		SetNquads: []byte(rdf),
+	})
+
```
mangalaman93, line 504: nit" don't need this newline

systest/unique_test.go
```diff
@@ -0,0 +1,551 @@
...
+	require.NoError(t, err)
+
+	require.NoError(t, txn1.Commit(ctx))
+	err = txn2.Commit(ctx)
```
mangalaman93, line 508: this and the next line could be combined

mangalaman93 (changes requested):

systest/unique_test.go
wisdomatom: Please take a review of the code above, infact i am introducing a method that can help to avoid duplicate insert for unique constrain under concurrency condition, it is not just a test case.

edgraph/server.go
```diff
@@ -1907,3 +2087,51 @@ func isDropAll(op *api.Operation) bool {
...
+					dql.TypeValFrom(pred1.ObjectValue).Value, pred1.Predicate)
+			}
+		}
+	}
+	return nil
+}
+
+// encodeBit two uint32 numbers by bit.
+// First 32 bits store k1 and last 32 bits store k2.
+func encodeIndex(k1, k2 uint32) uint64 {
```
mangalaman93, line 2118: put encodeIndex and decodeIndex close to each other and fix the comments. The comment still says encodeBit.

mangalaman93: I am also thinking that we should take int as an argument and check whether the conversion to uint32 is possible or not.

mangalaman93 (changes requested):
Looks pretty good to me now. 3 very minor comments. Please ask other people to review.

systest/unique_test.go
mangalaman93: Is it possible to give you some intro to the code? It's a lot of code to follow.

edgraph/server.go
```diff
@@ -68,6 +68,11 @@ const (
 
 type GraphqlContextKey int
 
+type uniquePredMeta struct {
+	queryVar string
```
harshil-goel, line 72: Add comment about what these values would hold and why we need them

edgraph/server.go
```diff
@@ -570,6 +575,10 @@ func (s *Server) doMutate(ctx context.Context, qc *queryContext, resp *api.Respo
 		return err
 	}
 
+	if err := verifyUniqueWithinMutation(qc); err != nil {
```
harshil-goel, line 578: Can we maybe update the function name? or Add a comment up here to as to show what we are trying to do? 
What is the behaviour today if the user gave two different mutations for the same predicate on the same uid? Have we handled that here? Is user giving two mutations, for same uid, for same predicate allowed for unique today? 
What I am thinking is this could be a in general function that validates duplicates in the same predicate. Irrespective of unique or not. For unique we could have one more check.

edgraph/server.go
```diff
@@ -823,6 +832,17 @@ func findMutationVars(qc *queryContext) []string {
 		varsList = append(varsList, v)
 	}
 
+	// We use certain variables to check uniqueness. To prevent errors related to unused
+	// variables, we explicitly add them to the 'varsList' since they are not used in the mutation.
+	for _, uniqueQueryVar := range qc.uniqueVars {
```
harshil-goel, line 837: Why aren't we update the mutation and letting dgraph doing it for us? We won't need this code then.

harshil-goel (commented):
I think I will need a meeting to go through the various design decisions.

edgraph/server.go
mangalaman93: We are literally trying to check unique within the mutation. What did you feel missing from the function name?

Why would we want to check for duplicate mutation if the predicate is not marked unique?

edgraph/server.go
mangalaman93: How will we check the data in badger before the mutation if we don't create these dummy variables?

systest/unique_test.go
```diff
@@ -0,0 +1,696 @@
...
+	require.ErrorContains(t, err, "could not insert duplicate value [1234567890] for predicate [mobile]")
+	query := `{
+		q(func: eq(mobile, 1234567890)) {
+			mobile
+		}
+    }`
+	resp, err := dg.Query(query)
+	require.NoError(t, err)
+	require.NoError(t, dgraphtest.CompareJSON(`{ "q": [ {	"mobile": 1234567890}]}`, string(resp.Json)))
+
```
mangalaman93, line 509: nit: extra newline

systest/unique_test.go
```diff
@@ -0,0 +1,696 @@
...
+	})
+	require.NoError(t, err)
+	rdf = `<0x33> <email> "aman@dgraph.io" .
+	       <0x22> <email> "shiva@dgraph.io" .`
+	_, err = dg.Mutate(&api.Mutation{
+		SetNquads: []byte(rdf),
+		CommitNow: true,
+	})
+	require.NoError(t, err)
+	// require.ErrorContains(t, err, "could not insert duplicate value [aman@dgraph.io] for predicate [email]")
```
mangalaman93, line 554: remove this

systest/unique_test.go
wisdomatom: @mangalaman93 
I feel like I understand what you mean. By reviewing the source code, I have an idea. Can we just ignore conflict detection for IdentTrigram type in the GetConflictKey function under upsert condition, like this:
```golang
func GetConflictKey(ctx context.Context, pk x.ParsedKey, key []byte, t *pb.DirectedEdge) uint64 {
	getKey := func(key []byte, uid uint64) uint64 {
		// Instead of creating a string first and then doing a fingerprint, let's do a fingerprint
		// here to save memory allocations.
		// Not entirely sure about effect on collision chances due to this simple XOR with uid.
		return farm.Fingerprint64(key) ^ uid
	}

	var conflictKey uint64
	switch {
	case schema.State().HasNoConflict(t.Attr):
		break
	case schema.State().HasUpsert(t.Attr):
		// if we have an predicate like "email @upsert @unique @index(hash, trigram) ."
		// for example, tom_1@gmail.com and tom_2@gmail.com will produce the same trigram token key like "om_", which create a conflict.
		t := ctx.Value("tokenIdentifier")
		token, ok := t.(uint8)
		if ok && token == tok.IdentTrigram {
			break
		}
		// Consider checking to see if a email id is unique. A user adds:
		// <uid> <email> "email@email.org", and there's a string equal tokenizer
		// and upsert directive on the schema.
		// Then keys are "<email> <uid>" and "<email> email@email.org"
		// The first key won't conflict, because two different UIDs can try to
		// get the same email id. But, the second key would. Thus, we ensure
		// that two users don't set the same email id.
		conflictKey = getKey(key, 0)

	case pk.IsData() && schema.State().IsList(t.Attr):
		// Data keys, irrespective of whether they are UID or values, should be judged based on
		// whether they are lists or not. For UID, t.ValueId = UID. For value, t.ValueId =
		// fingerprint(value) or could be fingerprint(lang) or something else.
		//
		// For singular uid predicate, like partner: uid // no list.
		// a -> b
		// a -> c
		// Run concurrently, only one of them should succeed.
		// But for friend: [uid], both should succeed.
		//
		// Similarly, name: string
		// a -> "x"
		// a -> "y"
		// This should definitely have a conflict.
		// But, if name: [string], then they can both succeed.
		conflictKey = getKey(key, t.ValueId)

	case pk.IsData(): // NOT a list. This case must happen after the above case.
		conflictKey = getKey(key, 0)

	case pk.IsIndex() || pk.IsCountOrCountRev():
		// Index keys are by default of type [uid].
		conflictKey = getKey(key, t.ValueId)

	default:
		// Don't assign a conflictKey.
	}

	return conflictKey
}
```

mangalaman93: @wisdomatom Is it possible for you to file a feature request for this? I get the idea but I am thinking whether this could have any bad consequences too. We can spend more time on it when we can.

harshil-goel (approved):

mangalaman93 (approved):

harshil-goel (approved):

rderbier:
What is the behavior if you add @unique on an existing predicate and deploy the schema?
Do we reject the schema or just ignore the unique directive?

mangalaman93:
we reject the schema and the Alter request should return an error.

PR #8829, opened 2023-05-17
dgraphtest: print container logs if the test fails

mangalaman93:


joshua-goldstein (approved):

.github/workflows/ci-dgraph-upgrade-tests.yml
```diff
@@ -18,7 +18,7 @@ on:
 jobs:
   dgraph-upgrade-and-integration-tests:
     if: github.event.pull_request.draft == false
-    runs-on: ubuntu-20.04
+    runs-on: [self-hosted, x64]
```
anurags92, line 21: why moving to self-hosted? This incurs cost.

anurags92: >@mangalaman93 try runing these jobs on our runners (more resources) 
?

mangalaman93: I think, the tests are failing on GitHub runners. What do you mean by "our runners"?

anurags92 (approved):

.github/workflows/ci-dgraph-upgrade-tests.yml
joshua-goldstein: "our runners" = self-hosted runners
@anurags92 this was my suggestion because there were intermittent failures in the tests that might be due to resource issues.  If we still see the same failures then at least we know this isn't the problem.

PR #8832, opened 2023-05-19
fix failing oss build

### LINKED ISSUE #8831: [BUG]: install_oss broken after commit c175edcaf4

_No response_

### Tell us a little more about your go-environment?

go version go1.20.2 darwin/arm64

Yes

### What is the hardware spec (RAM, CPU, OS)?

Apple M1 Max, 64GB RAM

### What steps will reproduce the bug?

`make install_oss`

### Expected behavior and actual result.

`make install_oss` is expected to succeed.

Instead, it fails because c175edcaf4 introduced a dependency on `github.com/dgraph-io/dgraph/ee/acl`, which is stubbed out in an OSS build.

### Additional information

_No response_

mangalaman93:
Fixes #8831

### DIFF
--- a/.github/workflows/ci-dgraph-tests.yml
+++ b/.github/workflows/ci-dgraph-tests.yml
@@ -81,10 +81,9 @@ jobs:
           export GOPATH=~/go
           # move the binary
           cp dgraph/dgraph ~/go/bin/dgraph
-          # run the tests
+          # run the unit and integration tests
           cd t; ./t
           # clean up docker containers after test execution
           ./t -r
           # sleep
           sleep 5
-          
--- a/ee/acl/run.go
+++ b/ee/acl/run.go
@@ -22,6 +22,7 @@ package acl
 import (
 	"github.com/spf13/cobra"
 
+	"github.com/dgraph-io/dgo/v230/protos/api"
 	"github.com/dgraph-io/dgraph/x"
 )
 
@@ -34,4 +35,15 @@ func init() {
 		Annotations: map[string]string{"group": "security"},
 	}
 	CmdAcl.Cmd.SetHelpTemplate(x.NonRootTemplate)
+}
+
+// CreateGroupNQuads cretes NQuads needed to store a group with the give ID.
+func CreateGroupNQuads(groupId string) []*api.NQuad {
+	return nil
+}
+
+// CreateUserNQuads creates the NQuads needed to store a user with the given ID and
+// password in the ACL system.
+func CreateUserNQuads(userId, password string) []*api.NQuad {
+	return nil
 }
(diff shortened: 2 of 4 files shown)


.github/workflows/ci-dgraph-tests.yml
```diff
@@ -81,10 +81,11 @@ jobs:
           export GOPATH=~/go
           # move the binary
           cp dgraph/dgraph ~/go/bin/dgraph
-          # run the tests
+          # Run oss unit tests
+          go test -timeout=60m -failfast -tags=oss ./...
```
joshua-goldstein, line 85: By adding this line you are running these tests twice (they should already get picked up by `./t`)

joshua-goldstein (changes requested):

.github/workflows/ci-dgraph-tests.yml
joshua-goldstein: If the goal is to make sure that OSS builds don't fail in the future, we should add another (parallel) workflow, so this job doesn't get even longer than it already is.

mangalaman93: These are the same unit tests but now running for OSS build. And these are only unit tests, they only take a few seconds to run. I want to make sure any PR we have in future does not break at least building the OSS release. I agree that we should have a separate workflow, but that may be a bigger effort where we should be able to run t binary tests for OSS build. I.can file a JIRA issue for it.

joshua-goldstein: On my local machine it takes
```
real	0m36.089s
```
to run these units tests, so a nontrivial amount of time IMO.  We are paying to run these machines, so ideally we should aim to reduce the time it takes to run these tests.  It is not a huge effort to create a new workflow to run these tests (on github hosted runners).  Also adding this line is confusing because as I mentioned we are already running these tests when we call `./t`.  I think someone seeing this would be confused.  

Otherwise this PR looks fine, we can merge it and then I can add the workflow.

joshua-goldstein: @mangalaman93 See https://github.com/dgraph-io/dgraph/pull/8834.  Once this is merged I can rebase and merge the other PR.

mangalaman93: cool, that makes sense. removed this change now. Thanks for the review and the PR, this PR is good to go now.

anurags92:
Still fails

joshua-goldstein (approved):

mangalaman93:
> 

I think he just means that the upgrade tests were failing. The test passes now.

anurags92 (approved):

PR #8833, opened 2023-05-20
Add a lambda-acl target

rderbier:
Create a docker image with lambda server and acl enabled. Add a sample lambda files with test functions.

### DIFF
--- a/contrib/local-test/Makefile
+++ b/contrib/local-test/Makefile
@@ -10,7 +10,7 @@ up-with-lambda: local-dgraph-data## Start the zero and alpha containers and a la
 	DGRAPH_VERSION=$(DGRAPH_VERSION) docker-compose -f docker-compose.yml -f docker-compose-lambda.yml up -d && docker-compose logs --follow alpha zero lambda keepalive
 up-with-lambda-acl: local-dgraph-data ## Start the zero and alpha containers and a lambda container
 	mkdir -p ~/local-dgraph-data/acl
-	openssl rand -hex 16 > ~/local-dgraph-data/acl/hmac_secret_file
+	test -f ~/local-dgraph-data/acl/hmac_secret_file || openssl rand -hex 16 > ~/local-dgraph-data/acl/hmac_secret_file
 	DGRAPH_VERSION=$(DGRAPH_VERSION) docker-compose -f docker-compose.yml -f docker-compose-acl.yml -f docker-compose-lambda.yml up -d && docker-compose logs --follow alpha zero lambda keepalive
 
 down: ## Stop the containers
--- a/contrib/local-test/Makefile
+++ b/contrib/local-test/Makefile
@@ -1,12 +1,17 @@
 DGRAPH_VERSION ?= local
 
 current_dir = $(shell pwd)
-
+local-dgraph-data: 
+	mkdir -p ~/local-dgraph-data
 up: ## Start the zero and alpha containers
 	DGRAPH_VERSION=$(DGRAPH_VERSION) docker-compose up -d && docker-compose logs --follow alpha zero keepalive
 
-up-with-lambda: ## Start the zero and alpha containers and a lambda container
+up-with-lambda: local-dgraph-data## Start the zero and alpha containers and a lambda container
 	DGRAPH_VERSION=$(DGRAPH_VERSION) docker-compose -f docker-compose.yml -f docker-compose-lambda.yml up -d && docker-compose logs --follow alpha zero lambda keepalive
+up-with-lambda-acl: local-dgraph-data ## Start the zero and alpha containers and a lambda container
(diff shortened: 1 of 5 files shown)


contrib/local-test/Makefile
```diff
@@ -1,12 +1,17 @@
...
+local-dgraph-data: 
+	mkdir -p ~/local-dgraph-data
 up: ## Start the zero and alpha containers
 	DGRAPH_VERSION=$(DGRAPH_VERSION) docker-compose up -d && docker-compose logs --follow alpha zero keepalive
 
-up-with-lambda: ## Start the zero and alpha containers and a lambda container
+up-with-lambda: local-dgraph-data## Start the zero and alpha containers and a lambda container
 	DGRAPH_VERSION=$(DGRAPH_VERSION) docker-compose -f docker-compose.yml -f docker-compose-lambda.yml up -d && docker-compose logs --follow alpha zero lambda keepalive
+up-with-lambda-acl: local-dgraph-data ## Start the zero and alpha containers and a lambda container
+	mkdir -p ~/local-dgraph-data/acl
```
joshua-goldstein, line 12: this kind of command can run into permission issues, e.g. on my local machine I saw
```
joshua@joshua-local:~/Developer/dgraph/contrib/local-test$ make up-with-lambda-acl
mkdir -p ~/local-dgraph-data
mkdir -p ~/local-dgraph-data/acl
mkdir: cannot create directory ‘/home/joshua/local-dgraph-data/acl’: Permission denied
make: *** [Makefile:12: up-with-lambda-acl] Error 1
```
which was because the docker container initially created the local-dgraph-data folder, so the owner is root:
```
drwxr-xr-x  6 root   root   4.0K May 23 17:40 local-dgraph-data
```
I will leave another comment with one way around this

contrib/local-test/Makefile
```diff
@@ -1,12 +1,17 @@
 DGRAPH_VERSION ?= local
 
 current_dir = $(shell pwd)
-
+local-dgraph-data:
```
joshua-goldstein, line 4: I think this target can be removed, see other comment

contrib/local-test/Makefile
```diff
@@ -1,12 +1,17 @@
...
-
+local-dgraph-data: 
+	mkdir -p ~/local-dgraph-data
 up: ## Start the zero and alpha containers
 	DGRAPH_VERSION=$(DGRAPH_VERSION) docker-compose up -d && docker-compose logs --follow alpha zero keepalive
 
-up-with-lambda: ## Start the zero and alpha containers and a lambda container
+up-with-lambda: local-dgraph-data## Start the zero and alpha containers and a lambda container
 	DGRAPH_VERSION=$(DGRAPH_VERSION) docker-compose -f docker-compose.yml -f docker-compose-lambda.yml up -d && docker-compose logs --follow alpha zero lambda keepalive
+up-with-lambda-acl: local-dgraph-data ## Start the zero and alpha containers and a lambda container
```
joshua-goldstein, line 11: nit: we can add a space above to keep the format consistent

contrib/local-test/Makefile
```diff
@@ -1,12 +1,17 @@
...
+	mkdir -p ~/local-dgraph-data/acl
+	test -f ~/local-dgraph-data/acl/hmac_secret_file || openssl rand -hex 16 > ~/local-dgraph-data/acl/hmac_secret_file
```
joshua-goldstein, line 13: main comment: I think it would be cleaner to avoid this kind of magic and simply check in a `hmac_secret_file` into this directory.  Then we can modify your `docker-compose-acl.yml` as follows:

```
  alpha:
    volumes:
      - ~/local-dgraph-data:/dgraph
      - type: bind
        source: ./hmac_secret_file
        target: /dgraph/acl/hmac_secret_file
        read_only: true

    environment:
      DGRAPH_ALPHA_ACL: secret-file=/dgraph/acl/hmac_secret_file
```

This is so we don't have to generate a new secret every time, and we don't have to worry about whether a directory exists on the local system or not.

joshua-goldstein:
Thanks @rderbier for looking into this, it looks good, I just had one thought above.  Also one question, this is not related to your PR since this also happens on the main branch, but when I run `make up-with-lambda` I see

```
$ make up-with-lambda
DGRAPH_VERSION=local docker-compose -f docker-compose.yml -f docker-compose-lambda.yml up -d && docker-compose logs --follow alpha zero lambda keepalive
[+] Building 0.0s (0/0)                                                                                                                              
[+] Running 6/6
...
no such service: lambda
make: *** [Makefile:9: up-with-lambda] Error 1
```

which seems to come down to the command `docker-compose logs --follow lambda` giving the error.   Are you running into this also?  

Again, not related to your PR, but if it is an issue I can look into it.

PR #8837, opened 2023-05-23
feat(graphql): adds @default directive to graphql (#8017)

mangalaman93:


gregpavl:
is there any update for this feature?

graphql/resolve/mutation_rewriter.go
```diff
@@ -1587,6 +1587,8 @@ func rewriteObject(
 		}
 	}
 
+	action := "update"
```
harshil-goel, line 1590: Make it a constant

shivaji-dgraph: We are changing the value of action based on whether the node is being created or already exists, so we can't make it a constant.

harshil-goel (approved):

PR #8838, opened 2023-05-23
Jassi/multi tenancy all version upgrade

jbhamra1:

### DIFF
--- a/systest/multi-tenancy/basic_test.go
+++ b/systest/multi-tenancy/basic_test.go
@@ -1,4 +1,5 @@
-//go:build integration
+//go:build (!oss && integration) || upgrade
+// +build !oss,integration upgrade
 
 /*
  * Copyright 2023 Dgraph Labs, Inc. and Contributors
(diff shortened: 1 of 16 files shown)


systest/multi-tenancy/basic_test.go
```diff
@@ -210,7 +210,7 @@ func (suite *MultitenancyTestSuite) TestPersistentQuery() {
 	require.NoError(t, err)
 
 	// Upgrade
-	suite.Upgrade(dstDB, dgraphtest.BackupRestore)
+	suite.Upgrade(dgraphtest.BackupRestore)
```
mangalaman93, line 213: do we need to pass strategy in the Upgrade function?

mangalaman93 (changes requested):

systest/multi-tenancy/basic_test.go
jbhamra1: Going forward we will have more than one strategy which we can specify as an argument here.
What are your thoughts ? Hardcode it as BackupRestore ?

jbhamra1: Yes, makes sense, I will change it and push the changes.

PR #8844, opened 2023-05-31
chore: add functions to kill alpha, improve restore check in dgraphtest

mangalaman93:


joshua-goldstein (approved):

all-seeing-code (approved):

dgraphtest/cluster.go
```diff
@@ -306,8 +307,31 @@ loop:
...
 			}
 		}
+		break
+	}
+
+	// sometimes it is possible that one alpha hasn't even started the restore process.
+	// In this case, the alpha will not show "opRestore" in the response to health endpoint
+	// and the "loop" will not be able to detect this case. The "loop2" below ensures
+	// by looking into the logs that the restore process has in fact been started and completed.
+loop2:
```
harshil-goel, line 317: Do we really need loops like this? Cant we do it with just continue statments?

harshil-goel (approved):

dgraphtest/cluster.go
mangalaman93: I can use a flag and write the code. But this is more readable.

PR #7639, opened 2021-03-23
fix(DGRAPH): fix @normalize response when multiple fields at different levels with same alias are selected.

JatinDevDG:
We are generating incorrect order of the fields in a result of the @normalize directive when there are multiple fields with same Alias at nested levels. As a result of that in resulting JSON, the nodes with the same Alias were not converted to list.
For example, if we have below data in dgraph .
```
{
      "_id": "Process",
      "Name": "Process",
      "Extends": {
        "_id": "Core/Namespace",
        "Name": "Namespace"        
      }
}

```
And if we do below query with @recurse and @normalize directive
```
{
  item as var(func:eq(_id,"Process"))
  b2(func:uid(item))@normalize@recurse {
    Name:Name id:_id  Extends
  }
}
```
We will get this response from Dgraph where fields with same alias are not together and while converting response to JSON  , we are not able to combine fields with same Alias in a list.
```
"b2": [
      {
        "Name": "Process",
        "id": "Process",
        "Name":"Namespace",
        "id": "Core/Namespace"
      }
    ]
```
In this PR we fix this in normalize, by putting all fields with the same Alias together. So now response of above query will be 
```
"b2": [
      {
        "Name": [
          "Process",
          "Namespace"
        ],
        "id": [
          "Process",
          "Core/Namespace"
        ]
      }
]
```
This behaviour is broken by PR https://github.com/dgraph-io/dgraph/pull/6362. Prior to this , we used sorting on fields which guarantee that fields with same name or Alias comes together.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/7639)

PR #8845, opened 2023-05-31
feat(query): add feature flag list-in-normalize

mangalaman93:
This feature flag allows configuring @normalize in the query language such that if there are multiple fields with same alias, they could be returned as a list or non-list.
    
There was a breaking change at some point with the PR https://github.com/dgraph-io/dgraph/pull/7639. Some customers do not prefer a list response in case of @normalize. We would like to not make any breaking changes going forward, and hence, introducing this feature flag enables users to make that choice.

### DIFF
--- a/dgraphtest/config.go
+++ b/dgraphtest/config.go
@@ -83,7 +83,8 @@ type ClusterConfig struct {
 	refillInterval time.Duration
 	uidLease       int
 	// exposed port offset for grpc/http port for both alpha/zero
-	portOffset int
+	portOffset   int
+	featureFlags string
 }
 
 func NewClusterConfig() ClusterConfig {
@@ -161,5 +162,10 @@ func (cc ClusterConfig) WithUidLease(uidLease int) ClusterConfig {
 // to fixed ports (port (5080, 6080, 8080 and 9080) + offset + id (0, 1, 2 ...)) on the host.
 func (cc ClusterConfig) WithExposedPortOffset(offset uint64) ClusterConfig {
 	cc.portOffset = int(offset)
+	return cc
+}
+
+func (cc ClusterConfig) WithFeatureFlags(val string) ClusterConfig {
+	cc.featureFlags = val
 	return cc
 }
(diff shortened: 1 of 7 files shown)


joshua-goldstein:
Which future release is this going into, since this is a breaking change?  Will we bring this into v23?

MichelDiz:
Just a question, are you sure about this?
I remember this behavior being a request from the community. I don't record exactly, but there was a discussion about this back then. Do we have a new discussion to revert this?
The previous behavior in my opinion is invalid even by JSON standards. Where two keys cannot have the same name in the same object.

e.g from the PR:
```
"b2": [
      {
        "Name": "Process",
        "id": "Process",
        "Name":"Namespace",
        "id": "Core/Namespace"
      }
    ]
```

```
"b2": [
      {
        "Name": [
          "Process",
          "Namespace"
        ],
        "id": [
          "Process",
          "Core/Namespace"
        ]
      }
]
```

mangalaman93:
Modified the PR to provide a feature flag which is enabled by default. It can be disabled if needed.

MichelDiz (approved):

dgraphtest/config.go
```diff
@@ -163,3 +164,8 @@ func (cc ClusterConfig) WithExposedPortOffset(offset uint64) ClusterConfig {
 	cc.portOffset = int(offset)
 	return cc
 }
+
+func (cc ClusterConfig) WithFeatureFlags(val string) ClusterConfig {
```
harshil-goel, line 168: For now its okay, but it would cause issues as we increase the number of feature flags. Can we do setters and getters for this? I am okay with it being a todo for now as we don't need it just yet

harshil-goel (approved):

dgraphtest/config.go
mangalaman93: oh, I see. This is part of dgraphtest and you are right, we should improve this. But as you said, we can do that whenever we have more flags. This is internal API. Let me see if I can do something now.

harshil-goel (approved):

PR #8860, opened 2023-06-07
chore: remove wiki,static directory completely & a few unused functions in x package

mangalaman93:


joshua-goldstein (approved):

joshua-goldstein (changes requested):
- CI is failing due to removal of some of these magic functions in `x/init.go`
- IMO it does not add much value to remove these functions at the moment, without understanding exactly why they were there in the first place
- Otherwise this looks fine

mangalaman93:
PR is good to go now. Fixed the CI issues, upgrade tests should pass after the other PR is merged.

wiki/content/graphql/admin/index.md
```diff
@@ -1,458 +0,0 @@
-+++
-title = "Admin"
```
harshil-goel, line 2: Why are we removing this?

mangalaman93: I think all docs moved to dgraph-docs. I don't know why this file remains.

harshil-goel (dismissed):

x/init.go
```diff
@@ -42,45 +39,13 @@ var (
 	lastCommitTime string
 )
 
-// SetTestRun sets a variable to indicate that the current execution is a test.
-func SetTestRun() {
```
MichelDiz, line 46: This looks ok to remove.
It is related to /root/dev/dgraph/worker/predicate_test.go at the line 124
The test was commented out by Martin with some TODOs

x/init.go
```diff
@@ -42,45 +39,13 @@ var (
...
 func checkDev() {
 	if dgraphVersion == "" {
 		dgraphVersion = "dev"
 	}
 }
 
-// IsTestRun indicates whether a test is being executed. Useful to handle special
-// conditions during tests that differ from normal execution.
-func IsTestRun() bool {
-	return isTest
```
MichelDiz, line 60: This looks ok to remove.

x/init.go
```diff
@@ -42,45 +39,13 @@ var (
...
-}
-
-// AddInit adds a function to be run in x.Init, which should be called at the
-// beginning of all mains.
-func AddInit(f func()) {
-	initFunc = append(initFunc, f)
-}
-
-// Init initializes flags and run all functions in initFunc.
-func Init() {
```
MichelDiz, line 70: This definitely shouldn't be removed.
If you remove this it will break the Alphas this is called at /root/dev/dgraph/dgraph/cmd/alpha/run.go


Also there is a bunch of tests the relies on this.
ALSO, Ratel depends on `checkDev()` if you remove this it will break any Ratel that is connecting to any cluster that doesn't has a version release. Which means all unreleased binaries, tests and so on.

If we need to change this behavior it should have a plan. As Joshua pointed out.

IMO, all PRs should be single purpose. Small things could be ignored(like identation, or grammar). But a PR shouldn't have more than one Goal. Unless all changes are releated. 

For example, accepting this PR just to remove the WIKI path would be accepted easily, no-brainer. It was supposed to be merged. But these second changes should be made in a new PR. To also be discussed and identified the side effects of removals.

shivaji-dgraph (dismissed):

MichelDiz (changes requested):
I choose to block this as I know it will affect Ratel. This needs to be resolved first. The cluster that doesn't have a release build needs to take "dev" in version key or it will break Ratel.

jbhamra1 (approved):

x/init.go
mangalaman93: Good catch Michel, I have fixed it now.

MichelDiz (approved):

PR #8863, opened 2023-06-09
chore: use `git` cmd tool for checkout a git hash

mangalaman93:
For some reason, when we use the worktree.Checkout, it seems to fail a lot in CI with errors
  test panicked: error while checking out git repo with hash [a77bbe8ae0d42697a38069a9749cfe71c2dafbe6]:
  open /home/ubuntu/actions-runner/_work/dgraph/repo/ee/updatemanifest: no such file or directory
Until we know why that is happening, I [Aman] have moved to use the git command line tool.

This PR also increases the timeout for the upgrade GitHub workflow to 2 hours and frequency of periodic runs to 4 hours.

### DIFF
--- a/dgraphtest/image.go
+++ b/dgraphtest/image.go
@@ -86,13 +86,22 @@ func openDgraphRepo() (*git.Repository, error) {
 }
 
 func checkoutGitRepo(repo *git.Repository, hash *plumbing.Hash) error {
-	worktree, err := repo.Worktree()
-	if err != nil {
-		return errors.Wrap(err, "error while getting git repo work tree")
+	// For some reason, when we use the worktree.Checkout, it seems to fail a lot in CI with errors
+	//   test panicked: error while checking out git repo with hash [a77bbe8ae0d42697a38069a9749cfe71c2dafbe6]:
+	//   open /home/ubuntu/actions-runner/_work/dgraph/repo/ee/updatemanifest: no such file or directory
+	// Until we know why that is happening, I [Aman] have moved to use the git command line tool.
+	cmd := exec.Command("git", "checkout", "-f", hash.String())
+	cmd.Dir = repoDir
+	if out, err := cmd.CombinedOutput(); err != nil {
+		return errors.Wrapf(err, "error while checking out hash [%v]\noutput:%v", hash.String(), string(out))
 	}
-	if err := worktree.Checkout(&git.CheckoutOptions{Hash: *hash, Force: true}); err != nil {
-		return errors.Wrap(err, fmt.Sprintf("error while checking out git repo with hash [%v]", hash.String()))
-	}
+	// worktree, err := repo.Worktree()
+	// if err != nil {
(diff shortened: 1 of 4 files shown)


joshua-goldstein (dismissed):
I think there is a fundamental architecture issue here that is causing problems:
- The workflow is checking out a particular SHA of dgraph, then we do the build of dgraph, then we do tests, but currently it looks like the test is doing another git checkout
- This is inviting a number of issues since the whole point of abstracting checkout as a step in the workflow means we do not have to worry about git, but by using git commands in the test we make this a concern inside the test itself

I realize that this workflow already is doing some kind of checkout inside the test, so I am late to raise the concern, but it is already causing problems (which is not surprising due to the now leaky abstraction / separation between checkout and test steps).  This change is simply inviting more problems.

If the goal is to make specific builds of Dgraph, we can do it in other steps of the workflow, or we can pull binaries from our releases on github.

mangalaman93:
@joshua-goldstein we do not touch the repo cloned by the GitHub workflow. That is crucial to upgrade testing because tests always are run from the main branch of the repo, no matter what version of Dgraph is running.

Now, the way we do it is by cloning repo again and checking out the version that we want to run and building the binary for that version. Now, as you said we are doing it in Go code and there are many other ways we could do this. Important goal that we wanted to achieve in upgrade testing is that, we can use go test command to run the test. Writing another script is always an option, though I think that is a poor option and doesn't provide us the flexibility one gets in through code. We wanted to avoid that. Let me know if you have more questions.

dgraphtest/image.go
```diff
@@ -86,13 +86,22 @@ func openDgraphRepo() (*git.Repository, error) {
...
-	if err != nil {
-		return errors.Wrap(err, "error while getting git repo work tree")
-	}
-	if err := worktree.Checkout(&git.CheckoutOptions{Hash: *hash, Force: true}); err != nil {
-		return errors.Wrap(err, fmt.Sprintf("error while checking out git repo with hash [%v]", hash.String()))
-	}
+	// For some reason, when we use the worktree.Checkout, it seems to fail a lot in CI with errors
+	//   test panicked: error while checking out git repo with hash [a77bbe8ae0d42697a38069a9749cfe71c2dafbe6]:
+	//   open /home/ubuntu/actions-runner/_work/dgraph/repo/ee/updatemanifest: no such file or directory
+	// Until we know why that is happening, I [Aman] have moved to use the git command line tool.
```
harshil-goel, line 92: Just make the comment as we, you don't have to take the responsibility. Maybe add a todo here saying we can upgrade this.

harshil-goel (approved):

dgraphtest/image.go
mangalaman93: It's no big deal. Thanks for the review.

shivaji-dgraph (approved):

jbhamra1 (commented):
Approved

PR #8868, opened 2023-06-09
chore: get rid of go-git dependency and use git cmd tool

mangalaman93:
go-git seems to be creating issue and not worth figuring out why. We have decided instead to just use the git command line tool.


dgraphtest/cluster.go
```diff
@@ -253,14 +253,14 @@ func (hc *HTTPClient) RunGraphqlQuery(params GraphQLParams, admin bool) ([]byte,
 
 // Backup creates a backup of dgraph at a given path
 func (hc *HTTPClient) Backup(c Cluster, forceFull bool, backupPath string) error {
-	// backup API was made async in the commit after this commit d3bf7b7b2786bcb99f02e1641f3b656d0a98f7f4
-	syncAPI, err := IsParent("caed2cb8c68b0ea8c69a3d71410a5fa6b0b94985", c.GetVersion())
+	// backup API was made async in the commit d3bf7b7b2786bcb99f02e1641f3b656d0a98f7f4
+	asyncAPI, err := IsParent("d3bf7b7b2786bcb99f02e1641f3b656d0a98f7f4", c.GetVersion())
```
harshil-goel, line 257: For different branches, the commit id might be different? Can we come up with a better way to check this?

dgraphtest/cluster.go
```diff
@@ -253,14 +253,14 @@ func (hc *HTTPClient) RunGraphqlQuery(params GraphQLParams, admin bool) ([]byte,
...
 	if err != nil {
 		return errors.Wrapf(err, "error checking incremental restore support")
 	}
 
 	var taskPart string
-	if !syncAPI {
+	if asyncAPI {
```
harshil-goel, line 263: did we change the output of IsParent? Seems like we flipped the results?

dgraphtest/cluster.go
mangalaman93: Yeah, we probably can but I am thinking let's improve this function when it stops working for us. Today it seems to work fine. When we see problems with this function, it will be easy to solve it too then.

dgraphtest/cluster.go
mangalaman93: Parent terminology in Git world is actually confusing. Parents are usually the older commits whereas children are usually the newer commit. In our head, we usually think in terms of versions. I am thinking that I should rename this whole thing to compare versions instead of the Git child parent ancestor descendant thing. Let me work on a separate PR to improve this interface.

harshil-goel (approved):

harshil-goel (approved):

shivaji-dgraph (approved):

PR #8872, opened 2023-06-13
refactor(dql): rename feature flag list-in-normalize to normalize-response

mangalaman93:


dgraph/cmd/alpha/run.go
```diff
@@ -263,8 +263,8 @@ they form a Raft group and provide synchronous replication.
 
 	flag.String("feature-flags", worker.FeatureFlagsDefaults, z.NewSuperFlagHelp(worker.FeatureFlagsDefaults).
 		Head("Feature flags to enable various experimental features").
-		Flag("list-in-normalize", "enables returning a list when there are multiple fields with same alias, "+
-			"see here for more details https://github.com/dgraph-io/dgraph/pull/7639").
+		Flag("normalize-response", "chose the kind of response for @normalize when there are multiple fields with same "+
```
rderbier, line 266: Doesn't Flag have an option to indicate what is the default value?

dgraph/cmd/alpha/run.go
```diff
@@ -263,8 +263,8 @@ they form a Raft group and provide synchronous replication.
...
+			"alias, could be `list` or `flat`. For more details, see https://github.com/dgraph-io/dgraph/pull/7639").
```
rderbier, line 267:
```suggestion
		Flag("normalize-response", "configure @normalize formatting. "+
			" `list`:  fields with the same alias are returned as a list; 'flat': returned  a single value for fields  with same alias  (default behavior)").
```

rderbier: Could we be more precise on which single value is returned? Is it the first one, the deepest level, the first level ?
If we can then change "a single value" by the correct description.

rderbier (approved):

dgraph/cmd/alpha/run.go
mangalaman93: Done

dgraph/cmd/alpha/run.go
mangalaman93: We return all of them, it it invalid json but has same key multiple times.

dgraph/cmd/alpha/run.go
mangalaman93:
```
      --feature-flags list         Feature flags to enable various experimental features
                                       normalize-response=list; configure @normalize formatting.  list: (default) fields with the same alias are returned as a list; 'flat': returns a single value for fields with same alias.For more details, see https://github.com/dgraph-io/dgraph/pull/7639
                                    (default "normalize-response=list")
```

PR #8873, opened 2023-06-13
fix(security): remove pull_request_target from coverage workflow

joshua-goldstein:
Description: By using a pull request target, we were exposing our testing environment to malicious actors that could exploit running unverified (unapproved) commands on our runners.  Note that while we could use `pull_request` instead of `pull_request_target`, such a workflow will fail on community PRs (i.e. PR's run from a fork) because the coveralls token (stored as a secret in github) is not accessible to community PR's even after explicit approval when using `pull_request`.  We will have to find a way around this issue.  For now, we resolve the security concern.
Fixes: n/a
Closes: n/a
Docs: n/a


mangalaman93 (approved):

.github/workflows/ci-dgraph-code-coverage.yml
```diff
@@ -22,82 +22,8 @@ on:
...
-  pull_request_target:
-    types:
-      - opened
-      - reopened
-      - synchronize
-      - ready_for_review
-    branches:
-      - main
-      - 'release/**'
-  schedule:
```
sanjayk-github-dev, line 34: What's the reason to remove the scheduled workflow execution?

joshua-goldstein: Code coverage results only change when the code changes (i.e. when code is pushed to main).  Running on a schedule would just send the same code coverage results repeatedly and waste CPU resources (since this is run on our self-hosted runners).

Not sure why this was already here, but presumably we copied it from another workflow and overlooked it.

joshua-goldstein: We could also change it so that is runs once a day on a schedule (instead of 4 times a day, which seems wasteful).

sanjayk-github-dev: Makes sense to only run when the code changes.

sanjayk-github-dev (approved):

nswamy:
>  Note that while we could use `pull_request` instead of `pull_request_target`, such a workflow will fail on community PRs (i.e. PR's run from a fork) 
Do you mean run on their fork or target to the main ? if its the former we don't need to expend resources on that.

.github/workflows/ci-dgraph-code-coverage.yml
nswamy: running once for a PR when someone has reviewed is sufficient IMO

nswamy (commented):
LGTM, thanks for fixing

PR #8877, opened 2023-06-15
Allow Dgraph to run badger in memory

harshil-goel:

### DIFF
--- a/dgraph/cmd/alpha/run.go
+++ b/dgraph/cmd/alpha/run.go
@@ -681,7 +681,7 @@ func run() {
 
 	raft := z.NewSuperFlag(Alpha.Conf.GetString("raft")).MergeAndCheckDefault(worker.RaftDefaults)
 	x.WorkerConfig = x.WorkerOptions{
-		TmpDir:              "./t",
+		TmpDir:              Alpha.Conf.GetString("tmp"),
 		ExportPath:          Alpha.Conf.GetString("export"),
 		ZeroAddr:            strings.Split(Alpha.Conf.GetString("zero"), ","),
 		Raft:                raft,
@@ -698,9 +698,11 @@ func run() {
 		Badger:              bopts,
 	}
 	x.WorkerConfig.Parse(Alpha.Conf)
-	fmt.Println("WORKER CONFIG ============", x.WorkerConfig)
 
+	// We need to set configurations now that we have updated TmpDir and other folders
+	// so that we validate the correct folders.
 	worker.SetConfiguration(&opts)
+
 	if telemetry.GetBool("reports") {
 		go edgraph.PeriodicallyPostTelemetry()
 	}
--- a/dgraph/cmd/alpha/run.go
+++ b/dgraph/cmd/alpha/run.go
@@ -671,8 +671,6 @@ func run() {
 		os.Exit(1)
 	}
 
-	worker.SetConfiguration(&opts)
-
 	ips, err := getIPsFromString(security.GetString("whitelist"))
 	x.Check(err)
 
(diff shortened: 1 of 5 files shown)


dgraph/cmd/alpha/run.go
```diff
@@ -683,7 +681,7 @@ func run() {
 
 	raft := z.NewSuperFlag(Alpha.Conf.GetString("raft")).MergeAndCheckDefault(worker.RaftDefaults)
 	x.WorkerConfig = x.WorkerOptions{
-		TmpDir:              Alpha.Conf.GetString("tmp"),
+		TmpDir:              "./t",
```
mangalaman93, line 684: do we need this?

dgraph/cmd/alpha/run.go
```diff
@@ -700,7 +698,9 @@ func run() {
 		Badger:              bopts,
 	}
 	x.WorkerConfig.Parse(Alpha.Conf)
+	fmt.Println("WORKER CONFIG ============", x.WorkerConfig)
```
mangalaman93, line 701: will this print correctly? Could you copy an example here?

dgraph/cmd/alpha/run.go
```diff
@@ -700,7 +698,9 @@ func run() {
...
 
+	worker.SetConfiguration(&opts)
```
mangalaman93, line 703: can you add a comment to why this needs to be moved down here?

raftwal/log.go
```diff
@@ -114,7 +114,7 @@ func openLogFile(dir string, fid int64) (*logFile, error) {
 			Dir:                           dir,
 			EncryptionKey:                 encryptionKey,
 			EncryptionKeyRotationDuration: 10 * 24 * time.Hour,
-			InMemory:                      false,
+			InMemory:                      true,
```
mangalaman93, line 117: we are doing this permanent?

worker/config.go
```diff
@@ -84,8 +84,6 @@ func (opt *Options) validate() {
 	x.Check(err)
 	td, err := filepath.Abs(x.WorkerConfig.TmpDir)
 	x.Check(err)
-	x.AssertTruef(pd != wd,
```
mangalaman93, line 87: we should add a special condition for in memory.

worker/server_state.go
```diff
@@ -118,19 +119,29 @@ func (s *ServerState) initStorage() {
...
+		}
 		s.WALstore, err = raftwal.InitEncrypted(Config.WALDir, x.WorkerConfig.EncryptionKey)
 		x.Check(err)
 	}
 	{
 		// Postings directory
 		// All the writes to posting store should be synchronous. We use batched writers
 		// for posting lists, so the cost of sync writes is amortized.
-		x.Check(os.MkdirAll(Config.PostingDir, 0700))
+		fmt.Println("Here", Config.PostingDir)
```
mangalaman93, line 132: remove this

mangalaman93 (changes requested):
What are our plans for this PR? Do we want to add a separate option in Dgraph for in memory badger?

dgraph/cmd/alpha/run.go
```diff
@@ -671,8 +671,6 @@ func run() {
 		os.Exit(1)
 	}
 
-	worker.SetConfiguration(&opts)
```
harshil-goel, line 674: Why?

mangalaman93 (changes requested):
We should add a test to ensure that we don't break it in future.

worker/server_state.go
```diff
@@ -118,19 +118,25 @@ func (s *ServerState) initStorage() {
 
 	{
 		// Write Ahead Log directory
-		x.Checkf(os.MkdirAll(Config.WALDir, 0700), "Error while creating WAL dir.")
+		if Config.WALDir != "" {
```
sanjayk-github-dev, line 121: @harshil-goel just trying to understand the changes. 

Is it the case that when badger is run in-memory, WALDir and PostingDir will be empty? And raftwal.InitEncrypted() handles this case properly?

PR #8883, opened 2023-06-23
chore(test): add support for Export Import upgrade strategy

mangalaman93:


dgraphtest/local_cluster.go
```diff
@@ -733,7 +773,7 @@ func (c *LocalCluster) getLogs(containerID string) (string, error) {
 
 	data, err := io.ReadAll(ro)
 	if err != nil {
-		log.Printf("[WARN] error in reading logs for [%v]: %v", containerID, err)
+		log.Printf("[WARNING] error in reading logs for [%v]: %v", containerID, err)
```
harshil-goel, line 776: Is there any way we can make sure that we are consistent with these?

harshil-goel (approved):

dgraphtest/local_cluster.go
mangalaman93: For now, it's all manually done. I do not want to use glog here because I do things before tests are run and glog needs to be initialized. I tried using it, it didn't work well.

shivaji-dgraph (approved):

PR #8897, opened 2023-07-05
chore(test): add workaround for multiple groot issue in export-import

mangalaman93:
When we perform an export and import into the cluster, this ends up creating two groot users in the cluster. This PR implements a workaround so that two groot users are not created to begin with. With this workaround, now we can also run upgrade tests using ExportImport.

The workaround is as follows. We find the UIDs of groot user and guardian group from the new cluster. We modify the output of the export and replace the UIDs of groot user and guardian group with the UIDs we found in the new cluster.

Closes: DGRAPHCORE-300

### DIFF
--- a/dgraphtest/load.go
+++ b/dgraphtest/load.go
@@ -18,8 +18,10 @@ package dgraphtest
 
 import (
 	"archive/tar"
+	"bytes"
 	"compress/gzip"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"log"
@@ -31,16 +33,24 @@ import (
 
 	"github.com/pkg/errors"
 
+	"github.com/dgraph-io/dgraph/ee/enc"
 	"github.com/dgraph-io/dgraph/x"
 )
 
+const (
+	groupOneRdfFile   = "g01.rdf"
+	groupOneRdfGzFile = "g01.rdf.gz"
+)
+
+// LiveOpts are options that are used for running live loader.
 type LiveOpts struct {
 	DataFiles      []string
 	SchemaFiles    []string
 	GqlSchemaFiles []string
 }
 
-func readGzFile(sf string) ([]byte, error) {
+// readGzFile reads the given file from disk completely and returns the content.
+func readGzFile(c *LocalCluster, sf string) ([]byte, error) {
 	fd, err := os.Open(sf)
 	if err != nil {
 		return nil, errors.Wrapf(err, "error opening file [%v]", sf)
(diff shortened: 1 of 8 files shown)


all-seeing-code:
> When we perform an export and import into the cluster, this ends up creating two groot users in the cluster. This PR implements a workaround so that two groot users are not created to begin with. With this workaround, now we can also run upgrade tests using ExportImport
> 
> Closes: DGRAPHCORE-300

What's the work around?

mangalaman93:
> > When we perform an export and import into the cluster, this ends up creating two groot users in the cluster. This PR implements a workaround so that two groot users are not created to begin with. With this workaround, now we can also run upgrade tests using ExportImport
> > Closes: DGRAPHCORE-300
> 
> What's the work around?

it fixes the UID in the export to the UIDs in the current cluster.

dgraphtest/load.go
```diff
@@ -51,77 +61,152 @@ func readGzFile(sf string) ([]byte, error) {
...
 
 	data, err := io.ReadAll(gr)
 	if err != nil {
-		return nil, errors.Wrapf(err, "error reading file content [%v]", sf)
+		return nil, errors.Wrap(err, "error reading data from io.Reader")
 	}
 	return data, nil
 }
 
+func writeGzData(c *LocalCluster, data []byte) (io.Reader, error) {
```
all-seeing-code, line 100: Could the following functions:
`writeGzData`,  `readGzData`, `setGraphQLSchema`, `findGrootAndGuardians` be methods on LocalCluster (eg `func (c *LocalCluster) LiveLoad`) ? 


More broadly any reason to have functions and methods interspersed?

dgraphtest/load.go
```diff
@@ -51,77 +61,152 @@ func readGzFile(sf string) ([]byte, error) {
...
 	if err != nil {
-		return errors.WithStack(err)
+		return errors.Wrap(err, "error creating grpc client")
 	}
 	defer cleanup()
-	if err := gc.LoginIntoNamespace(context.Background(),
-		DefaultUser, DefaultPassword, x.GalaxyNamespace); err != nil {
-		return errors.WithStack(err)
+
+	if c.conf.acl {
```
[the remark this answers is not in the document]
mangalaman93: we were either not using this operation or we were running the cluster with ACL. This case probably never happened.

dgraphtest/local_cluster.go
```diff
@@ -551,7 +551,7 @@ func (c *LocalCluster) Upgrade(version string, strategy UpgradeStrategy) error {
 				return errors.Wrapf(err, "error during login before upgrade")
 			}
 		}
-		if err := hc.Export(DefaultExportDir); err != nil {
+		if err := hc.Export(DefaultExportDir, -1); err != nil {
```
all-seeing-code, line 554: Can you add a comment on -1 as the namespace?

dgraphtest/load.go
all-seeing-code: I'd have personally liked them to be methods to be more readable in some sense, But it is a personal preference tbh.

all-seeing-code (approved):
Add a comment on `-1`. Looks good to me.

shivaji-dgraph (approved):

jbhamra1 (approved):

dgraphtest/local_cluster.go
mangalaman93: I will do it as a separate PR. I have a few other changes that I need to merge too.

PR #8898, opened 2023-07-05
cherry-pick from main

mangalaman93:


joshua-goldstein:
As we discussed there is one commit 
```
ca35c6a70 (HEAD -> release/v23, origin/release/v23) chore(changelog): update changelog in release branch (#8826)
```
on the release branch that is not in main.  This only updated the changelog and no changes to the source.  Ideally this wouldn't be there, so it would be easier to compare the main and release/v23 branches by just looking at the commit names, but I think this is okay and better than trying to rewrite the git history.

joshua-goldstein (approved):
Looks good - this is identical to the main branch as of commit `3de01e42b` (i.e. the latest commit).  Leaving the merge to you when it is ready.

PR #8903, opened 2023-07-12
fix(bulk): enable running bulk loader with only gql schema

mangalaman93:
If only gql schema is provided, and DQL schema is missing bulk loader used to fail. This PR adds the ability to auto generate the DQL schema from the gql schema and use it in the bulk loader.

This PR also adds support for running bulk loader in the dgraphtest package. Additionally, this PR also changes the dgraphtest package to always copy the dgraph binary into a temp folder even when the version of the dgraph used is set to "local".

Closes: DGRAPHCORE-324
Docs PR: NA

### DIFF
--- a/dgraph/cmd/bulk/run.go
+++ b/dgraph/cmd/bulk/run.go
@@ -215,12 +215,16 @@ func run() {
 	fmt.Printf("Encrypted input: %v; Encrypted output: %v\n", opt.Encrypted, opt.EncryptedOut)
 
 	if opt.SchemaFile == "" {
-		fmt.Fprint(os.Stderr, "Schema file must be specified.\n")
-		os.Exit(1)
-	}
-	if !filestore.Exists(opt.SchemaFile) {
-		fmt.Fprintf(os.Stderr, "Schema path(%v) does not exist.\n", opt.SchemaFile)
-		os.Exit(1)
+		// if only graphql schema is provided, we can generate DQL schema from it.
+		if opt.GqlSchemaFile == "" {
+			fmt.Fprint(os.Stderr, "Schema file must be specified.\n")
+			os.Exit(1)
+		}
+	} else {
+		if !filestore.Exists(opt.SchemaFile) {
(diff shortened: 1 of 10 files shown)


all-seeing-code (dismissed):

dgraph/cmd/bulk/run.go
```diff
@@ -215,12 +215,16 @@ func run() {
...
 	if opt.SchemaFile == "" {
-		fmt.Fprint(os.Stderr, "Schema file must be specified.\n")
-		os.Exit(1)
-	}
-	if !filestore.Exists(opt.SchemaFile) {
-		fmt.Fprintf(os.Stderr, "Schema path(%v) does not exist.\n", opt.SchemaFile)
-		os.Exit(1)
+		// if only graphql schema is provided, we can generate DQL schema from it.
+		if opt.GqlSchemaFile == "" {
+			fmt.Fprint(os.Stderr, "Schema file must be specified.\n")
```
harshil-goel, line 220: validate if gqlschema file is also present

systest/integration2/bulk_loader_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"testing"
+	"time"
+
+	"github.com/dgraph-io/dgraph/dgraphtest"
+	"github.com/dgraph-io/dgraph/x"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	gqlSchema = `[{"namespace": 0, "schema": "type Message ` +
```
harshil-goel, line 33: Can you add a test with multiple namespaces

dgraph/cmd/bulk/run.go
mangalaman93: that is what we are validating here.

systest/integration2/bulk_loader_test.go
mangalaman93: good point, I was thinking about it and then forgot. That is why I added multiple shards.

all-seeing-code (approved):

harshil-goel (approved):

Issue #8814, opened 2023-05-01
[FEATURE] ACL accepts externally generated JWT token


rderbier:
Currently ACL can be configured by entering a secret. Internally Dgraph is using a HS256 (symmetric algo) algo to generate JWT token. User have to use Dgraph login to obtain a JWT token (access token). The JWT token contains user id and role. It is used to apply 

We want to be able to accept RS256 externally generated tokens.

PR #8912, opened 2023-07-19
feat(acl): support more JWT algorithms for ACL

rderbier:
Description: Until now, we only supported HS256 algorithm for signing and verifying JWT algorithms. This PR adds support for longer keys (HS384, HS512) as well as asymmetric algorithms such as RS, RSS-PSS, ES for various key and shalengths.
Fixes: https://github.com/dgraph-io/dgraph/issues/8814
Closes: https://dgraph.atlassian.net/browse/DGRAPHCORE-95


dgraph/cmd/alpha/run.go
```diff
@@ -651,6 +651,7 @@ func run() {
 
 	if keys.AclKey != nil {
 		opts.HmacSecret = keys.AclKey
+		opts.UsePublicKey = keys.UsePublicKey
```
mangalaman93, line 654: I am thinking whether it would make more sense to take a more direct approach here. Usually, any JWT supported algorithm will either use a public-private key pair (RSA, ECDSA) or just one key (HMAC). We can use existing `secret-file` flag for HMAC and new flags for asymmetric case such as, "alg" flag for algorithm, "public-key-file" and "private-key-file" for the asymmetric key pair. The default algorithm will be HS256 unless specified. This would be backward compatible too.

edgraph/access_ee_test.go
```diff
@@ -21,17 +21,17 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/dgraph-io/dgraph/ee/acl"
-	"github.com/dgraph-io/dgraph/worker"
+	"github.com/dgraph-io/dgraph/x"
 )
 
-func generateJWT(namespace uint64, userId string, groupIds []string, expiry int64) string {
+func generateJWT(namespace uint64, userId string, groupIds []string, expiry int64, key string) string {
```
mangalaman93, line 27: we should take []byte as an argument directly here and avoid the conversion

ee/keys_ee.go
```diff
@@ -47,6 +47,16 @@ func GetKeys(config *viper.Viper) (*Keys, error) {
 		return nil, fmt.Errorf(
 			"ACL secret key must have length of at least 32 bytes, got %d bytes instead", l)
 	}
+	// read public key after the length test because public key can be longer - no limit
+	aclPublicKeyFile := aclSuperFlag.GetPath(flagAclKeyFile)
```
mangalaman93, line 51: I am a little confused, how are we using the same key as public key and also as private key.

x/jwt_helper.go
```diff
@@ -20,17 +20,31 @@ import (
...
-		if _, ok := token.Method.(*jwt.SigningMethodHMAC); !ok {
-			return nil, errors.Errorf("unexpected signing method: %v",
-				token.Header["alg"])
+		if WorkerConfig.UsePublicKey {
+			if _, ok := token.Method.(*jwt.SigningMethodRSA); !ok {
+				return nil, errors.Errorf("unexpected signing method: %v",
+					token.Header["alg"])
+			}
+
+			key, err := jwt.ParseRSAPublicKeyFromPEM([]byte(WorkerConfig.HmacSecret))
```
mangalaman93, line 35: we should do this parsing just one, not for every JWT token.

dgraph/cmd/alpha/run.go
mangalaman93: I just realized that we can derive the public key from the private key. We can skip taking public key as an argument altogether. It's possible you are already doing that, I am not so sure yet.

edgraph/access_ee_test.go
```diff
@@ -49,17 +49,43 @@ func sliceCompare(x, y []string) bool {
 
 	return true
 }
+func TestPublicKey(t *testing.T) {
+	tokenString := "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE3NzU3OTg5MDQsImdyb3VwcyI6WyI3MDEiLCI3MDIiXSwibmFtZXNwYWNlIjoxMjM0NTY3ODkwLCJ1c2VyaWQiOiJ1c2VyMSJ9.kVWrSQgWatrKToaS3DDxAqIx2QGx_6ubkuyCyTTLF8AjkaHLN2nDa-HhW5G_aGxh4-l_X6gCsyay2wUs9cn2R6VD53AUrEdEtF0davsTX0Q1ar-RRo8ShTHj0ufgLn7zRGC7grbZrNZJi1RtDWGDRKzzt67nG-a2CooyP40s-1I--g9qQJoWu69moWiDH6ac2tLo7-wBRmgtWn-SgXE-nJYVDOUcaNWSQBwEX5Pe5L2OUYiusCxnlBhpSSREGYcnXdx9jPKEhfl9atg9QkJSwGyRrCtGY_6gA8tvSZdB2rd6pgXamoxxBadEr_eURg0OeJAJ8wf_IlpS6boOhKNt5Q"
```
mangalaman93, line 53: we could and maybe we should use openssl to generate these keys on the fly. This way we will make sure that we are always compatible with openssl, which is the industry standard.

x/jwt_helper.go
mangalaman93: Also, I believe this could have different formats. I can find this out if needed. We should support or at least we should know what we do not support/support.

mangalaman93 (changes requested):
Thanks Raphael for doing this. We should also add integration test for public-private key case. Happy to help if you need any.

dgraph/cmd/alpha/run.go
mangalaman93: Done

rderbier:
Note: This PR is a first step to fix #8814.
The really get to the point where we accept an external login, we need to be able to start the alpha with an async algo (RS256 for example) and a public validation key for which we don't have the private key.
That means that login in Dgraph can be done by the Dgraph private key.
But JWT decoding is done using graph public key ( so user going through Dgraph login will be accepted) and the external public key ( so user doing an external login but presenting a JWT we can decode/ verify and in which we have the userid and groups info will also be accepted).

mangalaman93 (approved):

harshil-goel (dismissed):

mangalaman93 (approved):

mangalaman93 (dismissed):

harshil-goel (approved):

mangalaman93 (approved):

PR #8924, opened 2023-07-31
Added lang support

harshil-goel:
Added lang support to GraphQL

### DIFF
--- a/graphql/schema/gqlschema.go
+++ b/graphql/schema/gqlschema.go
@@ -872,7 +872,6 @@ func postGQLValidation(schema *ast.Schema, definitions []string,
 			}
 		}
 	}
-
 	errs = append(errs, applySchemaValidations(schema, definitions)...)
 
 	return errs
@@ -1206,7 +1205,7 @@ func addUnionMemberTypeEnum(schema *ast.Schema, defn *ast.Definition) {
 // it should be present in the addTypeInput as it should not be generated automatically by dgraph
 // but determined by the value of field in the GraphQL service where the type is defined.
 func addInputType(schema *ast.Schema, defn *ast.Definition, providesTypeMap map[string]bool) {
-	field := getFieldsWithoutIDType(schema, defn, providesTypeMap)
+	field := getFieldsWithoutIDType(schema, defn, providesTypeMap, true)
 	if hasExtends(defn) {
 		idField := getIDField(defn, providesTypeMap)
 		field = append(idField, field...)
@@ -1229,7 +1228,8 @@ func addReferenceType(schema *ast.Schema, defn *ast.Definition, providesTypeMap 
 		}
 		flds = append(getIDField(defn, providesTypeMap), getXIDField(defn, providesTypeMap)...)
 	} else {
-		flds = append(getIDField(defn, providesTypeMap), getFieldsWithoutIDType(schema, defn, providesTypeMap)...)
+		flds = append(getIDField(defn, providesTypeMap),
+			getFieldsWithoutIDType(schema, defn, providesTypeMap, true)...)
 	}
 
 	if len(flds) == 1 && (hasID(defn) || hasXID(defn)) {
@@ -1331,7 +1331,7 @@ func addFieldFilters(
 	for _, fld := range defn.Fields {
 		// Filtering and ordering for fields with @custom/@lambda directive is handled by the remote
 		// endpoint.
-		if hasCustomOrLambda(fld) {
+		if hasCustomOrLambda(fld) || isMultiLangField(fld, false) {
 			continue
 		}
 
@@ -1419,7 +1419,7 @@ func addTypeHasFilter(schema *ast.Schema, defn *ast.Definition, providesTypeMap 
 	}
 
 	for _, fld := range defn.Fields {
-		if isID(fld) || hasCustomOrLambda(fld) {
+		if isID(fld) || hasCustomOrLambda(fld) || isMultiLangField(fld, false) {
 			continue
 		}
 		// Ignore Fields with @external directives also excluding those which are present
@@ -1584,7 +1584,7 @@ func addFilterType(schema *ast.Schema, defn *ast.Definition, providesTypeMap map
 	}
 
 	// Has filter makes sense only if there is atleast one non ID field in the defn
-	if len(getFieldsWithoutIDType(schema, defn, providesTypeMap)) > 0 {
+	if len(getFieldsWithoutIDType(schema, defn, providesTypeMap, false)) > 0 {
 		filter.Fields = append(filter.Fields,
 			&ast.FieldDefinition{Name: "has", Type: &ast.Type{Elem: &ast.Type{NamedType: defn.Name + "HasFilter"}}},
 		)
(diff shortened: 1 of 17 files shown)


graphql/schema/gqlschema.go
```diff
@@ -2275,7 +2285,8 @@ func getNonIDFields(schema *ast.Schema, defn *ast.Definition, providesTypeMap ma
 	return append(fldList, pd)
 }
 
-func getFieldsWithoutIDType(schema *ast.Schema, defn *ast.Definition, providesTypeMap map[string]bool) ast.FieldList {
+func getFieldsWithoutIDType(schema *ast.Schema, defn *ast.Definition,
+	providesTypeMap map[string]bool, isAddingInput bool) ast.FieldList {
```
sanjayk-github-dev, line 2289: What does the parameter `isAddingInput` mean?

graphql/schema/gqlschema.go
```diff
@@ -2246,7 +2251,12 @@ func getNonIDFields(schema *ast.Schema, defn *ast.Definition, providesTypeMap ma
 		if hasCustomOrLambda(fld) {
 			continue
 		}
-
+		// We don't include fields in update patch, which corresponds to multiple language tags in dgraph
+		// Example, nameHi_En:  String @dgraph(pred:"Person.name@hi:en")
+		// We don't add above field in update patch because it corresponds to multiple languages
+		if isMultiLangField(fld, true) {
```
mangalaman93, line 2257: did we add a test for this case?

graphql/schema/gqlschema.go
```diff
@@ -2316,6 +2330,24 @@ func getFieldsWithoutIDType(schema *ast.Schema, defn *ast.Definition, providesTy
 	return append(fldList, pd)
 }
 
+// This function check if given gql field has multiple language tags
+func isMultiLangField(fld *ast.FieldDefinition, isMutationInput bool) bool {
+	dgDirective := fld.Directives.ForName(dgraphDirective)
+	if dgDirective != nil {
```
mangalaman93, line 2336: we should return early here if dgDirective is nil, will reduce indentation below

graphql/schema/gqlschema.go
```diff
@@ -2316,6 +2330,24 @@ func getFieldsWithoutIDType(schema *ast.Schema, defn *ast.Definition, providesTy
...
+		pred := dgDirective.Arguments.ForName("pred")
+		if pred != nil {
+			if strings.Contains(pred.Value.Raw, "@") {
+				langs := strings.Split(pred.Value.Raw, "@")[1]
```
mangalaman93, line 2340: what if there is nothing after @. We should use indexing than using strings.Split here

graphql/schema/gqlschema_test.yml
```diff
@@ -2889,6 +2889,50 @@ invalid_schemas:
...
+    ]
+
+  - name: "@id directive not supported on language tag field"
+    input: |
+      type Person  {
+        name: String!
+        nameHi: String! @dgraph(pred:"Person.name@hi") @id
+      }
+    errlist: [
+      { "message": "Type Person; Field nameHi: @id directive not supported on language tag fields",
```
mangalaman93, line 2910: why @id is not supported on language tagged fields?

graphql/schema/rules.go
```diff
@@ -1228,6 +1228,41 @@ func dgraphDirectiveValidation(sch *ast.Schema, typ *ast.Definition, field *ast.
...
+		}
+
+		if field.Directives.ForName(searchDirective) != nil && isMultiLangField(field, false) {
+			errs = append(errs, gqlerror.ErrorPosf(field.Directives.ForName(searchDirective).Position,
+				"Type %s; Field %s: @search directive not applicable"+
+					" on language tag field with multiple languages", typ.Name, field.Name))
+			return errs
+		}
+
+		tags := strings.Split(predArg.Value.Raw, "@")[1]
```
mangalaman93, line 1253: same

graphql/schema/rules.go
```diff
@@ -1228,6 +1228,41 @@ func dgraphDirectiveValidation(sch *ast.Schema, typ *ast.Definition, field *ast.
...
+		if tags == "*" {
+			errs = append(errs, gqlerror.ErrorPosf(dir.Position, "Type %s; Field %s: `*` language tag not"+
+				" supported in GraphQL", typ.Name, field.Name))
+			return errs
+		}
+		if tags == "" {
```
mangalaman93, line 1259: too late to check for this because we have already indexed at element 1 after strings.Split

graphql/schema/schemagen.go
```diff
@@ -725,7 +737,10 @@ func genDgSchema(gqlSch *ast.Schema, definitions []string,
 					sort.Strings(indexes)
 					indexStr = fmt.Sprintf(" @index(%s)", strings.Join(indexes, ", "))
 				}
-				fmt.Fprintf(&preds, "%s: %s%s %s%s.\n", fld.name, f.typ, indexStr, f.upsert,
+				if f.lang {
+					langStr = " @lang"
+				}
+				fmt.Fprintf(&preds, "%s: %s%s%s %s%s.\n", fld.name, f.typ, indexStr, langStr, f.upsert,
 					f.reverse)
```
mangalaman93, line 744: nit: this could in the same line before

mangalaman93 (changes requested):
Looks good, I see tests only for hash and exact index. We should add test for other indexes as well.

graphql/schema/gqlschema_test.yml
harshil-goel: You need to add it on the Person.name instead of Person.name@hi. You need to add one predicate without lang for it to work.

graphql/schema/rules.go
harshil-goel: For string like "abc@", when we do string.split, we return "abc", "". I have added a test case for "abc@ac@bc", which was convered here.

graphql/schema/rules.go
```diff
@@ -1228,6 +1228,48 @@ func dgraphDirectiveValidation(sch *ast.Schema, typ *ast.Definition, field *ast.
...
+			errs = append(errs, gqlerror.ErrorPosf(dir.Position, "Type %s; Field %s: empty language"+
+				" tag not supported", typ.Name, field.Name))
+			return errs
+		}
+		fmt.Println(field.Name)
```
mangalaman93, line 1265: do we need this print?

mangalaman93 (changes requested):
A couple of more comments and I see some old unresolved comments too. Please take a look. Thanks

jairad26 (approved):

graphql/schema/gqlschema.go
```diff
@@ -1905,7 +1910,7 @@ func addAggregationResultType(schema *ast.Schema, defn *ast.Definition, provides
 		}
 
 		// Adds titleMax, titleMin fields for a field of name title.
-		if isOrderable(fld, defn, providesTypeMap) {
+		if isOrderable(fld, defn, providesTypeMap) || isMultiLangField(fld, false) {
```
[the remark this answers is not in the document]
harshil-goel: Yes, fails some tests

graphql/schema/rules.go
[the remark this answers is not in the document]
harshil-goel: removed

graphql/schema/gqlschema.go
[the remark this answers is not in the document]
mangalaman93: why though?

mangalaman93 (approved):

PR #8931, opened 2023-08-04
sec(acl): convert x.Sensitive to string type for authn hash

mangalaman93:
we use a combination of namespace, start timestamp and ACL hmac secret key to make the transaction context more robust. Unfortunately, because the ACL secret is of type x.Sensitive, we ended up using "*****" as the key for computing hash. This commit fixes this issue. This is NOT a breaking change.


all-seeing-code (approved):

all-seeing-code:
Nice catch!

MichelDiz (approved):

PR #8933, opened 2023-08-04
test(upgrade): add upgrade tests for systest/ldbc

jbhamra1:
**Closes:** https://dgraph.atlassian.net/browse/DGRAPHCORE-26


all-seeing-code:
@jbhamra1 This is failing on CI. I also had questions:
1. I was under the impression that we were transitioning away from alpha.yml in favor of using the dgraphtest package to spin up the cluster. Could you please clarify?
2. Would you mind updating the description in the PR? It would be great to have a clearer understanding of the changes being made, the reasons behind them, and so on. I always find it helpful to grasp the overall intention and purpose of the PR in simple terms before diving into the code for a detailed review. Thank you!

PR #8935, opened 2023-08-06
chore(ci): add jepsen test workflow

joshua-goldstein:
Description: Run basic jepsen test on PR's and on cron schedule (daily).
Closes: DGRAPHCORE-275

### DIFF
--- a/.github/workflows/ci-dgraph-jepsen-tests.yml
+++ b/.github/workflows/ci-dgraph-jepsen-tests.yml
@@ -1,0 +1,67 @@
+name: ci-dgraph-jepsen-tests
+on:
+  push:
+    branches:
+      - main
+      - 'release/**'
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
+    branches:
+      - main
+      - 'release/**'
+  schedule:
+    - cron: "0 3 * * *" # every 8hrs
+env:
+    GOPATH: /home/ubuntu/go
+jobs:
+  dgraph-jepsen-tests:
+    if: github.event.pull_request.draft == false
+    runs-on: [self-hosted, x64]
+    steps:
+      - name: Checkout dgraph repo
+        uses: actions/checkout@v3
+      - name: Checkout jepsen repo
+        uses: actions/checkout@v3
+        with:
+          repository: joshua-goldstein/jepsen
+          path: jepsen
+          ref: master
+      - name: Get go version
+        run: |
+          #!/bin/bash
+          GOVERSION=$({ [ -f .go-version ] && cat .go-version; })
+          echo "GOVERSION=$GOVERSION" >> $GITHUB_ENV
+      - name: Set jepsen root
+        run: |
+          #!/bin/bash
+          cd jepsen
+          JEPSEN_ROOT=$(pwd)
+          echo "JEPSEN_ROOT=$JEPSEN_ROOT" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-dgraph-jepsen-tests.yml
```diff
@@ -0,0 +1,67 @@
+name: ci-dgraph-jepsen-tests
```
mangalaman93, line 1: I think we need to simplify this workflow:
1. We should probably run it once or twice in a week
2. Do we know that it passes today?
3. We don't need 3 different steps (cluster up, running the test, cluster down), just one command can do it all
4. We also may not want to run it on every PR, we should just run it periodically

mangalaman93 (changes requested):

mangalaman93:
When I follow these steps, I get the GLIBC errors:
```
STDERR:
./dgraph: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by ./dgraph)
./dgraph: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ./dgraph)
```

.github/workflows/ci-dgraph-jepsen-tests.yml
joshua-goldstein: 1. Agreed
2. Yes, you can see the job completed successfully
3. Not sure the reasoning here?  This is how the jepsen command is set up, and is similar to how we run other jobs
4. Agreed, will modify to only run on schedule

joshua-goldstein:
> When I follow these steps, I get the GLIBC errors:
> 
> ```
> STDERR:
> ./dgraph: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by ./dgraph)
> ./dgraph: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ./dgraph)
> ```

What steps are you using the run the tests?  Are you on ubuntu-22.04?  This looks related to the library linking issue we saw before for folks on 22.04.  We do not see this issue on CI.

sanjayk-github-dev (approved):
LGTM

.github/workflows/ci-dgraph-jepsen-tests.yml
mangalaman93: oh, wow! We still pass jepsen tests 🏖️ 
For 3rd, my reasoning was that when `main.go` handles setting up the cluster and destroying it, what is the benefit we get in separating it into 3 steps? when I run ` ./jepsen --jepsen-root ${{ env.JEPSEN_ROOT }} --web=false --test-all -l 300`, doesn't it set up and destroy the cluster?

Also, would it be a bad idea to keep the time limit to 600, the default?

.github/workflows/ci-dgraph-jepsen-tests.yml
```diff
@@ -0,0 +1,53 @@
...
+          #!/bin/bash
+          cd contrib/jepsen
+          ./jepsen --jepsen-root ${{ env.JEPSEN_ROOT }} --jaeger="" --web=false --test-all -l 300
+      - name: Stop jepsen cluster
+        run: |
+          #!/bin/bash
+          cd contrib/jepsen
+          ./jepsen --jepsen-root ${{ env.JEPSEN_ROOT }} --down-only
+
+
```
mangalaman93, line 53: nit: can remove the empty line

PR #8937, opened 2023-08-08
fix(acl): allow data deletion for non-reserved predicates

jbhamra1:
Description: when data (non-reserved predicates) is added on UIDs that belong to groot user or guardian group, it is allowed. But when the same data is deleted, that is not allowed. This PR allows deletion of non-reserved predicates on special UIDs.
Closes: https://dgraph.atlassian.net/browse/DGRAPHCORE-355
Docs: NA

### DIFF
--- a/query/mutation.go
+++ b/query/mutation.go
@@ -286,6 +286,9 @@ func checkIfDeletingAclOperation(ctx context.Context, edges []*pb.DirectedEdge) 
 
 	isDeleteAclOperation := false
 	for _, edge := range edges {
+		if !x.IsReservedPredicate(edge.Attr) {
+			continue
+		}
 		// Disallow deleting of guardians group
 		if edge.Entity == guardianUid && edge.Op == pb.DirectedEdge_DEL {
 			isDeleteAclOperation = true
@@ -298,7 +301,7 @@ func checkIfDeletingAclOperation(ctx context.Context, edges []*pb.DirectedEdge) 
 		}
 	}
 	if isDeleteAclOperation {
-		return errors.Errorf("Properties of guardians group and groot user cannot be deleted.")
+		return errors.Errorf("Reserved properties of guardians group and groot user cannot be deleted.")
 	}
 	return nil
 }
--- a/query/mutation_test.go
+++ b/query/mutation_test.go
@@ -20,6 +20,9 @@ package query
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
+	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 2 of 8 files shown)


query/mutation.go
```diff
@@ -286,6 +286,15 @@ func checkIfDeletingAclOperation(ctx context.Context, edges []*pb.DirectedEdge)
 
 	isDeleteAclOperation := false
 	for _, edge := range edges {
+		unReservedPredicate := false
```
mangalaman93, line 289: we should only look at edge.Predicate here

mangalaman93 (changes requested):

query/mutation.go
jbhamra1: Predicate is not a member of DirectedEdge{}

mangalaman93 (commented):
The change LGTM. We need to wait for the CI to pass. If it does, we should a few test cases, and then this PR would be ready to be merged.

query/mutation_test.go
```diff
@@ -70,3 +73,22 @@ func TestAlteringReservedTypesAndPredicatesShouldFail(t *testing.T) {
...
 }
+
+func TestUnreservedPredicateForDeletion(t *testing.T) {
+	dg, cleanup, err := dc.Client()
+	require.NoError(t, err)
+	defer cleanup()
+	ctx := context.Background()
+	require.NoError(t, dg.LoginIntoNamespace(ctx, dgraphtest.DefaultUser, dgraphtest.DefaultPassword, x.GalaxyNamespace))
+	namespace, err := x.ExtractNamespace(ctx)
+	grootsUid, ok := x.GrootUid.Load(namespace)
```
mangalaman93, line 84: we need to use a query here instead to figure out groot's UID

mangalaman93 (dismissed):

query/mutation_test.go
```diff
@@ -70,3 +76,48 @@ func TestAlteringReservedTypesAndPredicatesShouldFail(t *testing.T) {
...
+
+	resp, err := dg.Query(grootUserQuery)
+	require.NoError(t, err, "groot user query failed")
+
+	var userResp userQryResp
+	if err := json.Unmarshal(resp.GetJson(), &userResp); err != nil {
+		t.Fatal("Couldn't unmarshal response from groot user query")
+	}
+	grootsUidStr := userResp.GrootUser[0].Uid
+	re := regexp.MustCompile("^0[xX]")
```
mangalaman93, line 111: you don't need to do this. You need to parse this as a hex string.

query/mutation_test.go
```diff
@@ -70,3 +76,48 @@ func TestAlteringReservedTypesAndPredicatesShouldFail(t *testing.T) {
...
+	newGrootsUidStr := re.ReplaceAllString(grootsUidStr, "")
+	require.Greater(t, len(newGrootsUidStr), 0)
+	grootsUidInt, err := strconv.ParseUint(newGrootsUidStr, 16, 64)
+	require.NoError(t, err)
+	require.Greater(t, uint64(grootsUidInt), uint64(0))
+
+	err = addTriplesToCluster(fmt.Sprintf(`<%s> <dgraphcore_355> "Betelgeuse" .`, grootsUidStr))
+	if err != nil {
```
mangalaman93, line 119: use require.NoError
no panic, same below

mangalaman93 (changes requested):

mangalaman93 (approved):

harshil-goel (approved):

PR #8938, opened 2023-08-08
fix(snapshot): Enable snapshot streaming after bulkload

all-seeing-code:
**Description:** Enable snapshot streaming after bulkload

**Summary:** Previously, the bulkloaded `p` directory couldn't stream to a new alpha due to a commitTs of zero. In this PR, the commitTs is sourced from the `p` directory, allowing the alpha to create a snapshot and subsequently stream it to another alpha.

**Tests:**
- **TestBulkLoaderSnapshotPDirinAlpha0**: Load a `p` directory using bulkload. Then, initiate one alpha using the bulkloaded `p` directory and start a second alpha without any `p` directory. Query both alphas to ensure the snapshot has been successfully generated and that the data is accessible from both instances.
- **TestBulkLoaderSnapshotPDirinAll**: Load a `p` directory using bulkload. Then, copy bulkloaded `p` directory in all alphas and start the cluster. Query all alphas to ensure that the data is accessible from all instances.
- **TestBulkLoaderDataLoss**: Move the `zero` timestamp and load a `p` directory using bulkload. Then use this `p` directory on a fresh cluster (both zero and alpha are new). Validate that the query doesn't work without moving the timestamp of the new zero.

**Closes:** https://dgraph.atlassian.net/browse/DGRAPHCORE-214

**Docs:** NA

### DIFF
--- a/dgraphtest/load.go
+++ b/dgraphtest/load.go
@@ -493,3 +493,28 @@ func (c *LocalCluster) BulkLoad(opts BulkOpts) error {
 		return nil
 	}
 }
+
+func (c *LocalCluster) CopyBulkLoadDirsToAlphaMounts() error {
+	var src, target string
+	for _, alpha := range c.alphas {
+		if alpha.id == 0 {
+			// No need to copy
+			src = filepath.Join(c.conf.bulkOutDir, strconv.Itoa(alpha.id), "p")
+			continue
+		}
+		target = filepath.Join(filepath.Dir(c.conf.bulkOutDir), "copy", strconv.Itoa(alpha.id))
+		if err := copyFolder(src, target); err != nil {
+			return err
+		}
+		log.Printf("[INFO] copied bulk load dir from [%v] to [%v]", src, target)
+	}
+	return nil
+}
+
+func copyFolder(src, target string) error {
+	cmd := exec.Command("cp", "-r", src, target)
+	if out, err := cmd.CombinedOutput(); err != nil {
+		return errors.Wrapf(err, "error copying folder: %v", string(out))
+	}
+	return nil
+}
--- a/dgraphtest/local_cluster.go
+++ b/dgraphtest/local_cluster.go
@@ -343,6 +343,28 @@ func (c *LocalCluster) killContainer(dc dnode) error {
 	return nil
 }
 
+func (c *LocalCluster) HealthCheckAlpha(id int) error {
+	if id >= c.conf.numAlphas {
+		return fmt.Errorf("invalid id of alpha: %v", id)
+	}
+	log.Printf("[INFO] checking health of alpha-[%v]", id)
+	aa := c.alphas[id]
+	url, err := aa.healthURL(c)
+	log.Printf("[INFO] checking health of alpha-[%v] at: [%s]", id, url)
+	if err != nil {
+		return errors.Wrap(err, "error getting health URL")
+	}
+	if err := c.containerHealthCheckFor(id, url); err != nil {
+		return err
+	}
+	log.Printf("[INFO] container [%v] passed health check", aa.containerName)
+
+	if err := c.checkDgraphVersion(aa.containerName); err != nil {
+		return err
+	}
+	return nil
+}
+
 func (c *LocalCluster) HealthCheck(zeroOnly bool) error {
 	log.Printf("[INFO] checking health of containers")
 	for _, zo := range c.zeros {
@@ -378,6 +400,56 @@ func (c *LocalCluster) HealthCheck(zeroOnly bool) error {
 		}
 	}
 	return nil
+}
+
+func (c *LocalCluster) containerHealthCheckFor(alphaId int, url string) error {
+	for i := 0; i < 60; i++ {
+		time.Sleep(waitDurBeforeRetry)
+
+		req, err := http.NewRequest(http.MethodGet, url, nil)
+		if err != nil {
+			log.Printf("[WARNING] error building req for endpoint [%v], err: [%v]", url, err)
+			continue
+		}
+		body, err := doReq(req)
+		if err != nil {
+			log.Printf("[WARNING] error hitting health endpoint [%v], err: [%v]", url, err)
+			continue
+		}
+		resp := string(body)
+
+		// For Alpha, we always run alpha with EE features enabled
+		if !strings.Contains(resp, `"ee_features"`) {
+			continue
+		}
+		if !c.conf.acl {
+			return nil
+		}
+		if !c.conf.acl || !strings.Contains(resp, `"acl"`) {
+			continue
+		}
+
+		client, cleanup, err := c.ClientForAlpha(alphaId)
+		if err != nil {
+			return errors.Wrap(err, "error setting up a client")
+		}
+		defer cleanup()
+
+		ctx, cancel := context.WithTimeout(context.Background(), requestTimeout)
+		defer cancel()
+		for i := 0; i < 10; i++ {
+			if err = client.Login(ctx, DefaultUser, DefaultPassword); err == nil {
+				break
+			}
+			log.Printf("[WARNING] error trying to login: %v", err)
+			time.Sleep(waitDurBeforeRetry)
+		}
+		if err != nil {
+			return errors.Wrap(err, "error during login")
+		}
+		return nil
+	}
+	return fmt.Errorf("health failed, cluster took too long to come up [%v]", url)
 }
 
 func (c *LocalCluster) containerHealthCheck(url string) error {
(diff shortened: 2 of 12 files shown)


worker/draft.go
```diff
@@ -1837,6 +1853,11 @@ func (n *node) InitAndStartNode() {
 			n.SetRaft(raft.StartNode(n.Cfg, peers))
 			// Trigger election, so this node can become the leader of this single-node cluster.
 			n.canCampaign = true
+			// Also trigger a snapshot so that this node can take a snapshot if required
+			// Need a delay otherwise it interfers with starting of Raft loop
+			time.AfterFunc(1*time.Second, func() {
```
kevinmingtarja, line 1858: Can i ask how it might interfere with the starting of Raft loop?

dgraphtest/local_cluster.go
```diff
@@ -276,7 +276,11 @@ func (c *LocalCluster) StartAlpha(id int) error {
 	if id >= c.conf.numAlphas {
 		return fmt.Errorf("invalid id of alpha: %v", id)
 	}
-	return c.startContainer(c.alphas[id])
+	err := c.startContainer(c.alphas[id])
```
mangalaman93, line 279: this could go inside the if

dgraphtest/local_cluster.go
```diff
@@ -343,6 +347,28 @@ func (c *LocalCluster) killContainer(dc dnode) error {
 	return nil
 }
 
+func (c *LocalCluster) HealthCheckAlpha(id int) error {
```
mangalaman93, line 350: The way we should do this is, We should modify the `StartAlpha` and `StartZero` function to also perform the health check before returning. We should refactor the `HealthCheck` function on `*LocalCluster` type to be able to do the health check on a given alpha or zero. We should use the `dnode` interface as an argument to the health check function because it can call healthURL function independent of whether it is a zero or alpha.

dgraphtest/local_cluster.go
```diff
@@ -626,6 +702,32 @@ func (c *LocalCluster) Client() (*GrpcClient, func(), error) {
 	return &GrpcClient{Dgraph: client}, cleanup, nil
 }
 
+// Client returns a grpc client that can talk to any Alpha in the cluster
+func (c *LocalCluster) ClientForAlpha(id int) (*GrpcClient, func(), error) {
```
mangalaman93, line 706: We should define this function such that existing (*LocalCluster).Client function can use it too. Feels duplicated code.

systest/integration2/bulk_loader_snapshot_test.go
```diff
@@ -0,0 +1,138 @@
...
+package main
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/dgraph-io/dgraph/dgraphtest"
```
mangalaman93, line 27: separate imports, first non-dgraph imports and then, dgraph internal imports.

systest/integration2/bulk_loader_snapshot_test.go
```diff
@@ -0,0 +1,138 @@
...
+	require.NoError(t, c.StartAlpha(1))
+
+	// get gRPC client
+	gc2, cleanup2, err := c.ClientForAlpha(1)
+	require.NoError(t, err)
+	defer cleanup2()
+
+	// run some queries and ensure everything looks good
+	queryAlphaWith(t, query, expectedResp, gc2)
+
```
mangalaman93, line 137: what happened to Alpha 3, we created it but never started it.

systest/integration2/bulk_loader_test.go
```diff
@@ -72,7 +72,7 @@ const (
 )
 
 func TestBulkLoaderNoDqlSchema(t *testing.T) {
-	conf := dgraphtest.NewClusterConfig().WithNumAlphas(2).WithNumZeros(1).
+	conf := dgraphtest.NewClusterConfig().WithNumAlphas(3).WithNumZeros(1).
```
mangalaman93, line 75: curious why we needed to make this change?

worker/draft.go
mangalaman93: yeah, and we should avoid it. This will turn into a bad idea, soon enough.

worker/draft.go
```diff
@@ -1048,80 +1048,89 @@ func (n *node) updateRaftProgress() error {
...
+		// snapshotting.  We just need to do enough, so that we don't have a huge backlog of
+		// entries to process on a restart.
+		if calculate {
+			// We can set discardN argument to zero, because we already know that calculate
+			// would be true if either we absolutely needed to calculate the snapshot,
+			// or our checkpoint already crossed the SnapshotAfter threshold.
+			if err := n.proposeSnapshot(); err != nil {
+				glog.Errorf("While calculating and proposing snapshot: %v", err)
+			} else {
+				lastSnapshotTs = time.Now()
```
mangalaman93, line 1111: it used to be `lastSnapshotTime` in the earlier code.

worker/draft.go
```diff
@@ -1048,80 +1048,89 @@ func (n *node) updateRaftProgress() error {
 	return nil
 }
 
-func (n *node) checkpointAndClose(done chan struct{}) {
-	slowTicker := time.NewTicker(time.Minute)
-	lastSnapshotTime := time.Now()
-	defer slowTicker.Stop()
-
+func (n *node) takeSnapshot(lastSnapshotTs time.Time) (time.Time, error) {
+	glog.V(2).Info("Will attempt to do a snapshot now.")
```
mangalaman93, line 1052: "take a snapshot"

worker/draft.go
```diff
@@ -1048,80 +1048,89 @@ func (n *node) updateRaftProgress() error {
...
+			// or our checkpoint already crossed the SnapshotAfter threshold.
+			if err := n.proposeSnapshot(); err != nil {
+				glog.Errorf("While calculating and proposing snapshot: %v", err)
+			} else {
+				lastSnapshotTs = time.Now()
 			}
+		}
+		go n.abortOldTransactions()
+	}
+	return lastSnapshotTs, nil
```
mangalaman93, line 1117: nit: extra new line

worker/draft.go
```diff
@@ -1048,80 +1048,89 @@ func (n *node) updateRaftProgress() error {
...
-							// Save some cycles by only calculating snapshot if the checkpoint
-							// has gone quite a bit further than the first index.
-							calculate = calculate || chk >= first+snapshotAfterEntries
-							glog.V(3).Infof("Evaluating snapshot first:%d chk:%d (chk-first:%d) "+
-								"snapshotAfterEntries:%d snap:%v", first, chk, chk-first,
-								snapshotAfterEntries, calculate)
-						}
-					}
-				}
+func (n *node) checkpointAndClose(done chan struct{}) {
```
mangalaman93, line 1120: put this function before takeSnapshot, the diff will be slightly easier to review

worker/draft.go
```diff
@@ -1048,80 +1048,89 @@ func (n *node) updateRaftProgress() error {
...
+	slowTicker := time.NewTicker(time.Minute)
+	lastSnapshotTs := time.Now()
+	var err error
```
mangalaman93, line 1123: why do we need to declare it here?

mangalaman93 (changes requested):
We need more tests as we discussed, a few questions and minor comments are dgraphtest package

worker/draft.go
all-seeing-code: When the node initializes, it begins in a follower state, even in a single alpha cluster scenario. This pause ensures that the leader election process occurs, allowing the node to acquire leadership. Without this pause, the node tries to capture a snapshot but fails the check in [n.AmLeader()](https://github.com/dgraph-io/dgraph/blob/13e2f77075407437f0a7cb06df81f866644d28d2/worker/draft.go#L1062), resulting in no snapshot being taken.

worker/draft.go
all-seeing-code: Because of this just below it:
```
case <-slowTicker.C:
			// Do these operations asynchronously away from the main Run loop to allow heartbeats to
			// be sent on time. Otherwise, followers would just keep running elections.
			lastSnapshotTs, err = n.takeSnapshot(lastSnapshotTs)
			if err != nil {
				continue
			}
```

dgraphtest/local_cluster.go
all-seeing-code: which if?

dgraphtest/local_cluster.go
all-seeing-code: Done

dgraphtest/local_cluster.go
all-seeing-code: Defined one for `conn`, clients are little high level to be able to use one for the other. Let me know if the modified piece is an ok abstraction.

dgraphtest/local_cluster.go
all-seeing-code: oh, got it.

all-seeing-code:
> We need more tests as we discussed, a few questions and minor comments are dgraphtest package

I have added more tests and extended the dgraphtest package to support some of the functions required for the new tests. This PR is ready for another round of reviews.

dgraphtest/load.go
```diff
@@ -493,3 +493,28 @@ func (c *LocalCluster) BulkLoad(opts BulkOpts) error {
 		return nil
 	}
 }
+
+func (c *LocalCluster) CopyBulkLoadDirsToAlphaMounts() error {
```
mangalaman93, line 497: why are we doing the copy when we can just let the data stream?

mangalaman93 (changes requested):
As discussed, we need to think about more test cases. A few minor comments and questions.

dgraphtest/load.go
all-seeing-code: This is for the test when we want to copy `p` directories in each of the alphas in the cluster and start them. This is different from the case when we let the data stream from one alpha to all the other alpha (covered in: TestBulkLoaderSnapshotPDirinAlpha0)

PR #8941, opened 2023-08-09
perf(query): Improve Intersection with UID pack

harshil-goel:
This PR improves the performance of IntersectCompressedWithBin. LDBC09 Query went from 38 second -> 22 second. Benchmarks show up to 80% improvement in performance.

### DIFF
--- a/algo/uidlist.go
+++ b/algo/uidlist.go
@@ -112,9 +112,9 @@ func IntersectCompressedWithBin(dec *codec.Decoder, q []uint64, o *[]uint64) {
 				break
 			}
 			IntersectWithBin(blockUids, q, o)
-			last_uid := blockUids[len(blockUids)-1]
+			lastUid := blockUids[len(blockUids)-1]
 			qidx := sort.Search(len(q), func(idx int) bool {
-				return q[idx] >= last_uid
+				return q[idx] >= lastUid
 			})
 			if qidx >= len(q) {
 				return
--- a/algo/uidlist.go
+++ b/algo/uidlist.go
@@ -105,34 +105,43 @@ func IntersectCompressedWithBin(dec *codec.Decoder, q []uint64, o *[]uint64) {
 
 	// Pick the shorter list and do binary search
 	if ld < lq {
-		uids := dec.Uids()
-		for len(uids) > 0 {
-			for _, u := range uids {
-				qidx := sort.Search(len(q), func(idx int) bool {
-					return q[idx] >= u
-				})
-				if qidx >= len(q) {
-					return
-				}
-				if q[qidx] == u {
-					*o = append(*o, u)
-					qidx++
-				}
-				q = q[qidx:]
+		last_q := q[len(q)-1]
+		uids := make([]uint64, 0)
+		for {
+			n := dec.Uids()
+			if len(n) == 0 {
+				break
+			}
+			uids = append(uids, n...)
+			if uids[len(uids)-1] > last_q {
+				break
 			}
 			uids = dec.Next()
 		}
+		IntersectWithBin(uids, q, o)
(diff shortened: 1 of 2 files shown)


algo/uidlist.go
```diff
@@ -105,34 +105,43 @@ func IntersectCompressedWithBin(dec *codec.Decoder, q []uint64, o *[]uint64) {
...
-				})
-				if qidx >= len(q) {
-					return
-				}
-				if q[qidx] == u {
-					*o = append(*o, u)
-					qidx++
-				}
-				q = q[qidx:]
+		last_q := q[len(q)-1]
```
mangalaman93, line 108: We should either use lq everywhere or len(q) everywhere.

algo/uidlist.go
```diff
@@ -105,34 +105,43 @@ func IntersectCompressedWithBin(dec *codec.Decoder, q []uint64, o *[]uint64) {
...
+		uids := make([]uint64, 0)
+		for {
+			n := dec.Uids()
```
mangalaman93, line 111: choose a better name than `n`. For a second, I thought `n` was number of UIDs.

algo/uidlist.go
```diff
@@ -105,34 +105,43 @@ func IntersectCompressedWithBin(dec *codec.Decoder, q []uint64, o *[]uint64) {
...
+			if len(n) == 0 {
+				break
+			}
+			uids = append(uids, n...)
```
mangalaman93, line 115: We should avoid creating a big list here, that is not necessary. We could just keep doing intersection with one block at a time.

mangalaman93 (changes requested):
Please add unit tests for these functions too

algo/uidlist.go
```diff
@@ -105,34 +105,45 @@ func IntersectCompressedWithBin(dec *codec.Decoder, q []uint64, o *[]uint64) {
...
-			uids = dec.Next()
+			IntersectWithBin(block_uids, q, o)
+			last_uid := block_uids[len(block_uids)-1]
+			qidx := sort.Search(len(q), func(idx int) bool {
+				return q[idx] >= last_uid
+			})
+			if qidx >= len(q) {
+				return
+			}
+			q = q[qidx:]
```
mangalaman93, line 121: This is same as what you were doing earlier. The slice will keep increasing. I was hoping to avoid that.

all-seeing-code (dismissed):

algo/uidlist.go
```diff
@@ -105,34 +105,45 @@ func IntersectCompressedWithBin(dec *codec.Decoder, q []uint64, o *[]uint64) {
...
+			dec.Next()
 		}
 		return
 	}
 
+	uids := dec.Seek(q[0], codec.SeekStart)
```
mangalaman93, line 127: nit: we don't actually need it

algo/uidlist.go
```diff
@@ -105,34 +105,45 @@ func IntersectCompressedWithBin(dec *codec.Decoder, q []uint64, o *[]uint64) {
...
-				if qidx >= len(q) {
-					return
-				}
-				if q[qidx] == u {
-					*o = append(*o, u)
-					qidx++
-				}
-				q = q[qidx:]
+		for {
+			block_uids := dec.Uids()
```
mangalaman93, line 109: Go doesn't prefer underscore in vars.

mangalaman93 (changes requested):

mangalaman93 (approved):
Great work @harshil-goel. This should improve our query performance too.

all-seeing-code (approved):
Lgtm

joshua-goldstein (approved):
Looks good :+1:

PR #8951, opened 2023-08-15
chore(build): name codename configurable

joshua-goldstein:
Description: Currently `codename` ldflag` is not mutable and set to `dgraph`.  We would like to make it configurable for cloud builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@
 #
 
 BUILD          ?= $(shell git rev-parse --short HEAD)
-BUILD_CODENAME  = dgraph
+BUILD_CODENAME ?= dgraph
 BUILD_DATE     ?= $(shell git log -1 --format=%ci)
 BUILD_BRANCH   ?= $(shell git rev-parse --abbrev-ref HEAD)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -15,7 +15,7 @@
 #
 
 BUILD          ?= $(shell git rev-parse --short HEAD)
-BUILD_CODENAME  = dgraph
+BUILD_CODENAME ?= dgraph
```
sanjayk-github-dev, line 18: `BUILD_CODENAME` also defined in the `dgraph/Makefile`. Which one takes precedence?

joshua-goldstein: We define `BUILD_CODENAME ?= dgraph` in the nested makefile, which means that, if `BUILD_CODENAME` is defined at the root level, it will take precedence and be inherited at the lower level.  I've also confirmed this locally just to be sure.

sanjayk-github-dev: Thanks!

sanjayk-github-dev (approved):

mangalaman93 (approved):

PR #8961, opened 2023-08-18
fix(export): Escape MySQL column names

andres-lowrie:
Description: This change wraps all the column names with backticks so that MySQL won't complain when a table has columns that are also reserved keywords in MySQL.


MichelDiz (approved):

mangalaman93 (approved):

mangalaman93:
Thank you for the PR @andres-lowrie

PR #8965, opened 2023-08-19
test(upgrade): skip test for dgraph version when the bug is not fixed

mangalaman93:
TestUnreservedPredicateForDeletion should be run with dgraph versions in which the bug is fixed. This commit also updates the upgrade version to v23.1.0 in upgrade config.

### DIFF
--- a/dgraphtest/config.go
+++ b/dgraphtest/config.go
@@ -31,39 +31,40 @@ type UpgradeCombo struct {
 
 func AllUpgradeCombos(v20 bool) []UpgradeCombo {
 	fixedVersionCombos := []UpgradeCombo{
-		// OPEN SOURCE RELEASES
-		{"v21.03.0", "v23.0.1", BackupRestore},
-		{"v21.03.0-92-g0c9f60156", "v23.0.1", BackupRestore},
-		{"v21.03.0-98-g19f71a78a-slash", "v23.0.1", BackupRestore},
-		{"v21.03.0-99-g4a03c144a-slash", "v23.0.1", BackupRestore},
-		{"v21.03.1", "v23.0.1", BackupRestore},
-		{"v21.03.2", "v23.0.1", BackupRestore},
-		{"v22.0.0", "v23.0.1", BackupRestore},
-		{"v22.0.1", "v23.0.1", BackupRestore},
-		{"v22.0.2", "v23.0.1", BackupRestore},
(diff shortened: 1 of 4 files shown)


harshil-goel (approved):

harshil-goel (approved):

dgraphtest/config.go
```diff
@@ -31,39 +31,39 @@ type UpgradeCombo struct {
...
-		{"v21.03.0", "v23.0.1", BackupRestore},
-		{"v21.03.0-92-g0c9f60156", "v23.0.1", BackupRestore},
-		{"v21.03.0-98-g19f71a78a-slash", "v23.0.1", BackupRestore},
-		{"v21.03.0-99-g4a03c144a-slash", "v23.0.1", BackupRestore},
-		{"v21.03.1", "v23.0.1", BackupRestore},
-		{"v21.03.2", "v23.0.1", BackupRestore},
-		{"v22.0.0", "v23.0.1", BackupRestore},
-		{"v22.0.1", "v23.0.1", BackupRestore},
-		{"v22.0.2", "v23.0.1", BackupRestore},
+		// OPEN SOURCE RELEASES, 4fc9cfd => 23.1.0
```
all-seeing-code, line 34: I'd like to understand why we need to refer to the hash and not the release version. Also if you could add that comment in the code too, that'd be great.

query/mutation_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build integration || cloud || upgrade
+//go:build integration || upgrade
```
all-seeing-code, line 1: We don't want to run them in cloud?

all-seeing-code (changes requested):

dgraphtest/config.go
mangalaman93: The upgrade test fails for v23.1 because of a recent change that was made in an error. We will move it a newer release as soon as we have another.

query/mutation_test.go
mangalaman93: We don't run anything in cloud today. We don't use the cloud tag.

dgraphtest/config.go
mangalaman93: Added a comment.

all-seeing-code (approved):

harshil-goel (approved):

PR #8966, opened 2023-08-22
perf(query): Read just the latest value for scalar types

harshil-goel:
Description: Right now, when we read some data, we read the entire posting list. With this we populate List structure in memory, which has two layers. These two layers are then merge sorted. All of this takes a lot of time. In this diff, we are updating the query processor to just read the latest value in case of scalar types. 

LDBC test for IC09.
Before: 39 seconds
After: 30 seconds

### DIFF
--- a/posting/list.go
+++ b/posting/list.go
@@ -1397,52 +1397,51 @@ func (l *List) StaticValue(readTs uint64) (*pb.PostingList, error) {
 	l.RLock()
 	defer l.RUnlock()
 
-	return l.StaticValueWithLockHeld(readTs)
+	return l.findStaticValue(readTs), nil
 }
 
-func (l *List) StaticValueWithLockHeld(readTs uint64) (*pb.PostingList, error) {
-	val, found, err := l.findStaticValue(readTs, math.MaxUint64)
-	if err != nil {
-		return val, errors.Wrapf(err,
-			"cannot retrieve default value from list with key %s", hex.EncodeToString(l.key))
-	}
-	if !found {
-		return val, ErrNoValue
-	}
-	return val, nil
-}
-
-func (l *List) findStaticValue(readTs, uid uint64) (*pb.PostingList, bool, error) {
+func (l *List) findStaticValue(readTs uint64) *pb.PostingList {
 	l.AssertRLock()
 
-	mutation, ok := l.mutationMap[readTs]
-	if ok {
-		return mutation, true, nil
+	if l.mutationMap == nil {
+		// If mutation map is empty, check if there is some data, and return it.
+		if l.plist != nil && len(l.plist.Postings) > 0 {
+			return l.plist
+		}
+		return nil
 	}
 
+	// Return readTs is if it's present in the mutation. It's going to be the latest value.
+	mutation, ok := l.mutationMap[readTs]
+	if ok {
+		return mutation
+	}
+
+	// If maxTs < readTs then we need to read maxTs
 	if l.maxTs < readTs {
 		mutation, ok = l.mutationMap[l.maxTs]
 		if ok {
-			return mutation, true, nil
+			return mutation
 		}
 	}
 
-	if len(l.mutationMap) != 0 {
-		for ts, mutation_i := range l.mutationMap {
-			if ts <= readTs {
-				mutation = mutation_i
-			} else {
-				break
-			}
+	// This means that maxTs > readTs. Go through the map to find the closest value to readTs
+	mutation = nil
+	ts_found := uint64(0)
+	for ts, mutation_i := range l.mutationMap {
+		if ts <= readTs && ts > ts_found {
+			ts_found = ts
+			mutation = mutation_i
 		}
-		return mutation, true, nil
 	}
 
-	if len(l.plist.Postings) > 0 {
-		return l.plist, true, nil
+	if mutation != nil {
+		return mutation
 	}
 
-	return nil, false, nil
+	// If we reach here, that means that there was no entry in mutation map which is less than readTs. That
+	// means we need to return l.plist
+	return l.plist
 }
 
 // Value returns the default value from the posting list. The default value is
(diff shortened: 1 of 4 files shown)


posting/lists.go
```diff
@@ -195,6 +195,59 @@ func (lc *LocalCache) getInternal(key []byte, readFromDisk bool) (*List, error)
 	return lc.SetIfAbsent(skey, pl), nil
 }
 
+func (lc *LocalCache) GetSinglePosting(key []byte) (*pb.PostingList, error) {
```
jairad26, line 198: since its a public function could you add a description of what this function's doing/why its necessary? basically what you explained yesterday in text format would be helpful for future reference

posting/lists.go
```diff
@@ -195,6 +195,59 @@ func (lc *LocalCache) getInternal(key []byte, readFromDisk bool) (*List, error)
...
+		pl := &pb.PostingList{}
+		lc.RLock()
+		if delta, ok := lc.deltas[string(key)]; ok && len(delta) > 0 {
+			err := pl.Unmarshal(delta)
+			if err != nil {
+				lc.RUnlock()
+				return pl, nil
+			}
+		}
+		lc.RUnlock()
```
jairad26, line 210: why not defer unlock till the end of the function?

jairad26 (commented):
added inline comments

worker/task.go
```diff
@@ -376,6 +376,9 @@ func (qs *queryState) handleValuePostings(ctx context.Context, args funcArgs) er
 
 	outputs := make([]*pb.Result, numGo)
 	listType := schema.State().IsList(q.Attr)
+	hasLang := schema.State().HasLang(q.Attr)
+	getMultiplePosting := q.DoCount || q.ExpandAll || listType || hasLang
```
mangalaman93, line 380: add some explanation

posting/list_test.go
```diff
@@ -435,6 +435,55 @@ func TestAddMutation_mrjn1(t *testing.T) {
...
+		if j < int(ol.minTs) {
+			j = int(ol.minTs)
+		}
+		for ; j < i+6; j++ {
+			tx := NewTxn(uint64(j))
+			k, err := tx.cache.GetSinglePosting(key)
+			require.NoError(t, err)
+			checkValue(t, ol, string(k.Postings[0].Value), uint64(j))
+		}
+
```
mangalaman93, line 483: nit: remove newline

mangalaman93 (changes requested):

mangalaman93 (dismissed):

jairad26 (approved):

posting/list.go
```diff
@@ -1393,6 +1393,58 @@ func (l *List) GetLangTags(readTs uint64) ([]string, error) {
...
+	}
+	if !found {
+		return val, ErrNoValue
+	}
+	return val, nil
+}
+
+func (l *List) findStaticValue(readTs, uid uint64) (*pb.PostingList, bool, error) {
+	l.AssertRLock()
+
```
mangalaman93, line 1417: is it possible that the mutationMap may be nil here?

posting/list.go
```diff
@@ -1393,6 +1393,58 @@ func (l *List) GetLangTags(readTs uint64) ([]string, error) {
...
+	}
+
+	if l.maxTs < readTs {
+		mutation, ok = l.mutationMap[l.maxTs]
+		if ok {
+			return mutation, true, nil
+		}
+	}
+
+	if len(l.mutationMap) != 0 {
```
mangalaman93, line 1430: we don't need to check this

posting/list.go
```diff
@@ -1393,6 +1393,58 @@ func (l *List) GetLangTags(readTs uint64) ([]string, error) {
...
+		for ts, mutation_i := range l.mutationMap {
```
mangalaman93, line 1431: the iteration order is guaranteed to be not the same in map iteration in Go

posting/lists.go
```diff
@@ -322,6 +322,80 @@ func (lc *LocalCache) getInternal(key []byte, readFromDisk bool) (*List, error)
...
+// GetSinglePosting retrieves the cached version of the first item in the list associated with the
+// given key. This is used for retrieving the value of a scalar predicats.
+func (lc *LocalCache) GetSinglePosting(key []byte) (*pb.PostingList, error) {
+	getList := func() *pb.PostingList {
+		lc.RLock()
+
+		pl := &pb.PostingList{}
+		if delta, ok := lc.deltas[string(key)]; ok && len(delta) > 0 {
+			err := pl.Unmarshal(delta)
+			if err == nil {
```
mangalaman93, line 334: what if there is an error here?

posting/lists.go
```diff
@@ -322,6 +322,80 @@ func (lc *LocalCache) getInternal(key []byte, readFromDisk bool) (*List, error)
...
+				return pl
+			}
+		}
+
+		l := lc.plists[string(key)]
+		lc.RUnlock()
+
+		if l != nil {
+			pl, err := l.StaticValue(lc.startTs)
+			if err != nil {
```
mangalaman93, line 345: log the error?

mangalaman93: and fix this

posting/lists.go
```diff
@@ -322,6 +322,80 @@ func (lc *LocalCache) getInternal(key []byte, readFromDisk bool) (*List, error)
...
+	}
+
+	getPostings := func() (*pb.PostingList, error) {
+		pl := getList()
+		if pl != nil {
+			return pl, nil
+		}
+
+		pl = &pb.PostingList{}
+		txn := pstore.NewTransactionAt(lc.startTs, false)
```
mangalaman93, line 360: discard the txn?

posting/lists.go
```diff
@@ -322,6 +322,80 @@ func (lc *LocalCache) getInternal(key []byte, readFromDisk bool) (*List, error)
...
+		item, err := txn.Get(key)
+		if err != nil {
+			return nil, err
+		}
+
+		err = item.Value(func(val []byte) error {
+			if err := pl.Unmarshal(val); err != nil {
```
mangalaman93, line 367: can return directly and remove the if else

mangalaman93 (changes requested):

mangalaman93 (approved):

PR #8107, opened 2021-12-03
feat(DQL): Add keyword every to DQL

EnricoMi:
This adds the keyword `every` to DQL. This is similar to #7693, which picks a fixed number of nodes at random, whereas this picks equidistant nodes. This provides a deterministic sample of the result that is proportional in size. It can be combined with `first`, `offset`, and `after`.

There are tests that exemplify the functionality.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/8107)

PR #8976, opened 2023-08-25
feat(dql): add keyword "every" to DQL

EnricoMi:
This adds the keyword `every` to DQL, which picks equidistant nodes. This provides a deterministic sample of the result that is proportional in size. It can be combined with `first`, `offset`, and `after`. Pagination is applied **before** `every`.

This supersedes #8107

Docs PR: None

### DIFF
--- a/query/query.go
+++ b/query/query.go
@@ -2407,20 +2407,22 @@ func (sg *SubGraph) applyEvery(ctx context.Context) error {
 	offset := 0
 	every := sg.Params.Every
 	for i := 0; i < len(sg.uidMatrix); i++ {
-		uidList := codec.GetUids(sg.uidMatrix[i])
+		uidList := sg.uidMatrix[i].Uids
 
-		r := sroar.NewBitmap()
-		for j := offset; j < len(uidList); j += every {
-			r.Set(uidList[j])
+		// take every 'every'-th uid starting at offset
+		uidsLen := (len(uidList) - offset) / every
+		uids := make([]uint64, uidsLen)
+		for j := 0; j < uidsLen; j++ {
+			uids[j] = uidList[j*every+offset]
 		}
-		sg.uidMatrix[i].Bitmap = r.ToBuffer()
+		sg.uidMatrix[i].Uids = uids
 
 		// there might be some uids after the last uid picked from uidList
 		// consider these as skipped, so skip that number of uids fewer from next list
 		offset = (offset - len(uidList)%every) % every
 	}
-
-	sg.DestMap = codec.Merge(sg.uidMatrix)
+	// Re-merge the UID matrix.
+	sg.DestUIDs = algo.MergeSorted(sg.uidMatrix)
 	return nil
 }
 
--- a/query/query.go
+++ b/query/query.go
@@ -2399,8 +2399,8 @@ func ProcessGraph(ctx context.Context, sg, parent *SubGraph, rch chan error) {
 }
 
 // applies "every" to lists inside uidMatrix
-// the first node is selected from each of the uid lists, then every sg.Params.Every node
-// nodes are selected after applying first, offset and after
+// every 'every-th' uid is selected from the uid lists,
+// padding uids are considered when skipping uids from next list
 func (sg *SubGraph) applyEvery(ctx context.Context) error {
 	sg.updateUidMatrix()
 
@@ -2413,7 +2413,8 @@ func (sg *SubGraph) applyEvery(ctx context.Context) error {
 		uidsLen := (len(uidList) - offset) / every
 		uids := make([]uint64, uidsLen)
 		for j := 0; j < uidsLen; j++ {
-			uids[j] = uidList[j*every+offset]
+			// take the last uid of every-sized batches starting at offset
+			uids[j] = uidList[(j+1)*every-1+offset]
 		}
 		sg.uidMatrix[i].Uids = uids
 
--- a/query/query.go
+++ b/query/query.go
@@ -2407,15 +2407,20 @@ func ProcessGraph(ctx context.Context, sg, parent *SubGraph, rch chan error) {
 func (sg *SubGraph) applyEvery(ctx context.Context) error {
 	sg.updateUidMatrix()
 
+	offset := 0
 	every := sg.Params.Every
 	for i := 0; i < len(sg.uidMatrix); i++ {
 		uidList := codec.GetUids(sg.uidMatrix[i])
 
 		r := sroar.NewBitmap()
-		for j := 0; j < len(uidList); j += every {
+		for j := offset; j < len(uidList); j += every {
 			r.Set(uidList[j])
 		}
 		sg.uidMatrix[i].Bitmap = r.ToBuffer()
+
+		// there might be some uids after the last uid picked from uidList
+		// consider these as skipped, so skip that number of uids fewer from next list
+		offset = (offset - len(uidList) % every) % every
 	}
 
 	sg.DestMap = codec.Merge(sg.uidMatrix)
--- a/query/query.go
+++ b/query/query.go
@@ -1659,19 +1659,16 @@ func (sg *SubGraph) populateFacetVars(doneVars map[string]varValue, sgPath []*Su
 // recursiveFillVars fills the value of variables before a query is to be processed using the result
 // of the values (doneVars) computed by other queries that were successfully run before this query.
 func (sg *SubGraph) recursiveFillVars(doneVars map[string]varValue) error {
-	err := sg.fillVars(doneVars)
-	if err != nil {
+	if err := sg.fillVars(doneVars); err != nil {
 		return err
 	}
 	for _, child := range sg.Children {
-		err = child.recursiveFillVars(doneVars)
-		if err != nil {
+		if err := child.recursiveFillVars(doneVars); err != nil {
 			return err
 		}
 	}
 	for _, fchild := range sg.Filters {
-		err = fchild.recursiveFillVars(doneVars)
-		if err != nil {
+		if err := fchild.recursiveFillVars(doneVars); err != nil {
 			return err
 		}
 	}
@@ -2131,7 +2128,7 @@ func ProcessGraph(ctx context.Context, sg, parent *SubGraph, rch chan error) {
 		switch {
 		case isInequalityFn && sg.SrcFunc.IsValueVar:
 			// This is a ineq function which uses a value variable.
-			err = sg.applyIneqFunc()
+			err := sg.applyIneqFunc()
 			if parent != nil {
 				rch <- err
 				return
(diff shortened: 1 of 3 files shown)


query/query.go
```diff
@@ -2286,6 +2296,14 @@ func ProcessGraph(ctx context.Context, sg, parent *SubGraph, rch chan error) {
 		}
 	}
 
+	// We apply Every _after_ pagination, so it refers to the respective page
+	if sg.Params.Every > 1 {
+		if err = sg.applyEvery(ctx); err != nil {
```
mangalaman93, line 2301: we should use := here to reduce the scope of err

mangalaman93 (changes requested):
Thank you for updating the PR. This looks good to me, I have just one minor comment. I am also thinking that we should add test where every and after are used together. Thanks

query/query.go
EnricoMi: Many other place do not limit the scope of err, so I have reduced the scope in all those other places as well: eebba60020a5ce17be038a23bbb645ac57b5c896

EnricoMi:
> we should add test where every and after are used together.

Added in https://github.com/dgraph-io/dgraph/pull/8976/files#diff-99902c080bca80ed1ad52da34ea4041069d1c64a52317fc17f2285e8170ae6ddR1085

query/query.go
mangalaman93: Thanks

query/query.go
```diff
@@ -1654,13 +1664,13 @@ func (sg *SubGraph) recursiveFillVars(doneVars map[string]varValue) error {
 		return err
 	}
 	for _, child := range sg.Children {
-		err = child.recursiveFillVars(doneVars)
+		err := child.recursiveFillVars(doneVars)
```
mangalaman93, line 1667: nit: if you are doing this, you can also move the statement inside the if. Same for some of the statements below.

mangalaman93 (approved):
one minor comment, LGTM otherwise. Is it possible for you to raise a corresponding docs PR on https://github.com/dgraph-io/dgraph-docs

EnricoMi:
I'd love to run tests locally, but the instructions in CONTRIBUTING.md seem outdated and following the instructions from `ci-dgraph-tests.yml` don't work either:

```
cd t; ./t
testutil: "" localhost:9080 localhost:5080
...
Running tests for 121 packages.
Sent 1/121 packages for processing.
Bringing up cluster test-013-1 for package: ../dgraph/docker-compose.yml ...
CLUSTER UP: test-013-1. Package: ../dgraph/docker-compose.yml
Health for test-013-1_alpha6_1 failed: Get "http://localhost:49272/health": read tcp 127.0.0.1:35362->127.0.0.1:49272: read: connection reset by peer. Response: "". Retrying...
Health for test-013-1_alpha6_1 failed: Get "http://localhost:49272/health": dial tcp 127.0.0.1:49272: connect: connection refused. Response: "". Retrying...
Health for test-013-1_alpha6_1 failed: Get "http://localhost:49272/health": dial tcp 127.0.0.1:49272: connect: connection refused. Response: "". Retrying...
...
```

Running `dgraph version` in the `dgraph/dgraph:local` image returns:
```
dgraph: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by dgraph)
dgraph: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by dgraph)
```

EnricoMi:
I have reworked the notion of `every` in conjunction with `first`, `after` and `offset`. The `every` is applied before pagination, so that the result of `every` can be paginated, not the paginated is every-ed.

This required to change the way `every` works. It now picks the last uid from `every`-sized batches, so that paginating with `after` works: 2e4f90f2093b7a7b317d58dcfd35775a4a2c6f70

EnricoMi:
I have managed to get tests working by setting `contrib/Dockerfile` to `ubuntu:22.04` to match my dev machine.

query/query4_test.go
```diff
@@ -988,6 +989,246 @@ func TestCascadeSubQuery1(t *testing.T) {
...
+				"name": "Elizabeth"
+			  }
+			]
+		  }
+	}`, js)
+}
+
+func TestHasEveryWithFirstAfter(t *testing.T) {
+	query := `{
+		q(func:has(name), every: 10, first:2, after: 0x36) {
```
mangalaman93, line 1186: Shouldn't this return exactly 2 results?

mangalaman93 (changes requested):
Looks much better too. One question before it is ready to go.

query/query4_test.go
EnricoMi: It returns exactly two uids, some have multiple `<name>` values.

mangalaman93: Did you modify the test now?

EnricoMi: Yes, I found another triple in the test data with predicate `<name>`, which changed the sequence of results, so I had to update the expectation: https://github.com/dgraph-io/dgraph/compare/52d436be0c4589528b9dffb1c593fa0a13b1f4de..a8e9948eb6f95dbcb879831a83cff4b465913172

I think that should be the right response, but my code produces some other result.

EnricoMi: Is there a way to unit-test `SubGraph.applyEvery` with some test `uidMatrix`?

query/query.go
```diff
@@ -2383,6 +2398,35 @@ func ProcessGraph(ctx context.Context, sg, parent *SubGraph, rch chan error) {
...
+// applies "every" to lists inside uidMatrix
+// every 'every-th' uid is selected from the uid lists,
+// padding uids are considered when skipping uids from next list
+func (sg *SubGraph) applyEvery(ctx context.Context) error {
+	sg.updateUidMatrix()
+
+	offset := 0
+	every := sg.Params.Every
+	for i := 0; i < len(sg.uidMatrix); i++ {
+		uidList := sg.uidMatrix[i].Uids
```
EnricoMi, line 2410: @mangalaman93 assumption here is that uids in `sg.uidMatrix[i].Uids` are sorted ascending order, and all uids in `sg.uidMatrix[i].Uids` are smaller than in `sg.uidMatrix[i+1].Uids`. Is that true?

harshil-goel: No, i+1 doesn't have to be bigger than i. We merge these later using MergeSorted. Uids should be in the correct order.

harshil-goel (approved):

mangalaman93:
@EnricoMi Could you look at the CI failures? 

Also, I only have an incomplete version of this PR https://github.com/dgraph-io/dgraph/pull/8977. If you like, you could take over a complete it. It should be a very similar PR.

query/query.go
EnricoMi: So the individual `sg.uidMatrix[*].Uids` lists are arbitrary subsets of the entire uid space. This renders the "offset" logic obsolete.

Any idea how to efficiently sample every `every`-th uid deterministically?

EnricoMi: Given https://github.com/dgraph-io/dgraph/pull/8977/files#diff-c6a819ee05eb6643352fb889fdd48569cca191565c59c96ec849150a12bad2feR2417, it looks like then notion of having multiple `sg.uidMatrix[*].Uids lists is different to what I presume. What does it mean to have multiple `uidMatrix`?

EnricoMi:
@mangalaman93 happy to finish #8977 once I understand the notion of having multiple `sg.uidMatrix`.

query/query.go
EnricoMi: I have looked through the documentation of `subGraph` and I think I understand `sd.uidMatrix` now. Updating tests.

query/query4_test.go
EnricoMi: You were right, it should return only two results, as it selects the `name` property without language annotation per uid.

EnricoMi:
Tests should all pass now, will look into docs PR next.

mangalaman93:
@EnricoMi upgrade tests are failing but don't worry about those. They are failing because the older versions of Dgraph do not support "every" keyword. I will fix those.

query/query.go
```diff
@@ -2383,6 +2398,31 @@ func ProcessGraph(ctx context.Context, sg, parent *SubGraph, rch chan error) {
...
+		for j := 0; j < uidsLen; j++ {
+			// take the last uid of every-sized batches
+			// taking the last rather than the first allows us to paginate via after
+			uids[j] = uidList[(j+1)*every-1]
+		}
+		sg.uidMatrix[i].Uids = uids
+	}
+
+	// Re-merge the UID matrix.
+	sg.DestUIDs = algo.MergeSorted(sg.uidMatrix)
```
harshil-goel, line 2422: Can you also write unit test for applyEvery function? 
I think you need to first MergeSort  the entire uidMatrix and then take every ith element? Right now I don't think we will get the correct result.

query/query4_test.go
```diff
@@ -988,6 +989,142 @@ func TestCascadeSubQuery1(t *testing.T) {
 	}`, js)
 }
 
+func TestHasEvery(t *testing.T) {
+	query := `{
```
mangalaman93, line 993: add this line as the first line of the test in each new test that you have added. This should fix the upgrade tests. Thanks

```
dgraphtest.ShouldSkipTest(t, "071c85b59fdf80ade88f4571e9cf8c867fae1f6a", dc.GetVersion())
```

mangalaman93:
@EnricoMi one more question, what is the use case of "every" word?  Why can't we just use "random"?

EnricoMi:
@mangalaman93 the use case is to partition the result set (the entire graph) into equi-sized partitions with one query to get boundary uids that are then used with `after` to fetch results in parallel: https://discuss.dgraph.io/t/sample-your-result/9055/1 With the `random` key the size of partitions would not be as useful.

Using offset is not an option as it has `O(N)` time complexity, where `after` has constant time: https://discuss.dgraph.io/t/pagination-with-offset-should-scale-like-after/8577

Looking at the code, the `every` keyword seems of `O(N)` complexity as well. Maybe the sampling should happen remotely in `worker.ProcessTaskOverNetwork` to retrieve only the sample of uids (e.g. every 1,000,000th uid): https://github.com/dgraph-io/dgraph/blob/890c1f79b3afeeba7dd66fae54fa74b20b049a25/query/query.go#L2160

mangalaman93:
> Looking at the code, the `every` keyword seems of `O(N)` complexity as well. Maybe the sampling should happen remotely in `worker.ProcessTaskOverNetwork` to retrieve only the sample of uids (e.g. every 1,000,000th uid):

that does seem like an interesting idea. I just wonder whether it is feasible to do given that results may get filtered out later sometimes. If you want, you can look into it. Thanks for the explanation for adding 'every' to the query language.

EnricoMi:
> that does seem like an interesting idea

I may need some pointer to the code where that `worker.ProcessTaskOverNetwork` is being executed.

EnricoMi:
@mangalaman93 ping

PR #7693, opened 2021-04-07
feat(Query): Add random keyword in DQL

Samyak2:
Fixes: https://discuss.dgraph.io/t/pick-random-node/10815

## Changes

 - random nodes can be selected from the graph using `random: <number of nodes>`
 - random selection by replacement is used to select nodes

## Issues/future work

 - Since selection is by replacement, you can get duplicates
 - No error handling yet in the `applyRandom` function
 - Making of new slice in lines 2430-2432 looks a little inefficient, there might a better way to do this.

This is my first PR here (and I'm new to Go), there will be a lot more issues I'm not aware of.

## Example

I have only tested this on a basic example from Ratel. This will need a lot more testing.

<details>
  <summary>Mutation</summary>

  ```graphql
  {
    set {
     _:luke <name> "Luke Skywalker" .
     _:luke <dgraph.type> "Person" .
     _:leia <name> "Princess Leia" .
     _:leia <dgraph.type> "Person" .
     _:han <name> "Han Solo" .
     _:han <dgraph.type> "Person" .
     _:lucas <name> "George Lucas" .
     _:lucas <dgraph.type> "Person" .
     _:irvin <name> "Irvin Kernshner" .
     _:irvin <dgraph.type> "Person" .
     _:richard <name> "Richard Marquand" .
     _:richard <dgraph.type> "Person" .
  
     _:sw1 <name> "Star Wars: Episode IV - A New Hope" .
     _:sw1 <release_date> "1977-05-25" .
     _:sw1 <revenue> "775000000" .
     _:sw1 <running_time> "121" .
     _:sw1 <starring> _:luke .
     _:sw1 <starring> _:leia .
     _:sw1 <starring> _:han .
     _:sw1 <director> _:lucas .
     _:sw1 <dgraph.type> "Film" .
  
     _:sw2 <name> "Star Wars: Episode V - The Empire Strikes Back" .
     _:sw2 <release_date> "1980-05-21" .
     _:sw2 <revenue> "534000000" .
     _:sw2 <running_time> "124" .
     _:sw2 <starring> _:luke .
     _:sw2 <starring> _:leia .
     _:sw2 <starring> _:han .
     _:sw2 <director> _:irvin .
     _:sw2 <dgraph.type> "Film" .
  
     _:sw3 <name> "Star Wars: Episode VI - Return of the Jedi" .
     _:sw3 <release_date> "1983-05-25" .
     _:sw3 <revenue> "572000000" .
     _:sw3 <running_time> "131" .
[...]

PR #8977, opened 2023-08-25
feat(dql): add random keyword in DQL (#7693)

mangalaman93:
This change implements the `random` function argument. If an argument `random: k` is provided to a dql function then random k results will be returned.

For example:
`func(has(name), random:5)` will choose 5 random nodes having the name predicate.

### DIFF
--- a/query/query.go
+++ b/query/query.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"math/rand"
 	"sort"
 	"strconv"
 	"strings"
@@ -112,6 +113,8 @@ type params struct {
 	Count int
 	// Offset is the value of the "offset" parameter.
 	Offset int
+	// Random is the value of the "random" parameter
+	Random int
 	// AfterUID is the value of the "after" parameter.
 	AfterUID uint64
 	// DoCount is true if the count of the predicate is requested instead of its value.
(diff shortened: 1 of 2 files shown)


query/query.go
```diff
@@ -2383,6 +2402,43 @@ func ProcessGraph(ctx context.Context, sg, parent *SubGraph, rch chan error) {
...
+		rand.Shuffle(len(uidList), func(i, j int) {
+			uidList[i], uidList[j] = uidList[j], uidList[i]
+		})
+
+		numRandom := sg.Params.Random
+		if sg.Params.Random > len(uidList) {
+			numRandom = len(uidList)
+		}
+
+		sg.uidMatrix[i].Uids = uidList[:numRandom]
```
EnricoMi, line 2435: Uids in `sg.uidMatrix[i].Uids` are not sorted any more. I presume `algo.MergeSorted(sg.uidMatrix)` requires that.

query/query.go
```diff
@@ -2383,6 +2402,43 @@ func ProcessGraph(ctx context.Context, sg, parent *SubGraph, rch chan error) {
...
+	x int
+	y int
+}
+
+// applies "random" to lists inside uidMatrix
+// sg.Params.Random number of nodes are selected in each uid list
+// duplicates are avoided (random selection without replacement)
+// if sg.Params.Random is more than the number of available nodes
+// all nodes are returned
+func (sg *SubGraph) applyRandom(ctx context.Context) error {
```
EnricoMi, line 2417: Having `len(sg.uidMatics) == N` will produce `N * sg.Params.Random` nodes, not `sg.Params.Random`, right?

PR #8981, opened 2023-08-26
chore(ci): bring back embargo test

joshua-goldstein:
Description: While cleaning up TeamCity, we found an old test framework that still works and might be useful.  We bring back this test.
Remark: In this PR the job is configured to run only on a push to main and on a schedule, so it is not currently showing up in the jobs below.  However, it is passing on CI.  You can see a previous run e.g. [here](https://github.com/dgraph-io/dgraph/actions/runs/5983221790) and [here](https://github.com/dgraph-io/dgraph/actions/runs/5983254334/job/16233203923).


.github/workflows/ci-dgraph-embargo-tests.yml
```diff
@@ -0,0 +1,43 @@
+name: ci-dgraph-embargo-tests
+on:
+  push:
+    branches:
+      - main
```
mangalaman93, line 5: not on release branches?

.github/workflows/ci-dgraph-embargo-tests.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Move dgraph binary to gopath
+        run: mv dgraph/dgraph ~/go/bin/dgraph
+      - name: Get Python Version
+        run: |
+          #!/bin/bash
+          cd contrib/embargo
+          PYTHON_VERSION=$({ [ -f .python-version ] && cat .python-version; })
+          echo "PYTHON_VERSION=$PYTHON_VERSION" >> $GITHUB_ENV
+      - name: Set up Python
+        uses: actions/setup-python@v4
```
mangalaman93, line 35: why does it say @v4? Is that version of the GitHub action?

mangalaman93 (changes requested):

PR #8987, opened 2023-09-02
fix(upgrade): print logs when cluster startup fails

mangalaman93:


ee/acl/jwt_algo_test.go
```diff
@@ -42,7 +42,10 @@ func TestACLJwtAlgo(t *testing.T) {
 			c, err := dgraphtest.NewLocalCluster(conf)
 			require.NoError(t, err)
 			defer func() { c.Cleanup(t.Failed()) }()
-			require.NoError(t, c.Start())
+			if err := c.Start(); err != nil {
+				c.Cleanup(true)
```
kevinmingtarja, line 46: Do we still need to call Cleanup() here? Since we already do it in the defer block right above this.

mangalaman93: calling cleanup twice is fine, that's not a problem. But I know without this code the logs are not printed. Let me figure out why!

kevinmingtarja (approved):

ee/acl/jwt_algo_test.go
mangalaman93: you are right Kevin. Thanks for noticing, I'm missing something. Closing this PR, may open another if needed.

PR #8988, opened 2023-09-02
fix(restore): allow incrementalFrom to be 1 in restore API

mangalaman93:
This is a minor inconvenience while using the incremental restore API that when incrementalFrom is set to 1, we throw an error back. The restore API takes two backup numbers incrementalFrom & backupNum and restores all the backups including both the ends, i.e. following set notation all the backups in the set [incrementalFrom, backupNum] are restored. This should work fine when incrementalFrom is set to 1 which is a full backup.

### DIFF
--- a/worker/online_restore.go
+++ b/worker/online_restore.go
@@ -245,8 +245,14 @@ func handleRestoreProposal(ctx context.Context, req *pb.RestoreRequest, pidx uin
 		return errors.Errorf("nil restore request")
 	}
 
+	// This is a minor inconvenience while using the incremental restore API that
+	// when incrementalFrom is set to 1, we throw an error back. The restore API
+	// takes two backup numbers incrementalFrom & backupNum and restores all the
+	// backups including both the ends, i.e. following set notation all the backups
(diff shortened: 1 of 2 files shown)


kevinmingtarja (dismissed):

worker/online_restore.go
```diff
@@ -246,7 +246,7 @@ func handleRestoreProposal(ctx context.Context, req *pb.RestoreRequest, pidx uin
 	}
 
 	if req.IncrementalFrom == 1 {
-		return errors.Errorf("Incremental restore must not include full backup")
+		req.IncrementalFrom = 0
```
all-seeing-code, line 249: Could you illustrate how would this work if we need backups from 1 to say n. But not the full backup (which I guess would be from 0 to n).

mangalaman93: Full backup is number 1 and not 0. The PR should make sense to you now.

all-seeing-code (dismissed):

kevinmingtarja (dismissed):

kevinmingtarja (approved):
LGTM

PR #8994, opened 2023-09-06
Add debug build to load tests

harshil-goel:
Add debug build to load tests


mangalaman93 (approved):

joshua-goldstein (approved):

joshua-goldstein (commented):
For some reason it looks like this flag is not being picked up during the binary build step, not sure why though

gajanan-dgraph:
This was created only for debugging purpose and it is no longer needed.

PR #9002, opened 2023-09-24
ci(runners): use larger github hosted runners + reduce cron frequency

joshua-goldstein:
Description: Currently we use large 32gb self-hosted runners for jobs that require more resources.  Github now offers their own suite of hosted runners.  We can now transition to these new runners.  We also reduce the frequency of some scheduled jobs and remove some particular workflows from each PR run.

### DIFF
--- a/.github/workflows/ci-dgraph-code-coverage.yml
+++ b/.github/workflows/ci-dgraph-code-coverage.yml
@@ -24,8 +24,7 @@ on:
       - 'release/**'
 jobs:
   dgraph-code-coverage:
-    runs-on: [self-hosted, x64]
-    # runs-on: ubuntu-20.04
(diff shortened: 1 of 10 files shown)


all-seeing-code (approved):
Should we retire the auto scaler as we merge these PRs

joshua-goldstein:
> Should we retire the auto scaler as we merge these PRs

The coordinator ec2 instance is pretty small (I believe micro size) so it doesn't cost much to keep that infrastructure in place.  But yeah, it will certainly pick up less jobs and cost less on the ec2 side of things.

However there are no github hosted arm64 ubuntu runners available, so if we want to keep running the arm workflow we would have to keep that in place.

all-seeing-code (approved):

PR #9009, opened 2023-09-29
debug: fix debug tool to read WAL entries correctly

mangalaman93:
This commit fixes a bug in debug tool while reading WAL entries. Before this change, if WAL entries were bigger than 64 MB, the debug tool would go into a loop and keep reading the entry before the large entry.


dgraph/cmd/debug/wal.go
```diff
@@ -132,13 +132,14 @@ func printRaft(store *raftwal.DiskStorage) {
...
 	startIdx, lastIdx := printBasic(store)
 
-	for startIdx < lastIdx-1 {
+	for startIdx < lastIdx {
 		entries, err := store.Entries(startIdx, lastIdx+1, 64<<20)
 		x.Check(err)
 		for _, ent := range entries {
 			printEntry(ent, pending, isZero)
-			startIdx = x.Max(startIdx, ent.Index)
+			startIdx = ent.Index
```
harshil-goel, line 140: why remove max?

harshil-goel (dismissed):

dgraph/cmd/debug/wal.go
mangalaman93: Done

PR #8058, opened 2021-09-29
fix(restore): Do not retry restore proposal

ahsanbarkati:
There is an issue caused due to retry of restore proposal, consider the following scenario:
```
1. alpha-2 gets the restore request (leader is alpha-0)
2. alpha-2 sends the request to alpha-0 (leader).
3. alpha-0 called proposeAndWait which proposed the req (index 24) at time=15:56:10
4. alpha-0 was still waiting for the proposal to be applied and rpc call for `Restore` by alpha-2 got "transport closing error" at time=15:59:08
5. transport closing is a retriable error, so alpha-2 again tried to proposeoOrSend, this time leader was alpha-1, so it sent it to alpha-1
6. alpha-1 proposed the restore request (index 28) at time=15:59:09
```
This PR removes the retry logic. If the restore fails, it should be re-triggered manually.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/8058)

PR #9017, opened 2023-10-17
fix(restore): do not retry restore proposal (#8058)

mangalaman93:
Do not retry the restore proposal. It can cause issues in the edge case scenarios. Consider the following scenario:
  1. alpha-2 gets the restore request (leader is alpha-0)
  2. alpha-2 sends the request to alpha-0 (leader).
  3. alpha-0 called proposeAndWait which proposed the req (index 24) at time=15:56:10
  4. alpha-0 was still waiting for the proposal to be applied and RPC call for `Restore` by alpha-2 got "transport closing error" at time=15:59:08
  5. transport closing is a retriable error, so alpha-2 again tried to proposeoOrSend, this time leader was alpha-1, so it sent it to alpha-1
  6. alpha-1 proposed the restore request (index 28) at time=15:59:09

### DIFF
--- a/worker/online_restore.go
+++ b/worker/online_restore.go
@@ -22,7 +22,6 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
-	"time"
 
 	"github.com/golang/glog"
 	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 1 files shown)


harshil-goel:
What is the problem in the scenario you mentioned?

harshil-goel (approved):

PR #9053, opened 2024-03-14
Update math parsing function

harshil-goel:
Title format:

 - `Topic(Area): Feature`
 - `Topic` must be one of `build|ci|docs|feat|fix|perf|refactor|chore|test`
 - `Area` must be one of `acl|audit|backup|badger|cdc|dql|export|graphql|indexing|multi-tenancy|raft|restore|upgrade|zero`
 - * Add [BREAKING] if it is a breaking change

------------------
Body Format:

Description: <add description>
Fixes: <GitHub Issue>
Closes: <JIRA Issue>
Docs: <docs PR>


dql/math.go
```diff
@@ -347,6 +355,28 @@ loop:
...
+		}
+		var err error
+		t.Const, err = parseValue(va)
+		if err != nil {
+			return err
+		}
+		t.Var = ""
+	}
+	for _, i := range t.Child {
+		err := i.subs(vmap)
```
mangalaman93, line 372: nit: this could go inside if

dql/parser.go
```diff
@@ -303,6 +316,67 @@ type Request struct {
 	Variables map[string]string
 }
 
+func parseValue(v varInfo) (types.Val, error) {
+	var err error
```
mangalaman93, line 320: why are we declaring it here, we don't need it defined throughout the function

dql/parser.go
```diff
@@ -303,6 +316,67 @@ type Request struct {
...
+	typ := v.Type
+	if v.Value != "" {
+		switch typ {
+		case "int":
+			{
+				var i int64
```
mangalaman93, line 326: instead of this, mode the assignment outside the if with `:=`

dql/parser.go
```diff
@@ -303,6 +316,67 @@ type Request struct {
 	Variables map[string]string
 }
 
+func parseValue(v varInfo) (types.Val, error) {
+	var err error
+	typ := v.Type
+	if v.Value != "" {
```
mangalaman93, line 322: can we handle the opposite case first?

mangalaman93 (changes requested):
We should add unit tests for error cases. A few minor comments, looks good otherwise

mangalaman93 (dismissed):

PR #9058, opened 2024-03-22
upgrade(go): update go version to 1.22

harshil-goel:
fixes tests to use `docker compose` instead of `docker-compose`


graphql/schema/rules.go
```diff
@@ -32,8 +32,8 @@ import (
 )
 
 const (
-	baseRules         = 0
```
mangalaman93, line 35: We should add an explanation here

mangalaman93 (approved):

PR #9063, opened 2024-04-04
fix live loader and add tests for dropall, drop namespace, live load

shivaji-dgraph:


gajanan-dgraph (approved):
Parser changes are correct. Thanks for fixing tests.

PR #9074, opened 2024-04-17
Feat(graphql): Add vector support to graphql 

ghost:
Adds support for vector predicate in GraphQL. Introduced new queries like similar_to() in graphql


iyinoluwaayoola:
Defining a field in graphql type as the bellow doesn't work:
```graphql
embedding: [Float!] @embedding @search(by: ["hnsw(exponent: 4, metric: euclidian)"])
```
```error
{"errors":[{"message":"resolving updateGQLSchema failed because while validating GraphQL schema: input:8475: Undefined type HNSWSearchFilter. (Locations: [{Line: 3, Column: 4}])","extensions":{"code":"Error"}}]}
```

PR #9075, opened 2024-04-17
Adding codeql to release/v23.1

meghalims:
codeql CI job was added to main. Adding it to release/v23.1 branch as well so that we can see the set of current CVEs


gajanan-dgraph (approved):
Looks good.

PR #9078, opened 2024-04-24
fix(restore): fix incr restore and normal restore for vector predicates

shivaji-dgraph:


worker/restore_map.go
```diff
@@ -470,6 +472,15 @@ func (m *mapper) processReqCh(ctx context.Context) error {
 				}
 				return nil
 			}
+
+			isVectorSupportingPred := func(pred string) bool {
```
harshil-goel, line 476: Make it a function in schema. I think we are using something similar there also.

harshil-goel (approved):

worker/restore_map.go
shivaji-dgraph: In the schema, we are doing a different thing, not the same as above

worker/restore_map.go
```diff
@@ -470,6 +472,15 @@ func (m *mapper) processReqCh(ctx context.Context) error {
...
+				if strings.HasSuffix(pred, hnsw.VecEntry) || strings.HasSuffix(pred, hnsw.VecKeyword) ||
```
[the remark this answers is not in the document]
mangalaman93: we don't need this function. We can just check whether the condition is true and return from here itself

mangalaman93 (approved):

PR #9083, opened 2024-05-07
fix(vector): Update query_rewriter to fix dotproduct and cosine query conversion

rderbier:
fixes https://linear.app/hypermode/issue/DGR-315/querysimilarthingsbyembedding-not-working-with-dotproduct-or-cosine


rderbier:
Not sure why but if the $search_vector is not in (), we get an error.
Updating the math function solved the issue.

harshil-goel (approved):

harshil-goel (approved):

PR #9084, opened 2024-05-08
fix(vector):fix similar_to() error return when data is not present

shivaji-dgraph:


worker/task.go
```diff
@@ -381,7 +381,7 @@ func (qs *queryState) handleValuePostings(ctx context.Context, args funcArgs) er
 				int(numNeighbors), index.AcceptAll[float32])
 		}
 
-		if err != nil {
+		if err != nil && !strings.Contains(err.Error(), "HNSW tree has no elements: Key not found") {
```
harshil-goel, line 384: Lets make it a const string

harshil-goel (approved):

harshil-goel (approved):

PR #9122, opened 2024-07-31
perf(vector): Add heap to neighbour edges

ghost:
We now store the list of neighbours in a sorted distance from the starting node. This improves the accuracy and the time of the vector.
We are also using findStaticValue() function to get data faster.

Benchmarks:
1. Dataset 1 : 10k vector 100 dimension
  a. Time to ingest : 6 seconds
  b. Query time: 900us 
  c. Accuracy: 97.9%
  
2. Dataset 2: 1million vectors 1280 dimension
  a. Time to ingest: 7 mins
  b. Query time (without cache): 35 ms
  c. Query time (with cache): 5 ms

### DIFF
--- a/tok/hnsw/helper.go
+++ b/tok/hnsw/helper.go
@@ -603,8 +603,7 @@ func (ph *persistentHNSW[T]) addNeighbors(ctx context.Context, tc *TxnCache,
 		}
 		// This adds at most efConstruction number of edges for each layer for this node
 		allLayerEdges[level] = append(allLayerEdges[level], allLayerNeighbors[level]...)
-		m := 25
-		if len(allLayerEdges[level]) > m {
+		if len(allLayerEdges[level]) > ph.efConstruction {
 			err := ph.getVecFromUid(uuid, tc, &inVec)
 			if err != nil {
 				log.Printf("[ERROR] While getting vector %s", err)
@@ -621,7 +620,7 @@ func (ph *persistentHNSW[T]) addNeighbors(ctx context.Context, tc *TxnCache,
 					heap.Pop(h)
 				}
 			}
-			allLayerEdges[level] = allLayerEdges[level][:m]
+			allLayerEdges[level] = allLayerEdges[level][:ph.efConstruction]
 		}
 	}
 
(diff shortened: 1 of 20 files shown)


tok/hnsw/helper.go
```diff
@@ -518,6 +519,56 @@ func (ph *persistentHNSW[T]) newPersistentEdgeKeyValueEntry(ctx context.Context,
...
+
+	d, err := ph.simType.distanceScore(*inVec, *outVec, ph.floatBits)
+	if err != nil {
+		log.Printf("[ERROR] While getting vector %s", err)
+		return -1
+	}
+	return d
+}
+
+type CustomHeap struct {
```
jairad26, line 538: can you rename this so its a little more clear what its doing. is this a custom heap of all neighbors of a given node at a given layer? or is it essentially just holding a tuple and a compare function?

tok/hnsw/helper.go
```diff
@@ -544,13 +595,33 @@ func (ph *persistentHNSW[T]) addNeighbors(ctx context.Context, tc *TxnCache,
...
 		allLayerEdges[level] = append(allLayerEdges[level], allLayerNeighbors[level]...)
+		if len(allLayerEdges[level]) > ph.efConstruction {
+			err := ph.getVecFromUid(uuid, tc, &inVec)
+			if err != nil {
+				log.Printf("[ERROR] While getting vector %s", err)
+			} else {
+				h := &CustomHeap{
+					data: allLayerEdges[level],
+					compare: func(i, j uint64) bool {
+						return ph.distance_betw(ctx, tc, uuid, i, &inVec, &outVec) >
```
jairad26, line 614: we're still computing vectors distances at every insert addNeighbors call. instead, once the neighbors are calculated, the edges should now store the distance between the two nodes. so youll  need to modify what nodeAllEdges gives you.

jairad26: make that a heap instead that gets autosorted and you can just pop the worst one at addNeighbor time, and its O(1) cost.

jairad26 (dismissed):

jairad26 (dismissed):

PR #9113, opened 2024-07-18
fix(multi-tenancy): upsert groot for namespace after drop data

8i5dev:
Description: After executing drop data in a namespace, the namespace cannot be logged in anymore

1. drop data of a namespace

```bash
curl --location 'http://localhost:8080/alter' \
--header 'X-Dgraph-AccessToken: accesstoken' \
--header 'Content-Type: text/plain' \
--data '{"drop_op": "DATA"}
'
```

2. login the namespace again

```bash
curl --location 'http://127.0.0.1:8080/login' \
--header 'Content-Type: application/json' \
--data '{
    "userid": "groot",
    "password": "password",
    "namespace": 1
}'
```

3. get error message `Authentication from address10.42.0.1:55404 failed: unable to authenticate: invalidi credentials`

function `InitializeAcl` only upsert groot for namespace 0

PR #9159, opened 2024-09-09
fix(multi-tenancy): upsert groot for namespace after drop data

shivaji-dgraph:
Inspired from #9113

### DIFF
--- a/edgraph/server.go
+++ b/edgraph/server.go
@@ -458,8 +458,9 @@ func (s *Server) Alter(ctx context.Context, op *api.Operation) (*api.Payload, er
 
 		// just reinsert the GraphQL schema, no need to alter dgraph schema as this was drop_data
 		_, err = UpdateGQLSchema(ctx, graphQLSchema, "")
-		// recreate the admin account after a drop data operation
-		InitializeAcl(nil)
+
+		// Since all data has been dropped, we need to recreate the admin account in the respective namespace.
+		upsertGuardianAndGroot(nil, namespace)
 		return empty, err
 	}
 
--- a/systest/multi-tenancy/login_test.go
+++ b/systest/multi-tenancy/login_test.go
@@ -1,0 +1,123 @@
+//go:build integration
+// +build integration
+
+/*
+ * Copyright 2024 Dgraph Labs, Inc. and Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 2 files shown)


systest/integration2/multitenancy_test.go
```diff
@@ -0,0 +1,136 @@
...
+ */
+
+package main
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/dgraph-io/dgo/v240/protos/api"
```
mangalaman93, line 25: separate dgraph and non-dgraph imports

systest/integration2/multitenancy_test.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/dgraph-io/dgraph/v24/dgraphapi"
+	"github.com/dgraph-io/dgraph/v24/dgraphtest"
+	"github.com/dgraph-io/dgraph/v24/x"
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	gc *dgraphapi.GrpcClient
```
mangalaman93, line 33: You should pass them as args, avoid global vars

systest/integration2/multitenancy_test.go
```diff
@@ -0,0 +1,136 @@
...
+	require.NoError(t, hc.LoginIntoNamespace(dgraphapi.DefaultUser,
+		dgraphapi.DefaultPassword, x.GalaxyNamespace))
+	addData(t, gc)
+
+	RunAll(t)
+}
+
+func RunAll(t *testing.T) {
+	t.Run("login into the ns after drop data opration from the ns", testDropDataNSLogin)
+	t.Run("login into the default ns after drop data opration from default ns", testDropDataNSLogin)
```
mangalaman93, line 73: what about logging into ns after drop data operation from default ns?

systest/integration2/multitenancy_test.go
```diff
@@ -0,0 +1,136 @@
...
+	rdfs := `
+	_:a <name> "alice" .
+	_:b <name> "bob" .
+	_:c <name> "sagar" .
+	_:d <name> "ajay" .`
+	_, err := gc.Mutate(&api.Mutation{SetNquads: []byte(rdfs), CommitNow: true})
+	require.NoError(t, err)
+}
+
+func TestRunAllMultiTenancy(t *testing.T) {
```
mangalaman93, line 49: Is it possible to write an integration test instead of integration2?

mangalaman93 (changes requested):

systest/integration2/multitenancy_test.go
shivaji-dgraph: It is possible, but since we are moving from the existing test framework, I wrote these tests under the tag integration2.

systest/integration2/multitenancy_test.go
shivaji-dgraph: in the second test, I am logging into all namespaces after drop_data OP. I should rename the test.

systest/integration2/multitenancy_test.go
mangalaman93: We are thinking about moving compose tests to use dgraphtest, not planning to move integration tests to integration2 from what I remember.

harshil-goel (approved):

mangalaman93 (approved):

PR #9172, opened 2024-09-14
build: Fix OSS build

RJKeevil:
Perhaps my misunderstanding, but running "make install_oss" fails on the main repo, as e.g. on line 665 of dgraph/cmd/alpha/run.go calls ee.GetKeys, which will always return an error in OSS mode.  This change returns an empty struct instead of an error, allowing the build to complete.

### DIFF
--- a/ee/keys.go
+++ b/ee/keys.go
@@ -20,15 +20,12 @@
 package ee
 
 import (
-	"fmt"
-
 	"github.com/spf13/viper"
 )
 
 // GetKeys returns the ACL and encryption keys as configured by the user
 // through the --acl, --encryption, and --vault flags. On OSS builds,
-// this function always returns an error.
+// this function always returns an empty struct.
 func GetKeys(config *viper.Viper) (*Keys, error) {
-	return nil, fmt.Errorf(
-		"flags: acl / encryption is an enterprise-only feature")
+	return &Keys{}, nil
 }


RJKeevil:
Perhaps this change is not needed as OSS build was already in your tests, but now i'm confused how alpha/run.go could ever pass compilation without this?

mangalaman93:
> Perhaps this change is not needed as OSS build was already in your tests, but now i'm confused how alpha/run.go could ever pass compilation without this?

We do not support Encryption in OSS mode.

RJKeevil:
@mangalaman93 yes totally understand, I don't want encryption.  For me the default oss-install compilation failed trying to enable it.

mangalaman93:
`make oss` works for me. Could you show me the issue you are facing?

mangalaman93:
What is the command you are using to run zero?

mangalaman93:
@RJKeevil I can reproduce it now. This is the right fix for it. If you could include it along with a test so that it doesn't happen again, that would be acceptable. Thank you for filing this PR.

```
diff --git a/dgraph/cmd/alpha/run.go b/dgraph/cmd/alpha/run.go
index ec0d8d672..51ed3e347 100644
--- a/dgraph/cmd/alpha/run.go
+++ b/dgraph/cmd/alpha/run.go
@@ -663,9 +663,11 @@ func run() {
        }
 
        keys, err := ee.GetKeys(Alpha.Conf)
-       x.Check(err)
+       if err != nil && !strings.Contains(err.Error(), "enterprise") {
+               x.Check(err)
+       }
 
-       if keys.AclSecretKey != nil {
+       if keys != nil && keys.AclSecretKey != nil {
                opts.AclJwtAlg = keys.AclJwtAlg
                opts.AclSecretKey = keys.AclSecretKey
                opts.AclSecretKeyBytes = keys.AclSecretKeyBytes
@@ -705,17 +707,20 @@ func run() {
                Raft:                raft,
                WhiteListedIPRanges: ips,
                StrictMutations:     opts.MutationsMode == worker.StrictMutations,
-               AclEnabled:          keys.AclSecretKey != nil,
                AbortOlderThan:      abortDur,
                StartTime:           startTime,
                Security:            security,
                TLSClientConfig:     tlsClientConf,
                TLSServerConfig:     tlsServerConf,
-               AclJwtAlg:           keys.AclJwtAlg,
-               AclPublicKey:        keys.AclPublicKey,
                Audit:               opts.Audit != nil,
                Badger:              bopts,
        }
+       if keys != nil {
+               x.WorkerConfig.AclEnabled = keys.AclSecretKey != nil
+               x.WorkerConfig.AclJwtAlg = keys.AclJwtAlg
+               x.WorkerConfig.AclPublicKey = keys.AclPublicKey
+               x.WorkerConfig.EncryptionKey = keys.EncKey
+       }
        x.WorkerConfig.Parse(Alpha.Conf)
 
        if telemetry.GetBool("reports") {
@@ -725,8 +730,6 @@ func run() {
        // Set the directory for temporary buffers.
        z.SetTmpDir(x.WorkerConfig.TmpDir)
 
-       x.WorkerConfig.EncryptionKey = keys.EncKey
-
        setupCustomTokenizers()
        x.Config.PortOffset = Alpha.Conf.GetInt("port_offset")
        x.Config.LimitMutationsNquad = int(x.Config.Limit.GetInt64("mutations-nquad"))
```

RJKeevil:
Hi @mangalaman93, ee.GetKeys gets called in 9 different places (bulk loader, live loader etc etc) so i chose so solve it in get keys instead. Before i make the change, are you sure you prefer the fix on read of this?

mangalaman93:
@RJKeevil good point. Feel free to make changes that pass the test and add new tests for oss. I can review the PR once ready. Thanks

RJKeevil:
@mangalaman93 im unclear on what to do with the tests, it looks like in oss mode it runs some unit tests but never e.g. actually starts an alpha.  Do I e.g. add the ldbc tests to the oss script? Also not the failing load test above is nothing to do with this change, looks like some random instability unfortunately.

mangalaman93:
@RJKeevil you are right. We need to add tests for oss build and if you can help us with that, it would be pretty helpful. Happy to merge the PR once we have at least one test. You could pick one from the existing test and figure out a way to run it for oss build. I am busy with a few other things, but I could come back to it in some time if you can't figure it out.

mangalaman93:
closing this in lieu of https://hypermode.com/blog/the-future-of-dgraph-is-open-serverless-and-ai-ready

PR #9186, opened 2024-10-03
chore(repo): use go 1.22 int range loop

xqqp:
- use the int range loop introduced in go 1.22 so code is less verbose and easier to read
- add `intrange` linter to linter config


mangalaman93 (approved):

mangalaman93:
Thank you for the PR!

PR #9193, opened 2024-10-08
chore(repo): close response bodies to prevent resource leaks

xqqp:
- close response bodies to prevent resource leaks ("The caller must close the response body when finished with it", see https://pkg.go.dev/net/http)
- add `bodyclose` linter to linter config

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -38,3 +38,4 @@ linters:
     - unconvert
     - unused
     - intrange
+    - bodyclose
--- a/contrib/jepsen/main.go
+++ b/contrib/jepsen/main.go
@@ -43,6 +43,7 @@ import (
 	"sync"
 	"time"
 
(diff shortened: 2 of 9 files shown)


mangalaman93 (changes requested):
Could you log the error instead of ignoring it in the `defer`?

xqqp:
@mangalaman93 The errors are now logged outside of test files.

xqqp:
> Thank you for the update. Could you also update to use `%v` instead of `%s` in the glog so that it gets logged correctly. While you do that, you could also keep the structure like this (this is a nit):
> 
> ```
> 		if err := resp.Body.Close(); err != nil {
> 			glog.Errorf("Error while closing response body: %v", err)
> 		}
> ```

Done

mangalaman93 (approved):

PR #9206, opened 2024-10-29
fix(export): delete temporary export directory after completion of remote export

shivaji-kharse:


worker/export.go
```diff
@@ -571,7 +571,9 @@ func (r *remoteExportStorage) FinishWriting(w *Writers) (ExportedFiles, error) {
 			return nil, err
 		}
 	}
-
+	if err = os.RemoveAll(r.les.destination); err != nil {
```
mangalaman93, line 574: what if there is an error before this line? Should we delete this dir still?

worker/export.go
```diff
@@ -549,6 +549,20 @@ func (r *remoteExportStorage) OpenFile(fileName string) (*ExportWriter, error) {
 }
 
 func (r *remoteExportStorage) FinishWriting(w *Writers) (ExportedFiles, error) {
+	var err error
+
+	defer func() {
```
mangalaman93, line 554: we should just log a warning here and move on!

harshil-goel (approved):

PR #9214, opened 2024-11-11
Add index after adding data without vector schema for vector predicate

shivaji-kharse:


tok/hnsw/helper.go
```diff
@@ -378,9 +382,15 @@ func (ph *persistentHNSW[T]) getVecFromUid(uid uint64, c index.CacheType, vec *[
 		return err
 	}
 	if data != nil {
-		index.BytesAsFloatArray(data.([]byte), vec, ph.floatBits)
+		inVec, err := types.ParseVFloat(string(data.([]uint8)))
```
harshil-goel, line 385: We don't want to call ParseVFloat here. It's very slow

posting/index.go
```diff
@@ -187,6 +187,31 @@ func (txn *Txn) addIndexMutations(ctx context.Context, info *indexMutationInfo)
 			}
 			return pbEdges, nil
 		}
+
+		if info.op != pb.DirectedEdge_DEL &&
+			len(data) > 0 &&
+			len(info.factorySpecs) > 0 {
+			// how we can know is this is a schema mutation
+			inVec, err := types.ParseVFloat(string(data[0].Value.([]uint8)))
```
harshil-goel, line 195: Why are you doing this twice

posting/index.go
```diff
@@ -1387,6 +1386,37 @@ func rebuildTokIndex(ctx context.Context, rb *IndexRebuild) error {
...
+					return []*pb.DirectedEdge{}, err
+				}
+
+				if err := p.addMutation(ctx, txn, &edge); err != nil {
+					return []*pb.DirectedEdge{}, err
+				}
+			}
+		}
+
 		err := pl.Iterate(txn.StartTs, 0, func(p *pb.Posting) error {
```
harshil-goel, line 1420: No need to iterate when you already have the value in val. Just run the function inside iterate

harshil-goel (approved):

posting/index.go
```diff
@@ -1387,39 +1386,86 @@ func rebuildTokIndex(ctx context.Context, rb *IndexRebuild) error {
...
+				Value: p.Value,
+				Tid:   types.TypeID(p.ValType),
+			}
+			edge.Lang = string(p.LangTag)
+
+			newEdges, err := processAddIndexMutation(&edge, val)
+			if err != nil {
+				return err
+			}
+			if !runForVectors {
```
harshil-goel, line 1457: remove if

posting/index.go
```diff
@@ -1387,39 +1386,86 @@ func rebuildTokIndex(ctx context.Context, rb *IndexRebuild) error {
...
+				p, err := txn.Get(inKey)
+				if err != nil {
+					return []*pb.DirectedEdge{}, err
+				}
+
+				if err := p.addMutation(ctx, txn, &edge); err != nil {
+					return []*pb.DirectedEdge{}, err
+				}
+
+				_, err = processAddIndexMutation(&edge, val)
```
harshil-goel, line 1436: this should be after the if condition. it should get executed for both when val.td == types.vfloatid.

harshil-goel: write a test that fails because of this bug right now, then fix the bug

shivaji-kharse: TestVectorIndexDropPredicate and TestVectorIndexWithoutSchema have already been added, and they fail on the main branch

harshil-goel (approved):

PR #9230, opened 2024-12-01
fix(core): fix batching of running mutations in one transaction

harshil-goel:


mangalaman93 (changes requested):
Looks pretty good! One minor comment

PR #9236, opened 2024-12-07
fix(core): Improve error handling in hnsw module

xqqp:
Description: 
This PR improves error handling in the hnsw module. In some cases not all errors have to be propagated to the next caller, in this case the error string is currently compared with `strings.Contains(err.Error(), "some_error_text")` or directly `err.Error() == "some_error_string"`. This is bad practice. Instead errors should be directly compared with the native `errors.Is(err, target)` function, which traverses the tree of each error and detects if it contains the specified target error. 

This refactor also fixes two occurrences where the caller of a function compared the error string to a value which could never occur:

https://github.com/dgraph-io/dgraph/pull/9236/files#diff-3a3681ae8c7d51cf49f269f9a732cfe2941dee13903fed04a17a51d5ddff3540R323
https://github.com/dgraph-io/dgraph/pull/9236/files#diff-3a3681ae8c7d51cf49f269f9a732cfe2941dee13903fed04a17a51d5ddff3540R668

### DIFF
--- a/tok/hnsw/helper.go
+++ b/tok/hnsw/helper.go
@@ -43,8 +43,6 @@ const (
 	Euclidean            = "euclidean"
 	Cosine               = "cosine"
 	DotProd              = "dotproduct"
-	plError              = "\nerror fetching posting list for data key: "
-	dataError            = "\nerror fetching data for data key: "
 	EmptyHNSWTreeError   = "HNSW tree has no elements"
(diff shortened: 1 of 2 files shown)


mangalaman93 (approved):

mangalaman93:
Thank you for the PR @xqqp

PR #9252, opened 2025-01-06
add grpc API for creating/deleting/rename/listing namespaces

mangalaman93:


edgraph/server.go
```diff
@@ -1894,6 +1894,218 @@ func (s *Server) CheckVersion(ctx context.Context, c *api.Check) (v *api.Version
...
+		return nil, err
+	}
+
+	if _, err := getNamespaceID(ctx, in.NsName); err == nil {
+		return nil, errors.Errorf("namespace %q already exists", in.NsName)
+	} else if !strings.Contains(err.Error(), "not found") {
+		return nil, err
+	}
+
+	password := "password"
```
harshil-goel, line 1916: Why do we need a default password here?

edgraph/server.go
```diff
@@ -1894,6 +1894,218 @@ func (s *Server) CheckVersion(ctx context.Context, c *api.Check) (v *api.Version
...
+			},
+		}},
+		CommitNow: true,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	var data struct {
+		Data []struct {
```
harshil-goel, line 2013: We can make a global struct for it?

edgraph/server.go
```diff
@@ -1894,6 +1894,218 @@ func (s *Server) CheckVersion(ctx context.Context, c *api.Check) (v *api.Version
...
+	if err != nil {
+		return nil, err
+	}
+
+	glog.Infof("Created namespace [%v] with id [%d]", in.NsName, ns)
+	return &api.CreateNamespaceResponse{NsId: ns}, nil
+}
+
+func (s *Server) DropNamespace(ctx context.Context, in *api.DropNamespaceRequest) (
+	*api.DropNamespaceResponse, error) {
```
harshil-goel, line 1958: Do we need response if there's nothing in it?

edgraph/server.go
mangalaman93: In case someone passes ab empty string here.

edgraph/server.go
mangalaman93: This is usually a good idea for future extensibility so that we can modify the response if needed. This will be mostly hidden from the user through the client APIs that we expose.

dgraphtest/image.go
```diff
@@ -52,6 +61,9 @@ func (c *LocalCluster) setupBinary() error {
 }
 
 func ensureDgraphClone() error {
+	cloneMutex.Lock()
```
harshil-goel, line 64: shouldn't this be mutex.once()?

edgraph/server.go
```diff
@@ -2005,9 +2356,9 @@ func validateAndConvertFacets(nquads []*api.NQuad) error {
 	return nil
 }
 
-// validateForGraphql validate nquads for graphql
-func validateForGraphql(nq *api.NQuad, isGraphql bool) error {
-	// Check whether the incoming predicate is graphql reserved predicate or not.
+// validateForOtherReserved validate nquads for other reserved predicates
```
harshil-goel, line 2363: What are the others? Can you explain a bit here

edgraph/server.go
```diff
@@ -1883,6 +1884,356 @@ func (s *Server) CheckVersion(ctx context.Context, c *api.Check) (v *api.Version
...
+				Set: []*api.NQuad{
+					{
+						Subject:     "_:ns",
+						Predicate:   "dgraph.namespace.name",
+						ObjectValue: &api.Value{Val: &api.Value_StrVal{StrVal: nsName}},
+					},
+					{
+						Subject:     "_:ns",
+						Predicate:   "dgraph.namespace.id",
+						ObjectValue: &api.Value{Val: &api.Value_IntVal{IntVal: int64(nsID)}},
```
harshil-goel, line 2196: I think you can give StrVal itself instead of converting it youself. I think dgraph would be able to handle it.

edgraph/server.go
```diff
@@ -1883,6 +1884,356 @@ func (s *Server) CheckVersion(ctx context.Context, c *api.Check) (v *api.Version
...
+func getNamespaceName(ctx context.Context, nsID uint64) (string, error) {
+	const q = `{q(func: eq(dgraph.namespace.id, "%v")) { dgraph.namespace.name }}`
+	req := &api.Request{Query: fmt.Sprintf(q, nsID)}
+	resp, err := (&Server{}).doQuery(ctx, &Request{req: req, doAuth: NoAuthorize})
+	if err != nil {
+		return "", err
+	}
+
+	var data struct {
+		Data []struct {
```
harshil-goel, line 2163: Make it one function instead of two. get booth name and id at the same time. Then you can get name or id whatever you want out of it

edgraph/server.go
```diff
@@ -1883,6 +1884,356 @@ func (s *Server) CheckVersion(ctx context.Context, c *api.Check) (v *api.Version
...
+		return fmt.Errorf("namespace name [%v] cannot be a number", name)
+	}
+	return nil
+}
+
+func verifyDropNamespaceName(name string) error {
+	if name == "" {
+		return errors.Errorf("namespace name cannot be empty")
+	}
+	if name == "system" || name == "galaxy" || name == "dgraph-0" {
```
harshil-goel, line 2124: what is dgraph-0. Is it being used in hypermode? If so can we make some enum out of it so that it's easy to keep track

edgraph/server.go
```diff
@@ -1883,6 +1884,356 @@ func (s *Server) CheckVersion(ctx context.Context, c *api.Check) (v *api.Version
...
+	const q = `{q(func: type(dgraph.namespace)) {
+			dgraph.namespace.name
+			dgraph.namespace.id }}`
+	resp, err := (&Server{}).doQuery(x.AttachJWTNamespace(ctx),
+		&Request{req: &api.Request{Query: q}, doAuth: NoAuthorize})
+	if err != nil {
+		return nil, err
+	}
+	var data struct {
+		NsList []struct {
```
harshil-goel, line 2066: Lets just make a global structure. Why are we declaring it again and again.

edgraph/server.go
```diff
@@ -1883,6 +1884,356 @@ func (s *Server) CheckVersion(ctx context.Context, c *api.Check) (v *api.Version
...
+func (s *Server) ListNamespaces(ctx context.Context, in *apiv25.ListNamespacesRequest) (
+	*apiv25.ListNamespacesResponse, error) {
+
+	if err := AuthGuardianOfTheGalaxy(ctx); err != nil {
+		s := status.Convert(err)
+		return nil, status.Error(s.Code(),
+			"Non guardian of galaxy user cannot list namespaces. "+s.Message())
+	}
+
+	const q = `{q(func: type(dgraph.namespace)) {
```
harshil-goel, line 2057: Can we also make the query constant?

edgraph/server.go
mangalaman93: Defined here https://github.com/hypermodeinc/dgraph/blob/main/x/keys.go#L663

edgraph/server.go
mangalaman93: but we already have it as int, what would be the benefit of doing that?

edgraph/server.go
mangalaman93: That is the convention we are going to follow for pre v25 namespaces. Read here https://www.notion.so/hypermodeinc/Preview-Dgraph-v25-19de0dd4f736808bb1d7ffe98d38a3c8

edgraph/server.go
mangalaman93: Done

edgraph/server.go
mangalaman93: It is a constant. Because it is only used once, I have not defined it at the top.

harshil-goel (dismissed):

edgraph/server.go
harshil-goel: I meant, can you add a comment there for future?

edgraph/server.go
harshil-goel: copilot was crying over something there right?

edgraph/server.go
harshil-goel: Lets create the enum?

edgraph/server.go
mangalaman93: that's fixed now

harshil-goel (approved):

PR #9259, opened 2025-01-08
fix: ensure consistent time units and prevent erroneous cleanup in incrRollupi Process

gooohgb:
**Description**

This PR fixes a bug in the Process function of incrRollupi. The issue was caused by inconsistent time units between the doRollup and cleanupTick logic. Specifically, doRollup used time.Now().Unix() (seconds), while cleanupTick used time.Now().UnixNano() (nanoseconds). This mismatch caused cleanupTick to erroneously delete all entries in the map m during every cleanup cycle.

**Checklist**


mangalaman93 (approved):
Good catch, thank you for fixing it.

harshil-goel (approved):

PR #9271, opened 2025-01-09
Use Intersect to Narrow Iterate Range and Reduce Memory Allocation

gooohgb:
**Description**

In our index data, some hot keys are associated with a large number of UIDs, but the set filtered by our function is relatively small. During performance testing on this dataset, I noticed that pl.Uids consumes a significant amount of CPU time for slice memory allocation. Therefore, I propose an optimization in the Uids function to leverage the range provided by Intersect to reduce the scope of temporary result sets, thereby minimizing memory allocation.

### DIFF
--- a/posting/list.go
+++ b/posting/list.go
@@ -1706,10 +1706,15 @@ func (l *List) Uids(opt ListOptions) (*pb.List, error) {
 	}
 
 	if opt.Intersect != nil && len(opt.Intersect.Uids) < l.mutationMap.len()+codec.ApproxLen(l.plist.Pack) {
-		for _, uid := range opt.Intersect.Uids {
-			if uid <= opt.AfterUid {
-				continue
-			}
+		start := 0
+		if opt.AfterUid > 0 {
+			start = sort.Search(len(opt.Intersect.Uids), func(i int) bool {
+				return opt.Intersect.Uids[i] > opt.AfterUid
+			})
+		}
+
+		for i := start; i < len(opt.Intersect.Uids); i++ {
+			uid := opt.Intersect.Uids[i]
 
 			found, _, err := l.findPosting(opt.ReadTs, uid)
 			if err != nil {
--- a/posting/list.go
+++ b/posting/list.go
@@ -1685,7 +1685,11 @@ func (l *List) Uids(opt ListOptions) (*pb.List, error) {
 		return out, nil
 	}
 
-	preAllowcateLength := x.MinInt(opt.First, l.mutationMap.len()+codec.ApproxLen(l.plist.Pack))
+	absFirst := opt.First
+	if opt.First < 0 {
+		absFirst = -opt.First
+	}
+	preAllowcateLength := x.MinInt(absFirst, l.mutationMap.len()+codec.ApproxLen(l.plist.Pack))
 	if opt.Intersect != nil {
 		preAllowcateLength = x.MinInt(preAllowcateLength, len(opt.Intersect.Uids))
 	}
--- a/posting/list.go
+++ b/posting/list.go
@@ -1672,10 +1672,8 @@ func (l *List) Uids(opt ListOptions) (*pb.List, error) {
 	if opt.First == 0 {
 		opt.First = math.MaxInt32
 	}
-	// Pre-assign length to make it faster.
 	l.RLock()
-	// Use approximate length for initial capacity.
-	res := make([]uint64, 0, l.mutationMap.len()+codec.ApproxLen(l.plist.Pack))
+
 	out := &pb.List{}
 	if l.mutationMap.len() == 0 && opt.Intersect != nil && len(l.plist.Splits) == 0 {
 		if opt.ReadTs < l.minTs {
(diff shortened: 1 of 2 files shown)


harshil-goel:
Thanks a lot for the input. We have recently upgraded how the list works, and now we have an even better function that you could use instead. Right now there would be an issue if your uid range is too high, but still the numbers are too low. Now you can basically just call list.FindPosting(uid) to see if uid is present or not. But we shouldn't use it if uid intersection range is bigger than uids actually present. So we could even add a check using ApproxLen(). Let me know if you are willing to make the change, or should we?

gooohgb:
Thanks for your suggestion! I'm glad to contribute to this project. I’d like to make this change myself.

gooohgb:
> Thanks a lot for the input. We have recently upgraded how the list works, and now we have an even better function that you could use instead. Right now there would be an issue if your uid range is too high, but still the numbers are too low. Now you can basically just call list.FindPosting(uid) to see if uid is present or not. But we shouldn't use it if uid intersection range is bigger than uids actually present. So we could even add a check using ApproxLen(). Let me know if you are willing to make the change, or should we?

By the way, I’d like to ask if this PR (https://github.com/hypermodeinc/dgraph/pull/9218) is likely to be merged. The optimization of the eq execution plan seems to significantly improve performance.

gooohgb:
> Thanks a lot for the input. We have recently upgraded how the list works, and now we have an even better function that you could use instead. Right now there would be an issue if your uid range is too high, but still the numbers are too low. Now you can basically just call list.FindPosting(uid) to see if uid is present or not. But we shouldn't use it if uid intersection range is bigger than uids actually present. So we could even add a check using ApproxLen(). Let me know if you are willing to make the change, or should we?

I’ve pushed the updated changes to the PR. When you have time, could you kindly review it again? Let me know if there are any additional improvements you'd like to see.

x/x.go
```diff
@@ -619,6 +619,38 @@ func Max(a, b uint64) uint64 {
 	return b
 }
 
+// MinInt returns the smallest integer among the given numbers.
+// The first two arguments are mandatory, additional numbers are optional.
+func MinInt(a, b int, nums ...int) int {
```
xqqp, line 624: Since go 1.21 `min()` and `max()` are natively available. So no need to implement them yourself.

gooohgb: Thanks for your advice, I have updated the method to use the native implementation as suggested.

posting/list.go
```diff
@@ -1687,16 +1685,62 @@ func (l *List) Uids(opt ListOptions) (*pb.List, error) {
...
+				return opt.Intersect.Uids[i] > opt.AfterUid
+			})
+		}
+
+		for i := start; i < len(opt.Intersect.Uids); i++ {
+			uid := opt.Intersect.Uids[i]
+
+			found, _, err := l.findPosting(opt.ReadTs, uid)
+			if err != nil {
+				l.RUnlock()
```
xqqp, line 1725: Instead of manually calling `RUnlock()` each time a return happens, you could do `defer l.RUnlock()` at the start of the function for the same effect.

gooohgb: > Instead of manually calling `RUnlock()` each time a return happens, you could do `defer l.RUnlock()` at the start of the function for the same effect.

When implementing this code, I also considered making this change, but I referred to the previous implementation and believe it was designed to minimize the time spent holding the read lock. Therefore, I chose not to change this implementation.

harshil-goel: It's fine, you can do defer. Also I don't see any place where you have actually aquired the lock.

posting/list.go
```diff
@@ -1687,16 +1685,62 @@ func (l *List) Uids(opt ListOptions) (*pb.List, error) {
 		return out, nil
 	}
 
+	absFirst := opt.First
+	if opt.First < 0 {
+		absFirst = -opt.First
```
harshil-goel, line 1690: Isn't opt.first uint64? Making it negative doesn't seem like a good idea.

posting/list.go
```diff
@@ -1687,16 +1685,62 @@ func (l *List) Uids(opt ListOptions) (*pb.List, error) {
...
+	}
+	preAllowcateLength := min(absFirst, l.mutationMap.len()+codec.ApproxLen(l.plist.Pack))
```
harshil-goel, line 1692: I think we have a function for it

harshil-goel: Also the actual length could be way more for some postings (split postings). Lets use the exact length function here. It would only hurt in case of split postings. We can resolve that later.

posting/list.go
```diff
@@ -1687,16 +1685,62 @@ func (l *List) Uids(opt ListOptions) (*pb.List, error) {
...
+			if len(res) > -opt.First {
+				res = res[1:]
+			}
+		} else if len(res) > opt.First {
+			return true
+		}
+		return false
+	}
+
+	if opt.Intersect != nil && len(opt.Intersect.Uids) < l.mutationMap.len()+codec.ApproxLen(l.plist.Pack) {
```
harshil-goel, line 1712: same: use exact length. 
lets create another function inside list that intersects with the list of uids.

harshil-goel:
Merged as a part of the other pr. Thanks a lot for your contribution.

PR #9282, opened 2025-01-21
add internal predicate for string namespaces

mangalaman93:
This PR adds two internal predicates and one internal type needed for supporting namespaces with string names. We are creating a separate PR so that we can take this PR as model PR for adding more internal predicate in future.


.vscode/settings.json
```diff
@@ -2,5 +2,8 @@
   "editor.formatOnSave": true,
   "editor.defaultFormatter": "trunk.io",
   "editor.trimAutoWhitespace": true,
-  "trunk.autoInit": false
+  "trunk.autoInit": false,
+  "[go]": {
+    "editor.defaultFormatter": "golang.go"
```
ryanfoxtyler, line 7: this will remove some of Trunk's capabilities and shouldn't be necessary...are you seeing any issues?

mangalaman93: In local, it slows down saving files. Sometimes it even got stuck when my internet was slow.

worker/online_restore.go
```diff
@@ -412,6 +412,8 @@ func handleRestoreProposal(ctx context.Context, req *pb.RestoreRequest, pidx uin
 
 	posting.ResetCache()
 	ResetAclCache()
+	groups().applyInitialSchema()
+	groups().applyInitialTypes()
```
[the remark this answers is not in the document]
mangalaman93: because after restore from an older version, we need to ensure that initialTypes and predicates that are introduced in the latest version are applied.

PR #9289, opened 2025-01-24
perf(algo): single memory allocation in NewCountMinSketch

arturmelanchyk:
**Description**

Please explain the changes you made here.

**Checklist**

- [ ] Code compiles correctly and linting passes locally
      this PR
- [ ] For public APIs, new features, etc., PR on
      [docs repo](https://github.com/dgraph-io/dgraph-docs) staged and linked here

**Instructions**

- The PR title should follow the [Conventional Commits](https://www.conventionalcommits.org/)
  syntax, leading with `fix:`, `feat:`, `chore:`, `ci:`, etc.
- The description should briefly explain what the PR is about. In the case of a bugfix, describe or
  link to the bug.
- In the checklist section, check the boxes in that are applicable, using `[x]` syntax.
  - If not applicable, remove the entire line. Only leave the box unchecked if you intend to come
    back and check the box later.
- Delete the `Instructions` line and everything below it, to indicate you have read and are
  following these instructions. 🙂

Thank you for your contribution to Dgraph!

### DIFF
--- a/algo/cm-sketch.go
+++ b/algo/cm-sketch.go
@@ -78,8 +78,9 @@ func NewCountMinSketch(epsilon, delta float64) *CountMinSketch {
 		matrix = make([][]uint64, depth)
 	)
 
+	flatMatrix := make([]uint64, width*depth)
 	for i := uint(0); i < depth; i++ {
-		matrix[i] = make([]uint64, width)
+		matrix[i] = flatMatrix[i*width : (i+1)*width : (i+1)*width]
 	}
 
 	return &CountMinSketch{


harshil-goel:
Can you rebase to latest main?

algo/cm-sketch.go
```diff
@@ -67,8 +67,9 @@ func NewCountMinSketch(epsilon, delta float64) *CountMinSketch {
 		matrix = make([][]uint64, depth)
 	)
 
+	flatMatrix := make([]uint64, width*depth)
 	for i := uint(0); i < depth; i++ {
-		matrix[i] = make([]uint64, width)
+		matrix[i] = flatMatrix[i*width : (i+1)*width : (i+1)*width]
```
mangalaman93, line 72: why are we setting the capacity to (i+1)*width here? 
reference: https://tip.golang.org/doc/go1.2#three_index

mangalaman93 (changes requested):
Thank you for the PR. One comment.

algo/cm-sketch.go
arturmelanchyk: Hi @mangalaman93 
let me put answer from chatGPT here:
```
Why Use (i+1)*width for Capacity?

When slicing in Go, the syntax is:

slice[start:end:max]

    start: The starting index (inclusive).
    end: The ending index (exclusive) – determines the length of the slice.
    max: The maximum index (exclusive) – determines the capacity of the slice.

Now, if we used:

matrix[i] = rawData[i*width : (i+1)*width : width] // Incorrect

It would set the maximum capacity (cap(matrix[i])) to width, but this is not relative to i*width. This would cause an out-of-bounds error because Go calculates the capacity relative to start, meaning:

cap(matrix[i]) = max - start

So if max = width, and start = i*width, we'd get a negative capacity or incorrect slicing.
Correct Calculation:

matrix[i] = rawData[i*width : (i+1)*width : (i+1)*width]

    i*width → Start of row i
    (i+1)*width → End of row i
    (i+1)*width → Max capacity is exactly the next row's start, preventing access beyond row i.

What Happens if We Use Only width as Capacity?

If we wrote:

matrix[i] = rawData[i*width : (i+1)*width : width] // Incorrect

    Go would interpret width as an absolute index (not relative to start).
    Since width is smaller than i*width in most cases (for i > 0), it wouldn't work as intended.

Conclusion

We use (i+1)*width to set the capacity limit exactly at the end of the row and prevent accidental access to other rows' data while maintaining memory efficiency.
```

In order to prove this try running the following test (it will panic with `panic: runtime error: slice bounds out of range`):
```
func TestSlice(t *testing.T) {
	width := 2
	depth := 2

	matrix := make([][]uint64, depth)
	flatMatrix := make([]uint64, width*depth)
	for i := 0; i < depth; i++ {
		matrix[i] = flatMatrix[i*width : (i+1)*width : width]
	}
}
```

mangalaman93: Thanks for educating me. You are right.

mangalaman93 (approved):

PR #9315, opened 2025-02-20
Switch to s2 for bulk loader and backup/restore

RJKeevil:
**Description**

A small change, but i noticed https://github.com/klauspost/compress was already an indirect dependency of the repo, so this uses that version over the golang version as a drop in replacement. According to the repo it offers "better compression and concurrent streams".

**Checklist**

### DIFF
--- a/dgraph/cmd/bulk/mapper.go
+++ b/dgraph/cmd/bulk/mapper.go
@@ -20,7 +20,7 @@ import (
 
 	farm "github.com/dgryski/go-farm"
 	"github.com/golang/glog"
-	"github.com/golang/snappy"
+	"github.com/klauspost/compress/s2"
 	"google.golang.org/protobuf/proto"
 
 	"github.com/dgraph-io/dgo/v240/protos/api"
@@ -165,7 +165,7 @@ func (m *mapper) writeMapEntriesToFile(cbuf *z.Buffer, shardIdx int) {
 		x.Check(f.Close())
 	}()
 
-	w := snappy.NewBufferedWriter(f)
+	w := s2.NewWriter(f)
 	defer func() {
 		x.Check(w.Close())
 	}()
--- a/dgraph/cmd/bulk/reduce.go
+++ b/dgraph/cmd/bulk/reduce.go
@@ -24,7 +24,7 @@ import (
 
 	"github.com/dustin/go-humanize"
 	"github.com/golang/glog"
-	"github.com/golang/snappy"
+	"github.com/klauspost/compress/s2"
 	"google.golang.org/protobuf/proto"
 
 	"github.com/dgraph-io/badger/v4"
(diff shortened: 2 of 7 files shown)


mangalaman93:
Thank you for the PR @RJKeevil. I don't see a direct benefit of doing this. It doesn't change dependency tree much. Performance for snappy is not clear that it improves it significantly either. Am I missing something?

RJKeevil:
Yeh its a minor one, you had both dependencies anyway so I just thought it might be best to default to the one that claims to be faster and seems to receive more bug fixes and improvements.  Where it becomes interesting is it would allow you to consider more compression modes in future, such as s2, which is a more significantly improved snappy.

mangalaman93:
For Badger, we already use s2 instead of snappy. I think we can do that here too https://github.com/hypermodeinc/badger/blob/main/table/builder.go#L151

For connections, not sure what the best option is. I wouldn't change that unless we have a solid reasoning for it.

conn/snappy.go
```diff
@@ -18,7 +18,7 @@ import (
 	"io"
 	"sync"
 
-	"github.com/golang/snappy"
+	"github.com/klauspost/compress/snappy"
```
mangalaman93, line 21: undo this change.

dgraph/cmd/bulk/mapper.go
```diff
@@ -20,7 +20,7 @@ import (
 
 	farm "github.com/dgryski/go-farm"
 	"github.com/golang/glog"
-	"github.com/golang/snappy"
+	"github.com/klauspost/compress/snappy"
```
mangalaman93, line 23: we should move to s2 here. Badger already is using s2, we can align with that here.

dgraph/cmd/bulk/reduce.go
```diff
@@ -24,7 +24,7 @@ import (
 
 	"github.com/dustin/go-humanize"
 	"github.com/golang/glog"
-	"github.com/golang/snappy"
+	"github.com/klauspost/compress/snappy"
```
mangalaman93, line 27: move to s2

worker/backup_ee.go
```diff
@@ -20,7 +20,7 @@ import (
 	"time"
 
 	"github.com/golang/glog"
-	"github.com/golang/snappy"
+	"github.com/klauspost/compress/snappy"
```
mangalaman93, line 23: For backup, we could move to using s2 given s2 is backward compatible with snappy.

worker/restore_map.go
```diff
@@ -28,7 +28,7 @@ import (
 
 	"github.com/dustin/go-humanize"
 	"github.com/golang/glog"
-	"github.com/golang/snappy"
+	"github.com/klauspost/compress/snappy"
```
mangalaman93, line 31: Same here, we could move to using s2 here too.

worker/restore_reduce.go
```diff
@@ -22,7 +22,7 @@ import (
 
 	"github.com/dustin/go-humanize"
 	"github.com/golang/glog"
-	"github.com/golang/snappy"
+	"github.com/klauspost/compress/snappy"
```
mangalaman93, line 25: move to using s2 here.

RJKeevil:
@mangalaman93 made those changes

mangalaman93 (approved):

mangalaman93:
@RJKeevil could you rebase on latest main? We merged a fix so that the failing tests pass.

mangalaman93 (approved):

PR #9319, opened 2025-02-25
fix(core): change delete-on-updates to remove-on-update and set default false

harshil-goel:


dgraph/cmd/alpha/run.go
```diff
@@ -134,12 +134,15 @@ they form a Raft group and provide synchronous replication.
 		Flag("percentage",
 			"Cache percentages summing up to 100 for various caches (FORMAT: PostingListCache,"+
 				"PstoreBlockCache,PstoreIndexCache)").
-		Flag("delete-on-updates",
-			"When set as true, we would delete the key from the cache once it's updated. If it's not "+
-				"we would update the value inside the cache. If the cache gets too full, it starts"+
-				" to get slow. So if your usecase has a lot of heavy mutations, this should be set"+
-				" as true. If you are modifying same data again and again, this should be set as"+
-				" false").
+		Flag("remove-on-update",
```
mangalaman93, line 137: 1. I would prefer, `delete-on-update` instead of `remove-on-update` because that aligns with database terminology better and I think it is clearer.
2. This explanation does not make clear what happens on true or false
3. There is no need to say "during parallel transactions"
4. remove --> delete

mangalaman93 (changes requested):

mangalaman93 (approved):

mangalaman93 (approved):

PR #9343, opened 2025-03-12
Remove sentry integration

mangalaman93:
This PR removes the sentry integration from both alpha and zero. We keep the telemetry flag around so that existing deployment of dgraph do not break.


matthewmcneely:
Hey @mangalaman93 Happy to see this! Should we add a comment in flags.go to indicate that sentry is deprecated?

harshil-goel (approved):

mangalaman93:
> Hey @mangalaman93 Happy to see this! Should we add a comment in flags.go to indicate that sentry is deprecated?

Please check flags.go, it is already updated!

matthewmcneely:
Ah, you're right. Sorry I missed that!

PR #9375, opened 2025-04-14
fix(core): delete existing plists when doing a single UID update

xqqp:
If a single UID predicate (not a list) is updated, the old value is not deleted. This bug was introduced in https://github.com/hypermodeinc/dgraph/pull/9292. 

This PR adds the old behavior back.

Related: https://discuss.hypermode.com/t/inconsistent-query-result/19844/7

Note: I don't know why this was changed in the original PR as the existing comment in the code says that the loop is necessary, so please review carefully.

### DIFF
--- a/posting/list.go
+++ b/posting/list.go
@@ -773,7 +773,12 @@ func (l *List) updateMutationLayer(mpost *pb.Posting, singleUidUpdate, hasCountI
 		// Add the deletions in the existing plist because those postings are not picked
 		// up by iterating. Not doing so would result in delete operations that are not
 		// applied when the transaction is committed.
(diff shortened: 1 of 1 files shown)


harshil-goel:
Hey, thanks a lot for your diff. You saved me precious time trying to track down which change caused the issue. The reason I removed it in first place was that I thought iterate() would fix it. But I guess there were some corner cases. A simple fix was you could just iterate on m.startTs - 1. I checked, it's also passing all the tests. However we are going a different direction. I don't to keep thinking about the corner cases, we are just simply going to use deleteAll to delete all the old values fixing the issue.

PR #9382, opened 2025-04-15
fix(core): Add maxfrontiersize to shortest path query

harshil-goel:
Currently we use dijkstra's algorithm to find shortest path. Sometimes the heap could grow indefinitely, which would slow down the query significantly and cause dgraph to go out of memory. This would only happen in very specific instances depending upon the graph. To fix this, we are introducing a new argument, maxfrontiersize in shortest path query. This would limit the size of the heap fixing the issue. 
New query would look like:
```
{ 
  path as shortest(from: 0x2, to: 0x5, numpaths:5, maxfrontiersize: 10000) {
  connected_to
    @facets(weight)
 }
```
Fixes: https://github.com/hypermodeinc/dgraph/issues/9333


mangalaman93 (changes requested):
Please add a test for this change

matthewmcneely:
@harshil-goel Is "heap" the best term for this? Is it the max number of nodes to be considered? Maybe "maxnodes"?

harshil-goel:
@mangalaman93 @matthewmcneely What term do you think works well:
maxFrontierSize - "Frontier" is a common term in graph search for the set of nodes to be explored next
pathExplorationLimit - Describes the purpose rather than the implementation
maxCandidatePaths - Indicates you're limiting the number of candidate paths under consideration

matthewmcneely:
I like "frontier size", that lines up nicely with terminology that I've encountered over the years.

query/query.go
```diff
@@ -165,6 +165,8 @@ type params struct {
 	MaxWeight float64
 	// MinWeight is the min weight allowed in a path returned by the shortest path algorithm.
 	MinWeight float64
+	// MaxFrontierSize is the maximum size of the frontier that can be made while querying shortest path
```
mangalaman93, line 168: is it possible to provide a link to understand this better?

systest/shortest-path/shortest_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/hypermodeinc/dgraph/v25/dgraphapi"
+	"github.com/hypermodeinc/dgraph/v25/dgraphtest"
+	"github.com/hypermodeinc/dgraph/v25/x"
+	"github.com/stretchr/testify/require"
```
mangalaman93, line 18: nit: can separate non-dgraph and dgraph imports

mangalaman93 (approved):

PR #9386, opened 2025-04-16
fix(core): use passed context when joining cluster

xqqp:
This PR addresses an issue where the function does not propagate the passed context, resulting in the creation of a new context (`context.Background`). As a result, context timeouts and cancellations do not work. This update ensures proper context handling.

### DIFF
--- a/conn/node.go
+++ b/conn/node.go
@@ -771,7 +771,7 @@ func (n *Node) joinCluster(ctx context.Context, rc *pb.RaftContext) (*api.Payloa
 	}
 	n.Connect(rc.Id, rc.Addr)
 
-	err := n.addToCluster(context.Background(), rc)
+	err := n.addToCluster(ctx, rc)
 	glog.Infof("[%#x] Done joining cluster with err: %v", rc.Id, err)
 	return &api.Payload{}, err
 }


xqqp:
The CI fails with:
```
panic: runtime error: index out of range [1896] with length 1891
 
goroutine 1153 [running]:
github.com/hypermodeinc/dgraph/v24/tok/hnsw.decodeUint64MatrixUnsafe({0xc0028c2800, 0x763, 0xa?}, 0xc000a631e8)
```

I don't think the failure is related to this change.

conn/node.go
```diff
@@ -771,7 +771,7 @@ func (n *Node) joinCluster(ctx context.Context, rc *pb.RaftContext) (*api.Payloa
 	}
 	n.Connect(rc.Id, rc.Addr)
 
-	err := n.addToCluster(context.Background(), rc)
```
mangalaman93, line 774: I am not so sure that this is a good idea. The request from the user may time out  but we still would like to continue joining the cluster and not stop it in the middle of it.

mangalaman93 (changes requested):

conn/node.go
xqqp: Then how about removing the context from the function prototype? It is unused otherwise.

mangalaman93: It's not a big deal I believe. We can leave it as is unless it affects anything interesting.

PR #9400, opened 2025-04-30
Add implementation for AllocateIDs and use the API in live loader

mangalaman93:
This PR implements the AllocateIDs API for allocating UIDs, namespaces and timestamps for bulk and live loader. For now, this PR also updates the live loader to use this API and completely avoid talking to zero directly.

### DIFF
--- a/worker/mutation.go
+++ b/worker/mutation.go
@@ -609,7 +609,7 @@ func AssignNsIdsOverNetwork(ctx context.Context, num *pb.Num) (*pb.AssignedIds, 
 	return c.AssignIds(ctx, num)
 }
 
-// AssignUidsOverNetwork sends a request to assign UIDs to blank nodes to the current zero leader.
+// AssignUidsOverNetwork sends a request to assign UIDs from the current zero leader.
 func AssignUidsOverNetwork(ctx context.Context, num *pb.Num) (*pb.AssignedIds, error) {
 	// Pass on the incoming metadata to the zero. Namespace from the metadata is required by zero.
 	if md, ok := metadata.FromIncomingContext(ctx); ok {
--- a/xidmap/xidmap.go
+++ b/xidmap/xidmap.go
@@ -136,7 +136,10 @@ func New(opts XidMapOptions) *XidMap {
 		})
 		x.Check(err)
 	}
-	xm.zc = pb.NewZeroClient(opts.UidAssigner)
+
+	if opts.UidAssigner != nil {
+		xm.zc = pb.NewZeroClient(opts.UidAssigner)
+	}
 
 	go func() {
 		const initBackoff = 10 * time.Millisecond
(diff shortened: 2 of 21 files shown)


worker/mutation.go
```diff
@@ -609,8 +609,8 @@ func AssignNsIdsOverNetwork(ctx context.Context, num *pb.Num) (*pb.AssignedIds,
 	return c.AssignIds(ctx, num)
 }
 
-// AssignUidsOverNetwork sends a request to assign UIDs to blank nodes to the current zero leader.
-func AssignUidsOverNetwork(ctx context.Context, num *pb.Num) (*pb.AssignedIds, error) {
+// AssignIDsOverNetwork sends a request to assign UIDs/TSs from the current zero leader.
+func AssignIDsOverNetwork(ctx context.Context, num *pb.Num) (*pb.AssignedIds, error) {
```
harshil-goel, line 613: Maybe you need to remove num.TYpe = pb.Num_UID from this function? Also remove the AssignNsIdsOverNetwork function?

worker/mutation.go
```diff
@@ -609,8 +609,8 @@ func AssignNsIdsOverNetwork(ctx context.Context, num *pb.Num) (*pb.AssignedIds,
 	return c.AssignIds(ctx, num)
 }
 
-// AssignUidsOverNetwork sends a request to assign UIDs to blank nodes to the current zero leader.
-func AssignUidsOverNetwork(ctx context.Context, num *pb.Num) (*pb.AssignedIds, error) {
+// AssignIDsOverNetwork sends a request to assign UIDs/TSs from the current zero leader.
```
harshil-goel, line 612: Also NS

xidmap/xidmap.go
```diff
@@ -136,7 +136,10 @@ func New(opts XidMapOptions) *XidMap {
 		})
 		x.Check(err)
 	}
-	xm.zc = pb.NewZeroClient(opts.UidAssigner)
+
```
harshil-goel, line 139: Do we need this? I am guessing it might be needed for running bulk loader?

harshil-goel: Can we rewrite the code such that it's a function that gives uids, and it uses zero if it's present or alpha if it's present? Instead of adding the logic everywhere it's used.

graphql/admin/assign.go
```diff
@@ -39,7 +39,7 @@ func resolveAssign(ctx context.Context, m schema.Mutation) (*resolve.Resolved, b
 	num := &pb.Num{Val: input.Num}
 	switch input.What {
 	case uid:
-		resp, err = worker.AssignUidsOverNetwork(ctx, num)
+		resp, err = worker.AssignIDsOverNetwork(ctx, num)
```
harshil-goel, line 42: You will need to assign Type inside num, and now the whole switch case can be simplified

harshil-goel (approved):

PR #9403, opened 2025-05-08
fix(core): fix reverse edge loss when set and delete occur together

gooohgb:
This PR fixes a bug where a reverse edge (`@reverse` predicate) with a count index (`@count`) may lose values when the same predicate is set and deleted multiple times within a single DQL mutation.


fivemtebex:
Impressive project structure! The architecture really demonstrates solid design principles.

harshil-goel (approved):
Thanks a lot for the fix

PR #9431, opened 2025-06-02
perf(core): Add sharded map for post query speedup

harshil-goel:


types/sharded_map.go
```diff
@@ -0,0 +1,110 @@
...
+// Package types contains some very common utilities used by Dgraph. These utilities
+// are of "miscellaneous" nature, e.g., error checking.
+package types
+
+import (
+	"sync"
+)
+
+type ShardedMap struct {
+	Shards []map[uint64]Val
```
mangalaman93, line 15: can we not export it, it seems that we are only using it for finding the number of shards?

types/sharded_map.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func (s *ShardedMap) init() {
+	if s == nil {
+		*s = *NewShardedMap()
+	}
+}
+
+func (s *ShardedMap) getShard(key uint64) map[uint64]Val {
+	return s.Shards[key%uint64(len(s.Shards))]
```
mangalaman93, line 67: we can use the constant defined instead

types/sharded_map.go
```diff
@@ -0,0 +1,110 @@
...
+		wg.Add(1)
+		go func(i int) {
+			for k, v := range s.Shards[i] {
+				val, ok := other.Shards[i][k]
+				if !ok {
+					continue
+				}
+				s.Shards[i][k] = ag(val, v)
+			}
+			for k, v := range other.Shards[i] {
```
mangalaman93, line 40: can we not do just one iteration over other.Shards[i] and merge if found in s.Shards[i] otherwise, write in s.Shards[i]

mangalaman93 (changes requested):

PR #9439, opened 2025-06-12
perf(core): improve performance with has filter with order

ghost:


RJKeevil:
Note that github.com/golang/protobuf is deprecated, according to its page it's superceded google.golang.org/protobuf. I saw it becoming a direct dependency in this branch.

ghost:
@RJKeevil Yes that's why the tests are failing. We are trying to fix that.

PR #9469, opened 2025-07-16
feat(core): Add new index PartitionedHNSW for vectors

ghost:
## Summary by CodeRabbit

* **New Features**
  * Added support for partitioned vector indexes with scalable indexing and configurable search probing.
  * Added support for partitioned vector indexes in bulk loading, backups, restores, exports, restarts, and rollups.
  * Added persisted vector-dimension metadata and improved vector index lifecycle handling.

* **Bug Fixes**
  * Improved reliability when vectors are added, updated, or deleted during index rebuilds.
  * Prevented invalid vector dimensions and inconsistent schema updates.
  * Improved vector search recovery after transient storage errors and restarts.

* **Testing**
  * Expanded coverage for partitioned vector indexing, rebuilds, backups, restores, and migration scenarios.
  * Scheduled extended vector tests for nightly CI runs.


matthewmcneely:
Regarding the index building/mutation race issue:

Catching in-flight vector mutations and replaying them after the build completes is the right shape. It's essentially what Postgres does with `CREATE INDEX CONCURRENTLY`. But catching alone won't fix this. It has to be catch *and suppress*.

### Why capture without suppression doesn't help

The concurrent mutation isn't failing to reach the index today. It reaches it and corrupts it.

`AddMutationWithIndex` leaves `factorySpecs` nil, `addIndexMutations` fills it from the write context (`posting/index.go:89-95`), and `FactoryCreateSpec` prefers `mutSchema` (`schema/schema.go:369-382`). So a live mutation already knows about the new HNSW index and already writes graph edges for it, into a graph that `DropIndexes` just emptied. It hits `if data == nil { return create_edges(inUuid) }` in `createEntryAndStartNodes` (`tok/hnsw/helper.go:411-413`), installs itself as the sole entry point with an all-empty adjacency matrix, and commits at a `commitTs` above the rebuild's blob.

Note this is a lost update on an opaque blob, not a timestamp-ordering problem. Adjacency is read last-writer-wins, so on every key a live insert touched, the live version wins and it describes a disjoint second graph. The entry key is touched by essentially every early insert, so the surviving entry point reliably lands in a component the build never linked. That also rules out the "make the build merge-aware" family: there is no merge operation for two adjacency matrices.

If we capture the UID for replay but still let the insert happen, we keep the clobber. Suppressing the insert during the build is what actually closes it: the builder becomes the sole writer to the vector keyspace for its window, and `__vector_entry` stays untouched until the build establishes it.

### The seam is wider than it looks

It isn't the `indexer.Insert(ctx, tc, uid, inVec)` line at `posting/index.go:172`. `AddMutationWithIndex` makes **two** `addIndexMutations` calls: a DEL-op at `posting/index.go:619` whenever a prior posting existed, then the SET-op at `:633`. Both enter the vector block, and the DEL half has its own side effect on `__vector_dead` (see below). Capture and suppression have to wrap the whole block from `posting/index.go:103`, gating both branches.

There is also no way today to tell the builder's writes from live writes at that seam. Both arrive under identical context: `schema.GetWriteContext(ctx)` at `worker/mutation.go:59` for live mutations, `schema.GetWriteContext(context.Background())` at `worker/mutation.go:171` for the builder, whose ctx threads unbroken to `txn.addIndexMutations(ctx, ...)` at `posting/index.go:1455`. A naive "skip if this predicate is mid-rebuild" branch suppresses the build itself.

The near-misses don't work. Pre-populated `factorySpecs` looks like a discriminator, but `handleDeleteAll` pre-populates it on a live path too (`posting/index.go:396-403`), and `posting/index.go:89` overwrites the nil-ness before the vector block runs. Comparing `mutSchema`'s `IndexSpecs` against the predicate's can't work either: `GetQuerySchema` does `querySchema := *rb.CurrentSchema` at `posting/index.go:1134` and rewrites only `Tokenizer`, `Count`, and `Directive`, so `IndexSpecs` are identical throughout the build. The clean fix is a second context key stamped once at `worker/mutation.go:171`, which needs no signature changes. We'd also need a per-predicate signal, since `IndexingInProgress()` is global (`schema/schema.go:518`).

### Two things that make this cheaper than it sounds

- **Capture UIDs, not mutation payloads.** A UID set bounded by the predicate's cardinality. It dedupes repeated updates for free, and replay re-reads the authoritative value from the data posting list.
- **The list doesn't need to survive a crash.** The new schema only reaches disk in `updateSchema` (`worker/mutation.go:282-298`) after a successful build, and there's no resume path, so a crash mid-build already abandons the whole alter.

### Don't loop toward a drain

My first instinct was a converge loop with a bounded final drain. I think that's wrong. Capture is armed at apply time, but a mutation only becomes visible at commit time, and `detectPendingTxns` (`worker/draft.go:315`) is weaker than it looks: it rejects the alter only if a txn *already* holds a key on the predicate in the local oracle's `pendingTxns` (`posting/oracle.go:395-405`), and it explicitly refuses to wait. A txn that first touches the predicate after the alter is never caught. Its edge gets suppressed and captured at apply, but if replay reads before its commit lands, the vector is missing at replay time and never recomputed. That residue is sized by client think time, so a converge loop can't shrink it away.

Better: hold suppression armed across the entire replay, then quiesce pending txns on the predicate (the `tryAbortTransactions` path at `worker/draft.go:328` already exists) and flip once. Record under the `s.RLock()` that `FactoryCreateSpec` already takes on every live mutation (`schema/schema.go:369-372`), flip under the write lock in `DeleteMutSchema` (`schema/schema.go:212`). That makes the flip atomic against the local apply goroutine, with no drain mechanism and no termination question.

Two constraints on the replay itself. It can't flush through the rebuilder's `CommitToDisk` at `r.startTs`, because `posting/mvcc.go:300` skips any key whose recorded max version is at or above the commit ts, which is precisely the captured set. And a per-replica "fresh ts" makes index-key versions replica-dependent for the first time, which is what the incremental snapshot filter at `worker/snapshot.go:209` assumes away. Today every replica writes the whole graph at the same Raft-deterministic `r.startTs`. Route the replay through the ordinary `AddMutationWithIndex` path under a real transaction instead.

### Fix this first, whichever design we pick

`RunWithoutTemp` never calls `writer.Flush()`. It ends at the `ExponentialRetry(...CommitToDisk...)` at `posting/index.go:786-794`, while `rebuilder.Run` does `return writer.Flush()` at `posting/index.go:1111`. `TxnWriter.update` is an async `CommitAt(commitTs, w.cb)` whose errors only surface through `Wait()`. So `BuildIndexes` can return nil before the graph is durable, or after the commit failed. "Replay after indexing is complete" isn't a well-defined moment until that's fixed.

### The alternative worth weighing

A shadow generation keyspace: build into `__vector_g2_*`, have live mutations dual-write into both graphs, and flip a generation pointer at schema-flip time. More write amplification, but it keeps the old index queryable for the whole build, which capture-and-replay does not.

### Two adjacent bugs on main

**Updating a vector dead-marks its own uid.** `addMutationHelper` applies the mutation via `addMutationInternal` at `posting/index.go:559`, before the DEL-op index call at `:619`. So when that call runs `pl.AllValues(txn.StartTs)` at `posting/index.go:109`, the txn's own new posting is already visible via the `start == readTs` branch in `pickPostings` ("This mutation is by ME", `posting/list.go:1106-1108`). `data[0].Tid == types.VFloatID` holds, and the dead-node branch at `posting/index.go:114` fires. The mark is permanent (`// TODO add a path to delete deadNodes`, `tok/hnsw/helper.go:746`) and `removeDeadNodes` strips the uid from every neighbor list rewritten afterwards (`tok/hnsw/helper.go:670`).

Confirmed with a unit test in `posting/`: insert a vector for uid 1, then update it in place. First insert leaves `__vector_dead` empty; the update puts uid 1 in it.

```
Error:    []uint64{0x1} should not contain 0x1
Messages: updating a vector must not add its own uid to __vector_dead (got [1])
```

**`GetQuerySchema` never masks `IndexSpecs`.** It masks `Tokenizer`, `Count`, and `Directive` (`posting/index.go:1132`), but for a vfloat predicate `@index(hnsw)` populates `IndexSpecs` only. So readers see the index as live from the instant of the `Set` at `worker/mutation.go:251`, and `similar_to` runs against a graph under construction. The tokenizer equivalent would have errored out with "not indexed". `needsTokIndexRebuild` already computes `vectorIndexesToRebuild`, it just isn't consulted there.

.claude/agents/test-engineer.md
```diff
@@ -0,0 +1,467 @@
+---
+name: test-engineer
+description: Expert in writing Dgraph tests — knows all conventions, build tags, test placement, table-driven tests, testify assertions, dgraphtest/dgraphapi patterns, and best practices. Use for any task involving writing new tests or reviewing/improving existing tests in the Dgraph codebase.
```
matthewmcneely, line 3: @shiva-istari Did you mean to check in these claude defs? I'd prefer they get added in a separate PR

matthewmcneely (changes requested):

posting/oracle.go
```diff
@@ -77,11 +78,23 @@ func (vt *viTxn) StartTs() uint64 {
...
-	return vt.GetValueFromPostingList(pl)
+	val, err := vt.GetValueFromPostingList(pl)
+	if err == ErrNoValue {
+		// Genuine "no value" (e.g. a vector index whose centroids were never
+		// built). Surface the shared index.ErrNotFound sentinel so index-layer
+		// callers (kmeans centroid hydration) can tell it apart from the
+		// storage error above and cache the definitive miss instead of
+		// retrying. Wrap ErrNoValue so existing errors.Is(_, ErrNoValue)
+		// checks still hold.
+		return nil, fmt.Errorf("%w: %w", index.ErrNotFound, ErrNoValue)
```
matthewmcneely, line 95: **The `index.ErrNotFound` sentinel never reaches the query path.**

This wraps the miss in `viTxn`, but there are two production `CacheType` implementations and the search path uses the other one. `worker/task.go:370` builds `posting.NewViLocalCache(qs.cache)`, and `viLocalCache.GetValueFromPostingList` (posting/lists.go:112) still returns a bare `ErrNoValue` — a separate `errors.New`, so `errors.Is(err, index.ErrNotFound)` is false there. Confirmed with a scratch test against a real `viLocalCache`:

```
viLocalCache.Get err = No value found; Is(index.ErrNotFound)=false; Is(ErrNoValue)=true
```

Consequence on every query against a partitioned index whose centroids are not built yet: `maybeHydrate` takes the transient-failure branch, logs a `Warningf` (not `V(1)`), skips `vc.hydrated = true`, and re-reads the centroid key next call — all while holding `vc.mu.Lock()`, the write lock, across the read. Before this change it hydrated once and stayed quiet.

Apply the same wrap in `viLocalCache.GetValueFromPostingList`. Worth doing in both `GetWithLockHeld` variants too, for consistency.

Separately: `err == ErrNoValue` here is an equality check. `errors.Is` is safer if `GetValueFromPostingList` ever grows a wrapped return.

tok/kmeans/kmeans.go
```diff
@@ -0,0 +1,337 @@
...
+
+	vc.mu.Lock()
+	defer vc.mu.Unlock()
+	if vc.hydrated || len(vc.centroids) > 0 {
+		return nil
+	}
+	indexCountAttr := hnsw.ConcatStrings(vc.pred, CentroidPrefix)
+	key := x.DataKey(indexCountAttr, 1)
+	centroidsMarshalled, err := c.Get(key)
+	if err != nil && !errors.Is(err, index.ErrNotFound) {
```
matthewmcneely, line 206: Follow-on from the `viLocalCache` gap (see my comment on `posting/oracle.go`): on the query path this predicate is false for a genuine miss, so an unbuilt index takes the retry branch on every search.

`TestHydrationRetriesOnTransientError` passes because the same commit changed `fakeCache` to return `index.ErrNotFound` on a miss. The test encodes the intended contract; the query-path implementation does not honor it yet. Worth a test that exercises the real `viLocalCache` so the two stay in step.

tok/hnsw/persistent_hnsw.go
```diff
@@ -420,6 +483,63 @@ func (ph *persistentHNSW[T]) SearchWithUidAndOptions(
...
+
+	for i := range list {
+		var vec []T
+		err := ph.getVecFromUid(list[i], c, &vec)
+		if err != nil {
+			// A UID can remain in a shard's result set after its data key was
+			// removed (deleted but not yet cleaned from the graph). Skip it and
+			// keep returning the surviving neighbors rather than aborting the
+			// whole query; only genuine errors propagate.
+			if errors.Is(err, errNilVector) {
```
matthewmcneely, line 509: **This skips genuine storage errors, not just deleted UIDs.**

`errNilVector` is broader than the comment assumes:

- `getDataFromKeyWithCacheType` (helper.go:310) wraps **every** `c.Get` error as `errFetchingPostingList`
- `getVecFromUid` (helper.go:405) then wraps that as `errNilVector`

So a transient Badger read failure mid-merge now drops that neighbor and returns a short result set with no error — a silently wrong query instead of a failed one.

This is the same miss-vs-failure distinction `index.ErrNotFound` was introduced for in this PR; it just wasn't applied here. Either check the underlying cause, or have `getVecFromUid` return a distinct sentinel for "key absent" and match on that.

(Lines 331 and 784 already have this pattern, so it is consistent with existing code — but this one is on the result-correctness path.)

Minor: `if len(vec) == 0 { continue }` below becomes redundant once the error check is tightened.

posting/index.go
```diff
@@ -1412,9 +1491,506 @@ func (rb *indexRebuildInfo) prefixesForTokIndexes() ([][]byte, error) {
...
+				if err != nil {
+					return err
+				}
+				// Only genuinely float32vector-typed values encode a real
+				// dimension. A value written before the predicate was typed
+				// float32vector is stored as raw text; interpreting those bytes
+				// as packed float32 yields len(bytes)/4 — a bogus dimension that
+				// would then fail every vector in the conversion pre-pass. Skip
+				// them (mirrors ExistingVectorDimension).
+				if val.Tid != types.VFloatID {
```
matthewmcneely, line 1527: **The guard is right, but the failure mode it was meant to fix is unchanged.**

In the exact scenario from the original review — every value still raw text because the predicate was typed `float32vector` after the writes — this now skips *everything*. `dimension` stays `-1`, `SetDimension(schema, -1)` runs below, and the sampling scan at line 1654 then fails every vector with `vector dimension mismatch expected dimension -1`. Same aborted rebuild, different message.

The mixed case genuinely improves, so this is worth keeping. To close the all-text case, the sampling scan already converts text to vfloat inline (lines 1637-1650) — doing the same here, or re-inferring after the conversion pre-pass, would finish it.

posting/index.go
```diff
@@ -1412,9 +1491,506 @@ func (rb *indexRebuildInfo) prefixesForTokIndexes() ([][]byte, error) {
...
+					dimension = len(inVec)
+				}
+				numVectorsToCheck -= 1
+				if numVectorsToCheck <= 0 {
+					return ErrStopIteration
+				}
+				return nil
+			},
+			StartKey: x.DataKey(rb.Attr, 0),
+		})
```
matthewmcneely, line 1543: Still open from the original thread: the `IterateDisk` return value is discarded. A scan failure leaves `dimension` at `-1` and the rebuild proceeds to fail downstream with a confusing message instead of surfacing the real error. Bind it to `err` and return.

Also `val.Value.([]byte)` on line 1530 is an unchecked type assertion — a comma-ok would cost nothing.

posting/vector_rebuild_gate.go
```diff
@@ -0,0 +1,465 @@
...
+			}
+			deadline, ok := graceUntil[uid]
+			if !ok {
+				deadline = time.Now().Add(graceDuration)
+				graceUntil[uid] = deadline
+			}
+			if time.Now().Before(deadline) {
+				next[uid] = pm
+			} else {
+				delete(graceUntil, uid)
```
matthewmcneely, line 399: The wall-clock conversion is correct, and the cleanup on both the progress and expiry paths is right.

One thing the fix doesn't cover: this drop is silent. A captured uid whose transaction resolved but whose value never became readable within the grace window just never gets indexed — no log, no error. That's the part of the original finding about the vector being silently lost. A `glog.Warningf` here would at least make it diagnosable.

posting/vector_rebuild_gate.go
```diff
@@ -0,0 +1,465 @@
...
+			}
+		}
+		carry = next
+		if len(carry) > 0 && !progressed {
```
matthewmcneely, line 403: The sleep only runs when `progressed` is false. Under a steady arrival of new captures, a uid sitting in grace keeps this loop hot for the full 200ms with no sleep at all — the old round budget capped that at 40 iterations incidentally.

A short sleep when `len(graceUntil) > 0` restores the bound without giving up the wall-clock semantics.

tok/partitioned_hnsw/unified_factory.go
```diff
@@ -0,0 +1,191 @@
...
+// the numClusters option: absent => today's plain hnsw, unchanged;
+// numClusters > 1 => the partitioned (IVF-over-HNSW) implementation.
+type unifiedHNSWFactory[T c.Float] struct {
+	// mu serializes the whole select-then-delegate sequence. pick() removes a
+	// name's stale registration from the other child factory before the
+	// selected one creates it; the child locks protect each factory alone, not
+	// this cross-factory sequence, so without mu a concurrent monolithic and
+	// partitioned call for the same name could each create a registration and
+	// leave Find ambiguous. Find/Remove take the same lock.
+	mu   sync.Mutex
```
matthewmcneely, line 34: The mutex is the right fix, and taking it in `Find`/`Remove` too is correct.

The original review also asked for a concurrent mixed-configuration regression test, and `unified_factory_test.go` has no concurrency test. As it stands the invariant this lock protects is untested — a `-race` test that hammers `Create` (monolithic opts) and `FindOrCreate` (partitioned opts) on the same name would pin it.

.claude/agents/test-engineer.md
```diff
@@ -0,0 +1,467 @@
+---
```
matthewmcneely, line 1: Still here — `git ls-tree` confirms both `.claude/agents/test-engineer.md` and `test-runner.md` are added by this PR and absent from `main`, though the thread got resolved.

`t/benchmark_claude_diss.md` *was* removed in 677df14b, so the stray-file sweep just missed these two. Same ask as before: separate PR.

worker/task.go
```diff
@@ -370,7 +370,8 @@ func (qs *queryState) handleValuePostings(ctx context.Context, args funcArgs) er
 			posting.NewViLocalCache(qs.cache),
 			args.q.ReadTs,
 		)
-		indexer, err := cspec.CreateIndex(args.q.Attr)
+
+		indexer, err := cspec.FindOrCreateIndex(args.q.Attr)
```
matthewmcneely, line 374: On the resolved `nodeAllEdges` thread: skipping it was the right call, and I'd rather that reasoning lived here than in a silent resolve.

The premise doesn't hold. `persistentIndexFactory.FindOrCreate` deliberately delegates to `CreateOrReplace` (persistent_factory.go:198-206), so every caller gets a fresh instance, and `partitionedHNSW.subIndex` (partitioned_hnsw.go:314) builds a new sub-index per call. `nodeAllEdges` is per-call, so it can't go stale across read timestamps.

Worth a one-line reply on threads you close without a change — otherwise the next reviewer re-derives this.

systest/vector/backup_test.go
```diff
@@ -231,61 +199,57 @@ func TestVectorBackupRestoreDropIndex(t *testing.T) {
...
+	// range that references a non-existent second backup is a silent no-op:
+	// the test would then pass purely on the still-live data, never exercising
+	// restore at all.
+	t.Log("taking incremental backup \n")
+	require.NoError(t, hc.Backup(sharedCluster, false, backupDir))
+
 	t.Log("restoring backup \n")
-	require.NoError(t, hc.Restore(c, dgraphtest.DefaultBackupDir, "", 2, 1))
-	require.NoError(t, dgraphapi.WaitForRestore(c))
+	require.NoError(t, hc.Restore(sharedCluster, backupDir, "", 0, 0))
```
matthewmcneely, line 235: `0, 0` plus the incremental backup is better than what was suggested. One gap left, and it's the one your own comment above names: this restores over still-live data, so the assertions below pass whether or not the restore did anything.

Dropping the data (or restoring into a fresh namespace) before the restore would make the test actually exercise the path.

posting/index.go
```diff
@@ -1860,14 +2513,57 @@ func rebuildListType(ctx context.Context, rb *IndexRebuild) error {
...
+			}
+		}
+	}
+}
+
+// EvictVectorIndexCachesForNs is the namespace-scoped counterpart of
+// EvictVectorIndexCaches. It must run BEFORE the namespace's predicates are
+// removed from the schema (DeletePredsForNs), or the predicates can no longer
+// be found to evict — same ordering requirement as the DROP_ALL path.
+func EvictVectorIndexCachesForNs(ns uint64) {
```
matthewmcneely, line 2539: This is `EvictVectorIndexCaches` plus a predicate filter. Relocating the eviction into `DeleteAllForNs` rather than `worker/draft.go` is the better seam — but one function taking a `func(string) bool` would keep the two ordering-sensitive callers from drifting apart later.

matthewmcneely (changes requested):

posting/index.go
shiva-istari: Done. Extracted evictVectorIndexCaches(include func(string) bool); both
EvictVectorIndexCaches (nil filter) and EvictVectorIndexCachesForNs (namespace
filter) delegate to it, so the two ordering-sensitive callers stay identical.

systest/vector/backup_test.go
shiva-istari: Fixed. Drop all data (gc.DropAll) before the restore so the restore is the sole
source of the vectors; the assertions now actually exercise the restore path.

worker/task.go
shiva-istari: ack

.claude/agents/test-engineer.md
shiva-istari: Removed both test-engineer.md and test-runner.md

tok/partitioned_hnsw/unified_factory.go
shiva-istari: Added TestUnifiedConcurrentCreateFindOrCreate: it hammers Create (monolithic
opts) and FindOrCreate (partitioned opts) on the same name under -race, pinning
the factory mutex that keeps pick()'s cross-factory select-then-delegate from
leaving both children registered.

posting/vector_rebuild_gate.go
shiva-istari: Added a short sleep whenever captures are still waiting out their grace window
(len(graceUntil) > 0), so a steady arrival of new captures can't keep the loop
hot for the full 200ms. The wall-clock grace semantics are unchanged.

posting/vector_rebuild_gate.go
shiva-istari: Added a glog.Warningf on the grace-expiry path naming the predicate, uid, and
grace window, so a captured-but-never-readable vector is diagnosable instead of
being dropped silently.

posting/index.go
shiva-istari: Done. Bound the IterateDisk return and return it (IterateDisk maps
ErrStopIteration to nil internally, so the early stop is not treated as an
error), and switched to a checked (comma-ok) type assertion.

posting/index.go
shiva-istari: Fixed. The dimension scan now converts text to vfloat inline (mirroring the seed
scan at 1637-1650) before measuring, so an all-text predicate infers its real
dimension instead of leaving it at -1.

tok/hnsw/persistent_hnsw.go
shiva-istari: Fixed. getVecFromUid now skips only a genuinely absent key (index.ErrNotFound)
and propagates storage errors, so MergeResults no longer silently returns a
short result set on a transient read failure. This builds on the query-path
sentinel change. Updated the in-memory test doubles to emit index.ErrNotFound on
a miss too, so they match the real caches.

tok/kmeans/kmeans.go
shiva-istari: Added TestViLocalCacheGetReturnsErrNotFoundOnMiss in the posting package (tok
can't import posting, so the real-cache contract lives there). It pins that the
real viLocalCache returns index.ErrNotFound on a miss, which is exactly what the
fakeCache here assumes

posting/oracle.go
shiva-istari: confirmed. viLocalCache.GetValueFromPostingList returned a bare
ErrNoValue, and the query path (worker/task.go) uses viLocalCache, so
maybeHydrate never saw index.ErrNotFound there. Centralized the wrap in
GetValueFromPostingList for both viTxn and viLocalCache (so Get and
GetWithLockHeld both report a genuine miss as index.ErrNotFound) and dropped the
now-redundant wrap in viTxn.Get. Also replaced the `== ErrNoValue` equality
check with errors.Is.

PR #9484, opened 2025-08-15
perf(concurrency): cancel remaining goroutines when has error

eileenaaa:
**Description**

This PR optimizes goroutine execution in multi-task scenarios:

- Introduced `context.WithCancel` to propagate cancellation signals.
- When any goroutine encounters an error, all remaining goroutines are signaled to exit early.
- Reduces unnecessary computation and resource usage when partial failures occur.

This improves system responsiveness and reduces wasted CPU cycles in error cases.

**Checklist**

      this PR

### DIFF
--- a/worker/task.go
+++ b/worker/task.go
@@ -798,12 +798,19 @@ func (qs *queryState) handleUidPostings(
 	errCh := make(chan error, numGo)
 	outputs := make([]*pb.Result, numGo)
 
+	cctx, ccancel := context.WithCancel(ctx)
+	defer ccancel()
 	calculate := func(start, end int) error {
 		x.AssertTrue(start%width == 0)
 		out := &pb.Result{}
 		outputs[start/width] = out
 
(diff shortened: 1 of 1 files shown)


ghost:
Can you instead use errGroup to do the same thing?

eileenaaa:
> Can you instead use errGroup to do the same thing?

Yes, that works. I’ve implemented the same logic using an errgroup with context—please review code again.

Also, a quick question: since we’re using egCtx (from errgroup.WithContext(ctx)), 
it already inherits all features from the parent ctx, including cancellation and timeouts. 
In handleUidPostings, there’s still a check on the parent ctx every 100 iterations (around line 812 in worker/task.go). 
Given that, is this check still necessary, or is there a specific reason for keeping it?

ghost:
If we don't have that check, it could take a long time for a thread to finish as this one thread with no errors could take long when other threads have been cancelled.

PR #9485, opened 2025-08-15
chore(test): Increase performance of t-harness tests

matthewmcneely:
**Description**

This PR significantly improves performance of "t"-based testing in Dgraph

1. Removes Docker container address resolution from the testutil init() function (any package that imports would block waiting for container addresses that weren't established yet). Instead, the PR introduces a lazy-loading concept that sync.Once's the address resolution. This saves about 20 seconds per test invocation from `t`
2. Parallel-izes the resolution of expected container addresses
3. Parallel-izes the health and login flows
4. Suppresses health-check and login warnings until after a reasonable threshold of failures (these checks constantly pollute the logs)
5. On non-Linux system, informs the user about DGRAPH_BINARY settings that can let tests proceed.
6. Upgrades the warp runner for the core tests to the 16x instance (core tests are the most comprehensive tests)

In general (on my M4) many tests improve time to complete by a factor of 3.

**Checklist**

### DIFF
--- a/check_upgrade/check_upgrade_test.go
+++ b/check_upgrade/check_upgrade_test.go
@@ -13,6 +13,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"testing"
 	"time"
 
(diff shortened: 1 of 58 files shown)


check_upgrade/check_upgrade_test.go
```diff
@@ -26,6 +27,9 @@ import (
 )
 
 func TestCheckUpgrade(t *testing.T) {
+	if runtime.GOOS != "linux" {
```
harshil-goel, line 30: Maybe we need to find out ways to change it for other os?

matthewmcneely: I forgot about the DGRAPH_BINARY env variable... which can allow these tests to run... I updated all these to also check for that envvar.

harshil-goel (approved):

PR #9488, opened 2025-08-19
fix(core): use deep copy for committedUids to prevent mutation corruption

eileenaaa:
**Description**

This PR addresses a stability issue in mutation handling:

- Replaced shallow copy of committedUids with deep copy to prevent external
  modifications from corrupting committed state.

These changes make committedUids safer and preserve correctness during refresh.

**Checklist**

      this PR

### DIFF
--- a/posting/list.go
+++ b/posting/list.go
@@ -1010,7 +1010,7 @@ func (l *List) setMutationAfterCommit(startTs, commitTs uint64, pl *pb.PostingLi
 	if refresh {
 		newMap := make(map[uint64]*pb.Posting, len(l.mutationMap.committedUids))
 		for uid, post := range l.mutationMap.committedUids {
-			newMap[uid] = post
+			newMap[uid] = proto.Clone(post).(*pb.Posting)
 		}
 		l.mutationMap.committedUids = newMap
 	}


ghost:
Any reason for this? From what I understand, the underlying structures wont be changed. Did you notice any issue or just because of linters / safety?

eileenaaa:
> Any reason for this? From what I understand, the underlying structures wont be changed. Did you notice any issue or just because of linters / safety?

The change to proto.Clone(post).(*pb.Posting) was made because [PR 9480](https://github.com/hypermodeinc/dgraph/pull/9480) aims to fix a race condition in the mutation map.
Simply doing newMap[uid] = post only copies the pointer, meaning multiple goroutines could still share and mutate the same underlying object.
By using proto.Clone, each entry becomes an independent copy, which is necessary to fully prevent the race condition.

ghost:
@eileenaaa The reason I didn't do this in that PR was, even though map might get updated, the underlying values don't get updated. This is already committed data which is not going to get updated. Make a proto clone there only adds more memory without really giving us any benefit. Hence I asked did you see any error, which would mean that my theory was wrong.

eileenaaa:
@harshil-goel  I checked the subsequent code and confirmed that committedUids data is indeed not modified afterwards. I also ran tests with concurrent data, and it works fine, so making a deep copy here is not necessary.
Sorry for the confusion — I initially misunderstood the scope of PR 9480 and thought it was addressing the map race itself, which is why I raised the concern about not doing a deep copy here. Since this part is fine, I’ll go ahead and close the PR.

PR #9492, opened 2025-08-20
fix(query): return full float value in query results

### LINKED ISSUE #9491: Return full float value in query results
## Problem

`float` values are stored as `float64`, but returned truncated to 6 decimal digits. This means it's impossible to get the precise stored value in a query. Only an approximation is returned.

Data in:

```json
{ "float": 0.123456789012345 }
```

Data out:

```json
{ "float": 0.123457 }
```

## Solution

Outputting the full `float` value would provide full access to floating point data.

## Alternatives

This [forum post](https://discuss.hypermode.com/t/extend-number-precision-in-answer/19287) describes storing a string next to each float to enable reading the full value.

>Second is add to the schema one predicate with float type for comparison and each one with string type to return, but its kinda kludge.

## Discussion

The truncation happens in [`outputnode.go`](https://github.com/hypermodeinc/dgraph/blob/c2ccd6e243a41d1f9d5e13d199ecb125134384d0/query/outputnode.go#L673), in the `valToBytes` function.

```go
return []byte(fmt.Sprintf("%f", f)), nil
```

The [Golang docs](https://pkg.go.dev/fmt) specify that `%f` defaults to a precision of 6 digits.

>precision sets the number of places after the decimal

>The default precision for %e, %f and %#g is 6

qujoux:
**Description**

Updates `float` formatting to express full values. Switches from `%f` to `%g`, which according to [Golang docs](https://pkg.go.dev/fmt) will use as many digits as necessary to express the value.

>for %g [the default precision] is the smallest number of digits necessary to identify the value uniquely.

I had trouble getting `t` to run. A new unit test passes with:

```sh
go test github.com/hypermodeinc/dgraph/v25/query
```

Closes #9491.

**Discussion**

This changes the output format but not semantics of float values in result data. Currently floats have a fixed precision of 6, filled with trailing zeros as necessary.

```json
{ "float": 1.000000 }
```

The new specifier `%g` drops unnecessary trailing zeros.

```json
{ "float": 1 }
```

I think this OK as they are semantically equivalent. Since unnecessary characters are dropped, it also saves bytes on the wire. It also aligns with the `bigfloat` format, which never gives trailing zeros.

**Checklist**

- [x] For all _code_ changes, an entry added to the `CHANGELOG.md` file describing and linking to this PR


mangalaman93 (approved):

mangalaman93 (approved):

qujoux:
Thank you for reviewing. Just a note that I've got `t` running and I'm working on getting these tests cleaned up.

matthewmcneely:
> Thank you for reviewing. Just a note that I've got `t` running and I'm working on getting these tests cleaned up.

@qujoux Be aware I'm in the middle of improving the `t` harness, not sure what you're planning when you say "cleaned up", but wanted to avoid duplicate efforts: https://github.com/hypermodeinc/dgraph/pull/9485

qujoux:
Thank you! I'm just aiming to get tests passing for this PR, I see there are JSON comparisons scattered around that expect the old float format. Faster testing will be a nice improvement.

qujoux:
Some tests are giving Docker errors I don't understand, but they seem to have no float comparisons in them. I think all the relevant tests are passing now.

qujoux:
Sorry about the bumpy road, and thank you for reviewing!

PR #9516, opened 2025-10-19
fix(dev): Add jemalloc package to local docker image

xqqp:
**Description**

The jemalloc package is missing, so Dgraph fails to build. This PR adds it.


darkcoderrises:
Can you also send the changes to predictable-labs?

xqqp:
Done

PR #9515, opened 2025-10-19
fix(cache): Estimate size of posting lists

xqqp:
**Description**

The maximum cost of the Ristretto cache is set to a number of megabytes, but when inserting a posting list item into the cache, the cost is set to 1. Because of this the cache only grows and no items get evicted. This results in high memory consumption, especially with the new UID cache (see #9513). 

This PR introduces a cost function which estimates the memory size of each item. For more details see https://github.com/predictable-labs/dgraph/issues/6. Credits to @darkcoderrises for the implementation.

The default cache size is changed from 1024 to 4096, to reflect the more accurate cost estimation. See below for how the size of the cache relates to the occupied memory by the Dgrpah process, when this PR is applied.

size-mb=1024

Dgraph process occupies up to 3.7 GiB of memory

heap:

```
  344.49MB 30.34% 30.34%   407.26MB 35.87%  github.com/hypermodeinc/dgraph/v25/posting.(*List).calculateUids
  130.37MB 11.48% 41.83%   130.37MB 11.48%  github.com/dgraph-io/ristretto/v2.newCmRow
   83.20MB  7.33% 49.15%    83.20MB  7.33%  github.com/dgraph-io/badger/v4/skl.newArena
   69.50MB  6.12% 55.28%    69.50MB  6.12%  github.com/hypermodeinc/dgraph/v25/posting.(*List).Uids.func1
   65.16MB  5.74% 61.02%    65.16MB  5.74%  github.com/dgraph-io/ristretto/v2/z.(*Bloom).Size
   53.27MB  4.69% 65.71%    53.27MB  4.69%  github.com/hypermodeinc/dgraph/v25/posting.(*List).calculateUids.func1
```
   
   
size-mb=2048

Dgraph process occupies up to 5.4 GiB of memory

heap:
```
  637.07MB 33.68% 33.68%   699.33MB 36.97%  github.com/hypermodeinc/dgraph/v25/posting.(*List).calculateUids
  260.63MB 13.78% 47.46%   260.63MB 13.78%  github.com/dgraph-io/ristretto/v2.newCmRow
  130.04MB  6.88% 54.34%   130.04MB  6.88%  github.com/dgraph-io/ristretto/v2/z.(*Bloom).Size
   99.01MB  5.23% 59.57%    99.01MB  5.23%  github.com/hypermodeinc/dgraph/v25/posting.(*List).Uids.func1
[...]

PR #9526, opened 2025-11-16
fix(cache): make updating the max cost of posting cache work again

xqqp:
Updating the max cost of the posting list cache is not wired up. It just calls an empty function. This PR adds a method to `MemoryLayer`, that allows updating the max cost of its cache. The old function is deleted and the function call is replaced with a call to the new method.

### DIFF
--- a/posting/lists.go
+++ b/posting/lists.go
@@ -46,9 +46,6 @@ func SetEnabledDetailedMetrics(enableMetrics bool) {
 	EnableDetailedMetrics = enableMetrics
 }
 
-func UpdateMaxCost(maxCost int64) {
-}
-
 // Cleanup waits until the closer has finished processing.
 func Cleanup() {
 	closer.SignalAndWait()
--- a/posting/mvcc.go
+++ b/posting/mvcc.go
@@ -405,6 +405,13 @@ func (ml *MemoryLayer) del(key []byte) {
 	ml.cache.del(key)
 }
 
+func (ml *MemoryLayer) UpdateMaxCost(maxCost int64) {
+	if ml.cache == nil || ml.cache.data == nil {
+		return
+	}
+	ml.cache.data.UpdateMaxCost(maxCost)
+}
+
 type IterateDiskArgs struct {
 	Prefix         []byte
 	Prefetch       bool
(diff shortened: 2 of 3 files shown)


matthewmcneely:
Oh wow, good catch. So the `updateConfig` in the GraphQL admin api was basically a no-op?

Edit: I see now that it was max cost that was not getting updated, not the other configs.

RJKeevil:
My app has a mode where it can do a full linear scan of all nodes to check consistency etc. I've notice that even with this change, if I set posting cache to anything greater than 0, my alphas mem usage goes from a steady 1-2G to 10G (container limit) and then restarts.  My cache is set to 4G so I beleive there is one more issue hiding here preventing the cache to be sized appropriately?

matthewmcneely:
> even with this change

@RJKeevil Just to be clear, you're trying from the current head of the main branch, right? #9515 (merged into main a few days ago) comprises the actual fix.

RJKeevil:
@matthewmcneely Yep im running direct from main, hash 68c24d5.  This isnt an immediate problem for me as i run with the posting cache off currently, but just a heads up that I think theres another leak in there somewhere in this cache.

The code that triggers it runs through all nodes of a type in ascending uid order, in batches of 250. The uid batches are passed to another routine that fetches all predicates for this uid, returning the complete node .

matthewmcneely:
> I think theres another leak

Right, but as you probably know the posting cache is only part of the memory that Dgraph allocates. I've ran tests that show the fix in #9515 is making things better from that perspective. But it would be good to understand your "full linear scan" details. Maybe open up a new topic thread in our new, spiffy [Discussions](https://github.com/orgs/dgraph-io/discussions)?

xqqp:
@RJKeevil Can you collect and share a heap profile of an alpha node when it has high memory consumption? Below an example request for that.
```sh
curl http://<alpha_host>:8080/debug/pprof/heap > heap.out
```
Please also post the output of the command below.
```sh
dgraph version
```
Also, memory calculation on Linux is not straightforward, your container runtime might determine used memory different from other tools. I suggest you experiment with different cache sizes, for example `size-mb=2048`.

RJKeevil:
for future reference discussion moved to here and solved https://github.com/orgs/dgraph-io/discussions/9535

PR #9537, opened 2025-12-04
chore(test): unskip previously skipped tests

shiva-istari:


matthewmcneely:
@shiva-istari You missed two added skips in systest/backup/nfs-backup/backup_test.go, lines 41 and 47

shiva-istari:
> nfs-backup/backup_test.go

@matthewmcneely What is the context behind skipping them? Are they flaky? I haven’t seen them failing recently.

matthewmcneely:
> > nfs-backup/backup_test.go
> 
> @matthewmcneely What is the context behind skipping them? Are they flaky? I haven’t seen them failing recently.

@shiva-istari They would fail often when we were on the shared runners between the move.

shiva-istari:
> > > nfs-backup/backup_test.go
> > 
> > 
> > @matthewmcneely What is the context behind skipping them? Are they flaky? I haven’t seen them failing recently.
> 
> @shiva-istari They would fail often when we were on the shared runners between the move.
 @matthewmcneely  When I unskipped these tests, they started failing because they’re unable to find the mapped port of the container. Surprisingly, they pass every time when run locally. I added retry logic to fetch the container info, but they’re still failing. Should I dig into this further, or leave them skipped for now? This will require more work.

matthewmcneely:
> Should I dig into this further, or leave them skipped for now? This will require more work.

I think it's worth spending a little time on this. Why do all the other tests that rely on getting the mapped port of the container pass? What about your local env (OSX?) is different from the runner?

shiva-istari:
> > Should I dig into this further, or leave them skipped for now? This will require more work.
> 
> I think it's worth spending a little time on this. Why do all the other tests that rely on getting the mapped port of the container pass? What about your local env (OSX?) is different from the runner?
@matthewmcneely , Blacksmith runners lack NFS support, leading to container failures. We should keep these tests skipped as NFS is rarely used.

PR #9550, opened 2026-01-03
chore(core): use unsafe.Slice instead of deprecated reflect.SliceHeader

xqqp:
**Description**

- `reflect.SliceHeader` has been deprecated. This PR uses `unsafe.Slice` instead, as recommended by the documentation
- Removed type hints when calling generic functions
- removed the `floatBits` argument from `dotProductT()`, because it was unused
- added a test for `BytesAsFloatArray()` that validates that the conversion works

### DIFF
--- a/tok/index/helper.go
+++ b/tok/index/helper.go
@@ -8,14 +8,13 @@ package index
 import (
 	"encoding/binary"
 	"math"
-	"reflect"
 	"unsafe"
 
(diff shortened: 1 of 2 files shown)


xqqp:
@matthewmcneely there are two functions named BytesAsFloatArray, one in index/helper.go and the other in types/value.go. Both functions convert a byte slice into a float slice. What do you think about refactoring the code so we only have one function?

PR #9557, opened 2026-01-05
chore(core): remove slash grpc handling

xqqp:
**Description**

Dgraph Cloud is gone, so remove Slash GRPC endpoint handling from Dgraph.


shiva-istari:
Thanks for opening this PR. Since Dgraph Cloud has been shut down, it makes sense to question whether the /slash endpoint is still needed. However, we haven’t made a final decision yet on whether we want to remove it entirely, as we may still need it in the future.

For now, I’m closing this PR until we have clearer direction. We can revisit and reopen it once a decision is made.

PR #9593, opened 2026-02-07
chore(core): remove net/trace event log

xqqp:
In c4bb4114c0e6bb6eb0eac5b5053b2e11b43555b0 there was an effort to replace net/trace with spans from OpenCensus. A few remnants of net/trace still remain, this PR removes them. 

The trace package exposes two endpoints: `/debug/requests` and `/debug/events`. They contain some incomplete data regarding raft indexes and schema loading. The documentation also does not refer to the endpoints or the event log at all. 

The documentation however incorrectly refers to the flag `expose_trace` (https://docs.dgraph.io/learn/howto/retrieving-debug-information#metrics-information), that is meant for the event log, but describes the flag as controlling the `/debug/vars`endpoint.

This PR does the following:

- replace golang.org/x/net/context with stdlib context package
- remove all uses of golang.org/x/net/trace
- remove `expose_trace` flag from code and docker compose files
- demote golang.org/x/net to an indirect dependency 

Regardless of this PR, the documentation should be updated to not associate the `expose_trace` config option with the `/debug/vars` endpoint

### DIFF
--- a/dgraph/cmd/root.go
+++ b/dgraph/cmd/root.go
@@ -96,8 +96,6 @@ func initCmds() {
 			"overridden to values set with environment variables and flags.")
 	RootCmd.PersistentFlags().Bool("bindall", true,
 		"Use 0.0.0.0 instead of localhost to bind to all addresses on local machine.")
-	RootCmd.PersistentFlags().Bool("expose_trace", false,
-		"Allow trace endpoint to be accessible from remote")
 	x.Check(rootConf.BindPFlags(RootCmd.PersistentFlags()))
 
 	// Add all existing global flag (eg: from glog) to rootCmd's flags
--- a/raftwal/storage.go
+++ b/raftwal/storage.go
@@ -9,13 +9,11 @@ import (
 	"math"
 	"sync"
 
+	"github.com/dgraph-io/dgraph/v25/x"
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
 	"go.etcd.io/etcd/raft/v3"
 	"go.etcd.io/etcd/raft/v3/raftpb"
-	"golang.org/x/net/trace"
-
-	"github.com/dgraph-io/dgraph/v25/x"
 )
 
 // DiskStorage handles disk access and writing for the RAFT write-ahead log.
@@ -58,8 +56,7 @@ import (
 // HardSync is set, msync is called after every write, which flushes those
 // writes to disk.
 type DiskStorage struct {
-	dir  string
-	elog trace.EventLog
+	dir string
 
 	meta *metaFile
 	wal  *wal
@@ -91,8 +88,6 @@ func InitEncrypted(dir string, encKey x.Sensitive) (*DiskStorage, error) {
 	if w.wal, err = openWal(dir); err != nil {
 		return nil, err
 	}
-
-	w.elog = trace.NewEventLog("Badger", "RaftStorage")
 
 	snap, err := w.meta.snapshot()
 	if err != nil {
@@ -150,8 +145,6 @@ func (w *DiskStorage) InitialState() (hs raftpb.HardState, cs raftpb.ConfState, 
 	w.lock.Lock()
 	defer w.lock.Unlock()
 
-	w.elog.Printf("InitialState")
-	defer w.elog.Printf("Done")
 	hs, err = w.meta.HardState()
 	if err != nil {
 		return
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -15,7 +15,6 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
-	"golang.org/x/net/trace"
 	"google.golang.org/protobuf/proto"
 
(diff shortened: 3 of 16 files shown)


xqqp:
The Trunk Code Quality check complains about formatting in the compose files. I don't have trunk installed. Can a maintainer take a look?

schema/schema.go
```diff
@@ -552,7 +549,6 @@ func Load(predicate string) error {
 		return err
 	}
 	State().Set(predicate, &s)
-	State().elog.Printf(logUpdate(&s, predicate))
 	glog.Infoln(logUpdate(&s, predicate))
```
matthewmcneely: This is a duplicate now. `State().Set()` on the line above does its own `glog.Infoln(logUpdate(...))` as of this PR, so the same message lands twice per load. The old pairing here was `elog` and `glog` going to two different sinks; now it's just the log file twice. Drop this line.

schema/schema.go
```diff
@@ -198,7 +195,7 @@ func (s *state) Set(pred string, schema *pb.SchemaUpdate) {
 	s.Lock()
 	defer s.Unlock()
 	s.predicate[pred] = schema
-	s.elog.Printf(logUpdate(schema, pred))
+	glog.Infoln(logUpdate(schema, pred))
```
matthewmcneely: Worth thinking about the volume here, and in `SetType` below.

`elog` was a bounded in-memory ring buffer. `glog.Infoln` is an unconditional synchronous write, and it happens while `s.Lock()` is held. `LoadFromDb` calls `Set` and `SetType` once per predicate and per type from the parallel Badger stream workers (`schema.go:609` and `:621`), so on a large multi-tenant cluster this becomes thousands of Info lines serialized under the schema write lock on every startup and after every restore. `SetType` is the worse of the two since it `%v`-formats the entire `pb.TypeUpdate`.

`glog.V(2).Infoln` would keep the diagnostic value without the unconditional cost, and it's closer to what the event log was there for in the first place.

Minor, while you're in here: `logUpdate` and `logTypeUpdate` both already end in `\n` and `Infoln` appends another, so every one of these gets a trailing blank line.

dgraph/cmd/root.go
```diff
@@ -96,8 +96,6 @@ func initCmds() {
 			"overridden to values set with environment variables and flags.")
 	RootCmd.PersistentFlags().Bool("bindall", true,
 		"Use 0.0.0.0 instead of localhost to bind to all addresses on local machine.")
-	RootCmd.PersistentFlags().Bool("expose_trace", false,
-		"Allow trace endpoint to be accessible from remote")
```
matthewmcneely, line 100: This is a persistent root flag that's been around a long time, and pflag errors on unknown flags (we don't set `ParseErrorsWhitelist` anywhere in the repo), so anything still passing `--expose_trace` on the command line will fail to start rather than warn. Config-file users are fine, since viper ignores unknown keys.

There's precedent for a softer landing a few lines down at `root.go:106`, where `MarkDeprecated("stderrthreshold", ...)` keeps the flag accepted and hidden while printing a warning. Given the docs still ship a copy-paste alpha command with `--expose_trace` in it, I'd rather deprecate for one release and delete in the next.

raftwal/storage.go
```diff
@@ -9,13 +9,11 @@ import (
 	"math"
 	"sync"
 
+	"github.com/dgraph-io/dgraph/v25/x"
```
matthewmcneely: Convention nit, and the same thing happens in `testutil/docker.go`: we keep `github.com/dgraph-io/*` in its own trailing import group rather than mixed in with the other third-party imports. See `worker/draft.go` or `posting/lists.go` for the prevailing shape. It isn't lint-enforced (no gci local-prefix in `.trunk/configs/.golangci.json`), which is why trunk is green, so this just needs `x` moved back down into its own block.

matthewmcneely (changes requested):

schema/schema.go
xqqp: remove the line

schema/schema.go
xqqp: Added the log level and removed `\n` from `logUpdate` and `logTypeUpdate`.

dgraph/cmd/root.go
xqqp: Added the deprecation warning

raftwal/storage.go
xqqp: Adjusted the import order

matthewmcneely:
@xqqp Can you please run `go mod tidy`? golang.org/x/net should move to indirect (grpc pulls it in)

Also, please update the PR description to say the command line flag is deprecated, not removed.

xqqp:
> @xqqp Can you please run `go mod tidy`? golang.org/x/net should move to indirect (grpc pulls it in)
> 
> Also, please update the PR description to say the command line flag is deprecated, not removed.

Done, I also updated the code rabbit release notes in the PR description accordingly.

PR #9599, opened 2026-02-15
fix(query): correct frontier trimming in shortest path

### LINKED ISSUE #9577: k shortest path does not return the shortest path when using MaxFrontierSize
## Describe the bug

In https://github.com/dgraph-io/dgraph/issues/9333 a new Feature got introduced called MaxFrontierSize, which limits memory usage in case of high density graphs. This causes a algorithmic issue, because the "best" node is droped from the priority queue when the MaxFrontierSize is triggered.
This causes a nearly 100% to not get the shortest path ever.

## To Reproduce

1. run shortest path query with numpath = 1 to get the shortest path
2. run the same query with numpath > 1 with MaxFrontierSize low enough to get triggered on your data set
3. Compare Paths

## Expected behavior

numpaths > 1 should return the somewhat shortest paths, possibly even the correct shortest paths

Schwarf:
## Description

Fix incorrect frontier trimming in shortest path queries when `maxfrontiersize`
is set.

Previously, the code used `pq.Pop()` to trim the priority queue. This removes
the last element of the heap rather than the highest-cost
element, which can discard valid low-cost candidates and lead to incorrect
shortest-path results.

The fix replaces slice-based removal with `heap.Remove` of the maximum-cost
element.

Applied in both:
- `shortestPath`
- `runKShortestPaths`

A regression unit test has been added to ensure the highest-cost element is
removed when trimming.

Fixes #9577


themavik (commented):
TrimToMax runs after heap.Push so the new node stays eligible before you drop the worst frontier entry; same max-eviction idea as #9607 but ordered differently. nit: if both PRs land, reconcile with the pre-push removeMax variant to avoid two competing policies.

Schwarf:
Hi @themavik,
Thanks for pointing this out.

From what I see, both PRs fix the same underlying issue (removing an arbitrary element instead of the highest-cost one), but they differ slightly in eviction policy:

- This PR uses *push-then-trim*, so the newly discovered node is considered before evicting the worst element.
- #9607 keeps the existing structure and does *trim-before-push* with `removeMax()`.

My understanding is that push-then-trim better preserves the intended frontier under a size cap, because it effectively keeps the best `M` elements from `old_frontier ∪ {new_node}`. A trim-before-push policy makes the eviction decision without considering the new candidate, which can drop a better existing entry while keeping a worse new one.

In addition, this PR includes a regression test (`query/shortest_test.go`) to guard this behavior.

That said, I’m happy to align with whichever eviction policy you prefer, or close this PR and contribute the test to #9607 if that’s the direction you’d like to take.

Schwarf:
Hi @themavik , 
just a gentle ping for review.

PR #9603, opened 2026-02-18
fix(test): resolve dgraph binary selection for macOS in dgraphtest and testutil

shiva-istari:
On macOS, setupBinary() now copies both the Linux binary (for Docker containers) and the host-native binary (for local bulk/live loader commands) into tempBinDir. On Linux, a single binary serves both purposes. BulkLoad() and LiveLoad() use hostDgraphBinaryPath() to pick the correct one based on the platform. Removed macOS skip guards from 13 test files so all tests run on both platforms after make install.

### DIFF
--- a/check_upgrade/check_upgrade_test.go
+++ b/check_upgrade/check_upgrade_test.go
@@ -10,11 +10,8 @@ package checkupgrade
 import (
 	"context"
 	"fmt"
-	"os"
 	"os/exec"
-	"path/filepath"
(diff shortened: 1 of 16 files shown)


mlwelles:
Thanks @shiva-istari! Incorporated your changes into #9606 with an additional fix for a data clobbering bug in `t/t.go` that surfaces when running `--suite=all` on macOS. Closing in favor of #9606.

Issue #9333, opened 2025-03-05
Possible Memoryleak with k shortest path


hr-alebel:
## Describe the bug

We are currently making a PoC of an App using dGraph as the graphdb. We will have a heavy usage of the k-shortest-path method. 

Dataset is ~9000 nodes with 600000 edges, with barely any data in predicates.

While testing it with a sample dataset we found that our k8s nodes went out of memory when using anything, but k=1.

After even more debugging and testing different setups (and increasing resources on the nodes) we found the following:

A) using k=1 and no weights in facets -> dGraph behaves normal (normal = query is very fast ~1sec and memory footprint is low ~512MB)

B) using k=1 and weights in facets -> dGraph behaves same as A

C) using k>1 and no weights on facets -> dGraph gets a bit slower ~2sec for numpath 100 and uses slightly more memory, but very much in acceptable ranges

D) using k>1 and weights in facets -> dGraph eventually crashes with OOM while using >8GB of RAM

## To Reproduce

1. Create a dataset with ~9000 nodes with ~600000 edges in between them with a float facet that will be used for weights
2. Create Queries for scenarios A-D
3. observe memory footprint

## Expected behavior

Looking at the code, the aproximate resource usage between scenario C and D should be somewhat similar, in our case ~1GB of RAM and return the answer in a reasonable  time

## Environment

- OS: linux docker
- query using ratel
- Version v21.3.2 and v24-latest (had to rollback due to https://github.com/hypermodeinc/dgraph.net/issues/56 )

hr-alebel:
![Image](https://github.com/user-attachments/assets/25148a2f-965d-462b-af9d-adc766d847c2)
Thats the heap from the broken queries

hr-alebel:
Almost all of the memory traces back to 

`newSlice := make([]pathInfo, len(*item.path.route)+1)` (~Line 400 in query/shortest.go)

In scenario C) the memory footprint of this line is ~200MB (at k=500, as its difficult to get the footprint at how fast i get the result)
In scenario D) the foot print is basically 2-6GB (depending on when i manage to get the heap before OOM )

harshil-goel:
Let me take a look if it's a memory leak or if the algorithm is not best that we can use.

harshil-goel:
@hr-alebel Could you help me reproduce the issue? 
```
import random

def generate_rdf(nodes=9000, edges=600000, output_file="dataset.rdf"):
    rdf_data = []
    
[...]

PR #9607, opened 2026-02-23
fix(query): evict highest-cost node when MaxFrontierSize exceeded in shortest path

### LINKED ISSUE #9577: k shortest path does not return the shortest path when using MaxFrontierSize

veeceey:
Fixes #9577

The `MaxFrontierSize` feature introduced in #9333 has a bug in how it evicts nodes from the priority queue when the size limit is exceeded.

The current code calls `pq.Pop()` directly on the slice (not `heap.Pop`), which:
1. Removes the last element of the underlying array rather than the highest-cost item. In a min-heap, the last element's cost is arbitrary, so we might drop a low-cost node that's actually on the shortest path.
2. Doesn't maintain the heap invariant since it bypasses `container/heap`, corrupting the queue for subsequent `heap.Push`/`heap.Pop` calls.

The fix adds a `removeMax()` method that scans for the highest-cost item and removes it using `heap.Remove()`, which properly rebalances the heap. This ensures we always evict the least promising node while preserving the priority queue invariant and keeping shortest-path correctness.

Affects both the single shortest path (`shortestPath`) and k-shortest-path (`runKShortestPaths`) code paths.

All existing query package tests pass.

### DIFF
--- a/query/shortest.go
+++ b/query/shortest.go
@@ -88,6 +88,22 @@ func (h *priorityQueue) Pop() interface{} {
 	return val
 }
 
+// removeMax removes and returns the highest cost item from the priority queue.
+// This is used to evict the least promising node when the frontier exceeds its
+// size limit, preserving the lowest cost nodes needed for shortest paths.
+func (h *priorityQueue) removeMax() {
+	if len(*h) == 0 {
+		return
+	}
+	maxIdx := 0
+	for i := 1; i < len(*h); i++ {
+		if (*h)[i].cost > (*h)[maxIdx].cost {
+			maxIdx = i
+		}
+	}
+	heap.Remove(h, maxIdx)
+}
+
 type mapItem struct {
 	attr  string
 	cost  float64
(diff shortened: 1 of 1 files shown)


veeceey:
### Test Results

Query package tests all pass:
```
$ go test ./query/ -count=1
ok  github.com/dgraph-io/dgraph/v25/query  4.521s
```

Package builds cleanly:
```
$ go build ./query/...
# no errors
```

The systest/shortest-path tests require a running dgraph instance so I couldn't run those locally, but the fix is logically straightforward - replacing direct slice manipulation with proper heap removal of the max-cost node.

veeceey:
gentle ping — would love to get some eyes on this when you get a chance

veeceey:
hey, any chance this could get a look?

themavik (commented):
removeMax walks the heap and heap.Remove-s the highest-cost node when MaxFrontierSize is exceeded, which matches the min-heap semantics Pop violated. Cost is O(n) whenever the frontier sits at the cap.

PR #9610, opened 2026-02-25
fix(test): macOS local image, plugin tests, and test suite restructuring

mlwelles:
## Summary

This PR fixes macOS compatibility for local image builds and plugin tests, restructures the test suite system for better ergonomics and resource management, and fixes a flaky integration2 test.

### macOS / Local Build Fixes
- Enable CGO cross-compilation for plugin support on macOS (Darwin → Linux)
- Fix integration test routing, `gotestsum` PATH resolution, and cross-compilation issues
- Add cluster pause/resume and resilient restore waits for stability
- Resolve shellcheck SC2312 in `check-cross-compiler.sh`

### Test Suite Restructuring
- **New `integration` suite** (new default): runs everything except `ldbc`, `load`, and `systest-heavy` — replaces old `unit,systest,core` default
- **New `unit` suite behavior**: true unit tests only — no Docker cluster, no `--tags=integration`, skips custom-cluster packages
- **New `systest-baseline` / `systest-heavy` sub-suites**: separates lightweight systests from resource-intensive ones (minio, encryption, tracing, online-restore) that can OOM on Docker
- `systest` still runs both sub-suites for backward compatibility
- **New Make targets**: `test-integration`, `test-integration-heavy`
- **Renamed**: `test-full` → `test-all`
- **Removed**: `test-suite`, `test-ldbc`, `test-load` (use `make test SUITE=...` instead)

### Import Client Retry Fix
- **Fixed flaky `TestImportApis/SingleGroupShutOneAlpha`** in the `dgraph-integration2-tests` CI job
- Root cause: `initiateSnapshotStream` made a single attempt to start the snapshot stream, but the Dgraph server can return "overloaded with pending proposals" during Raft membership changes (e.g., when an alpha is shut down and the cluster is rebalancing)
- Fix: Added exponential backoff retry (1s → 10s cap, 60s max total) for transient "overloaded" errors
- Only retries on "overloaded" — connectivity errors ("connection refused", "unable to connect to the leader") still fail immediately so negative test cases don't block unnecessarily

### Quick Reference

| Command | What it runs |
|---------|-------------|
| `make test` | `integration` suite + `integration2` (~30 min) |
| `make test-unit` | True unit tests — no Docker |
| `make test-integration` | Integration tests via t/ runner with Docker |
| `make test-integration-heavy` | systest-heavy + ldbc + load |
| `make test-all` | Every test: all suites + integration2 + upgrade + fuzz |

## Test plan

- [x] `make test-unit` runs without starting Docker cluster and discovers only non-integration tests
- [x] `make test` (integration suite) correctly excludes heavy/ldbc/load packages
- [x] `make test-integration-heavy` (`systest-heavy`) correctly runs only heavy packages (11 packages)
- [x] `make test SUITE=systest` runs both systest-baseline and systest-heavy (30 packages)
- [x] `make test SUITE=systest-baseline` runs only lightweight systests (18 packages)
- [x] `TestImportApis/SingleGroupShutOneAlpha` passes reliably with retry logic (CI)
- [x] Existing CI workflows continue to pass

### DIFF
--- a/dgraph/docker-compose.yml
+++ b/dgraph/docker-compose.yml
@@ -9,10 +9,6 @@ services:
     labels:
       cluster: test
       service: zero
-    deploy:
-      resources:
-        limits:
-          memory: 512M
     volumes:
       - type: bind
         source: ${LINUX_GOBIN:-$GOPATH/bin}
@@ -34,10 +30,6 @@ services:
     labels:
       cluster: test
       service: zero
-    deploy:
-      resources:
-        limits:
-          memory: 512M
     volumes:
       - type: bind
         source: ${LINUX_GOBIN:-$GOPATH/bin}
@@ -58,10 +50,6 @@ services:
     labels:
       cluster: test
       service: zero
-    deploy:
-      resources:
-        limits:
-          memory: 512M
     volumes:
       - type: bind
         source: ${LINUX_GOBIN:-$GOPATH/bin}
@@ -74,10 +62,6 @@ services:
   alpha1:
     image: dgraph/dgraph:local
     working_dir: /data/alpha1
-    deploy:
-      resources:
-        limits:
-          memory: 2048M
     volumes:
       - type: bind
         source: ${LINUX_GOBIN:-$GOPATH/bin}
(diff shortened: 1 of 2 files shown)


xqqp:
> Add Docker memory limits to all docker-compose services to prevent OOM kills on macOS Docker Desktop (zeros: 512M, alphas: 2-4GB, minio: 512M)

This just shifts OOM kills from the OS to docker, not sure if this really accomplishes much. Setting a memory limit on docker containers does not make applications use less memory.

mlwelles:
@xqqp: Good point.  While they did *seem* to improve stability, it was a purely subjective *seeming*, and you're right -- they really just shift where the OOM kill comes from, and the amount of noise they add to the compose files is icky.  Removed.

PR #9633, opened 2026-03-04
fix(worker): protect CDC pendingTxnEvents map read with mutex

shaunpatterson:
## Summary
- Direct map read without holding mutex while addToPending/removeFromPending hold it
- Can cause `fatal error: concurrent map read and map write`
- Added `getFromPending` method that acquires the lock before reading

## Test plan
- [x] `go build ./worker/...` passes
- [ ] `go test -race ./worker/...`


shaunpatterson:
Closing — consolidating PRs.

PR #9634, opened 2026-03-04
fix(edgraph): remove auth token from error message to prevent credential leak

shaunpatterson:
## Summary
- Auth token was included verbatim in error message: `"Provided auth token [%s] does not match"`
- Leaked credentials to logs and potentially to clients
- Removed the token from the error message

## Test plan
- [x] `go build ./edgraph/...` passes


shaunpatterson:
Closing — consolidating PRs.

PR #9636, opened 2026-03-04
fix(posting): use math.MaxInt consistently for length sentinel

shaunpatterson:
## Summary
- `length` initialized to `math.MaxInt` but checked against `math.MaxInt64`
- On 32-bit systems these differ, causing the sentinel check to fail

## Test plan
- [x] `go build ./posting/...` passes


shaunpatterson:
Closing — consolidating PRs.

PR #9637, opened 2026-03-04
fix(worker): fix goroutine leak in monitorRaftMetrics

shaunpatterson:
## Summary
- `time.Tick` in infinite loop with no exit condition leaked goroutine and ticker permanently
- Replaced with `time.NewTicker` + `defer Stop()` and integrated with `n.closer` for clean shutdown
- Bumped closer count from 4 to 5 to account for the new goroutine

## Test plan
- [x] `go build ./worker/...` passes


shaunpatterson:
Closing — consolidating PRs.

PR #9639, opened 2026-03-04
fix(query): add overflow check for uint16 attribute ID mapping

shaunpatterson:
## Summary
- Attribute IDs stored as `uint16` silently wrap at 65,535 unique predicates
- Different attributes could map to the same ID, corrupting JSON output
- Added bounds check that logs error and returns 0 on overflow

## Test plan
- [x] `go build ./query/...` passes


shaunpatterson:
Closing — consolidating PRs.

PR #9640, opened 2026-03-04
fix(posting): use time.NewTicker instead of time.Tick in metrics goroutine

shaunpatterson:
## Summary
- The metrics goroutine in `initMemoryLayer` had no shutdown path — it ran forever with no way to exit
- `defer ticker.Stop()` was unreachable dead code since the goroutine never returned
- Added a `z.Closer` to `MemoryLayer` so the goroutine can be signaled to stop during `Cleanup()`
- The goroutine now uses `select` to listen for both ticker events and the close signal, allowing proper cleanup

## Test plan
- [x] `go build ./posting/...` passes
- [x] `go test ./posting/...` passes

### DIFF
--- a/posting/lists.go
+++ b/posting/lists.go
@@ -48,6 +48,9 @@ func SetEnabledDetailedMetrics(enableMetrics bool) {
 
 // Cleanup waits until the closer has finished processing.
 func Cleanup() {
+	if MemLayerInstance != nil && MemLayerInstance.closer != nil {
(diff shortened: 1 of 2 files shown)


xqqp:
> time.Tick creates a ticker that can never be garbage collected

This is wrong

shaunpatterson:
Closing — consolidating PRs.

PR #9641, opened 2026-03-04
fix(conn): don't return snappy reader to pool on EOF while caller holds reference

shaunpatterson:
## Summary
- On `io.EOF`, snappy reader was immediately returned to `sync.Pool`
- If recycled to another goroutine, `Reset()` would corrupt the reader while the original caller still references it
- Removed eager pool return; GC reclaims when no references remain

## Test plan
- [x] `go build ./conn/...` passes


shaunpatterson:
Closing — consolidating PRs.

PR #9642, opened 2026-03-04
fix(posting): add missing lock in GetWithLockHeld

shaunpatterson:
## Summary
- `GetWithLockHeld` in both `viLocalCache` and `viTxn` accesses `pl.cache` (read and write) without holding the posting list lock
- The "lock held" in the name refers to a higher-level key lock (`txn.LockKey`), not the posting list mutex
- Added `pl.Lock()/defer pl.Unlock()` to match the `Get()` implementations and prevent data races on `pl.cache`

## Test plan
- [ ] Verify `go vet ./posting/` shows no new warnings
- [ ] Run race detector on HNSW vector index tests that exercise `GetWithLockHeld`

### DIFF
--- a/posting/lists.go
+++ b/posting/lists.go
@@ -103,8 +103,7 @@ func (vc *viLocalCache) GetWithLockHeld(key []byte) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	pl.Lock()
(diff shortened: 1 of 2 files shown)


shaunpatterson:
Closing — consolidating PRs.

PR #9643, opened 2026-03-04
fix(raftwal): reorder WAL Save to match etcd Raft protocol

shaunpatterson:
## Summary
- WAL `Save()` was persisting entries → hard state → snapshot
- The etcd Raft protocol expects snapshot → entries → hard state
- Reordered to snapshot → entries → hard state so recovery can determine the correct starting point before replaying entries
- A crash between steps could previously leave the WAL inconsistent

## Test plan
- [ ] Verify `go build ./raftwal/` succeeds
- [ ] Verify `go vet ./raftwal/` shows no new warnings
- [ ] Run Raft integration tests

### DIFF
--- a/raftwal/storage_test.go
+++ b/raftwal/storage_test.go
@@ -196,6 +196,65 @@ func TestStorageCreateSnapshot(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


shaunpatterson:
Closing — consolidating PRs.

PR #9644, opened 2026-03-04
fix(posting): hold LocalCache lock when iterating deltas in ProcessDelta

shaunpatterson:
## Summary
- `oracle.ProcessDelta` iterates `txn.cache.deltas` under `oracle.Lock()` but not the `LocalCache` lock
- Concurrent mutation application can modify the deltas map, risking a `concurrent map iteration and map write` panic
- Snapshots the keys under `cache.RLock` into a local slice, then iterates the copy
- Avoids holding both `oracle.Lock` and `cache.Lock` simultaneously to prevent lock ordering issues

## Test plan
- [ ] Run race detector on transaction commit tests


shaunpatterson:
Closing — consolidating PRs.

PR #9645, opened 2026-03-04
fix(posting): deep copy committedEntries map in MutableLayer.clone()

shaunpatterson:
## Summary
- `clone()` was copying the `committedEntries` map by reference
- When `setMutationAfterCommit` is called with `refresh=false`, it writes directly into the shared map, racing with reads from any clone
- Deep copy the map so each clone has an independent copy

## Test plan
- [ ] Run race detector on posting list mutation tests

### DIFF
--- a/posting/list_test.go
+++ b/posting/list_test.go
@@ -1792,6 +1792,50 @@ func TestRecursiveSplits(t *testing.T) {
 	}
 }
 
+func TestCloneIndependentCommittedEntries(t *testing.T) {
+	original := newMutableLayer()
+	original.committedEntries[100] = &pb.PostingList{
+		Postings: []*pb.Posting{{Uid: 1}},
+	}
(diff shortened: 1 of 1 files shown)


posting/list_test.go
```diff
@@ -1792,6 +1792,50 @@ func TestRecursiveSplits(t *testing.T) {
...
+	require.Nil(t, cloned.committedEntries[300])
+}
+
+func TestCloneNilMutableLayer(t *testing.T) {
+	var mm *MutableLayer
+	cloned := mm.clone()
+	require.Nil(t, cloned)
+}
+
+func TestNewMutableLayerSentinelValues(t *testing.T) {
```
xqqp: What does this test achieve?

shaunpatterson:
Closing — consolidating PRs.

PR #9646, opened 2026-03-04
docs(worker): clarify DROP ALL idempotency on Raft apply path

shaunpatterson:
## Summary
- The sequential operations in DROP ALL are not atomic, but run on the Raft apply path
- If the node crashes mid-operation, Raft re-applies the entry on restart
- Each sub-step (ResetTxns, DeleteAll, ResetCache, applySchema, updateType) is idempotent
- Added a comment to clarify this is safe by design

## Test plan


shaunpatterson:
Closing — consolidating PRs.

PR #9647, opened 2026-03-04
fix(conn): log warning when Raft messages are dropped on full channel

shaunpatterson:
## Summary
- When the per-peer message channel (capacity 100) is full, messages were silently dropped
- This is especially problematic for `MsgProp` (follower→leader proposal forwarding) which is only sent once
- Added a warning log so operators can detect and investigate message drops

## Test plan


shaunpatterson:
Closing — consolidating PRs.

PR #9648, opened 2026-03-04
fix(worker): replace panic with glog.Fatalf in commitOrAbort disk write

shaunpatterson:
## Summary
- When `CommitToDisk` exhausts all retries, the code panicked with an unclear stack trace
- Replaced with `glog.Fatalf` which flushes logs before exiting, providing better diagnostics
- A fatal exit is correct here — continuing after failing to persist a committed Raft entry would risk state divergence

## Test plan
- [ ] Verify `go vet ./worker/` shows no new warnings


shaunpatterson:
Closing — consolidating PRs.

PR #9649, opened 2026-03-04
fix(conn): clean up confChanges map entries on timeout and error

shaunpatterson:
## Summary
- When `proposeConfChange` times out (returning `errInternalRetry`) or errors, the conf change entry was never removed from `n.confChanges`
- Over repeated timeouts, this map grows without bound (acknowledged by existing TODO comment)
- Added cleanup on all error/timeout paths
- `DoneConfChange` is already tolerant of missing entries, so a late-arriving apply is handled gracefully

## Test plan
- [ ] Run Raft membership change tests

### DIFF
--- a/conn/node_test.go
+++ b/conn/node_test.go
@@ -48,6 +48,67 @@ func (n *Node) run(wg *sync.WaitGroup) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


shaunpatterson:
Closing — consolidating PRs.

PR #9650, opened 2026-03-04
fix(conn): add exponential backoff to addToCluster conf change retry

shaunpatterson:
## Summary
- `addToCluster` retried `proposeConfChange` indefinitely with no backoff and no context cancellation check
- During network partitions, this wasted CPU in a tight loop
- Added exponential backoff (100ms to 10s cap) with context cancellation checks between retries

## Test plan
- [ ] Run Raft membership change tests


shaunpatterson:
Closing — consolidating PRs.

PR #9651, opened 2026-03-04
fix(worker): add exponential backoff to oracle delta proposal retry

shaunpatterson:
## Summary
- The oracle delta proposal loop retried forever with no backoff when proposals kept failing
- This burned CPU and blocked shutdown since there was no closer check
- Added exponential backoff (100ms to 30s cap) and a closer check so the goroutine can exit cleanly on shutdown

## Test plan
- [ ] Verify `go vet ./worker/` shows no new warnings
- [ ] Run Raft delta proposal tests

### DIFF
--- a/worker/groups.go
+++ b/worker/groups.go
@@ -1058,18 +1058,33 @@ func (g *groupi) processOracleDeltaStream() {
 					}
 				}
 			}
-			for {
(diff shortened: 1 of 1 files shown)


shaunpatterson:
Closing — consolidating PRs.

PR #9652, opened 2026-03-04
feat(search): add BM25 ranked text search

shaunpatterson:
## Summary

- Adds BM25 relevance-ranked text search to Dgraph as a new query function and filter
- New `@index(bm25)` schema directive and `bm25(predicate, "query")` DQL syntax
- Results sorted by BM25 relevance score (IDF-weighted term frequency with document length normalization)
- Supports custom k/b tuning parameters: `bm25(pred, "query", "1.5", "0.5")`
- Works as both root function and `@filter`

## Changes

| File | Description |
|------|-------------|
| `tok/tok.go` | BM25Tokenizer with duplicate-preserving tokens and TokensWithFrequency |
| `tok/tokens.go` | GetBM25QueryTokens with deduplication and encoding |
| `x/keys.go` | BM25IndexKey, BM25DocLenKey, BM25StatsKey helpers |
| `posting/index.go` | addBM25IndexMutations, updateBM25Stats write path |
| `worker/task.go` | handleBM25Search with full BM25 scoring engine |
| `worker/tokens.go` | BM25 verification and token generation wiring |
| `dql/parser.go` | Register "bm25" as valid function name |
| `tok/tok_test.go` | 12 unit tests for tokenizer, frequencies, stemming, stopwords |
| `query/query_bm25_test.go` | 12 integration tests for queries, ordering, filters, pagination |
| `query/common_test.go` | BM25 test schema and 7 test documents |

## BM25 Formula

```
score(doc, term) = IDF(term) * (k+1) * tf / (k * (1 - b + b * docLen/avgDL) + tf)
IDF(term) = log1p((N - df + 0.5) / (df + 0.5))
```

Default: k=1.2, b=0.75 (Lucene/Elasticsearch variant with non-negative IDF)

## Test plan

- [x] Unit tests pass: `go test ./tok/... -run TestBM25 -v` (12 tests)
- [x] `go vet` clean on all modified packages
- [x] All existing tests continue to pass
- [x] Integration tests: `go test -tags integration -run TestBM25 ./query/ -v` (requires Docker cluster)

### DIFF
--- a/dql/parser.go
+++ b/dql/parser.go
@@ -1701,7 +1701,7 @@ func validFuncName(name string) bool {
 
 	switch name {
 	case "regexp", "anyofterms", "allofterms", "alloftext", "anyoftext", "ngram",
-		"has", "uid", "uid_in", "anyof", "allof", "type", "match", "similar_to":
+		"has", "uid", "uid_in", "anyof", "allof", "type", "match", "similar_to", "bm25":
 		return true
 	}
 	return false
--- a/posting/bm25block/bm25block.go
+++ b/posting/bm25block/bm25block.go
@@ -1,0 +1,261 @@
+/*
+ * SPDX-FileCopyrightText: © 2017-2025 Istari Digital, Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// Package bm25block provides block-based storage for BM25 index data.
+//
+// Instead of storing all postings for a term in a single blob, this package
+// splits them into fixed-size blocks (~128 entries). Each block is stored as
+// a separate Badger KV entry, and a lightweight directory indexes the blocks.
+//
+// This enables:
+//   - Selective I/O: queries only read blocks they need
+//   - WAND/Block-Max WAND: per-block upper bounds enable early termination
+//   - Efficient mutations: only the affected block is rewritten
+package bm25block
+
+import (
+	"encoding/binary"
+	"math"
+	"sort"
+
+	"github.com/dgraph-io/dgraph/v25/posting/bm25enc"
+)
+
+const (
+	// TargetBlockSize is the ideal number of entries per block.
+	TargetBlockSize = 128
+	// MaxBlockSize is the threshold at which a block is split.
+	MaxBlockSize = 256
+	// DocLenBlockSize is the target entries per document-length block.
+	DocLenBlockSize = 512
+
+	// dirHeaderSize is 4 (blockCount) + 4 (nextID).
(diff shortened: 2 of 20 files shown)


matthewmcneely:
Shaun, thank you for this. The depth of work here is obvious: WAND/Block-Max, length normalization, the test coverage, the careful concurrency notes.

I'm going to decline it, and I want to be transparent about why so the time isn't lost.

The decision isn't really about BM25 as a feature. It's that landing it as proposed would mean carrying a parallel storage and retrieval stack inside Dgraph, alongside the one we already have:

* A second write path on LocalCache (bm25Writes) that bypasses posting lists, deltas, and rollup, with its own BitBM25Data user-meta and its own commit branch in mvcc.go.
* Two new packages (posting/bm25block, posting/bm25enc) that re-implement sorted-UID blocks, splits, and value encoding — work the posting list already does.
* A 600+ line WAND engine inside worker/, plus a bm25_score pseudo-predicate threaded through the query planner via ParentVars.

The note in lists.go is telling: "two transactions updating different UIDs that share a common term could theoretically race… If higher write concurrency is needed, blocks should be integrated into the posting list delta mechanism." That's a real concurrency regression relative to every other tokenizer, mitigated only by leaning on Raft serialization. And the cleanest fix is exactly the integration this PR chose not to do.

Every other tokenizer (term, fulltext, trigram, ngram, geo, hash) lands as a Tokenizer impl in tok/ plus a FuncType case in worker/task.go. The reason this PR is +3471 / −4 is that it isn't really shaped like a tokenizer, it's a retrieval engine sitting next to the existing one. Maintaining two indexing pipelines, two write paths, and two concurrency stories is the cost we'd be signing up for in perpetuity.

If you're up for it, the version we'd be down to review would:

- Express BM25 postings as a normal posting list — term frequency as the value, document length stored next to the predicate's existing length/count machinery.
- Drop bm25block/bm25enc in favor of the existing block format and let MVCC do the delta merging.
- Land scoring as a tokenizer + a query-side ranker, without a pseudo-predicate or new ParentVars channel.
- Keep WAND if it pays off, but as an optimization on top of the standard posting list iterator, not a replacement for it.

That's a much smaller PR — and it inherits all of Dgraph's existing concurrency, snapshot, backup, and rollup behavior for free. Happy to sketch the integration points if useful.

Either way, thank you again — this is clearly a lot of careful work, and the analysis you did (especially around the IDF variant, length normalization, and the block-max upper bounds) is exactly the kind of rigor we want on a feature like this.

PR #9658, opened 2026-03-16
fix(test): resolve rate-limit failures when downloading  datasets

shiva-istari:
CI workflows on forked PRs fail with HTTP 429 (Too Many Requests) when downloading benchmark datasets from GitHub, because shared-IP runners hit rate limits on unauthenticated requests. This fix replaces GitHub web URLs with direct CDN URLs, adds retry logic with exponential backoff, limits concurrent downloads to 5, and cleans up partial files on failure. It also adds GitHub Actions caching to avoid re-downloading data files on repeat runs. Together these changes make benchmark data downloads reliable for both forked and internal CI workflows.


t/t.go
```diff
@@ -1159,7 +1159,27 @@ var rdfFileNames = [...]string{
 	"workAt_0.rdf"}
 
 var ldbcDataFiles = map[string]string{
-	"ldbcTypes.schema": "https://github.com/dgraph-io/dgraph-benchmarks/blob/main/ldbc/sf0.3/ldbcTypes.schema?raw=true",
+	"ldbcTypes.schema": "https://media.githubusercontent.com/media/dgraph-io/dgraph-benchmarks/refs/heads/main/ldbc/sf0.3/ldbcTypes.schema",
+}
+
+func wgetWithRetry(fname, url, dir string) error {
```
mwelles-istari: This function is a near-exact duplicate of `downloadFile` in `dgraphtest/load.go`. Is there a reason it needs to be redefined here instead of moving it to a shared location and reusing it?

This is especially concerning here because the function contents have clearly had to evolve over time as different failure causes were found and more guards were added. Every place it's duplicated is one more place to find and fix (or forget to fix) when the next issue arises.

The duplication is compounded by the double-retry approach — both copies pass `--tries=3 --waitretry=5 --retry-connrefused` to wget AND wrap that in a Go-level 3-attempt loop with exponential backoff, meaning up to **9 wget invocations** per file with cumulative delays that could reach several minutes before a genuine failure is reported. If this retry logic needs tuning in the future (and it likely will), having it in exactly one place is essential.

mwelles-istari (changes requested):

.github/workflows/ci-dgraph-integration2-tests.yml
```diff
@@ -29,6 +29,15 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Restore benchmark dataset cache
+        uses: actions/cache/restore@v4
+        with:
+          path: dgraphtest/datafiles
+          key: dataset-dgraphtest-v1
```
mwelles-istari: @shiva-istari one thing I'm concerned about here is that what version of the dataset fixtures is used is non-deterministic, and there's no way to know for sure which version of which fixture was used in a particular build if the cache was used. Which build seeded that cache? Which version of which fixture was cached when it did?

There's a hardcoded "-v1" suffix on the key, which AFAIK doesn't refer to any version of the test data being used — just the version of the key name itself.

I think instead either:
- The release tag of the test fixtures to use should be configured in an env var or project file, such that:
  - Only the versions of the fixtures at that tag can be downloaded.
  - The cache key suffix used must match the release tag (e.g. `dataset-dgraphtest-<release-tag>`).
  - If possible, make it controllable via a test execution runtime flag as well.

This will ensure builds get the same version of the test data they would on a fresh checkout even when a cache copy is used, that builds are idempotent and test results are reproducible. And it ensures that any version of a given test data asset is downloaded only once, then cached and reused by every subsequent build that uses it.

**Additional concern:** The file-exists check that guards re-downloading (`fi.Size() > 0`) won't catch corrupted or truncated files from a previous failed run. Combined with replacing `MakeDirEmpty` with `os.MkdirAll` (which preserves existing content), there's no mechanism to detect or recover from a bad cached file. A checksum validation or known-size check would improve reliability here.

mwelles-istari (changes requested):

dgraphtest/load.go
```diff
@@ -604,11 +605,22 @@ func (d *Dataset) ensureFile(filename string) string {
...
-
-	if _, err := cmd.CombinedOutput(); err != nil {
-		return fmt.Errorf("error downloading file %s: %w", fname, err)
+	const maxRetries = 3
+	fpath := filepath.Join(datasetFilesPath, fname)
+	for attempt := 1; attempt <= maxRetries; attempt++ {
+		cmd := exec.Command("wget", "--tries=3", "--waitretry=5", "--retry-connrefused", "-O", fname, url)
+		cmd.Dir = datasetFilesPath
+
+		if out, err := cmd.CombinedOutput(); err != nil {
```
mlwelles: The Go-level retry loop (3 attempts) wraps a wget invocation that itself retries 3 times (`--tries=3 --waitretry=5`). This means up to **9 wget requests** per file, with cumulative backoff delays that could reach several minutes before a genuine failure is reported.

Consider either:
- Removing the wget-level retries (`--tries=1`) and letting the Go loop handle all retry logic, or
- Removing the Go-level loop and relying solely on wget's built-in retry mechanism.

Having both layers is redundant and makes the failure timeline hard to reason about.

t/t.go
```diff
@@ -1449,7 +1478,6 @@ func main() {
 	procId = rand.Intn(1000)
 
 	err := run()
-	_ = os.RemoveAll(*tmp)
 	if err != nil {
 		os.Exit(1)
 	}
```
mlwelles, line 1407: Removing `os.RemoveAll(*tmp)` is necessary for the caching strategy, but it silently changes behavior for local development — test data will now accumulate in the temp directory across runs and never be cleaned up.

Rather than removing this unconditionally, add a new flag to control it:

```go
keepData = pflag.Bool("keep-data", false,
    "Preserve downloaded test data after run (for CI caching). Default cleans up.")
```

Then in `main()`, restore the cleanup but make it conditional:

```go
err := run()
if !*keepData {
    _ = os.RemoveAll(*tmp)
}
```

The CI workflows would then pass `--keep-data` explicitly:

```yaml
cd t; ./t --suite=load --tmp=${{ github.workspace }}/test-data --keep-data
```

This fits the existing flag pattern (similar to `--keep` for clusters, `--download` for resources), is explicit and discoverable via `--help`, and avoids coupling behavior to the environment.

t/t.go
```diff
@@ -1189,20 +1210,26 @@ func downloadLDBCFiles(dir string) {
...
-				panic(fmt.Sprintf("error downloading a file: %s", string(out)))
+			sem <- struct{}{}
+			defer func() { <-sem }()
+
+			dlStart := time.Now()
+			if err := wgetWithRetry(fname, link, dir); err != nil {
+				panic(fmt.Sprintf("error downloading %s: %v", fname, err))
 			}
-			fmt.Printf("Downloaded %s to %s in %s \n", fname, dir, time.Since(start))
-		}(fname, link, &wg)
```
mlwelles, line 1205: Calling `panic()` inside a goroutine will crash the entire program without unwinding other goroutines or running deferred cleanup. The original code had this same issue, but since this PR is improving error handling, it would be worth fixing.

Consider using `errgroup.Group` instead of `sync.WaitGroup` — it propagates errors cleanly from goroutines and would allow `downloadLDBCFiles` to return an error rather than panicking.

mlwelles (changes requested):
Additional findings beyond the code duplication and cache key concerns already noted:

**Note:** `t/t.go:1133` still has `var suffix = "?raw=true"` which gets appended to `media.githubusercontent.com` URLs via `baseUrl + name + suffix`. The `?raw=true` parameter is a GitHub blob-URL convention and is unnecessary for direct CDN URLs. Not harmful, but misleading — should be removed for consistency with the other URL updates in this PR.

.github/workflows/ci-dgraph-integration2-tests.yml
shiva-istari: now we fetches the LFS pointer from raw.githubusercontent.com, parses the oid sha256: and size, then verifies the downloaded content matches both. Any mismatch deletes the file and returns an error.

t/t.go
```diff
@@ -1121,18 +1128,17 @@ func isHeavyPackage(pkg string) bool {
...
-var datafiles = map[string]string{
-	"1million-noindex.schema": "https://raw.githubusercontent.com/dgraph-io/dgraph-benchmarks/refs/heads/main/data/1million-noindex.schema",
-	"1million.schema":         "https://raw.githubusercontent.com/dgraph-io/dgraph-benchmarks/refs/heads/main/data/1million.schema",
-	"1million.rdf.gz":         "https://media.githubusercontent.com/media/dgraph-io/dgraph-benchmarks/refs/heads/main/data/1million.rdf.gz",
-	"21million.schema":        "https://raw.githubusercontent.com/dgraph-io/dgraph-benchmarks/refs/heads/main/data/21million.schema",
-	"21million.rdf.gz":        "https://media.githubusercontent.com/media/dgraph-io/dgraph-benchmarks/refs/heads/main/data/21million.rdf.gz",
+// datafilePaths maps filenames to their paths inside the dgraph-benchmarks repo.
+// URLs are constructed at runtime from BenchmarkDataRef().
+var datafilePaths = map[string]string{
+	"1million-noindex.schema": "data/1million-noindex.schema",
```
mwelles-istari: This map is is duplicated in two different places.  DRY:  Define it in once and have both files use the same map.

t/t.go
```diff
@@ -1156,56 +1162,81 @@ var rdfFileNames = [...]string{
...
-		cmd := exec.Command("wget", "-O", fname, link)
-		cmd.Dir = *tmp
-
-		if out, err := cmd.CombinedOutput(); err != nil {
-			fmt.Printf("Error %v\n", err)
-			panic(fmt.Sprintf("error downloading a file: %s", string(out)))
+	ref := testutil.BenchmarkDataRef(*dataRef)
+	fmt.Printf("Using benchmark data ref: %s\n", ref)
+	for fname, repoPath := range datafilePaths {
+		fpath := filepath.Join(*tmp, fname)
```
mwelles-istari: This downlodFiles() function duplicates alot of the same code in the ensureFile() logic in load.go.  Why not refactor and have them use shared functions instead?    This logic is going to need to be maintained going forward and having to keep it current in two places will add tech debt.

t/t.go
```diff
@@ -1156,56 +1162,81 @@ var rdfFileNames = [...]string{
...
+		if testutil.FileExistsAndValid(fpath) {
+			fmt.Printf("Skipping %s (already exists)\n", fname)
+			continue
+		}
+		var err error
+		if strings.HasSuffix(fname, ".rdf.gz") {
+			err = testutil.DownloadLFSFile(fname, ref, repoPath, *tmp)
```
mwelles-istari: This logic is brittle -- while *right now* all of and only the "rdf.gz" suffixed files are LFS tracked, there's no guarantee that they will remain so.   If you need to detect if a file is LFS tracked and handle it differently, then instead of assuming it's so based on the suffix, actually check to see if it is tracked in LFS 

Also, as implemented now, you're requiring every caller to know which file is tracked in LFS and which aren't and to know how to handle each. That's t.go is the test runner, it's responsibility it running the tests.  load.go is responsible for running the load tests.   Neither of them is where one would expect to need to manage (duplicated) logic  to determine what  download mechanism to use to for which kinds of data file   That responsibility falls inside the package that's responsible for downloading test data.   The call sites should only need to call a single DownloadTestData function, and have *it* determine if the file is LFS tracked or not and how to handle the download, calling the appropriate  *internal* helper functio for doing so. 

"testutil" isn't a vary helpful package name to have these defined in  All of these download helpers are used for test data.   That's a clear area of concern  package.   Anytime there's a "utill" package it's usually short hand for "I didn't know where to put this" -- "testutil" is a grab bag.   Putting all of these in a "testdata" or "test/fixtures" module and having it provide a clear and simple API surface for its consumers would be a lot better. 

General rule of thumb:

> - Package by Domain/Feature, not Type: Avoid generic package names like models, utils, or controllers. Instead, group related functionality around specific domains or features (e.g., auth, payments, user). This makes the package's purpose clear and reduces circular dependencies.
> - Keep Packages Flat and Simple: Start with a flat structure and only introduce new packages or subdirectories when a clear need arises, such as isolating a significant feature or a reusable piece of code. Big files or packages are not inherently bad if they are cohesive.
> - Name Packages Carefully: Use short, lowercase, and descriptive names that clearly convey the package's purpose. Avoid underscores or plural forms. Good examples from the standard library include fmt, http, and json.
> - Minimize Exported Surface Area: Export only the types, functions, and variables that are explicitly intended for use by other packages (start identifiers with a capital letter). Everything else should be unexported (lowercase identifier) to hide internal implementation details. 

So, create a separate package for the test data file download helpers and file path / url maps, and you can have the call sites just call a single fixtures.DownloadTestFIxtures() or testdata.DownloadTestDataFiles() or whatever you want to call it.  and download everthing and return a slice of paths to the local data files,  and either have it accept optional varargs of the basenames of the data files  if the caller only needs specific files. 

If you wanted to make it more robust you could define a type alias for the test data file names so that they get registered and you're using the type system instead of requiring the caller to know the names in advance and not typo.  

The above is a lot, I know.  I took a pass at an partial  implementation so that what I'm describing is more clear.  

See this gist: 

https://gist.github.com/mlwelles/fc1323b93f34e6d349d62cb1607791cc

mwelles-istari (changes requested):

benchdata/benchdata.go
```diff
@@ -0,0 +1,454 @@
...
+	"log"
+	"net/http"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"time"
+
+	"golang.org/x/sync/errgroup"
+)
```
mlwelles, line 23: risk: no `_test.go` in this package. The hand-rolled LFS pointer parser, retry/backoff, gzip-magic check, and `DataRef` precedence are pure functions ideal for table tests. Without them, any regression to (e.g.) `parseLFSPointer` for a pointer with CRLF line endings, or a `DataRef` precedence change, will fail silently in CI under cache-hit and surface only on the next miss — which by then nobody is watching for.

At minimum: `parseLFSPointer`, `validateFile`, `verifyChecksum`, `DataRef`, and an httptest-backed `resolveAndDownload` round-trip.

benchdata/benchdata.go
```diff
@@ -0,0 +1,454 @@
...
+	if err := os.MkdirAll(destDir, 0o755); err != nil {
+		return nil, fmt.Errorf("creating dest dir: %w", err)
+	}
+
+	paths := make([]string, len(files))
+	for i, f := range files {
+		paths[i] = filepath.Join(destDir, string(f))
+	}
+
+	g, _ := errgroup.WithContext(context.Background())
```
mlwelles, line 180: risk: `ctx` is discarded with `_`. With `errgroup.WithContext`, that context is cancelled when any goroutine returns a non-nil error — but `downloadToFile` calls `http.NewRequest`, not `http.NewRequestWithContext`, so a sibling failure won't abort in-flight downloads of large LFS blobs. On a 21M.rdf.gz failure you'll still wait for the other four to finish (or time out).

Fix: capture the ctx, plumb it into `downloadToFile` via `http.NewRequestWithContext(ctx, ...)`, and into `fetchRawBlob` similarly. `resolveAndDownload` should accept ctx as its first arg.

benchdata/benchdata.go
```diff
@@ -0,0 +1,454 @@
...
+func downloadToFile(url, destPath, token string) error {
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return fmt.Errorf("creating request: %w", err)
+	}
+	if token != "" && strings.Contains(url, "githubusercontent.com") {
+		req.Header.Set("Authorization", "Bearer "+token)
+	}
+
+	resp, err := http.DefaultClient.Do(req)
```
mlwelles, line 367: risk: `http.DefaultClient` has no timeout. A stalled TCP connection (no FIN, no RST) can wedge a download forever — the CI job hangs until the workflow-level timeout, far past where a useful error would appear.

Also — `http.DefaultClient` is shared global state; if anything else in the binary tweaks it, this download silently inherits that. Use a package-private client:

```go
var httpClient = &http.Client{Timeout: 10 * time.Minute}
```

Or better, derive per-attempt timeout from a context with deadline.

benchdata/benchdata.go
```diff
@@ -0,0 +1,454 @@
...
+	n, err := io.ReadFull(resp.Body, buf)
+	if err == io.ErrUnexpectedEOF {
+		// File is smaller than 1KB -- that's fine, use what we got.
+		return buf[:n], nil
+	}
+	if err != nil {
+		return nil, err
+	}
+	// Read exactly 1KB. If there's more data, it's not an LFS pointer.
+	return buf[:n], nil
```
mlwelles, line 260: nit: the comment on L259 says "if there's more data, it's not an LFS pointer" but the code returns `buf[:n]` (1KB) regardless — `parseLFSPointer` will then check the prefix, find no LFS magic in the first 1KB of (e.g.) a small JSON schema, and correctly return `nil`. That works, but it's coincidental, not enforced. If a non-LFS file's first 1KB ever happens to start with the LFS magic prefix (extremely unlikely but not impossible for a corrupted commit), we'd misidentify it.

Consider: peek with `bufio.Reader.Peek(maxPointerSize)` and explicitly require the response body to be exactly the pointer length when LFS-detected. Or check `Content-Length` header, which the Contents API sets correctly.

benchdata/benchdata.go
```diff
@@ -0,0 +1,454 @@
...
+// detecting LFS), downloads it with retry, and validates the result.
+func resolveAndDownload(file TestDataFile, ref, destPath string) error {
+	path, ok := repoFilePath[file]
+	if !ok {
+		return fmt.Errorf("unknown test data file: %s", file)
+	}
+
+	lfsInfo, err := detectLFS(path, ref)
+	if err != nil {
+		// GitHub API unavailable (rate limit, network). Try media URL first
```
mlwelles, line 296: risk: when `detectLFS` fails (rate limit / network), the code falls back to *trying both* CDN URLs sequentially. That doubles the request count on the very condition (rate limiting) the PR is trying to fix. If the GitHub Contents API is rate-limited, `media.githubusercontent.com` may also be — you'll burn 2× retries × 3 attempts × N files before failing.

Safer: distinguish 403/429 (rate limit, abort) from 5xx/network (transient, fall back). A 403 from the API means you don't have an authoritative answer about LFS-tracking; falling back is wrong on first principles.

benchdata/benchdata.go
```diff
@@ -0,0 +1,454 @@
...
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
+		return nil, fmt.Errorf("GitHub Contents API raw %d: %s", resp.StatusCode, body)
```
mlwelles, line 244: nit: error message format — `"GitHub Contents API raw %d: %s"` — `%s` on a `[]byte` body is fine but if the API returns binary garbage in a 5xx response you'll dump it to stderr. Either truncate (`body[:min(256, len(body))]`) or extract the `message` field from the JSON error envelope GitHub returns.

benchdata/benchdata.go
```diff
@@ -0,0 +1,454 @@
...
+			_ = os.Remove(destPath)
+			if attempt < maxRetries {
+				time.Sleep(time.Duration(attempt*5) * time.Second)
+				continue
+			}
+			return fmt.Errorf("failed after %d attempts: %w", maxRetries, err)
+		}
+		return nil
+	}
+	return nil
```
mlwelles, line 354: nit: dead `return nil` on L354. The for loop's only paths are `return nil` (success), `continue`, or the final `return fmt.Errorf(...)` after `attempt == maxRetries`. The outer `return nil` is unreachable. Either remove it or convert the loop to use a `var lastErr error` accumulator and return that.

mlwelles (commented):
Re-review against `4ee619c`. All eight threads from prior rounds are resolved cleanly — the new `benchdata` package matches what the last comment described, errors now propagate via `errgroup`, the `--keep-data` flag is wired through, the cache key is bound to `benchmark-data-version`, and LFS detection is dynamic via the Contents API. Nice work.

A few new findings on the introduced code; nothing blocking, but worth addressing before merge:

- 🟡 No tests on `benchdata` despite hand-rolled LFS pointer parsing, retry/backoff, gzip-magic validation, and ref resolution priority — these are exactly the helpers that silently break.
- 🟡 `EnsureFiles` ignores the context returned by `errgroup.WithContext` — first failure won't cancel sibling downloads.
- 🟡 No per-attempt timeout on `http.DefaultClient` — a stalled connection hangs CI indefinitely.
- 🔵 A few smaller consistency / correctness issues inline.

Inline comments follow.

benchdata/benchdata.go
```diff
@@ -0,0 +1,454 @@
...
+	n, err := f.Read(header)
+	if err != nil {
+		return fmt.Errorf("reading file header: %w", err)
+	}
+	header = header[:n]
+
+	if bytes.HasPrefix(header, lfsPointerPrefix) {
+		return fmt.Errorf("file is a Git LFS pointer, not actual content")
+	}
+	if strings.HasSuffix(fpath, ".gz") && (n < 2 || !bytes.HasPrefix(header, gzipMagic)) {
```
mlwelles, line 419: nit: gzip-magic check is `strings.HasSuffix(fpath, ".gz")`. The TestDataFile constants encode this perfectly — a check on the filename string is one indirection too many. Consider exposing a `(f TestDataFile) IsGzip() bool` method or registering content-type metadata next to `repoFilePath`. This also lets you validate `.rdf` files (they should at least be UTF-8) and `.schema` files separately, instead of "everything not-gz is fine".

benchdata/benchdata.go
```diff
@@ -0,0 +1,454 @@
...
+	if v := os.Getenv(envDataRef); v != "" {
+		return v
+	}
+	versionFile := filepath.Join(pkgDir(), dataVersionFile)
+	if data, err := os.ReadFile(versionFile); err == nil {
+		if ref := strings.TrimSpace(string(data)); ref != "" {
+			return ref
+		}
+	}
+	return "main"
```
mlwelles, line 156: nit: silent fallback to `"main"` defeats the reproducibility goal of having a `benchmark-data-version` file. If the file exists but is empty/whitespace, falling back to `main` will silently use whatever HEAD points to. Prefer logging a warning when the file exists but yields no usable ref, so cache invalidation surprises (cache key from the file, but downloads from `main`) are discoverable.

benchdata/benchdata.go
```diff
@@ -0,0 +1,454 @@
...
+		return fmt.Errorf("computing SHA256: %w", err)
+	}
+	actual := hex.EncodeToString(h.Sum(nil))
+	if actual != info.OID {
+		return fmt.Errorf("SHA256 mismatch: local %s != expected %s", actual, info.OID)
+	}
+	return nil
+}
+
+func pkgDir() string {
```
mlwelles, line 451: nit: `pkgDir()` uses `runtime.Caller` to find `benchmark-data-version` next to the source file. This breaks when the binary is built with `-trimpath` or shipped without source — `thisFile` becomes `"benchdata/benchdata.go"` (relative), and `os.ReadFile` will resolve it against CWD, which is whatever shell happened to invoke `./t`.

Safer: `//go:embed benchmark-data-version` and read from the embedded FS. That also pins the version into the binary, which is exactly what you want for reproducibility.

t/t.go
```diff
@@ -1449,7 +1399,9 @@ func main() {
 	procId = rand.Intn(1000)
 
 	err := run()
-	_ = os.RemoveAll(*tmp)
+	if !*keepData && *tmp != "" {
```
mlwelles, line 1402: nit: `*tmp != ""` guard is correct but the contract is murky — if `run()` populated `*tmp` then errored out *before* downloading anything, we still skip cleanup with `--keep-data` but otherwise remove it. That's the right behavior. However, if `pflag.Parse()` panics (malformed flag) the cleanup never runs at all. Consider `defer` after parse:

```go
pflag.Parse()
defer func() {
    if !*keepData && *tmp != "" {
        _ = os.RemoveAll(*tmp)
    }
}()
```

Minor since the panic path is rare, but the `defer` form is also more idiomatic.

dgraphtest/load.go
```diff
@@ -590,25 +584,9 @@ func (d *Dataset) GqlSchemaPath() string {
...
-	return fullPath
-}
-
-func downloadFile(fname, url string) error {
-	cmd := exec.Command("wget", "-O", fname, url)
-	cmd.Dir = datasetFilesPath
-
-	if _, err := cmd.CombinedOutput(); err != nil {
-		return fmt.Errorf("error downloading file %s: %w", fname, err)
+	paths, err := benchdata.EnsureFiles(datasetFilesPath, benchdata.DataRef(""), benchdata.TestDataFile(filename))
```
mlwelles, line 587: nit: passing `benchdata.DataRef("")` here means `dgraphtest` callers can never override the ref, and silently get whichever ref the env var or version file resolves to. That's fine for now — but worth a comment noting the ref source isn't pluggable from this entry point, so future readers don't assume it is.

PR #9664, opened 2026-03-20
perf(hnsw): deduplicate neighbor updates and fix error handling

shaunpatterson:
## Summary
- **Deduplicate `addNeighbors` calls**: Replace `nnUidArray` (which accumulated duplicate UIDs when the same neighbor appeared across multiple HNSW layers) with direct iteration over `inboundEdgesAllLayersMap` keys, ensuring each neighbor is processed exactly once
- **Fix error handling for `addNeighbors(inUuid, ...)`**: The error was previously checked *after* the neighbor loop instead of immediately, allowing unnecessary work on failure
- **Fix silent error swallowing**: `searchPersistentLayer` errors were returned as `layerErr` (always `nil`) instead of `err`, masking real failures

## Test plan
- [ ] Verify existing HNSW vector index tests pass (`go test ./tok/hnsw/...`)
- [ ] Confirm `addNeighbors` is called once per unique neighbor UID (no redundant writes)
- [ ] Validate error propagation from `searchPersistentLayer` is no longer swallowed

### DIFF
--- a/tok/hnsw/persistent_hnsw.go
+++ b/tok/hnsw/persistent_hnsw.go
@@ -593,8 +593,6 @@ func (ph *persistentHNSW[T]) insertHelper(ctx context.Context, tc *TxnCache,
 	// var nns []persistentHeapElement[T]
 	visited := []persistentHeapElement[T]{}
 	inLevel := getInsertLayer(ph.maxLevels) // calculate layer to insert node at (randomized every time)
-	var layerErr error
-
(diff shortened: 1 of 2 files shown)


matthewmcneely:
@shaunpatterson The err swallowing was definitely an issue. However, could you add a test that verifies what this PR is claiming to solve?

matthewmcneely (approved):

PR #9667, opened 2026-03-24
fix(dql): support IRI-ref at query start

### LINKED ISSUE #9666: Bug: Lexer error in v25.3.0: Unexpected character while lexing DQL: U+003C '<' when query starts with IRI-ref
## Describe the bug

After upgrading from Dgraph v24.1.2 to v25.3.0, DQL queries using angle-bracket-wrapped names (e.g. `<caseNodeDel_xxx>`) as the first token after `{` fail with:

```
Unexpected character while lexing DQL: U+003C '<'
```

This is a regression introduced by commit [`6c31f0475`](https://github.com/dgraph-io/dgraph/pull/9355/changes/a2492ca65bf098650cbf34fa2194880db37f0e7e) ("Add implementation for RunDQL for v25 [#9355](https://github.com/dgraph-io/dgraph/pull/9355)"). In v24, `lexTopLevel` routed `{` directly to `lexQuery`, which supports `<` (IRI-ref style names). In v25, the new `lexIdentifyMutationOrQuery` state function was inserted between `lexTopLevel` and `lexQuery` to distinguish mutations (`set`/`delete`) from queries, but it does not handle the `<` character, causing the lexer to reject valid DQL.

## To Reproduce

1. Deploy Dgraph v25.3.0
2. Send the following DQL query:

```dql
{
  <caseNodeDel_xxx> as var(func: eq(caseXid, "xxx")) @filter(type(Case))

  check(func: uid(caseNodeDel_xxx)) {
    uid
    caseXid
  }
}
```

3. Observe error response:

```json
{
  "errors": [
    {
      "message": "while lexing {\n\n  <caseNodeDel_xxx> as var(func: eq(caseXid, \"xxx\")) @filter(type(Case))\n\n  check(func: uid(caseNodeDel_xxx)) {\n    uid\n    caseXid\n\n  }\n\n} at line 3 column 2: Unexpected character while lexing DQL: U+003C '<'",
      "extensions": {
        "code": "ErrorInvalidRequest"
      }
    }
  ]
}
```

## Expected behavior

The query should execute without lexing errors, as it did in v24.1.2. Supporting the <identifier> syntax is necessary to avoid breaking existing DQL workflows that rely on IRI-ref style naming to handle special characters or to maintain RDF-standard compatibility.

## Environment

- OS: Windows
- Language: Go
- Version: v25.3.0 (worked correctly in v24.1.2)

### Root Cause

In `dql/state.go`, commit [`6c31f0475`](https://github.com/dgraph-io/dgraph/pull/9355/changes/a2492ca65bf098650cbf34fa2194880db37f0e7e)  changed `lexTopLevel` to route `{` to the new `lexIdentifyMutationOrQuery` instead of directly to `lexQuery`. The new function handles `isNameBegin` characters (to detect `set`/`delete`/`del` for mutations vs query names) but omits the `case r == lsThan` branch, so `<` falls through to the `default` error case.

### Proposed Fix

Add a `case r == lsThan` branch in `lexIdentifyMutationOrQuery` (`dql/state.go`) that parses the IRI ref and then transitions to `lexQuery`:

```go
case r == lsThan:
    if err := lex.IRIRef(l, itemName); err != nil {
        return l.Errorf(err.Error())
    }
    return lexQuery
```

The proposed fix ensures that when the lexer encounters a < at the start of a block, it correctly identifies it as the start of an IRI-ref and proceeds to parse it as a query, maintaining the same behavior as v24.

Detailed fix implementation can be reviewed in this pr:
https://github.com/dgraph-io/dgraph/pull/9667

eileenaaa:
**Description**

This PR fixes a regression in DQL lexing where queries starting with an IRI-ref (names wrapped in `< >`, e.g., `<caseNodeDel_xxx>`) would fail with an `Unexpected character: U+003C '<'` error.

**Root Cause:**
In v25, the state function `lexIdentifyMutationOrQuery` was introduced to distinguish between mutations and queries. However, it lacked a case for the `lsThan` (`<`) character, causing the lexer to fail when a query block started with an IRI-ref instead of a standard name or keyword.

**Changes:**
- Updated `lexIdentifyMutationOrQuery` in `dql/state.go` to handle the `lsThan` (`<`) case.
- Ensures the lexer correctly calls `lex.IRIRef` and transitions to `lexQuery`.

Fixes https://github.com/dgraph-io/dgraph/issues/9666

**Checklist**

- [x] The PR title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) syntax, leading with `fix:`, `feat:`, `chore:`, `ci:`, etc.


dql/state_test.go
```diff
@@ -193,6 +193,24 @@ func TestVariablesDefault(t *testing.T) {
...
+	}`
+	l := lex.Lexer{
+		Input: input,
+	}
+	l.Run(lexTopLevel)
+	it := l.NewIterator()
+	for it.Next() {
+		item := it.Item()
+		require.NotEqual(t, item.Typ, lex.ItemError, "Error: %v", item.String())
+		t.Log(item.String())
```
shiva-istari, line 210: The test only asserts absence of errors but doesn't verify the IRI ref was actually parsed. A broken implementation that silently drops the token would still pass. Add a positive assertion that "user-query" was emitted as a token value (similar to TestNumberInLang), and drop the t.Log.
```suggestion
	var foundIRI bool
	for it.Next() {
		item := it.Item()
		require.NotEqual(t, item.Typ, lex.ItemError, "Error: %v", item.String())
		if item.Val == "user-query" {
			foundIRI = true
		}
	}
	require.True(t, foundIRI, "expected IRI ref 'user-query' to be lexed")```

shiva-istari (approved):

matthewmcneely (approved):

PR #9679, opened 2026-04-07
fix(backup): prevent duplicate predicates in backup manifest for vector indexes

shiva-istari:
**Description**

Fixed a bug in ProcessBackupRequest where vecPredMap[gid] was being built using predMap[gid] as the append base instead of vecPredMap[gid]. This caused all base predicates in the group to be duplicated in the manifest every time a vector predicate was found, making manifest.json grow larger and larger with each backup.
With multiple vector predicates in the same group, each iteration also overwrote the previous result, so only the last vector predicate's supporting entries (__vector_entry, __vector_, __vector_dead) were kept  the rest were silently lost.

### DIFF
--- a/dgraphtest/local_cluster.go
+++ b/dgraphtest/local_cluster.go
@@ -1397,6 +1397,27 @@ func (c *LocalCluster) GetAlphaGrpcEndpoint(id int) (string, error) {
 
 // CopyExportToHost copies exported files from the container to a host directory.
 // It returns the paths to RDF/JSON files and schema files on the host.
+func (c *LocalCluster) ReadFileFromContainer(containerPath string) ([]byte, error) {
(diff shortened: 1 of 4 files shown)


matthewmcneely (changes requested):
Looks good but there's no test covering the scenario where a group has multiple vector predicates with indexes. A test should:

1. Create a schema with 2+ float32vector predicates with HNSW indexes in the same group
2. Run ProcessBackupRequest
3. Verify the backup request's Predicates list includes the supporting predicates (VecEntry, VecKeyword, VecDead) for all vector predicates, not just the last one

Without such a test, this exact bug pattern (appending to the wrong map key) could easily regress.

matthewmcneely (approved):

PR #9686, opened 2026-04-22
feat: buildvars registry + extension hooks + FIPS-awareness + test-infra resilience

mlwelles:
## Summary

Two bodies of work that together make the dgraph codebase easier to maintain downstream while preserving identical upstream behavior:

1. **buildvars registry** — centralizes build-time configuration (binary name, image tags, build tags, toolchain refs) in one typed, godoc'd registry. Replaces scattered hardcoded literals and ad-hoc `os.Getenv` calls. 5 commits.
2. **extension hooks + resilience** — adds public function-var extension hooks to four test-infra packages (`testutil`, `t`, `compose`, `dgraphtest`), plus a minimal FIPS-awareness surface in `x/` (var + env helper), plus universally-beneficial observability and resilience improvements (stderr capture on failure, volume-race retry, test-skip helpers). 7 commits.

All 12 commits preserve upstream behavior byte-for-byte: every hook defaults to the exact code it replaces; every widened test secret is still valid under stock HMAC; every FIPS signal is opt-in via env and defaults off. The net effect on an upstream consumer is **no behavior change**; the net effect on a downstream fork is that none of these concerns require patching upstream files.

### buildvars registry (cmds 1–5, unchanged from prior review)

1. **`feat(buildvars): introduce buildvars registry`** — new `buildvars/` package (363 LOC + 324 LOC tests) + `buildvars/cmd/buildvars` CLI with `shell`/`make`/`plain` output.
2. **`refactor(buildvars): replace hardcoded binary/image literals`** — migrates 4 Go files and 3 shell scripts to read from the registry. Includes an incidental nil-panic fix in `getPortMappingsOnMac`.
3. **`fix(makefile): make BIN overridable`** — `dgraph/Makefile` uses `BIN ?= dgraph`, `go build -o $(INSTALL_TARGET)` (respects renames), and broader jemalloc detection (`/usr/lib`, `libjemalloc_pic.a` for Debian/Alpine/Wolfi). `t/Makefile`'s `check` target threads `$BIN`.
4. **`test(restore): skip on non-Linux`** — `namespace-aware/restore_test.go` skipped locally; CI unchanged.
5. **`docs(buildvars): drop fork-specific names from godoc + tests`** — env var renamed to `DGRAPH_COMPOSE_BUILD_DIR`; generic language throughout.

### Extension hooks + FIPS surface + resilience (6–12, new in this update)

6. **`refactor(buildvars): rename cmd/buildvarprint → cmd/buildvars; align Go ident ExtraBuildTags → PrivateBuildTags`** — two consistency-focused renames. The CLI dir now matches the top-level package name. The Go identifier `PrivateBuildTags` now pairs with env var `PRIVATE_BUILD_TAGS`, matching every other Var in the registry (Bin/BIN, etc.).

7. **`feat(x): add FIPSEnabled var + FIPSBinary() runtime helper`** — introduces two minimal, coordinated FIPS-awareness signals in package `x`:
   - `FIPSEnabled` (var, default false): compile-time signal. Declared as a var (not a const) so a build-tag-guarded sibling file in a downstream fork can flip it via `init()` before any test or main body runs. For all practical purposes behaves as a compile-time constant for `if x.FIPSEnabled { ... }` guards.
   - `FIPSBinary()` (function): runtime signal via `DGRAPH_FIPS_BINARY=1`. Covers the case where the test binary is built without FIPS tags but spawns a FIPS-enforcing dgraph subprocess.
   - Includes tests covering the env-var parsing (only exact `"1"` is true) and the default-false invariant.

8. **`refactor(testutil): extract ComposeArgs, EnvForCompose, BuildPlugins hooks`** — three public function-var hooks in new `testutil/hooks.go`. `StartAlphas`/`StopAlphasForCoverage` call the compose-args and env-for-compose hooks. `GeneratePlugins` delegates to the `BuildPlugins` hook (default is the verbatim previous implementation). Upstream behavior byte-exact.

9. **`test(t): add EnvForCompose + ComposeFileArgs hooks; resilience improvements`** — hook surface in `t/hooks.go` + threaded-through the test runner. Plus three universally-beneficial improvements:
   - **stderr capture**: `startCluster` captures docker-compose stderr and prints on failure. Previous `cmd.Stderr = nil` suppressed actionable error text (volume races, port conflicts, pull failures).
   - **volume-race retry**: `startCluster` retries `docker compose up` once after `down -v` on failure. Covers the docker volume-init race when multiple services share a named volume. One retry catches the class; anything persistent is a real error.
   - **container-log iteration**: `stopCluster` iterates every container in the project prefix rather than guessing `zero0..zero3, alpha0..alpha6`. Output is a fenced block per container for easier CI-log triage.
   - **defer-err bugfix**: `defer stopCluster(..., err)` was capturing err at defer-statement time (always nil), silently suppressing container-log dumps on test failure. Wrap in a closure to pick up the final err value.
   - `testTags(baseTag)` helper replaces inline `"-tags=integration"` strings and composes with `buildvars.GoRunTags`.

10. **`refactor(compose,dgraphtest): add extension hooks; rename local copy → copyFile`** — one hook in `compose/hooks.go` (`EmitUser`, wired into service generation), six hooks in `dgraphtest/hooks.go` (`ApplyContainerUser`, `WidenTempDirPerms`, `WidenSecretFilePerms`, `GeneratePlugins`, `SetupLinuxBinaries`, `HostBinaryName`, wired into `local_cluster.go`/`image.go`/`load.go`). All hooks default to no-op / upstream-pristine behavior. Also renames the local `func copy(src, dst string) error` in `dgraphtest/image.go` to `copyFile` (the name shadowed the Go builtin `copy`).

11. **`test: widen HMAC/JWT/MinIO test secrets >= 14 bytes; add FIPS-aware skip helpers`** — widens every HMAC/JWT/MinIO test secret to the 14-byte minimum NIST SP 800-131A requires and some FIPS-validated crypto providers enforce at `EVP_MAC_init`:
   - `"secret"` → `"secret-long-enough"` (schema VerificationKey, JWT signer)
   - `"secret1"` → `"secret1-long-enough"` (multi-tenancy)
   - `"secretkey"` → `"secretkey-long-enough"` (schema VerificationKey, MinIO secret key)
   - 9 hardcoded HS256 JWT fixtures in `graphql/resolve/auth_test.go` and `graphql/e2e/auth/auth_test.go` re-signed with the new key (verified against old key before replacing signature; no stray tokens modified).
   - Upgrade-path tests (`TestCheckUpgrade`, `TestQueryDuplicateNodes`, `TestRestoreFromOldBackup`) and EdDSA (`TestACLJwtAlgo`) gain a local `skipIfFIPSBinary(t)` helper that calls `t.Skip` when `x.FIPSEnabled || x.FIPSBinary()` is true. No-op on stock upstream builds.

12. **`style(admin): gofmt admin_auth_test.go`** — `gofmt -w`. Collapsed alignment in a map literal + trailing newline. No functional change.

## Why

All items from the prior PR body still apply. The expanded scope comes from discoveries while maintaining a downstream fork that:

- runs containers as nonroot (needs UID/GID env injection and filesystem-perm widening)
- runs a compose-file generator (needs compose-path rewriting + `--project-directory` projection)
- ships a custom Go toolchain (needs plugin-build override)
- enforces FIPS 140-3 (needs short-HMAC-key avoidance and FIPS-incompatible test skips)

Every single one of these concerns was reachable by patching upstream files with `//go:build` guards or inline `if os.Getenv(...)` checks. Each such patch is a maintenance burden: the fork tracks upstream changes to the patched file and must re-apply edits on every rebase. Extracting extension hooks eliminates that burden — the fork owns a single registration file per package and never touches upstream source again. Upstream sees only: neutral public vars with upstream-pristine defaults, and resilience improvements (stderr capture, volume-race retry, fenced log output) that help every user.

## API

```go
// buildvars
Bin = buildvars.NewVar("BIN", "dgraph")
path := filepath.Join(gopath, "bin", buildvars.Bin.Get())
buildvars.Bin.SetDefault("custom-binary-name") // or: os.Setenv("BIN", ...)

// x
if x.FIPSEnabled || x.FIPSBinary() {
    t.Skip("test requires features not available under FIPS")
}

// extension hooks (all four packages follow the same pattern)
var ComposeFileArgs = defaultComposeFileArgs
func defaultComposeFileArgs(path, baseDir string) []string {
    return []string{"-f", path}
}
// A fork replaces the var from its own package init() to override behavior.
```

## Tests

- `buildvars/buildvars_test.go`: 18 unit tests (Get/Default/SetDefault/Export/ExportAll, env-override precedence, derived Vars, concurrent access).
- `x/fips_base_test.go`: `TestFIPSEnabledDefault` (invariant) + `TestFIPSBinaryEnv` (env-var parsing across 6 states).
- `graphql/resolve/auth_test.go`: `TestStringCustomClaim`, `TestAudienceClaim`, `TestJWTExpiry` all pass with re-signed JWT fixtures.
- Full module build (`go build ./...`, excluding `testutil/custom_plugins/*` plugin-mode sources and `graphql/bench` test-only package) clean.
- `go vet` clean on all touched packages (pre-existing warnings in `conn/pool.go`, `schema/schema.go`, `graphql/admin/restore.go` untouched).
- `gofmt -l` clean on all touched files (the one remaining `gofmt` hit is generated `protos/pb/pb.pb.go`).

## Backwards compatibility

- All hook defaults are the verbatim previous implementations. Upstream behavior byte-for-byte identical.
- Widened HMAC keys are strictly-longer; any HMAC implementation accepts longer keys. Re-signed JWT fixtures verify against the new keys.
- FIPS skip helpers are no-ops unless `DGRAPH_FIPS_BINARY=1` is set or the binary is built with a FIPS-enforcing tag.
- `copy` → `copyFile` rename in `dgraphtest/image.go` is a local function; no exported API touched.
- `buildvars.ExtraBuildTags` → `buildvars.PrivateBuildTags` Go ident rename does not change the env var name (`PRIVATE_BUILD_TAGS` unchanged).

## Notes for review

- Copyright headers use `© 2017-2026 Istari Digital, Inc. SPDX-License-Identifier: Apache-2.0` on new files. Matches the repository convention for externally-contributed Apache-2.0 additions.
- The hook-pattern godoc in each `hooks.go` file names the upstream-pristine default and a concrete downstream use case, so a reader unfamiliar with the pattern can follow it from the declaration alone.
- `x/fips_base.go` is intentionally minimal: two symbols, no FIPS-specific logic. The actual FIPS cipher/curve/JWT-algorithm helpers stay in the fork (they are specific to Chainguard go-fips and MS-Go requirefips toolchains, not generally applicable upstream).


mlwelles:
Moved to #9690 which is based on a branch in this repo (dgraph-io/dgraph/mlwelles/buildvars-registry-and-hooks) rather than my fork. This lets the labeler workflow apply PR labels (forked-repo PRs lack pull-requests:write permission) and is consistent with other maintainer branches in this repo. All commits, history, and description transferred verbatim.

PR #9691, opened 2026-04-24
feat: buildvars registry + extension hooks + FIPS-awareness + test-infra resilience

mlwelles:
## Summary

Two refactors that make dgraph easier to maintain downstream and preserve upstream behavior byte-for-byte:

1. **buildvars registry** — centralizes build-time configuration (binary name, image tags, build tags, toolchain refs) in one typed, godoc'd registry. Replaces scattered hardcoded literals and ad-hoc `os.Getenv` calls. 2 commits.
2. **extension hooks + FIPS surface + resilience** — adds public function-var extension hooks to four test-infra packages (`testutil`, `t`, `compose`, `dgraphtest`), a minimal FIPS-awareness surface in `x/` (var + env helper), and resilience improvements that help every user (volume-race retry, resume-grace delay, test-skip helpers, widened test secrets). 7 commits.
3. **systest helper-image cleanup** — replaces the `nfs-backup` helper container's external `Dockerfile` with an inline `dockerfile_inline:` block, decoupling cluster bring-up from `make local-image` ordering. 1 commit.

Every hook defaults to the exact code it replaces. Every widened test secret stays valid under stock HMAC. Every FIPS signal is opt-in via env and defaults off. The nfs-backup change keeps the test behavior identical — the dgraph binary still arrives via the `/gobin` bind mount the compose declares. Net effect on upstream: **no behavior change**. Net effect on a downstream fork: none of these concerns require patching upstream files.

Supersedes #9690 with clean logical commit history. See [Relationship to #9690](#relationship-to-9690) below.

## Why

Each concern below surfaced while maintaining a downstream fork that:

- runs containers as nonroot (needs UID/GID env injection and filesystem-perm widening)
- runs a compose-file generator (needs compose-path rewriting and `--project-directory` projection)
- ships a custom Go toolchain (needs plugin-build override)
- enforces FIPS 140-3 (needs short-HMAC-key avoidance and FIPS-incompatible test skips)
- doesn't produce the `dgraph/dgraph:local` image tag (needs helper-image bases decoupled from that tag)

All of these previously required patching upstream files with `//go:build` guards or inline `os.Getenv` checks. Each patch is a maintenance burden — the fork re-applies edits on every rebase. Extension hooks eliminate the burden: the fork owns one registration file per package and leaves upstream source untouched.

Upstream sees only neutral public vars with upstream-pristine defaults, plus resilience improvements (volume-race retry, resume-grace delay, test-skip helpers) that help every user.

### 1. `feat(buildvars): introduce typed build-config registry`

New `buildvars/` package (363 LOC + 324 LOC tests) and `buildvars/cmd/buildvars` CLI with `shell`/`make`/`plain` output formats. Migrates hardcoded call sites in `contrib/jepsen`, `testutil/exec`, `dgraphtest/dgraph`, and systest scripts to read from the registry. Includes an incidental nil-panic fix in `getPortMappingsOnMac`.

### 2. `fix(makefile): make BIN overridable, expand jemalloc detection, use go build -o`

`dgraph/Makefile` now uses `BIN ?= dgraph`, `go build -o $(INSTALL_TARGET)` (respects renames), and broader jemalloc detection (`/usr/lib`, `libjemalloc_pic.a` for Debian/Alpine/Wolfi). `t/Makefile`'s `check` target threads `$BIN`.

### 3. `test(restore): skip namespace-aware restore tests on non-Linux`

Guards `systest/online-restore/namespace-aware/restore_test.go` with `//go:build linux` so Darwin CI does not fail on Linux-only kernel features.

### 4. `feat(x): add FIPSEnabled var + FIPSBinary() runtime helper`

Two minimal, coordinated FIPS-awareness signals in package `x`:

- `FIPSEnabled` (var, default false): compile-time signal. Declared as a var, not a const, so a build-tag-guarded sibling file in a downstream fork can flip it via `init()` before any test or main body runs. For `if x.FIPSEnabled { ... }` guards it behaves as a compile-time constant.
- `FIPSBinary()` (function): runtime signal via `DGRAPH_FIPS_BINARY=1`. Covers the case where the test binary builds without FIPS tags but spawns a FIPS-enforcing dgraph subprocess.

Tests cover env-var parsing (only exact `"1"` is true) and the default-false invariant.

### 5. `refactor(testutil): extract ComposeArgs, EnvForCompose, BuildPlugins hooks`

Three public function-var hooks in new `testutil/hooks.go`. `StartAlphas` and `StopAlphasForCoverage` call the compose-args and env-for-compose hooks. `GeneratePlugins` delegates to `BuildPlugins` (default is the verbatim previous implementation). Upstream behavior byte-exact.

### 6. `refactor(compose,dgraphtest): add extension hooks; rename copy → copyFile`

One hook in `compose/hooks.go` (`EmitUser`, wired into service generation). Six hooks in `dgraphtest/hooks.go` (`ApplyContainerUser`, `WidenTempDirPerms`, `WidenSecretFilePerms`, `GeneratePlugins`, `SetupLinuxBinaries`, `HostBinaryName`, wired into `local_cluster.go` / `image.go` / `load.go`). All hooks default to no-op or upstream-pristine behavior. Also renames `func copy(src, dst string) error` in `dgraphtest/image.go` to `copyFile`, since the name shadowed the Go builtin `copy`.

### 7. `test: widen HMAC/JWT/MinIO test secrets + add FIPS-aware skip helpers`

Widens every HMAC/JWT/MinIO test secret to the 14-byte minimum NIST SP 800-131A requires (which some FIPS-validated crypto providers enforce at `EVP_MAC_init`):

- `"secret"` → `"secret-long-enough"` (schema VerificationKey, JWT signer)
- `"secret1"` → `"secret1-long-enough"` (multi-tenancy)
- `"secretkey"` → `"secretkey-long-enough"` (schema VerificationKey, MinIO secret key)
- Re-signs 9 hardcoded HS256 JWT fixtures in `graphql/resolve/auth_test.go` and `graphql/e2e/auth/auth_test.go` with the new key (verified against the old key before replacing the signature).
- Adds a local `skipIfFIPSBinary(t)` helper to upgrade-path tests (`TestCheckUpgrade`, `TestQueryDuplicateNodes`, `TestRestoreFromOldBackup`) and EdDSA (`TestACLJwtAlgo`) that calls `t.Skip` when `x.FIPSEnabled || x.FIPSBinary()` is true. No-op on stock upstream builds.

Also aligns three env files (`dgraph/minio.env`, `systest/backup.env`, `systest/export/export.env`) so docker-compose test harnesses see the same widened secrets as Go code.

### 8. `refactor(t): extract EnvForCompose + ComposeFileArgs hooks; add resilience`

Hook surface in `t/hooks.go` threaded through the test runner. Plus two resilience improvements that help every user:

- **3-attempt volume-race retry.** `startCluster` retries `docker compose up` up to 3 times with 2s delay between attempts after `down -v` on failure. Covers the docker volume-init race when multiple services share a named volume. Prior retry-once was insufficient — 300ms is not long enough for docker to release mount points.
- **5s resume grace.** `resumeDefault` waits 5 seconds after the health-check wait-group before returning. Alpha `/health` returns OK before the alpha→zero connection pool is fully re-established, which causes "No connection exists" on the first post-resume test.

### 9. `style(admin): gofmt admin_auth_test.go`

`gofmt -w`. Collapsed alignment in a map literal plus a trailing newline. No functional change.

### 10. `fix(systest): inline nfs-backup Dockerfile, decouple from dgraph:local`

The `nfs-backup` helper container's external `Dockerfile` started `FROM dgraph/dgraph:local`, creating an implicit ordering dependency on `make local-image` and breaking any downstream build matrix that doesn't produce that tag. The helper image only needs glibc and NFS-client tooling — the dgraph binary already arrives via the `/gobin` bind mount the compose declares (`${LINUX_GOBIN:-$GOPATH/bin}` → `/gobin`) and is invoked as `/gobin/dgraph` regardless of what's baked into the image.

Replaces `build: { context: ., dockerfile: ./Dockerfile }` on alpha1 with Compose's `dockerfile_inline:` form basing on `debian:bookworm-slim` and installing `nfs-common` (provides `mount.nfs4`) + `util-linux` (provides `mountpoint`). Other services keep `image: dgraph-nfs-client:local` and reuse the built tag. Deletes the now-orphaned `systest/backup/nfs-backup/Dockerfile`. Test behavior is identical: the cluster brings up the same set of services, mounts the same NFS share, and runs the same backup/restore tests against the bind-mounted dgraph binary.

## API

```go
// buildvars
Bin = buildvars.NewVar("BIN", "dgraph")
path := filepath.Join(gopath, "bin", buildvars.Bin.Get())
buildvars.Bin.SetDefault("custom-binary-name") // or: os.Setenv("BIN", ...)

// x
if x.FIPSEnabled || x.FIPSBinary() {
    t.Skip("test requires features not available under FIPS")
}

// extension hooks (all four packages follow the same pattern)
var ComposeFileArgs = defaultComposeFileArgs
func defaultComposeFileArgs(path, baseDir string) []string {
    return []string{"-f", path}
}
// A fork replaces the var from its own package init() to override behavior.
```

## Relationship to #9690

The final content is **byte-identical** to #9690's last green tip (`d63438211`) through commit 9, plus commit 10 (the nfs-backup inline-Dockerfile change). Only the commit structure changed.

#9690 accumulated 17 commits through CI debugging and review feedback, including a merge from `upstream/main`. Reviewers had to reconstruct intent across commits that touched the same area of concern multiple times. This PR groups the same content into 10 commits by area of concern.

#9690 will be closed once this PR is green. Review comments do not migrate across GitHub PRs — **please re-state any unresolved feedback here if it still applies**.


buildvars/buildvars.go
```diff
@@ -0,0 +1,363 @@
...
+	// compile-time code paths set this to space-separated tag names
+	// (e.g. "myfork requirefips"). The Makefile concatenates
+	// PrivateBuildTags onto BuildTags, so callers typically read BuildTags
+	// to see the final tag set.
+	//
+	// "Private" here refers to fork-local build choices (tags that enable
+	// private/downstream code paths), not access control. The name is
+	// retained for compatibility with the established PRIVATE_BUILD_TAGS
+	// environment variable that Makefile and CI have long used.
+	PrivateBuildTags = newVar("PRIVATE_BUILD_TAGS", "")
```
mwelles-istari: Let's rename this CustomBuildTags,  CUSTOM_BUILD_TAGS

compose/compose.go
```diff
@@ -154,6 +155,9 @@ func initService(basename string, idx, grpcPort int) service {
 	svc.Image = opts.Image + ":" + opts.Tag
 	svc.ContainerName = containerName(svc.name)
 	svc.WorkingDir = fmt.Sprintf("/data/%s", svc.name)
+	if u := EmitUser(); u != "" {
```
mwelles-istari, line 165: Add a comment explaining what EmitUser() hook  call here provides (when digraph runs as non root user in compose tests, hook overrides default return of  "").

dgraphtest/load.go
```diff
@@ -24,20 +24,31 @@ import (
...
 	"github.com/dgraph-io/dgo/v250/protos/api"
+	"github.com/dgraph-io/dgraph/v25/buildvars"
 	"github.com/dgraph-io/dgraph/v25/dgraphapi"
 	"github.com/dgraph-io/dgraph/v25/enc"
 	"github.com/dgraph-io/dgraph/v25/x"
 )
 
+// hostBinaryFileName is the conventional filename for the host-OS-native
+// dgraph binary when it differs from the container binary (non-Linux). On
+// Linux the host and container binaries are the same file (named
```
mwelles-istari: s/file (named/file name returned by/

dgraphtest/local_cluster.go
```diff
@@ -118,11 +118,17 @@ func (c *LocalCluster) init() error {
 	if err != nil {
 		return errors.Wrap(err, "error while creating tempBinDir")
 	}
+	if err := WidenTempDirPerms(c.tempBinDir); err != nil {
```
mwelles-istari, line 126: Add  explaining is a hook, default is no-op,  can be overridden on forks where dgraph running as non-root user in compose test containers needs widened perms to read directories mounted from container host.

dgraphtest/local_cluster.go
```diff
@@ -306,6 +312,7 @@ func (c *LocalCluster) createContainer(dc dnode) (string, error) {
 	}
 
 	cconf := &container.Config{Cmd: cmd, Image: image, WorkingDir: dc.workingDir(), ExposedPorts: dc.ports()}
+	ApplyContainerUser(cconf)
```
mwelles-istari, line 327: Again add comment about hook here.

dgraphtest/local_cluster.go
```diff
@@ -1312,13 +1319,19 @@ func (c *LocalCluster) setupSecrets() error {
 		if err := os.WriteFile(c.encKeyPath, encKey, 0600); err != nil {
 			return err
 		}
+		if err := WidenSecretFilePerms(c.encKeyPath); err != nil {
```
mwelles-istari, line 1341: Add comment explaining hook and default behavior

dgraphtest/local_cluster.go
```diff
@@ -1375,6 +1388,12 @@ func runOpennssl(args ...string) error {
 }
 
 func (c *LocalCluster) GeneratePlugins(raceEnabled bool) error {
+	if tokenizers, handled, err := GeneratePlugins(raceEnabled, c.tempBinDir, baseRepoDir); handled {
```
mwelles-istari, line 1418: add comment explaining purpose

graphql/e2e/auth/auth_test.go
```diff
@@ -623,7 +623,7 @@ func TestQueryWithStandardClaims(t *testing.T) {
 					name
 				}
 			}`,
-			jwt:    "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiZXhwIjozNTE2MjM5MDIyLCJlbWFpbCI6InRlc3RAZGdyYXBoLmlvIiwiVVNFUiI6InVzZXIxIiwiUk9MRSI6IkFETUlOIn0.cH_EcC8Sd0pawJs96XPhpRsYVXuTybT1oUkluBDS8B4",
+			jwt:    "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiZXhwIjozNTE2MjM5MDIyLCJlbWFpbCI6InRlc3RAZGdyYXBoLmlvIiwiVVNFUiI6InVzZXIxIiwiUk9MRSI6IkFETUlOIn0.pZ2-Dib2lXrCeXghCoPD7CnZ8GUGXhv1WbbRQ7mhPnM",
```
mwelles-istari, line 635: Add comment explaining length change, minimum length required for FIPS

systest/21million/test-21million.sh
```diff
@@ -124,7 +124,7 @@ if [[ ${LOADER} == bulk ]]; then
             mkdir -p /data/alpha1
             mkdir -p /data/alpha2
             mkdir -p /data/alpha3
-            /gobin/dgraph bulk --schema=${SCHEMA_FILE} --files=${DATA_FILE} \
+            /gobin/${BIN:-dgraph} bulk --schema=${SCHEMA_FILE} --files=${DATA_FILE} \
```
mwelles-istari: Shouldn't this be ${GOBIN_DGRAPH_PATH:-/gobin/dgraph} ? not /gobin/${BIN:-dgraph}:

buildvars/buildvars.go
mlwelles: Done in 352ff5afe — renamed PrivateBuildTags → CustomBuildTags (and PRIVATE_BUILD_TAGS → CUSTOM_BUILD_TAGS) at the Var declaration, every reference (BuildTags composer, All slice, test names), and the test env-cleanup. `go test ./buildvars/...` passes.

compose/compose.go
mlwelles: Done in 352ff5afe — added a paragraph above the call explaining the hook surface: default returns "" (service inherits image USER), downstream consumers running dgraph as a non-root compose-test user override to return e.g. "${UID:-65532}" so bind-mounted host paths are readable by the in-container uid.

dgraphtest/load.go
mlwelles: Done in 352ff5afe — "the same file (named buildvars.Bin.Get())" → "the same file name (returned by buildvars.Bin.Get())".

dgraphtest/local_cluster.go
mlwelles: Done in 352ff5afe — added a paragraph at the first call site explaining: public hook, default no-op, downstream consumers running dgraph as non-root inside compose-test containers override to widen perms on host-side temp dirs (which are bind-mounted into containers) so the in-container uid can read/write those paths. The second call site below points back to the same hook with a one-line note.

dgraphtest/local_cluster.go
mlwelles: Done in 352ff5afe — comment explains: public hook, default no-op, downstream consumers running dgraph as a non-root user inside the test container override to set cconf.User to host uid:gid so files written by the container are readable on the host (and vice-versa for bind-mounts).

dgraphtest/local_cluster.go
mlwelles: Done in 352ff5afe — comment at the top of setupSecrets explains both call sites: public hook, default no-op; secret files write at 0600 (correct upstream); downstream consumers running the container as a non-root user that differs from the host owner override to widen perms (e.g. group/world-read) so the in-container uid can read the bind-mounted secrets.

dgraphtest/local_cluster.go
mlwelles: Done in 352ff5afe — comment explains: hook returns (nil, false, nil) upstream so the host-side fallback path runs; downstream consumers whose toolchain isn't directly host-invokable (e.g. forks pinning a Docker-only build image) override to compile inside that image and return the .so paths, in which case handled=true short-circuits the host fallback.

graphql/e2e/auth/auth_test.go
mlwelles: Done in 352ff5afe — added a paragraph at the top of the testCases block explaining: pre-signed JWTs use the widened secret ("secret-long-enough"); secret + fixture JWTs were re-signed when secrets were widened to 14+ bytes (NIST SP 800-131A HMAC minimum, enforced by FIPS-validated providers at EVP_MAC_init). Signatures verified against the previous secret before re-signing so claim payloads are unchanged.

systest/21million/test-21million.sh
mlwelles: Good catch — done in 352ff5afe. Switched all three sites (test-21million.sh, test-reindex.sh, loader-benchmark.sh) to ${GOBIN_DGRAPH_PATH:-/gobin/dgraph}. Both forms resolve to the same string (GoBinDgraphPath's default is /gobin/$Bin), but using the registry's named Var matches the rest of the codebase's conventions and keeps the Var ↔ shell mapping discoverable.

matthewmcneely (approved):

PR #9702, opened 2026-05-13
fix(graphql): allow update mutations to set an @id field to its existing value

shiva-istari:
**Description**

When a GraphQL update mutation included an id directive field in its set patch with the same value already stored on the node, the resolver incorrectly returned null instead of applying the update.
The rewriter ran an existence query for the id value before execution. When it found a matching UID it assumed a conflict and aborted rewriting it had no way to know at that point whether the existing UID was the same node being updated (no-op, should succeed) or a different node (genuine conflict, should error).

The same early-abort also caused a false-positive error in a second case: when a filter matched no nodes at all, the existence of a matching id directive value elsewhere in the database was being flagged as a conflict even though no mutation ran.
A third related bug: the remove patch was issuing unnecessary existence queries for top-level id fields. Removing a scalar value from a node can never steal uniqueness from another node, so those queries were wasted work and could generate spurious conflict errors.

 Fix
Deferred conflict detection  a two-phase approach:

1. Rewrite time: when an id directive value is found to exist during UpdateWithSet, instead of aborting, record a xidConflict{existenceUID, xidField, xidValue, typeName} entry and let the mutation proceed.
2. Post-execution (VerifyXIDConflicts): after the upsert runs, compare each recorded existenceUID against the UIDs that were actually mutated. If the existence UID is among the mutated UIDs → same node → not a
  conflict, succeed. If it is not → different node → return the conflict error. If mutatedUIDs is empty (filter matched nothing) → no verification needed, return nil.

The new XIDConflictVerifier interface on MutationRewriter keeps the post-execution check opt-in only UpdateRewriter implements it.

Additionally, the remove patch path strips top-level id fields before generating existence queries, and rewriteObject skips id processing entirely for UpdateWithRemove at top level.


matthewmcneely:
@shiva-istari Can you add a proper PR description covering the changes in detail? It's not clear looking at the code what the exact problem this is fixing.

graphql/e2e/auth/add_mutation_test.go
```diff
@@ -1225,29 +1225,39 @@ func TestUpdateMutationWithIDFields(t *testing.T) {
...
-                       "set": {
-                           "company": "ABC tech"
-                       }
-                   }
-               }`,
-		error: "couldn't rewrite mutation updateEmployer because failed to rewrite mutation" +
-			" payload because GraphQL debug: id ABC tech already exists for field company" +
-			" inside type Employer",
 	},
+		// TODO(reviewer): This test case is commented out because the scenario it tests is
```
matthewmcneely: I think it's correct to remove. Was definitely a false positive (the test case sets the company to its own current value, right?).

graphql/e2e/auth/debug_off/debugoff_test.go
```diff
@@ -259,26 +259,33 @@ func TestUpdateMutationWithIDFields(t *testing.T) {
...
-                               "in": "ABC"
-                           }
-                       },
-                       "set": {
-                           "company": "ABC tech"
-                       }
-                   }
-               }`,
 	},
+		// TODO(reviewer): Commented out for the same reason as the corresponding case in
```
matthewmcneely: Same response as in `add_mutation_test.go`, safe to remove

graphql/e2e/common/common.go
```diff
@@ -939,6 +939,20 @@ func RunAll(t *testing.T) {
...
+	t.Run("update mutation errors on @id conflict with different node", updateMutationXIDConflictStillErrors)
+	t.Run("update mutation Int @id unchanged succeeds (B1)", updateMutationIntXIDUnchanged)
+	t.Run("update mutation Int @id conflict errors (B2)", updateMutationIntXIDConflictErrors)
+	t.Run("update mutation Int64 @id unchanged succeeds (B3)", updateMutationInt64XIDUnchanged)
+	t.Run("update mutation nullable @id unchanged succeeds (C1)", updateMutationNullableXIDUnchanged)
+	t.Run("update mutation nullable @id set for first time succeeds (C2)", updateMutationNullableXIDFirstTimeSet)
+	t.Run("update mutation nullable @id cleared via remove patch succeeds (C3)", updateMutationNullableXIDClearedViaRemove)
+	t.Run("update mutation remove patch on non-@id field succeeds (C3-contrast)", updateMutationRemovePatchNonXIDField)
+	t.Run("update mutation delete and recreate same @id succeeds (F1)", updateMutationDeleteRecreateXID)
+	t.Run("update mutation concurrent unchanged @id all succeed (H1)", updateMutationConcurrentUnchangedXID)
```
matthewmcneely, line 954: Unlikely, but this only asserts that no goroutine sees "already exists for field". If all 10 goroutines abort due to MVCC contention (again unlikely), the test passes — but a real regression (unchanged-@id silently failing) would also pass. Suggest asserting that at least one update returned NumUids == 1.

matthewmcneely (changes requested):

PR #9732, opened 2026-06-08
feat(hnsw): batch neighbor vector reads via badger Txn.MultiGet

shaunpatterson:
> **Draft / dependent PR.** Requires badger `Txn.MultiGet` — dgraph-io/badger#2297. CI will not compile until a badger release with `MultiGet` is available and `go.mod` is bumped. To review/test locally:
> ```
> go mod edit -replace github.com/dgraph-io/badger/v4=/path/to/badger   # checked out at sp/badger_multiget
> ```

## What

HNSW vector search reads each candidate's neighbor vectors **one key at a time** (`getVecFromUid` → `CacheType.Get`), and every `Get` becomes a full single-key `NewKeyIterator(AllVersions)` in badger. For a fixed candidate those sibling reads are independent — this batches them into a single read.

## Changes

- **`tok/index`**: add `MultiGet(keys) (vals, errs)` to `CacheType` / `Txn` / `LocalCache` (batched counterpart of `Get`).
- **`posting`**:
  - `ReadPostingListFromVersions` — folds a key's version chain from a badger `[]ItemVersion` exactly as `ReadPostingList` does from an iterator (delta-on-complete, newest-first, stop on complete/empty/deleted).
  - `MemoryLayer.ReadManyData` — resolves many keys in one `badger.Txn.MultiGet`; warm keys served from the global cache; two-phase read mirroring `ReadData`.
  - `LocalCache.MultiGet` — adds the per-txn cache/delta layer; `viLocalCache`/`viTxn.MultiGet` expose resolved values.
- **`tok/hnsw`**: `getVecsFromUids` batch-fetches a frontier's vectors; `searchPersistentLayer` collects a candidate's unvisited neighbors and reads their vectors in one `MultiGet`. **Traversal/heap logic is unchanged** — only the vector fetch is batched.
- `TxnCache`/`QueryCache` and the test mocks implement `MultiGet`.

## Testing (with the local badger replace)

- **Differential test** (`posting`): `ReadManyData` and `viLocalCache.MultiGet` return identical results to the per-key `ReadPostingList`/`Get` path over an on-disk round-trip of complete/delta/empty/absent posting shapes.
- `tok/hnsw` and `posting` suites pass; full dgraph build clean.

## Benchmarks (cold cache, real badger; benchstat n=6)

- badger `MultiGet` vs per-key `NewKeyIterator`: **−5…−16% time, −35…−40% allocs** (grows with frontier size).
- posting frontier read (`Get` loop vs `MultiGet`, `BenchmarkHNSWFrontierRead`): **−12…−18% time, −15…−21% allocs** for K=16/64/256.

This speeds up the neighbor-vector read path; whole-query latency improves by that fraction of time spent reading vectors.

### DIFF
--- a/posting/lists.go
+++ b/posting/lists.go
@@ -98,6 +98,31 @@ func (vc *viLocalCache) Get(key []byte) ([]byte, error) {
 	return vc.GetValueFromPostingList(pl)
 }
 
+// MultiGet resolves many keys to their values in one batched read (see
+// LocalCache.MultiGet). vals and errs are aligned with keys; errs[i] is
+// ErrNoValue when keys[i] has no value, matching Get's per-key semantics.
+func (vc *viLocalCache) MultiGet(keys [][]byte) ([][]byte, []error) {
+	vals := make([][]byte, len(keys))
+	errs := make([]error, len(keys))
+	lists, err := vc.delegate.MultiGet(keys)
+	if err != nil {
+		for i := range errs {
+			errs[i] = err
+		}
+		return vals, errs
+	}
+	for i, pl := range lists {
+		if pl == nil {
+			errs[i] = ErrNoValue
+			continue
+		}
+		pl.Lock()
+		vals[i], errs[i] = vc.GetValueFromPostingList(pl)
+		pl.Unlock()
+	}
+	return vals, errs
+}
+
 func (vc *viLocalCache) GetWithLockHeld(key []byte) ([]byte, error) {
 	pl, err := vc.delegate.Get(key)
 	if err != nil {
@@ -388,6 +413,61 @@ func (lc *LocalCache) GetSinglePosting(key []byte) (*pb.PostingList, error) {
 // Get retrieves the cached version of the list associated with the given key.
 func (lc *LocalCache) Get(key []byte) (*List, error) {
 	return lc.getInternal(key, true, false)
+}
+
+// MultiGet is the batched form of Get(readFromDisk=true): it resolves many keys
+// to their *List in one shared read. Keys already in the per-txn cache (plists)
+// are served from there; the cold remainder is read from disk in a single
+// MemoryLayer.ReadManyData (one badger MultiGet), then has any pending txn delta
+// applied and is interned via SetIfAbsent — mirroring getInternal per key.
+// Returned lists are aligned with keys.
+func (lc *LocalCache) MultiGet(keys [][]byte) ([]*List, error) {
+	lists := make([]*List, len(keys))
+
+	lc.RLock()
+	plistsNil := lc.plists == nil
+	lc.RUnlock()
+
+	var missIdx []int
+	var missKeys [][]byte
+	for i, key := range keys {
+		if !plistsNil {
+			lc.RLock()
+			l, ok := lc.plists[string(key)]
+			lc.RUnlock()
+			if ok {
+				lists[i] = l
+				continue
+			}
+		}
+		missIdx = append(missIdx, i)
+		missKeys = append(missKeys, key)
+	}
+	if len(missKeys) == 0 {
+		return lists, nil
+	}
+
+	fetched, err := MemLayerInstance.ReadManyData(missKeys, pstore, lc.startTs)
+	if err != nil {
+		return nil, err
+	}
+	for j, idx := range missIdx {
+		pl := fetched[j]
+		if plistsNil {
+			// Cache disabled: return the freshly read list directly, exactly as
+			// getInternal's plists==nil path returns getNew without caching.
+			lists[idx] = pl
+			continue
+		}
+		skey := string(missKeys[j])
+		lc.RLock()
+		if delta, ok := lc.deltas[skey]; ok && len(delta) > 0 {
+			pl.setMutation(lc.startTs, delta)
+		}
+		lc.RUnlock()
+		lists[idx] = lc.SetIfAbsent(skey, pl)
+	}
+	return lists, nil
 }
 
 func (lc *LocalCache) GetUids(key []byte) (*List, error) {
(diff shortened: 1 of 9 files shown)


shaunpatterson:
Closing this. A second performance pass shows the change makes HNSW queries **slower**, and the premise it rests on doesn't hold up.

## End-to-end HNSW query latency

Index built once (5000 x 384-d vectors, `exponent:3` -> `efConstruction` 150, `efSearch` 90) into a single on-disk store; both binaries query a private copy of *that same store* through the `worker/task.go` stack (`NoCache` + global memory layer), 40 fixed queries, k=10, interleaved process runs, 6 samples of 200 queries each:

| config | baseline | this PR | time | allocs |
|---|---|---|---|---|
| posting cache off, block cache 256M | **4.46 ms** | **8.05 ms** | **+80%** | -24% |
| posting cache 1G, block cache 256M | **6.12 ms** | **9.06 ms** | **+48%** | -23% |

Sample ranges don't overlap (baseline 4.39-4.49 vs 7.93-8.33).

It *is* semantically correct: on an identical graph all 40 queries return byte-identical top-10 uid lists, recall@10 is 0.6875 in both arms, and read counts match exactly (baseline 1462.6 `Get`/query; this PR 4.0 `Get` + 19.7 `MultiGet` covering 1458.6 keys). So this is purely a perf question.

## Why: iterator construction was never the cost

`readFromDisk` uses `txn.NewKeyIterator(key, ...)`, which sets `prefixIsKey`. That lets `pickTables` reject tables by bloom filter (`t.DoesNotHave(y.Hash(key))`) and bound the table range. `ReadManyData`'s shared `txn.NewIterator(...)` sets no `Prefix`, so the merge iterator is built over every table at every level and every `Seek` descends into all of them.

Isolated in pure badger (200k keys x 1536 B values, snappy, 256 MB block cache, medians of 6):

| frontier | `NewKeyIterator` (today) | shared txn, per-key iter | shared prefix-less iter (this PR) | + predicate prefix | badger `Txn.MultiGet` |
|---|---|---|---|---|---|
| scattered K=8 | 35.5 us | **+4%** | +67% | +76% | +71% |
| scattered K=16 | 77.4 us | **-11%** | +101% | +63% | +53% |
| scattered K=64 | 291 us | **+1%** | +56% | +53% | +53% |
| scattered K=150 | 686 us | **-0%** | +74% | +54% | +56% |

Sharing the transaction while keeping per-key iterators is **free**. There was no construction cost to amortize; the pruning that gets discarded is what actually costs. Scoping the shared iterator to the predicate prefix recovers range pruning but not per-key bloom rejection, so it's still +53...+76% - no cheap fix.

Batch sizes weren't the limiter either: mean 74, max 150 (out-degree cap is `efConstruction`, `helper.go:693`). 1462 keys in ~20 batches is a 74x amortization, and it still loses.

## Why the benchmark in the description showed a win

`BenchmarkHNSWFrontierRead` writes only K keys (16/64/256, <1 MB) under one fresh predicate into an otherwise empty store - one memtable, contiguous keys, nothing to prune. I reproduced its -8...-15%. Sweeping store size (`ReadManyData` vs `ReadData` loop, K=64 scattered, 256 MB block cache):

| store | bytes | time | allocs |
|---|---|---|---|
| n=500, dim=384 | ~0.8 MB | **-13%** | -27% |
| n=5 000, dim=384 | ~7.7 MB | **-13%** | -27% |
| n=50 000, dim=384 | ~77 MB | **-1%** | -33% |
| n=200 000, dim=64 | ~51 MB | **-14%** | -38% |
| n=200 000, dim=384 | ~307 MB | **+52%** | -18% |
| n=200 000, dim=1536 | ~1.2 GB | **+72%** | +8% |

The predictor is total predicate bytes vs the badger block cache, not key count - note `n=200 000/dim=64` (51 MB) wins while `n=50 000/dim=384` (77 MB) is already flat. Resident working set: the extra table descents hit memory and amortization nets -13%. Beyond the cache: every extra descent is a real block decode and the sign flips. Production embeddings are 384-3072 dims over millions of vectors, so real deployments sit permanently on the losing side and degrade as the corpus grows.

## Also worth noting

The description is stale: commit 6a9679969 replaced the badger `Txn.MultiGet` call with a plain shared iterator, so this branch has no badger dependency at all - `go.mod` is stock v4.9.4 and `go build ./posting/... ./tok/...` succeeds with no `replace`. `ReadPostingListFromVersions` doesn't exist on the branch either.

## Where the headroom actually is

The floor is ~3.05 us per warm read x ~1462 reads/query at ef=90. Batching reduces neither factor. Better levers, in order:

1. **Parallelize the frontier across goroutines** rather than funnelling it into one iterator. Independence of the reads is the correct observation in this PR; serializing them is what destroys it. On the read-only query path this needs no locking work - `NoCache` leaves `plists`/`deltas` nil so nothing is interned, and `readFromDisk` already creates its own `pstore.NewTransactionAt` per key. (This does **not** hold for `Insert`, where `viTxn` mutates interned `*List` objects under a live `LocalCache`.) Per-key bloom pruning is preserved.
2. **Cut the read count.** Out-degree is `efConstruction` - 150 by default, 250 at the systest's `exponent:"5"` - where conventional HNSW uses M=16-32. This is the single biggest win available.
3. **Look at the posting-list cache on this path.** In this harness the baseline was 37% *slower* with a 1 GiB posting cache than with none (6.12 vs 4.46 ms): ~4.18 us per ristretto hit (`readFromCache` -> `copyList` + `mutationMap.clone()`) vs ~3.05 us for a full badger read against a warm block cache. Production hands it 40% of 4 GiB. Side observation from this harness, not a conclusion - but it deserves a dedicated look.

Unrelated but worth flagging: `getInsertLayer` uses the auto-seeded global `math/rand` (`tok/hnsw/helper.go:352`), so two processes build different graphs from identical input. Index builds aren't reproducible, which matters for anyone benchmarking this area - it's why paired runs must share a prebuilt corpus rather than each building their own.

PR #9734, opened 2026-06-08
fix(query): handle pagination and offset for ordered cascade queries

tomhollingworth:
## Description

 Implement missing first/offset pagination propagation from cascade directives to the parent query when ordering is applied and resolve an existing TODO comment. Effectively allowing `@cascade` and ordered queries to resolve for data sets with more than 1000 entries. The change also replaces hardcoded result limit of 1000 with a named constant. 

Test cases were also added to prove out solution.

### Background

GraphQL queries with `@cascade` and a `order: { .. }` only return the first [1000](https://github.com/dgraph-io/dgraph/blob/main/query/query.go#L2535) results. This is due to the query Count being 0 - even if `first: ....` is set to something else, because[ the first / offset are moved into the query CascadeArgs and the deleted](https://github.com/dgraph-io/dgraph/blob/main/query/query.go#L664) - effectively appearing as 0 and [getting overridden to 1000](https://github.com/dgraph-io/dgraph/blob/main/query/query.go#L2535).

**Checklist**

- [] Code compiles correctly and linting (via trunk) passes locally

### DIFF
--- a/query/query.go
+++ b/query/query.go
@@ -32,6 +32,9 @@ import (
 	"github.com/dgraph-io/dgraph/v25/x"
 )
 
+// defaultRetrieveCount is the default number of results to retrieve for a query when pagination arguments are not specified.
+const defaultRetrieveCount = 1000
+
 /*
  * QUERY:
  * Let's take this query from GraphQL as example:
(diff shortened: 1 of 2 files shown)


tomhollingworth:
I'm keeping this in draft until I've confirmed the additional test cases run successfully.

tomhollingworth:
This is still on my todo list!

PR #9746, opened 2026-06-15
fix: S390x compatability

navaneeswar1011:
**Description**

This PR fixes a build failure encountered while compiling Dgraph on **IBM s390x architecture** caused by a datatype mismatch in the disk metrics module.

## Problem

During compilation, the build fails in `x/disk_metrics_linux.go` due to architecture-specific type differences in the Linux `syscall.Statfs_t` structure.

The existing implementation performs arithmetic operations using mixed types:

```go id="7d3k1q"
disk.All = s.Frsize * int64(s.Blocks-reservedBlocks)
disk.Free = s.Frsize * int64(s.Bavail)
```

On **s390x architecture**, the datatype definitions for `Frsize`, `Blocks`, and `Bavail` differ from architectures such as `x86_64`, causing a compile-time type mismatch.

## Fix

Added explicit type conversion for `Frsize` to ensure consistent datatype handling during arithmetic operations.

Updated code:

```go id="1f8m9r"
disk.All = int64(s.Frsize) * int64(s.Blocks-reservedBlocks)
disk.Free = int64(s.Frsize) * int64(s.Bavail)
```

## Validation Performed

The fix was validated by building and testing on multiple architectures:

* **s390x (RHEL VM)** → Build completed successfully
* **x86_64 (Ubuntu VM)** → Build completed successfully
* **ARM64 (macOS)** → Build completed successfully

Additionally verified:

* Successful startup of both **Zero** and **Alpha** services
* CRUD operations executed successfully after build validation

## Impact

* Resolves architecture-specific build failure on **IBM s390x**
* Improves cross-platform compatibility
* No functional logic changes introduced
* No impact observed on existing supported architectures

**Checklist**


x/disk_metrics_linux_test.go
```diff
@@ -0,0 +1,112 @@
+//go:build linux
```
matthewmcneely: I'd prefer to drop these tests. The one-line fix is self-evidently correct and the tests add maintenance surface while testing nothing. If we want a guard, it belongs in CI as an s390x cross-compile step, not here -- but as a rule we don't do that (our testing and CI is already waaaaay too long).

matthewmcneely (changes requested):

x/disk_metrics_linux_test.go
navaneeswar1011: Deleted the file

.github/workflows/cd-dgraph.yml
matthewmcneely, line 307: These changes are out of scope. Please revert. I will address in another PR

matthewmcneely (changes requested):

matthewmcneely (approved):

PR #761, opened 2017-03-28
Language list handling in orderasc and orderdesc

tzdybal:
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/761)

PR #307, opened 2016-11-01
Bench/gensubgraph

jchiu0:
1) We move benchmarks from query_test into benchmark_test as query_test.go is getting long.
2) We refactor benchmarking code a bit to make it more concise.
3) We make UIDList, task.ValueList, task.CountList serializable.
4) We export SrcUIDs and DestUIDs in SubGraph so that gob will encode/decode them.
5) We add a flag in cmd/dgraph to output subgraphs into a directory, if the flag is specified.
6) We regenerate these binaries.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/307)

Issue #290, opened 2016-10-20
Dgraph server crashes with Unable to parse posting error.


pawanrawal:
While trying to update tags on Gru, there is an instance when Dgraph server crashes. This could possibly be a bug or something wrong that we might be doing. This happens when we delete an existing tag related to the question and try to add a new one in the same mutation.

**To reproduce try.**

```
// Assume two tags exist with some uids. We create a new question and associate
// the tags with it.
curl localhost:8080/query -XPOST -d $'mutation {
        set {
                <root> <question> <_new_:qn> .
                <_new_:qn> <name> "qn1" .
                <_new_:qn> <question.tag> <_uid_:0x2b693088816b04b7> .
                <_uid_:0x2b693088816b04b7> <tag.question> <_new_:qn> .
                <_new_:qn> <question.tag> <_uid_:0x29bf442b48a772e0> .
                <_uid_:0x29bf442b48a772e0> <tag.question> <_new_:qn> .
        }
}'
```

Say this replies with `{"code":"ErrorOk","message":"Done","uids":{"qn":"0x5b6a2f404da65153"}}`
Then

```
// Export the qn returned as part of response so that it can be used in the cURL request.
export qn=0x5b6a2f404da65153

// Deleting a tag and attaching a new one.
curl localhost:8080/query -XPOST -d $"mutation {
        set {
                <_uid_:$qn> <question.tag> <_uid_:0x29bf442b48a772e0> .
                <_uid_:0x29bf442b48a772e0> <tag.question> <_uid_:$qn> .
                <_uid_:$qn> <question.tag> <_new_:tag1> .
                <_new_:tag1> <tag.question> <_uid_:$qn> ..
        }
        delete {
                <_uid_:$qn> <question.tag> <_uid_:0x2b693088816b04b7> .
                <_uid_:0x2b693088816b04b7> <tag.question> <_uid_:$qn> .
        }
}"
```
Returns `{"code":"ErrorOk","message":"Done","uids":{"tag1":"0x5bfe99d56d35be27"}}` successfully but then server crashes with 

`2016/10/20 11:06:52 list.go:705: Idx: 1. Unable to parse posting.`

This is related to the new code which calculates and stores the checksum for a posting.

manishrjain:
I don't see the crash. What version of Dgraph are you using?

pawanrawal:
This was fixed and merged to master.

PR #9747, opened 2026-06-17
feat(txn): surface transaction abort reasons to clients

rahst12:
## Problem

When Dgraph aborts a transaction, Zero knows *why* — but the client never finds out.
Zero distinguishes at least three causes internally:

- a write-write **conflict**,
- a **predicate move** (the tablet relocated mid-transaction, or commits are blocked
  while a move is in flight), and
- a **stale start-ts** (the transaction predates the current Zero leader — i.e. a leader
  change, not a real conflict).

All three collapse into `src.Aborted = true` inside `Server.commit()`
(`dgraph/cmd/zero/oracle.go`), and by the time the error reaches the caller it is one
opaque string:

```
Transaction has been aborted. Please retry
```

Worse, the `checkPreds` path builds specific, human-readable messages for predicate
moves and then throws them away. With no discriminator, a client cannot make the one
decision the category enables: **retry immediately** (conflict) vs. **back off**
(predicate is moving).

### Why we can't just name the conflicting predicate

For conflicts, you might expect the abort to name the contended predicate/UID. It can't —
not without extra bookkeeping. For every mutated edge, Alpha computes a conflict key in
`GetConflictKey()` (`posting/list.go`) as a one-way fingerprint:

```go
farm.Fingerprint64(key) ^ uid
```

Only this `uint64` is retained (`conflicts map[uint64]struct{}` in `posting/oracle.go`);
the predicate name and UID are dropped immediately. It is then base-36 encoded into
`TxnContext.Keys` and shipped to Zero, which detects a conflict purely by matching
fingerprints — it never sees a predicate name. Because the value is a fingerprint XORed
with a UID, it **cannot be inverted** to recover either input. Naming the predicate would
require *remembering* what was hashed (a per-transaction reverse map on Alpha), which is
larger scope. So this PR delivers the **category**, which needs no reverse mapping, and
leaves predicate/UID fidelity to follow-up work.

## Fix

Thread a categorized reason out of Zero to the client. No change to conflict *detection*,
and no proto change — the reason rides on the existing `codes.Aborted` gRPC status as
`"<reason>: <detail>"`.

- **`dgraph/cmd/zero/oracle.go`** — `commit()` tags each abort with a category
  (`conflict`, `stale-startts`, `predicate-move`); predicate-move now reuses the existing
  `checkPreds` messages instead of swallowing them.
- **`worker/mutation.go`** — `CommitOverNetwork` forwards the reason and records the
  abort metric, instead of flattening it to the reasonless `dgo.ErrAborted`.
- **`edgraph/server.go`** — both commit paths (`CommitNow` and explicit `CommitOrAbort`)
  pass the reason through and set `Aborted`.

Backward compatible: the existing message is preserved and the status code stays
`codes.Aborted`, so existing client retry logic is unaffected. This phase is
**category-only**; naming the contended predicate/UID is planned follow-up.

## Future work

This PR is deliberately scoped to the **category** only — the cheapest, lowest-risk slice
that is immediately useful to clients — because it needs no reverse mapping, no new proto
fields, and no change to conflict detection. All correctness risk stays in *what we
report*, not *what we decide*. Richer fidelity is layered, additive work that can land
independently:

- **Name the contended predicate(s).** Retain a small per-transaction reverse map on
  Alpha (`fingerprint → {predicate, uid, kind}`), populated where the conflict key is
  already computed. The aborting transaction's own Alpha still holds this metadata, so
  Zero need only echo back *which* fingerprint matched — no hash inversion. The cost lands
  only on the abort path. Target the single-request `CommitNow` path first.
- **Structured, machine-readable detail.** Carry category + predicate + uid/token via the
  gRPC rich-error model (`ErrorInfo` details) so clients get typed fields instead of
  parsing a message string. Still no proto change.
- **First-class API field.** Eventually promote the reason/conflict detail to a real
  `TxnContext` field in the `dgo` proto, giving uniform structured access across all
  language clients — at the cost of a coordinated cross-repo release.
- **Refine the `predicate-move` category.** Today all `checkPreds` failures report as
  `predicate-move`, but some are not moves (e.g. a malformed/internal predicate key, or a
  predicate not currently served by any group). A later phase can split these into honest
  categories (`predicate-unavailable`, `internal`) and revisit retryability — an
  `internal`/malformed abort is not retryable, unlike a move.

## Checklist

- [ ] Code compiles correctly and linting (via trunk) passes locally
- [ ] Refer to dgraph4j, pydgraph, and dgraph-docs as corresponding updates are made.


shiva-istari (commented):
Thanks for putting this together @rahst12 surfacing why a transaction aborted is a really useful addition, and I appreciate the care here: keeping codes.Aborted for backward compatibility, and adding tests across all the repos.

I went through it in depth alongside the dgraph4j / pydgraph / docs PRs, and I have a few design-level thoughts I'd like to raise.

A few things I noticed:

No single source of truth for the categories. The reason strings (conflict / stale-startts / predicate-move) are defined in 4 places — server, dgraph4j, pydgraph, docs so they can drift independently over time.
The reason travels inside the message string and is recovered via split(":"). It works, but it's a bit brittle to depend on the message format as a contract.
A few categories may not always match what actually happened e.g. a ctx cancel/timeout currently reports as conflict:, the late (post-lease) abort is always conflict: even when it's really a stale start-ts, and predicate-move: also catches some non-move errors (malformed key / tablet-not-served).
Observability (future scope). The reason is currently invisible in metrics TxnAborts is a bare counter, so we can only see the total, not the breakdown by category. Since the server already computes the reason, it'd be a nice low-effort win down the line to add it as a metric label (e.g. txn_aborts_total{reason="conflict"}) so the split shows up in dashboards. Not needed for this PR more of a follow-up thought.

rahst12:
Thanks for the feedback.  My goal is to keep this first pass super minimalistic, small in scope, and non-impacting to exisiting users.  With these goals, there's some things that don't get addressed, however, I feel like the surfacing of this information is still better than dropping it altogether, recognizing that more work is needed for a comprehensive solution.

> No single source of truth for the categories. The reason strings (conflict / stale-startts / predicate-move) are defined in 4 places — server, dgraph4j, pydgraph, docs so they can drift independently over time.

Yes this is true and by design.  I want to keep the changes very minimal and ensure it's fully backwards compatible.  

> The reason travels inside the message string and is recovered via split(":"). It works, but it's a bit brittle to depend on the message format as a contract.

Agreed, I have some future phase thoughts that would make this better, but would require changes to the proto fields.

> Observability (future scope). The reason is currently invisible in metrics TxnAborts is a bare counter, so we can only see the total, not the breakdown by category. Since the server already computes the reason, it'd be a nice low-effort win down the line to add it as a metric label (e.g. txn_aborts_total{reason="conflict"}) so the split shows up in dashboards. Not needed for this PR more of a follow-up thought.

Acknowledged and agreed it would be a future change.

For this piece of feedback:

A few categories may not always match what actually happened e.g. a ctx cancel/timeout currently reports as conflict:, the late (post-lease) abort is always conflict: even when it's really a stale start-ts, and predicate-move: also catches some non-move errors (malformed key / tablet-not-served).

I don't think we should assert a category that can't substantiated.  There is an UNKNOWN category for things like this.  I'm aiming to capture anything unsubstantiated into the UNKNOWN category, then in future PRs introduce greater fidelity.  I'm working on disambiguating some things.  My thoughts of the categories:

| # | Cause | Category | Detail string at | Selected at | Wire message | dgraph4j |
|--:|:---|:---|:---|:---|:---|:---|
| 1 | Write-write conflict (early) | conflict | oracle.go:368 | :484 → :473 | conflict: Transaction has been aborted. Please retry | CONFLICT |
| 2 | Stale start timestamp (early) | stale-startts | oracle.go:375 | :484 → :473 | stale-startts: Transaction start timestamp is older than the oldest timestamp Zero can still validate (Zero leader change, or its conflict map was trimmed at a snapshot). Please retry | STALE_STARTTS |
| 3 | Move in flight (isBlocked) | predicate-move | oracle.go:438 | :487 → :473 | predicate-move: Commits on predicate `<p>` are blocked due to predicate move | PREDICATE_MOVE |
| 4 | Move completed (group mismatch) | predicate-move | oracle.go:448 | :487 → :473 | predicate-move: Mutation done in group: 1. Predicate `<p>` assigned to 2 | PREDICATE_MOVE |
| 5 | Predicate served by no group | withheld | oracle.go:442 | :487 → :473 | Tablet for `<p>` is nil | UNKNOWN |
| 6 | Malformed key — no separator | withheld | oracle.go:420 | :487 → :473 | Unable to find group id in `<pkey>` | UNKNOWN |
| 7 | Malformed key — non-numeric group id | withheld | oracle.go:425 | :487 → :473 | unable to parse group id from `<pkey>`: `<strconv err>` | UNKNOWN |
| 8 | Late conflict (keyCommit) | conflict | oracle.go:368 | :525 → :550 | conflict: Transaction has been aborted. Please retry | CONFLICT |
| 9 | Late stale start timestamp | stale-startts | oracle.go:375 | :527 → :550 | same as row 2 | STALE_STARTTS |
| 10 | Context cancelled / timed out | withheld | Go stdlib, via oracle.go:541 | :550 | context canceled / context deadline exceeded | UNKNOWN |

rahst12:
Added some clarity to the aborts and looked for any lingering ones and performed some organization with the error messages.  This PR is backwards compatible with all clients, but now also providing transaction abort details when possible.  Attached is the thoughts on the future PRs/phases (attached plan doc)
[transaction-abort-reason-plans.md](https://github.com/user-attachments/files/30574599/transaction-abort-reason-plans.md)

| Plan | Status |
|:---|:---|
| A — category, message-based | ✅ 100% + extras |
| B — conflicting predicate (R1 + hasConflict returns keys) | ❌ 0% |
| C — structured via error-details (T3) | ❌ 0% |
| D — proto field (T4) | ❌ 0% |

New table:
| # | Cause | Category | Detail declared | Selected → emitted | Wire message | dgraph4j |
|--:|:---|:---|:---|:---|:---|:---|
| 1 | Write-write conflict (early) | conflict | `:405` | `:550 → :539` | conflict: Transaction has been aborted. Please retry. Another transaction committed to one of the same keys. The conflicting key cannot be identified: it may be the data key written directly, or an index or count key derived from it. On an @upsert predicate the uid is excluded, so any two transactions writing the same value conflict | CONFLICT |
| 2 | Stale start timestamp (early) | stale-startts | `:415` | `:550 → :539` | stale-startts: Transaction start timestamp is older than the oldest timestamp Zero can still validate (Zero leader change, or its conflict map was trimmed at a snapshot). Please retry | STALE_STARTTS |
| 3 | Move in flight (isBlocked) | predicate-move | `:444` (used `:504`) | `:553 → :539` | predicate-move: Commits on predicate `<p>` are blocked due to predicate move | PREDICATE_MOVE |
| 4 | Move completed (group mismatch) | predicate-move | `:445` (used `:513`) | `:553 → :539` | predicate-move: Mutation done in group: 1. Predicate `<p>` assigned to 2 | PREDICATE_MOVE |
| 5 | Predicate served by no group | withheld | `:443` (used `:508`) | `:553 → :539` | Tablet for `<p>` is nil | UNKNOWN |
| 6 | Malformed key — no separator | withheld | `:441` (used `:488`) | `:553 → :539` | Unable to find group id in `<pkey>` | UNKNOWN |
| 7 | Malformed key — non-numeric group id | withheld | `:442` (used `:492`) | `:553 → :539` | unable to parse group id from `<pkey>`: `<strconv err>` | UNKNOWN |
| 8 | Late conflict (keyCommit) | conflict | `:405` | `:591 → :616` | same as row 1 | CONFLICT |
| 9 | Late stale start timestamp | stale-startts | `:415` | `:593 → :616` | same as row 2 | STALE_STARTTS |
| 10 | Context cancelled / timed out | withheld | Go stdlib, via `:607` | `:616` | context canceled / context deadline exceeded | UNKNOWN |
| 11 | Aborted out of band before commit was decided | withheld | `:434` | `:624 → :627` | Transaction has been aborted. Please retry. It was already aborted before this commit was decided, which happens when a schema update or a drop-predicate cancels pending transactions on a predicate it touched, or when the server ages out transactions idle for longer than --limit "txn-abort-after" | UNKNOWN |

dgraph/cmd/zero/oracle.go
```diff
@@ -338,56 +340,220 @@
...
+		if s.isBlocked(pred) {
+			return abortReasonPredicateMove, errors.Errorf(abortDetailPredicateBlockedFmt, pred)
+		}
+		tablet := s.ServingTablet(pred)
+		if tablet == nil {
+			return abortReasonUncategorized, errors.Errorf(abortDetailTabletNilFmt, pred)
+		}
+		// The predicate finished moving to another group while this transaction was open: the
+		// mutation was written against group gid, but the predicate now belongs elsewhere.
+		if tablet.GroupId != uint32(gid) {
```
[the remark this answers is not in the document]
rahst12: Fixed in c7f2f3549 by switching to `strconv.ParseUint(splits[0], 10, 32)`, which bounds the value to 32 bits before the `uint32(gid)` comparison and rejects negatives at parse time (they now fall into the existing "unable to parse group id" abort path).

Note: this pattern is pre-existing — `main` has the identical `strconv.Atoi` + `uint32(gid)` code today, inside the `checkPreds` closure in `commit()`. The alert fired on this PR only because these lines moved. In practice no legitimate value could overflow: the `preds` entries are built by the Alpha from a `uint32` group id.

rahst12:
Dropping a polite ping for a review or an estimate for when it can be reviewed.  Thanks

shiva-istari (commented):
Latest changes look good to me. One ask before this merges: since centralising the reason strings across the server, client libraries (dgraph4j, pydgraph), and docs is deferred to phases C and D, please file tracking issues for those phases and link them in the PR description so the follow-up work outlives this thread. Otherwise, LGTM!

dgraph/cmd/zero/oracle.go
```diff
@@ -338,56 +340,223 @@ func (s *Server) proposeTxn(ctx context.Context, src *api.TxnContext) error {
...
+		"cancels pending transactions on a predicate it touched, or when the server ages out " +
+		"transactions idle for longer than --limit \"txn-abort-after\""
+
+	// Details produced by checkPreds. Wording is unchanged from before the abort-reason work; only
+	// the declaration moved here, so existing log-scrapers and docs still match.
+	abortDetailMissingGroupIDFmt   = "Unable to find group id in %s"
+	abortDetailBadGroupIDFmt       = "unable to parse group id from %s"
+	abortDetailTabletNilFmt        = "Tablet for %s is nil"
+	abortDetailPredicateBlockedFmt = "Commits on predicate %s are blocked due to predicate move"
+	abortDetailGroupMismatchFmt    = "Mutation done in group: %d. Predicate %s assigned to %d"
```
matthewmcneely, line 445: These three details become a **client-visible wire contract** as of this PR, and they still carry the internal predicate encoding.

`posting/lists.go:437` builds preds as `fmt.Sprintf("%d-%s", gid, pk.Attr)`, and `pk.Attr` is already `x.NamespaceAttr` — `hex(ns) + "-" + attr`, with `NsSeparator = "-"`. So after `SplitN(pkey, "-", 2)` below, `pred` is `0-name`, and a client now receives:

```
predicate-move: Commits on predicate 0-name are blocked due to predicate move
Tablet for 0-name is nil
Mutation done in group: 3. Predicate 0-name assigned to 1
```

The byte-for-byte compat argument that (correctly) freezes `abortDetailConflict` doesn't apply to these. The old `checkPreds` closure built them and threw them away, so nothing has ever scraped them. This is the one chance to print the predicate the way the user wrote it, before #761 and the client parsers freeze `0-name` as the published form.

Don't reach for `x.ParseAttr` to do it — it's `SplitN(...)[1]` with no bounds guard, and `Preds` is client-supplied, so a hand-crafted `Preds: ["1-nodash"]` would panic Zero. A guarded `strings.Cut(pred, x.NsSeparator)` falling back to the raw string is the safe version.

worker/mutation.go
```diff
@@ -854,6 +856,16 @@ func CommitOverNetwork(ctx context.Context, tc *api.TxnContext) (uint64, error)
 	h := hooks.GetHooks()
 	tctx, err := h.CommitOrAbort(ctx, tc)
 	if err != nil {
+		// Zero signals a server-decided abort as a codes.Aborted status carrying a categorized
+		// reason (e.g. "conflict: ...", "predicate-move: ..."). Record the abort metric and
+		// forward the reason rather than treating it as a generic transport error or flattening
+		// it to the reasonless dgo.ErrAborted.
+		if status.Code(err) == codes.Aborted {
```
matthewmcneely, line 863: This early return skips both `span.AddEvent("Error in CommitOrAbort", ...)` and `span.AddEvent("Commit status", ...)`.

Before this change every abort emitted the second one with `committed=false`. Now a server-decided abort emits neither, so the transactions you'd most want to trace are the ones that go dark. Worth adding an event carrying the reason before returning.

edgraph/server.go
```diff
@@ -2139,6 +2143,12 @@ func (s *Server) CommitOrAbort(ctx context.Context, tc *api.TxnContext) (*api.Tx
 
 		return tctx, status.Error(codes.Aborted, err.Error())
 	}
+	if status.Code(err) == codes.Aborted {
+		// Server-decided abort carrying a categorized reason; err is already a codes.Aborted
+		// status (e.g. "conflict: ...", "predicate-move: ...") — surface it unchanged.
+		tctx.Aborted = true
+		return tctx, err
+	}
```
matthewmcneely, line 2151: `tctx.Aborted = true` here (and the matching line in `doMutate`) is inert as far as I can tell.

gRPC drops the response message when a unary handler returns a non-nil error, so no gRPC client sees the flag. The in-process callers don't read it either: `handleCommit` (`dgraph/cmd/alpha/http.go:605`) returns on `err != nil` before touching `tc`, `handleAbort` is the client-discard path, and both `graphql/resolve/mutation.go` call sites discard the context on error.

Harmless and consistent with the existing `dgo.ErrAborted` branch above, so I'm not asking you to pull it — but the PR description sells "set `Aborted`" as part of the fix and it isn't observable anywhere.

dgraph/cmd/zero/oracle.go
```diff
@@ -338,56 +340,223 @@ func (s *Server) proposeTxn(ctx context.Context, src *api.TxnContext) error {
...
+// Abort-reason codes. When Zero decides to abort a transaction it surfaces the category to the
+// client as the prefix of a codes.Aborted gRPC status message, formatted as "<code>: <detail>".
+// api.TxnContext (external dgo module) has no field for the reason, so the reason rides on the
+// error instead; the status code stays codes.Aborted, so existing abort handling is unaffected.
+// The dgraph4j and pydgraph clients parse these prefixes into their AbortReason enums — keep all
+// three in sync.
+//
+// These three codes are the entire published vocabulary. A cause that cannot be mapped onto one of
+// them emits the detail with *no* prefix (abortReasonUncategorized) rather than borrowing a code
+// that would imply the wrong remedy. Both dgraph4j and pydgraph already degrade an absent or
```
matthewmcneely, line 352: Both of these are true only with dgraph4j #290 and pydgraph #307, and both are still open.

Present tense in a merged comment ("already degrade") will read as shipped behavior to the next person in this file. Either name the PRs or the client versions, or land the clients first.

dgraph/cmd/zero/oracle.go
```diff
@@ -338,56 +340,223 @@ func (s *Server) proposeTxn(ctx context.Context, src *api.TxnContext) error {
...
+	//   - the Alpha leader ageing out transactions idle longer than --limit "txn-abort-after"
+	//     (abortOldTransactions, worker/draft.go)
+	//
+	// Before this was handled, commit() returned nil here and the abort reached the client as a
+	// bare dgo.ErrAborted with no reason at all.
+	abortDetailPreAborted = "Transaction has been aborted. Please retry. It was already aborted " +
+		"before this commit was decided, which happens when a schema update or a drop-predicate " +
+		"cancels pending transactions on a predicate it touched, or when the server ages out " +
+		"transactions idle for longer than --limit \"txn-abort-after\""
+
```
matthewmcneely, line 438: Two things here.

The cause list reads as exhaustive ("which happens when...") but misses a fourth: `commit()`'s own `src.Aborted` early return proposes an abort for that start ts, so a client `Discard` racing a `CommitNow` on the same transaction lands here too.

Separately, `proposeTxn`'s comment ~100 lines up says "A predicate move, triggered by Zero, would abort all pending transactions," which contradicts "none of them is a predicate *move*" here. I traced it and I think **your** version is the correct one: a move blocks commits at `checkPreds` before the lease is taken, and the source Alpha's move path uses `posting.Oracle().WaitForTs` rather than `TryAbort`. Worth correcting the old comment in passing so the two aren't sitting in the same file disagreeing.

systest/integration2/txn_abort_reason_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	// Commit the stale transaction via the raw stub to observe the categorized status.
+	port, err := c.GetAlphaGrpcPublicPort(0)
+	require.NoError(t, err)
+	conn, err := grpc.NewClient("0.0.0.0:"+port, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	require.NoError(t, err)
+	defer func() { _ = conn.Close() }()
+	raw := api.NewDgraphClient(conn)
+
+	_, err = raw.CommitOrAbort(ctx, tc)
```
matthewmcneely, line 80: This can fail on the status code rather than the prefix.

`defaultZeroHooks.CommitOrAbort` (`worker/default_zero_hooks.go:65`) does `groups().Leader(0)` and returns `conn.ErrNoConnection` when that's nil, which arrives as `codes.Unknown`. `GetZeroLeader` polls Zero's own `/state`, not the Alpha's membership view, so there's a window where Zero has a leader and the Alpha hasn't caught up.

Suggest polling the commit itself: `require.Eventually` around `raw.CommitOrAbort` until `status.Code(err) == codes.Aborted`, then assert the prefix. That also lets you drop the separate leader poll.

systest/integration2/txn_abort_reason_test.go
```diff
@@ -0,0 +1,88 @@
+//go:build integration2
+
+/*
+ * SPDX-FileCopyrightText: © 2017-2025 Istari Digital, Inc.
```
matthewmcneely, line 4: Nit: prevailing header in the repo is `© 2017-2026` (568 files, vs 7 still on 2025). Same in `dgraph/cmd/zero/oracle_reason_test.go`.

matthewmcneely (changes requested):

PR #9748, opened 2026-06-17
edgraph: add AlterNoAuth for trusted in-process schema callers

matthewmcneely:
## What

Adds `AlterNoAuth` — `Alter` without the admin-IP-whitelist and ACL authorization checks — mirroring the existing `Query`/`QueryNoAuth` pair, for trusted in-process callers.

- `validateAlterOperation` gains an `AuthMode`: the structural checks (field set, health, mutations-allowed, reserved-namespace) always run; the IP-whitelist and ACL checks run only under `NeedAuthorize`.
- `Alter` and `AlterNoAuth` share a private `alter()`.
- `AlterNoAuth` is restricted to schema operations — `isDropOperation` refuses every drop form — so bypassing auth can never be used to remove data. It must not be exposed to network clients.

## Why

An in-process caller using `context.Background()` carries no gRPC peer, so `x.HasWhitelistedIP` fails with `"unable to find source ip"` on every config — not only under `--security` whitelist. `QueryNoAuth` already solves this for queries; `AlterNoAuth` is its schema-write counterpart.

## Tests

`edgraph/alter_noauth_test.go` (white-box) exercises the `validateAlterOperation` auth gate — a background context is rejected under `NeedAuthorize`, accepted under `NoAuthorize`, and the structural/health checks still fire — plus `AlterNoAuth`'s drop refusal. No cluster needed.

### DIFF
--- a/edgraph/alter_noauth_test.go
+++ b/edgraph/alter_noauth_test.go
@@ -1,0 +1,89 @@
+/*
+ * SPDX-FileCopyrightText: © 2017-2025 Istari Digital, Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package edgraph
+
+import (
+	"context"
+	"net"
+	"testing"
+
+	"github.com/dgraph-io/dgo/v250/protos/api"
+	"github.com/dgraph-io/dgraph/v25/x"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc/peer"
+)
+
+// TestValidateAlterOperationAuthGate isolates the authorization gate that an
+// in-process Alter caller hits. It proves the bug (a context with no gRPC peer
+// is rejected at the IP-whitelist check) and the fix (NoAuthorize skips that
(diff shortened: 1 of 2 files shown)


edgraph/alter_noauth_test.go
```diff
@@ -0,0 +1,89 @@
+/*
+ * SPDX-FileCopyrightText: © 2017-2025 Istari Digital, Inc.
```
amalistari, line 2: 2026

matthewmcneely: Let's keep this PR focused on the features. I've opened #[9750](https://github.com/dgraph-io/dgraph/issues/9750) to accomplish these changes (there will be many).

mlwelles (approved):
LGTM. The `NoAuthorize` early-return sits after every structural check — only the IP-whitelist and ACL checks are skipped — and `AlterNoAuth` refuses each drop form through `isDropOperation`. The white-box tests assert both the bug and the fix. Clean.

PR #9749, opened 2026-06-17
x,edgraph,worker: add a reserved-namespace plugin registry

matthewmcneely:
## What

Adds `x.RegisterReservedNamespace`: a plugin can claim ownership of a sub-namespace under the reserved `dgraph.` prefix — a dynamic predicate prefix and/or an exact predicate/type allowlist — so its names can be created at runtime via `Alter` even though they aren't part of the pre-defined initial schema, and optionally value-lock a subset of its predicates to a trusted in-process writer.

The three reserved-namespace enforcement points consult the registry:
- `parseSchemaFromAlterOperation` — a registered predicate/type passes the reserved-but-not-pre-defined check (`IsRegisteredReservedPredicate` / `IsRegisteredReservedType`).
- `worker.proposeAndWait` — a registered predicate may be mutated before its schema exists.
- the mutation value guard (`newReservedPredicateGuard`) — a write to a value-locked predicate is rejected unless the request context carries the namespace's `TrustMarker` (`ReservedPredicateValueLock`).

## Why

There is currently no way for a downstream consumer to own names under `dgraph.` — the prefix is all-or-nothing reserved. This adds a narrow, explicit extension point rather than a broad carve-out.

## Safety

With no registration the registry is empty and every check reports false, so a stock build keeps the pristine behavior: nothing under `dgraph.` may be created except the pre-defined names, and the GraphQL reserved-value path (`IsGraphql` / `IsOtherReservedPredicate`) is unchanged.

## Notes for review

- **Stacked on #9748** (AlterNoAuth) — both touch `edgraph/server.go` in separate regions, so this targets that branch; please merge #9748 first. The diff here is registry-only.
- `ParseMutationObject(ctx, mu)` is an exported signature change (was `(mu, isGraphql bool)`) so the value guard can read the request context. The only in-tree caller is `parseRequest`; happy to add a compatibility shim if you'd prefer.

## Tests

`x/reserved_namespace_test.go` covers dynamic-prefix and exact membership, the predicate/type split, and value-lock lookup against a dummy registered namespace.

### DIFF
--- a/x/keys.go
+++ b/x/keys.go
@@ -12,6 +12,7 @@ import (
 	"math"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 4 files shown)


x/keys.go
```diff
@@ -644,6 +645,108 @@ var aclPredicateMap = map[string]struct{}{
...
+		reservedNsPrefixes = append(reservedNsPrefixes, strings.ToLower(ns.PredicatePrefix))
+	}
+	for _, p := range ns.Predicates {
+		reservedNsPredicates[strings.ToLower(p)] = struct{}{}
+	}
+	for _, t := range ns.Types {
+		reservedNsTypes[strings.ToLower(t)] = struct{}{}
+	}
+	for _, p := range ns.ValueLocked {
+		reservedNsValueLocked[p] = ns.TrustMarker
```
mlwelles: **Value locks match case-sensitively, but ownership does not — so a value lock can be bypassed by changing case.**

`Predicates`, `Types`, and `PredicatePrefix` are all stored lowercased ([L700-L708](https://github.com/dgraph-io/dgraph/blob/e997efd01214c7c53602e943debd846b9a893f98/x/keys.go#L700-L708)), and `IsRegisteredReservedPredicate` looks them up with `strings.ToLower(ParseAttr(pred))`. `ValueLocked` is the exception: it is stored raw here, and `ReservedPredicateValueLock` looks it up raw ([L745-L747](https://github.com/dgraph-io/dgraph/blob/e997efd01214c7c53602e943debd846b9a893f98/x/keys.go#L745-L747)).

So if a namespace value-locks a name with any uppercase letter (say `dgraph.foo.Secret`), the variant `dgraph.foo.secret` is still owned — the prefix and predicate checks lowercase it — yet not locked. A mutation to that variant skips the trust-marker check and goes through. The lock holds only for the exact spelling registered.

Every reserved name today is lowercase, so nothing breaks in practice. But the two halves of the registry disagree on case, which is the kind of gap that bites a future caller.

Fix — lowercase both sides, matching the loops just above:
- here: `reservedNsValueLocked[strings.ToLower(p)] = ns.TrustMarker`
- in `ReservedPredicateValueLock`: `marker, locked = reservedNsValueLocked[strings.ToLower(pred)]`

One caution: don't also add `ParseAttr` to mirror `IsRegisteredReservedPredicate`. The guard passes the bare predicate (no `-` namespace separator), so `ParseAttr` would panic at `SplitN(...)[1]` — which is exactly why the neighboring `IsOtherReservedPredicate` matches raw. Only the case folding is missing.

x/keys.go
```diff
@@ -644,6 +645,108 @@ var aclPredicateMap = map[string]struct{}{
...
+	Types []string
+	// ValueLocked is the subset of owned predicates whose stored *value* may
+	// only be written by a request whose context carries TrustMarker. Use it
+	// for predicates whose value is authoritative state the owner must control.
+	// Names must be the exact (bare) predicate form, matching how
+	// IsOtherReservedPredicate is consulted on the mutation path.
+	ValueLocked []string
+	// TrustMarker is a context key the owner's trusted in-process caller sets,
+	// via context.WithValue(ctx, TrustMarker, true), to authorize writing
+	// ValueLocked predicates. Required (non-nil) iff ValueLocked is non-empty.
```
mlwelles: **Optional / non-blocking.** Consider enforcing this invariant rather than only documenting it.

The comment says `TrustMarker` is "Required (non-nil) iff `ValueLocked` is non-empty," but `RegisterReservedNamespace` ([L694-L708](https://github.com/dgraph-io/dgraph/blob/e997efd01214c7c53602e943debd846b9a893f98/x/keys.go#L694-L708)) never checks it. If a caller fills `ValueLocked` but leaves `TrustMarker` nil, the guard's `marker != nil` test stays false, and the predicate becomes unwritable by everyone — including its owner.

That fails closed, so it is safe rather than urgent. But it surfaces only at mutation time, far from the registration that caused it. Since registration runs in `init()`, a `panic` (or a returned error) here would catch the mistake at startup instead.

mlwelles (commented):
Two notes on the reserved-namespace registry, both in `x/keys.go`: one correctness issue in the value-lock matching, plus one optional hardening idea.

mlwelles (approved):
LGTM. The two threads I left — the value-lock case-folding and the optional `TrustMarker` invariant — are non-blocking; address them here or in a follow-up as you prefer.

x/keys.go
matthewmcneely: OK, cool. This one addressed in [855e2eb](https://github.com/dgraph-io/dgraph/pull/9749/commits/855e2eba56b46b5463b178351f3bd9f0f1ecd755). Also updated the upstream PR.

x/keys.go
matthewmcneely: Same here. Addressed in [ff30800](https://github.com/dgraph-io/dgraph/pull/9749/commits/ff30800bfe7b282a50c5582a788f70b035f165ba) and the upstream PR is updated

matthewmcneely:
Auto-closed when the stacked base `oss-alter-noauth` (#9748) was deleted on merge. Reopened as #9753, now based on `main` with both review items addressed (case-fold value-lock lookups + TrustMarker enforcement). Continuing there.

PR #9760, opened 2026-06-20
fix(backup): contain restore file paths within the handler root

alhudz:
**Description**

The `path` field of a backup manifest is read back from the backup location (`readMasterManifest`) and used as-is during restore.

1. `restore_map.go` joins `manifest.Path` with the backup file name and streams it through `fileHandler.Stream` -> `os.Open(filepath.Join(rootDir, prefix, path))`.
2. `getManifestsToRestore` joins the same `manifest.Path` for `FileExists`, and the s3/minio handler builds object keys the same way in `getObjectPath`.

`JoinPath`/`getObjectPath` do not contain the relative path, so a manifest left in the backup target with `path` set to `../../..` steers those reads outside the handler root. The fix sits in the handlers because that is the single choke point every backup/restore file op goes through.

`Repro:` `worker/backup_handler_test.go` (fails before, passes after).
`Expected:` the resolved path stays under the handler root.
`Actual:` `JoinPath("../../../../etc/passwd")` returns `/etc/passwd`; `getObjectPath` returns `../../etc/passwd`.
`Fix:` `cleanRelPath` anchors the path at the root and drops a leading separator / `..` before the join; `JoinPath` (file + s3) and `getObjectPath` all route through it.

**Checklist**

### DIFF
--- a/worker/backup_handler.go
+++ b/worker/backup_handler.go
@@ -13,6 +13,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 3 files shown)


alhudz:
gentle ping

shiva-istari (commented):
Thanks for the fix, @alhudz !

I went through the changes and they look good to me. It would be nice to add a table-driven test for cleanRelPath covering a few edge cases like "", ".", a/../b, absolute paths, and paths with trailing or duplicate separators. I think that would make future regressions less likely.

In addition, an integration test would be valuable to verify the end-to-end behavior and help prevent regressions in the future.

Other than that, LGTM. Happy to see this merged.

alhudz:
Both added. `TestCleanRelPath` is table-driven and covers the cases you listed (`""`, `.`, `a/../b`, absolute paths, trailing and duplicate separators). For end-to-end, `TestRestorePathContainmentEndToEnd` builds a real backup layout on disk, plants a `manifest.json` whose `path` points at `../outside`, and drives `getManifestsToRestore` through the file handler: the planted file stays unreachable and the manifest is rejected, while a legitimate backup streams back unchanged. I checked it against the pre-fix handler as well: fails before the patch, passes after. It runs in the plain unit suite so it gates every CI run; if you'd rather have a cluster-level version under `systest/backup` I can port it there too.

shiva-istari (approved):

PR #9761, opened 2026-06-21
fix(graphql): enforce update auth on existing nodes linked via @hasInverse

kike-loco:
## Summary

Fixes a **false accept** security gap in GraphQL add mutations: when a nested insert links to an **existing parent** via `@hasInverse`, Dgraph mutates the parent's inverse predicate but previously only ran `@auth add` checks on newly allocated UIDs. Protected parents were never validated, allowing unauthorized users to modify admin-only nodes indirectly.

Example bypass (now blocked):

```graphql
mutation {
  addFooItem(input: [{ parent: { id: "foo1" } }]) { numUids }
}
```

…when `ProtectedFoo` requires admin for both add and update.

- Track **affected existing nodes** during mutation rewrite when `asIDReference()` adds an inverse link to an existing UID (`mutation_rewriter.go`)
- After `authorizeNewNodes()`, run **`authorizeAffectedNodes()`** using **update rules** on those nodes (`mutation.go`)
- Scope affected-node auth to **add mutations only** to avoid breaking update mutation flows

Related discussion: https://discuss.dgraph.io/t/bug-auth-rules-of-parent-not-respected-when-child-with-hasinverse-is-added/12955

## Test plan

- [x] Unit rewrite tests: `go test ./graphql/resolve -run 'TestAddChildRecordsAffectedProtectedParent|TestNestedAddRecordsDeepNewNodes' -count=1`
- [x] Auth query rewriting: `go test ./graphql/resolve -run TestAuthQueryRewriting/Add_Mutation -count=1`
- [x] E2E integration: `go test -tags=integration -run TestNestedAdd ./graphql/e2e/auth/...` (requires Docker cluster with patched binary)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/NESTED_AUTH_FIX.md
+++ b/NESTED_AUTH_FIX.md
@@ -1,0 +1,99 @@
+# Nested Insert @auth Fix
+
+This branch fixes a **false accept** security gap in Dgraph GraphQL mutations: nested inserts and `@hasInverse` edge linking could modify protected nodes without running the parent type's update authorization rules.
+
+## Problem
+
+When a mutation links a new child to an **existing parent** via `@hasInverse`, Dgraph mutates the parent's inverse predicate but only ran `@auth add` checks on **newly allocated UIDs**. Existing parents were never validated, so non-admin users could run mutations like:
+
+```graphql
+mutation {
+  addFooItem(input: [{ parent: { id: "foo1" } }]) { numUids }
+}
+```
+
+…even when `ProtectedFoo` requires admin for both add and update.
+
+Deep nested adds could also create leaf nodes whose stricter `@auth add` rules were not enforced consistently.
+
+## Fix
+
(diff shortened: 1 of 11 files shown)


docker-compose.override.example.yml
```diff
@@ -0,0 +1,5 @@
+# Example override for projects using Dgraph standalone (e.g. raggen).
```
matthewmcneely, line 1: Please don't add fix-specific artifacts in your PR

NESTED_AUTH_FIX.md
```diff
@@ -0,0 +1,99 @@
+# Nested Insert @auth Fix
```
matthewmcneely, line 1: Please don't add fix-specific artifacts in your PR

matthewmcneely (changes requested):
Core tests are failing and seem related to this change.

PR #2330, opened 2018-04-11
Allow filtering inside of lists for regexp tokenizers

gpahal:
Fixes https://github.com/dgraph-io/dgraph/issues/2271 for regexp

To be resolved:
- [ ] Check if list supports lang tags

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/2330)

PR #9762, opened 2026-06-23
perf(core): per-predicate mutation pipeline with intra-mutation parallel apply (revives #9467)

rahst12:
**Scope, stated plainly:** Every execution change on this branch parallelizes work inside a single transaction's apply, within one iteration of the serial processApplyCh loop. The per-predicate pipeline (from #9467) and your entire optimization series (proportional budgets, parallel reverse/forward writes, conflict-key batching, map sizing) all fan out inside one mutation's pre-write and join before the loop moves on. The loop itself stays serial across entries; no two transactions ever execute concurrently anywhere on this branch. Today's two additions don't change that — txns_per_delta is observability (it measures cross-transaction batching but executes nothing differently), and the help-text deletion removed the description of a cross-transaction feature that never existed.

## What

The complete per-predicate mutation pipeline, now targeting `main`: the pipeline implementation revived from #9467, its correctness-fix series, and a measured intra-mutation parallelism series on top (+36.8% sustained ingest on a 32-core EC2 box, +17.6% on a production-replica predicate mix; re-validated 2026-08-08 after the main merge at **+39.8% — 1.40× — over current main**, table below).

The legacy apply path processes all edges of a Raft proposal serially. The pipeline groups a mutation's edges by predicate and applies them in parallel — one runner per predicate, with this PR adding a proportional worker budget across predicates, per-CPU (`auto`) sizing, and a parallel forward data-write in `ProcessList` so `[uid] @reverse` predicates parallelize too (their forward write was serial and measured 1.00× before that change).

## Why the base changed

This PR previously targeted `harshil-goel/mutation-pipeline` (#9467), but to help keep this branch and PR up to date with main, it's repointed to main. This also helps make benchmarking the two PRs easier.

## What's included

**The pipeline series (from #9467, authorship preserved):**
- The per-predicate mutation pipeline, gated behind a feature flag
- Correctness fixes found while reviving it (all were in the original branch, none in main): key-buffer reuse silently corrupting unrelated reverse posting lists; `InsertTokenizerIndexes` deadlock on UIDs ≥ 2⁶³; `IterateDisk.IsEmpty` stubbed to `false` (broke `has()` after star-deletes); HNSW vector-insert races (now serialized); star-delete skipping index deletion during background indexing; scalar Del-of-old-value wiping a new Set in `ProcessCount`; nil-deref in `ProcessSingle` on multi-Del-per-uid; six further gaps surfaced by alpha integration tests
- Regression harnesses: four `[uid] @reverse @count` tests, a 592-genre reverse-count systest

**The intra-mutation parallelism series (this branch, each commit benchmarked on i4i.16xlarge):**

| commit | change | MB/s (badger_write_bytes_user) |
|---|---|---:|
| — | branch base | 9.63 |
| parallel reverse write | lock-free store at any worker grant | 10.58 |
| lock-free forward write | at a one-worker grant | 11.44 |
| conflict-key batching | one lock acquisition per batch | 12.45 |
| uid-lease snapshot | once per mutation, not per nquad | 12.53 |
| `LockedShardedMap` sizing | per machine; drop per-call hash alloc | **13.17** |

**+36.8% cumulative** (reverse-heavy corpus: 25 `[uid] @reverse` predicates, 20k triples/txn, 64 writer threads, 300 s runs). A production-replica predicate mix measured **+17.6%**.

## Review-critical facts, stated up front

1. **The pipeline becomes the default apply path.** `intra-mutation-min-edges` defaults to `1`, so every mutation ≥1 edge routes through the pipeline. The single kill switch is `intra-mutation-min-edges=0`, which restores the legacy serial path entirely.
2. **The flag surface was renamed** from #9467's `mutations-pipeline-threshold` to `intra-mutation-*`. The names say what they scope: this parallelizes a *single* mutation. Transactions still apply serially in `processApplyCh`, so these flags will not relieve a many-concurrent-writers bottleneck.

## How to configure

All keys live under the existing `--feature-flags` superflag. Every key is live at every setting:

| Key | Default | Meaning |
|---|---|---|
| `intra-mutation-min-edges` | `1` | Takes the parallel apply path when `>0` **and** `len(edges) >= it`. `0` = legacy path entirely, and is the single kill switch. |
| `intra-mutation-parallelism` | `auto` | Worker goroutines inside one mutation. `off` = one per predicate, no fan-out. `N` = exactly N. `Fx` = F per CPU available to Go (e.g. `1.5x`). `auto` = `1x`. |
| `intra-mutation-edges-per-worker` | `256` | Caps the pool at `totalEdges / this` so small mutations don't over-spawn. Applies to **every** sizing mode. |

```bash
# Sized to the host (the default):
dgraph alpha ... --feature-flags "intra-mutation-parallelism=auto"

# Fixed 64-worker pool:
dgraph alpha ... --feature-flags "intra-mutation-parallelism=64"

# No intra-predicate fan-out, but still the per-predicate pipeline:
dgraph alpha ... --feature-flags "intra-mutation-parallelism=off"

# Legacy path entirely:
dgraph alpha ... --feature-flags "intra-mutation-min-edges=0"
```

Note which term binds: at the default `edges-per-worker=256`, a 20k-edge mutation caps at **78 workers however many cores exist** — above ~78 CPUs the cap governs, not the multiplier. Lowering `intra-mutation-edges-per-worker` is what lifts that ceiling. The `V(2)` log line reports which term won (`boundBy=`).

## Correctness verification

- `intra-mutation-parallelism=off` is byte-identical to the one-goroutine-per-predicate path; byte-identical committed output **and** identical conflict-key set verified under `-race` for scalar, `[uid] @reverse`, and per-CPU-vs-fixed sizing.
- A schema/index-type matrix verifies byte-identical committed state across 12 schema shapes.
- `ProcessReverse` stays serial deliberately (many-to-one, hot-target-prone — parallelizing it measured as a net loss); all workers join before the reverse/index/count passes, preserving MVCC order.
- Integration suite: 39 packages, zero failures (beyond flakes that also fail on the base).
- Merged with current `main` (2026-08-08): zero conflicts; full build and the `posting/`, `worker/`, `x/` unit suites pass on the merged tree.

## Three-arm benchmark against current main (2026-08-08)

Re-validated after the merge with main, on an i4i.8xlarge (32 vCPU, NVMe instance store): reverse-heavy corpus (25 `[uid] @reverse` predicates, 20k triples per transaction), 64 writer threads, 300 s sustained `commitNow` ingest, fresh store per run, three interleaved rounds, every arm at its shipping defaults. Metric: `badger_write_bytes_user` delta over elapsed.

| Arm | MB/s (mean of 3) | vs main | triples/s (mean) | vs main |
|---|---:|---:|---:|---:|
| `main` @ `64804bd28` (control) | 9.06 | — | 90,998 | — |
| #9467 tip @ `918262cdc` | 9.49 | +4.7% (1.05×) | 94,490 | +3.8% |
| this branch @ `5ce99cdca` | 12.67 | **+39.8% (1.40×)** | 125,447 | +37.9% |

Zero failed batches and zero aborts in every run; control spread across rounds was under 0.3%. This branch over the #9467 tip directly: +33.5% MB/s — noting #9467's tip still sits on its May 2026 base, so that direct delta includes 40 commits of newer main; the control arm anchors both to today's main.

## Known open item

On an `@lang`-heavy workload the pipeline path emits fewer conflict keys than the legacy path (divergence measured at ~600 keys on a synthetic corpus). This is pre-existing pipeline behavior, not introduced by this series, and is documented as the highest-priority follow-up since the risk direction (lost update under SI) matters once the pipeline is the default. Flagging it here rather than leaving it for reviewers to find.

## Benchmarks (8-core dev box — a floor; the win scales with cores)

~20k-edge batch, 5 hot reverse targets. `edges/s`, speedup vs `budget=0`:

| Workload | budget=0 | budget=8 | budget=32 | auto |
|---|---|---|---|---|
| **`[uid] @reverse` dominant** | 1.00× | **1.35×** | 1.29× | 1.15× |
| **50% reverse / 50% scalar** | 1.00× | 1.42× | **1.82×** | 1.39× |
| scalar `@index` dominant (control) | 1.00× | 1.23× | 1.44× | 1.19× |

`BenchmarkReverseDominant` / `BenchmarkReverseFiftyFifty` are included so this can be re-swept on production-class hardware.

## At-scale benchmark: does the improvement survive LSM compaction? (2026-08-09)

The empty-store numbers above leave one fair question: does the gain hold once Badger has a deep LSM tree and compaction competes with ingest? Measured directly:

**Test scale:** a **34 GB seeded store** (≈1 billion distinct triples ingested via a bulk pass, plus the measured corpus resident so every benchmark write overwrites live keys), i4i.8xlarge (32 vCPU, NVMe), 64 writer threads, `commitNow` ingest. Every run restores the identical seed snapshot; runs are interleaved; compaction is continuously active during measurement (120–203 slow compactions per 30-minute window, levels 4→5 and 5→6). Verdict rendered against pre-registered acceptance criteria (8/9 pass; full protocol, per-run data and formal verdict archived).

| Scenario | main | this branch | **advantage** |
|---|---:|---:|---:|
| Empty store, 300 s × 3 (series 06) | 9.06 MB/s | 12.67 MB/s | **+39.8% (1.40×)** |
| 34 GB store, 600 s × 3 | 8.41 MB/s | 11.58 MB/s | **+37.7% (1.38×)** |
| 34 GB store, 1800 s authoritative | 8.38 MB/s | 11.32 MB/s | **+35.1% (1.35×)** |

**Stability under compaction, beyond the headline:**

- **No decay over 30 sustained minutes**: the branch's advantage was +35.9% in the first 10 minutes and +35.1% in the last 10 — compaction debt does not erode it.
- **Faster and smoother**: rate variability CV 0.20 vs main's 0.44. Both arms show a periodic ingest dip anti-correlated with compaction bursts (an intrinsic Badger ingest/compaction rhythm, present on stock main); this branch **dampens** it — 97% vs 155% peak-to-trough, dips half as frequent.
- **No pathology from the higher write rate**: zero L0 stalls, zero aborts, zero failed batches in every run on both arms; the ~+69% extra compaction work the higher throughput generates is fully absorbed (disk never exceeded 49% utilization).
- **Cross-checked on badger v4.9.6**: both arms rebuilt against the newer badger (with the #2330 compaction base-level fix) reproduce the margin (+35.3%), ruling out the result being an artifact of a known v4.9.4 compaction regression.

One follow-up this testing surfaced (will fix in this PR): the parallel apply path does not record `num_edges_total`, so operators on the pipeline path lose the nquads/sec metric — the recording sites currently live only on the legacy path.

https://claude.ai/code/session_011uGZrf7FExkfwT5cWowrb5

## Summary by CodeRabbit

- **New Features**
  - Added configurable intra-mutation parallelism, including automatic, fixed-worker, and disabled modes.
  - Improved mutation throughput for indexed, reverse, list, language, scalar, count, and vector data.
  - Added metrics for mutation parallelism and transactions per committed delta.

- **Bug Fixes**
  - Improved handling of schema subscription postings and delete-all mutations.
  - Preserved posting-list state more reliably across updates and commits.
  - Strengthened UID validation during mutation processing.

- **Documentation**
  - Updated the documented project module path.


rahst12:
TODO: The flags need consolidated to something more manageable.

Issue #491, opened 2017-01-16
ICU_DATA breaks Java on OSX


nlothian:
After installing the ICU data on OSX following the instructions at https://wiki.dgraph.io/Deploy#String_indexing Java stopped working with the following error:

`~ java`

` 2017-01-16 14:29:16.193 java[1110:14026] [JVM Detection] FillMatcher: failed to create CFNumberFormatter`

Un-setting the ICU_DATA environment variable fixes the problem.

jchiu0:
Related issue: https://github.com/dgraph-io/dgraph/issues/429

I would have to revert the changes for #429 and look into why it doesn't work for the embedded binary.

jchiu0:
Give it a try. It should be fixed now. Be sure to pull the latest from dgraph-io/goicu.

jtblin:
> Give it a try. It should be fixed now. Be sure to pull the latest from dgraph-io/goicu.

What does that mean exactly? I still have the issue with the latest dgraph-io/goicu. I have an environment variable set as `export ICU_DATA=/usr/local/share/icudt58l.dat` from a previous installation I guess so I'm not too sure. Unsetting the environment variable seems to be the only way to fix that right now. Is that expected?

pawanrawal:
Hi @jtblin 

As of v0.7.3, you don't need to set the ICU_DATA environment variable. You can unset it. Make sure to install the latest version using https://wiki.dgraph.io/Get_Started#System_Installation and let us know if something doesn't work.

jtblin:
OK, thanks. Yes I've removed the environment variable, I guess it was from an old installation.

manishrjain:
Note that from the coming release, we would get rid of ICU in favor of Bleve, which would work without any of these env variable hacks.

PR #9765, opened 2026-06-29
fix: handle mremap size mismatch on Linux s390x

Spratima:
### **Description**

This PR fixes a runtime crash in Dgraph Alpha on IBM s390x architecture
that prevents the service from restarting after any unclean shutdown.

### **Problem**

After any unclean exit of Alpha (SSH session close, kill -9, OOM kill, VM
reboot), a stale memtable file (00001.mem, always 128 MB) is left in the
postings directory. On restart, BadgerDB calls mremap(2) to shrink the
mapping from 128 MB down to the actual data size.

On s390x, the Linux kernel does not support shrinking a mapping via mremap
— it returns the original size (134,217,728 bytes) instead of the requested
smaller size. Ristretto asserts returned == requested and calls log.Fatalf,
crashing Alpha before it can serve any requests:

    while opening memtables err: while opening fid: 3 err:
      while updating skiplist err:
        mremap size mismatch: requested: 2628 got: 134217728
    Error while creating badger KV posting store

Ristretto already has this fix for ARM64 (identical kernel behavior) via
mremap_nosize.go. The fix was never extended to s390x.

### **Fix**

This PR updates the vendored Ristretto files to add s390x to the build tags
of the existing ARM64 safe path. The upstream fix has been raised at:
https://github.com/dgraph-io/ristretto/pull/491

vendor/github.com/dgraph-io/ristretto/v2/z/mremap_nosize.go
- //go:build (arm64 || arm) && linux && !js
+ //go:build (arm64 || arm || s390x) && linux && !js
- // +build arm64 arm
+ // +build arm64 arm s390x

vendor/github.com/dgraph-io/ristretto/v2/z/mremap_size.go
- //go:build linux && !arm64 && !arm && !js
+ //go:build linux && !arm64 && !arm && !s390x && !js
- // +build linux,!arm64,!arm,!js
+ // +build linux,!arm64,!arm,!s390x,!js

**Note: The vendored files are not tracked in upstream main, so the diff
shows the full file content as additions. The only meaningful change is
the two build tag lines at the top of each file.**

### **Validation Performed**

- s390x (IBM Fyre VM, RHEL on IBM Z) - Native full runtime test
  - Alpha restarts cleanly after unclean shutdown with stale 128 MB .mem present
  - Full CRUD operations verified on Dgraph v25.3.4 and v25.3.7

- linux/arm64, linux/amd64, x86_64 - go build ./... - Build successful on all three

### **Impact**

- Resolves Alpha startup crash on IBM s390x after any unclean shutdown
- Build-tag change only — no functional logic changes introduced
- No impact on x86_64, ARM64, ARM, or any other supported architecture

### **Related**

- Upstream Ristretto fix: https://github.com/dgraph-io/ristretto/pull/491
- Follows fix: S390x compatibility (#9746) which fixed the s390x compilation blocker

### **Checklist**

- [x] The PR title follows the Conventional Commits syntax, leading with fix:
- [x] Code compiles correctly on s390x, x86_64, linux/arm64, linux/amd64

Note: A unit test for this fix is not feasible — the mremap shrink behavior
is specific to the Linux s390x kernel and cannot be reproduced or mocked on
other architectures. Runtime validation was performed directly on an IBM Z
(s390x) RHEL VM as described in the Validation section above.

### DIFF
--- a/vendor/github.com/dgraph-io/ristretto/v2/z/mremap_nosize.go
+++ b/vendor/github.com/dgraph-io/ristretto/v2/z/mremap_nosize.go
@@ -1,0 +1,46 @@
+//go:build (arm64 || arm || s390x) && linux && !js
+// +build arm64 arm s390x
+// +build linux
+// +build !js
+
+/*
+ * SPDX-FileCopyrightText: © 2017-2025 Istari Digital, Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package z
+
+import (
+	"reflect"
+	"unsafe"
+
+	"golang.org/x/sys/unix"
+)
+
+// mremap is a Linux-specific system call to remap pages in memory. This can be used in place of munmap + mmap.
+func mremap(data []byte, size int) ([]byte, error) {
+	//nolint:lll
+	// taken from <https://github.com/torvalds/linux/blob/f8394f232b1eab649ce2df5c5f15b0e528c92091/include/uapi/linux/mman.h#L8>
+	const MREMAP_MAYMOVE = 0x1
+
+	header := (*reflect.SliceHeader)(unsafe.Pointer(&data))
+	// For ARM64, the second return argument for SYS_MREMAP is inconsistent (prior allocated size) with
+	// other architectures, which return the size allocated
+	mmapAddr, _, errno := unix.Syscall6(
+		unix.SYS_MREMAP,
+		header.Data,
+		uintptr(header.Len),
+		uintptr(size),
+		uintptr(MREMAP_MAYMOVE),
+		0,
+		0,
+	)
+	if errno != 0 {
+		return nil, errno
+	}
+
+	header.Data = mmapAddr
+	header.Cap = size
+	header.Len = size
+	return data, nil
+}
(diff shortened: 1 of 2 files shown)


Spratima:
Hello @matthewmcneely,

Could you please review this PR as well? It contains the same fix required for the s390x compatibility issue.

Since you've already reviewed and merged the corresponding change in Ristretto (https://github.com/dgraph-io/ristretto/pull/491), your review on this PR would be greatly appreciated. Thanks!

matthewmcneely:
@Spratima These are vendored files (vendor/[github.com/dgraph-io/ristretto/v2/z/](http://github.com/dgraph-io/ristretto/v2/z/)), not committed in the repo. Modifying vendored code directly diverges from the upstream module and will be overwritten on the next `go mod vendor`/dependency bump. The change should land in upstream Ristretto (PR #491 is referenced) and be pulled in via go.mod, otherwise the fix will silently regress when dependencies are updated.

Spratima:
@matthewmcneely Understood, thank you for the clarification. The upstream fix is merged at 
https://github.com/dgraph-io/ristretto/pull/491 and will be picked up when the Ristretto dependency is bumped.

Issue #9791, opened 2026-07-27
Predicate moves: shrink the commit-abort window (two-phase bulk + delta move)


matthewmcneely:
## Current behavior

For the entire duration of a predicate move, Zero aborts every commit that touches the predicate: `movePredicate` calls `blockTablet` before streaming starts and releases it only when the move returns, and the oracle rejects blocked predicates at commit time (dgraph/cmd/zero/oracle.go: "Commits on predicate %s are blocked due to predicate move"). With multi-hour moves of large tablets, that is a multi-hour write outage on the predicate. A failed move also discards all streamed work; there is no resumability.

## Proposed direction

Two-phase move:

1. Bulk phase: stream the tablet at ts1 without blocking commits.
2. Delta phase: block commits, take ts2, stream only keys changed since ts1 (badger's stream framework supports `SinceTs`), then flip the tablet to the destination group.

The write-abort window shrinks from O(move duration) to O(delta), minutes instead of hours. An alternative or complement is moving the bulk data outside Raft entirely, the way snapshot streaming already transfers group data.

Progress visibility belongs in this redesign too. `MovePredicate` is a unary RPC (protos/pb.proto), so Zero is blind while a move runs. A pollable move-status RPC on the source Alpha would let Zero log progress and ETA, and eventually replace the wall-clock timeout with stall detection: cancel only when bytes sent stops advancing.

## Context

Surfaced while fixing #9784 (size-aware move timeout). Contained throughput work in the receive path is tracked in #9790; this issue covers the design-level availability fix.

Jira: DGR-309 (https://istari.atlassian.net/browse/DGR-309)

Issue #9790, opened 2026-07-27
Predicate moves: pipeline destination Raft proposals and batch KV writes


matthewmcneely:
## Current behavior

The destination side of a predicate move funnels all streamed data through the destination group's Raft, one chunk at a time. `batchAndProposeKeyValues` (worker/predicate_move.go) accumulates 32MB of KVs, then calls `proposeAndWait` and blocks until that chunk is replicated to quorum, written to the Raft WAL, and applied on the leader before the next chunk is proposed. The apply path (`populateKeyValues` -> `posting.TxnWriter`) opens one badger transaction per KV, so a 32MB chunk of small posting lists commits tens of thousands of individual managed transactions.

Net effect: every byte moved is written roughly 6x on the destination group (Raft WAL plus badger, times 3 replicas), and throughput is bounded by one quorum round trip per 32MB chunk. The report in #9784 works out to about 0.35 MB/s on-disk (20GB in roughly 16 hours).

## Proposed change

Two contained optimizations in the receive path:

1. Pipeline the proposals. After the initial `CleanPredicate` barrier, chunks are order-independent: badger's stream framework partitions key ranges so keys are disjoint, and every KV is written at the same version (`MoveTs`). Keep a bounded number of proposals in flight (4-8) and wait for all of them before acking EOF.
2. Batch the writes. All KVs in a move share the same commit timestamp, so `populateKeyValues` can write them through shared transactions or a write batch instead of one `CommitAt` per KV.

Plausibly a 4-8x improvement for changes confined to `batchAndProposeKeyValues` and `populateKeyValues`.

## Context

Surfaced while fixing #9784 (size-aware move timeout). That fix stops large moves from being cancelled but does not make them faster.

Jira: DGR-308 (https://istari.atlassian.net/browse/DGR-308)

PR #6388, opened 2020-09-03
fix(DGraph): Increased timeout for rebalancing alpha nodes from 20m to 2h

jgoodall:
DGraph zero always times out when it tries to rebalance across alpha nodes for very large nodes. This PR increases the timeout from 20 minutes to 2 hours.

---
This change is [Reviewable](https://reviewable.io/reviews/dgraph-io/dgraph/6388)

PR #9792, opened 2026-07-27
fix(zero): size-aware predicate move timeout and rebalancer backoff

### LINKED ISSUE #9784: Unblock large (>15GB) predicate moves by removing hardcoded 2-hour timeout
## Describe the bug

Moving predicates as small as 15GB to 20GB fails because Dgraph Zero hardcodes a strict 120-minute execution timeout in `dgraph/cmd/zero/tablet.go`. 

## To Reproduce

1. Set up a Dgraph cluster with 1 Zero and 3 Alphas (each in its own group).
2. Load a predicate containing ~15GB to 20GB+ of data.
3. Trigger a predicate move from one Alpha group to another.
4. Wait 120 minutes and inspect the Zero and Alpha logs.
5. Validate the move has been cancelled.

## Expected behavior

Operators should be able to configure the predicate move timeout via a Dgraph Zero CLI flag (e.g., `--predicate_move_timeout` or `--predicate-move-timeout`), or set it to `0` to disable the timeout entirely, allowing transfers of any size to complete successfully.

## Screenshots

N/A

## Environment

- OS: Linux RHEL 9 / Containerized
- Dgraph Version: v25.3.7
- Cluster Setup: 1 Zero, 3 Alphas (each in its own group)

## Additional context

- Code references:
  - [`dgraph/cmd/zero/tablet.go#L25`](https://github.com/dgraph-io/dgraph/blob/main/dgraph/cmd/zero/tablet.go#L25)
  - [`dgraph/cmd/zero/tablet.go#L121`](https://github.com/dgraph-io/dgraph/blob/main/dgraph/cmd/zero/tablet.go#L121)
- Historical discussion & references:
  - Discuss Thread 1: https://discuss.dgraph.io/t/support-setting-predicatemovetimeout-value/8513
  - Discuss Thread 2: https://discuss.dgraph.io/t/hardcoded-predicatemovetimeout-set-to-2-hours/14557
  - Previous related issue: [[#6388](https://github.com/dgraph-io/dgraph/commit/213f2b3fd3f52bc127d15f534d838b02f03c7f5e)](https://github.com/dgraph-io/dgraph/pull/6388)
  - Previous related commit: [`213f2b3`](https://github.com/dgraph-io/dgraph/commit/213f2b3fd3f52bc127d15f534d838b02f03c7f5e)

matthewmcneely:
## Summary

Zero cancels every predicate move at a hardcoded 120 minutes (`predicateMoveTimeout` in dgraph/cmd/zero/tablet.go), so moves of large tablets can never complete. The report in #9784 is a 20GB predicate that needed about 16 hours. This constant has been raised once before for the same reason, from 20 minutes to 2 hours in #6388, so any fixed wall-clock value is eventually wrong for someone's data size.

The failure mode is worse than a cancelled move. While a move runs, Zero aborts every commit on the predicate (`blockTablet`, enforced in oracle.go). On timeout, all streamed work is discarded, and the auto-rebalancer deterministically re-picks the same largest tablet every `--rebalance_interval` (default 8m). A cluster with one oversized predicate ends up in a permanent loop: two hours of write aborts and streaming, cancel, discard, repeat.

## Changes

- Scale the move timeout with tablet size. `moveTimeout()` grants at least size / 256 KiB-per-second, using the larger of on-disk and uncompressed bytes, with the previous 2h constant as the floor. Tablets under ~1.8GiB behave exactly as before; the 20GB tablet from #9784 gets ~23h.
- Back off after a failed move. A failure that ran at least a minute puts the tablet on a cooldown that doubles per consecutive failure (1h up to 24h) and is never shorter than the failed attempt itself. `chooseTablet` skips tablets on cooldown, a successful move clears the state, and manual moves via `/moveTablet` bypass the backoff.
- Log the computed timeout in the move announcement line so operators can see the budget each move was given.

A configurable flag was considered and dropped. A floor-style flag reads ambiguously against a "timeout" name, and a cap-style flag recreates the original bug; the size-derived value needs no configuration.

## Why 256 KiB/s

The receive path applies the whole stream through serial 32MB Raft proposals (`batchAndProposeKeyValues` in worker/predicate_move.go): each chunk is replicated to quorum, written to the Raft WAL, and fully applied before the next is proposed. Sustained rates in the field are therefore low; the #9784 report works out to ~0.35 MB/s on-disk. The floor rate is deliberately conservative. The throughput problem itself is out of scope here and tracked in #9790 (pipeline the receive path) and #9791 (two-phase move to shrink the commit-abort window).

## Testing

- New unit tests for the timeout computation and backoff bookkeeping in dgraph/cmd/zero/tablet_test.go (`TestMoveTimeout`, `TestMoveCooldown`, `TestMoveBackoff`).
- New end-to-end test: `TestLargePredicateMove` in systest/predicate-move runs the full scenario against a real two-group cluster. It loads several GiB of incompressible data (self-calibrating the size to the host's ingest rate so the move reliably outlasts the kill point), kills the destination Alpha mid-stream, asserts the failed move triggers the rebalancer backoff and leaves the source group intact, then restarts the Alpha, retries, and asserts the move completes with both announcements carrying the size-scaled timeout. It sits behind the `largemove` build tag, which no CI workflow compiles; run it manually per the package README (a run loads GiBs of data and takes 15+ minutes).
- Verified on GCE (c3-standard-8, Ubuntu 24.04, pd-ssd): PASS in 16m at the default MOVE_TEST_GB=8. Zero announced `timeout: 9h11m59s` for the 8.68GB tablet on both attempts, the mid-stream kill produced the expected failure plus the "Skipping automatic rebalancing" backoff line, and the retry completed with all 131,296 triples intact on the destination. Observed move throughput was ~19 MiB/s against 42-68 MiB/s raw ingest even with single-replica groups, which is the receive-path bottleneck tracked in #9790.
- `go build`, gofmt, and trunk pass.
- Existing move coverage (`systest/group-delete` TestNodes, `systest/integration2` TestUniqueMultipleGroups) moves small tablets, so the 2h floor applies and behavior there is unchanged.

Fixes #9784

---
<a href="https://app.blacksmith.sh/dgraph-io/codesmith/dgraph/pr/9792"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787764968&installation_model_id=8575&pr_number=9792&repository=dgraph-io%2Fdgraph&return_to=https%3A%2F%2Fgithub.com%2Fdgraph-io%2Fdgraph%2Fpull%2F9792&signature=e0c0612e5b196baf0194fe9aa23f39c51e3cde8a3eb1dbb9af389d3dfbcd54eb"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/dgraph/cmd/zero/tablet.go
+++ b/dgraph/cmd/zero/tablet.go
@@ -22,8 +22,33 @@ import (
 )
 
 const (
-	predicateMoveTimeout = 120 * time.Minute
+	// predicateMoveTimeout is the minimum time a predicate move may run before Zero cancels it.
+	// moveTimeout extends it for tablets too large to transfer within it at minMoveRate.
+	predicateMoveTimeout = 2 * time.Hour
+
+	// minMoveRate is the assumed floor throughput of a predicate move. It converts tablet size
+	// into time when computing the move timeout, so that a large tablet is not cancelled by a
+	// wall-clock limit it could never meet. The rate is deliberately conservative: the receiving
+	// group applies the whole stream through serial Raft proposals, so sustained rates are low.
+	minMoveRate = 256 << 10 // bytes per second
+
+	// moveFailureMinElapsed separates real move failures from quick validation ones: only
+	// attempts that ran at least this long feed the rebalancer's backoff.
+	moveFailureMinElapsed = time.Minute
+
+	// moveBackoffBase and moveBackoffMax bound the cooldown during which the automatic
+	// rebalancer skips a tablet whose last move failed. The cooldown doubles per consecutive
+	// failure, and is never shorter than the failed attempt itself.
+	moveBackoffBase = time.Hour
+	moveBackoffMax  = 24 * time.Hour
 )
+
+// moveTimeout returns how long a move of tab may run before Zero cancels it: at least base,
+// extended for large tablets assuming they transfer no faster than minMoveRate.
+func moveTimeout(base time.Duration, tab *pb.Tablet) time.Duration {
+	size := max(tab.OnDiskBytes, tab.UncompressedBytes)
+	return max(base, time.Duration(size/minMoveRate)*time.Second)
+}
 
 /*
 Steps to move predicate p from g1 to g2.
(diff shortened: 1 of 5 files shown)


shiva-istari (approved):

dgraph/cmd/zero/tablet.go
```diff
@@ -112,22 +137,34 @@ func (s *Server) MoveTablet(ctx context.Context, req *pb.MoveTabletRequest) (*pb
...
+
+	timeout := moveTimeout(predicateMoveTimeout, tab)
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	span := trace.SpanFromContext(ctx)
+	defer span.End()
 
 	// Ensure that I'm connected to the rest of the Zero group, and am the leader.
 	if _, err := s.latestMembershipState(ctx); err != nil {
```
xqqp, line 170: Should be `_, err = s.latestMembershipState(ctx)`, so the `err` variable is not shadowed and the error is recorded in the deferred statement (L159).

matthewmcneely: The shadow is harmless here: return errors.Wrapf(err, ...) assigns the wrapped error to the named result before any defers run, so the deferred recordMoveResult sees exactly what the return statement produced. The scenario that would really be a problem is a bare return, which would leave the named `err` nil and record a false success; every exit in movePredicate returns an explicit value, and any bare return added here later should be treated as a bug.

See https://go.dev/play/p/dOPXDjTQ4Py

Issue #9807, opened 2026-08-05
calculatedUids materializes the full uid list on the read path, ungated by the posting-list cache


matthewmcneely:
## Summary

`List.calculateUids()` materializes the complete uid set of a posting list as an
uncompressed `[]uint64` on the read path, inside `readFromDisk`, before the caller's
`ListOptions` are known. Three problems follow from doing it there:

1. It is not gated on the posting-list cache being enabled, so with
   `--cache percentage=0,...` the array is built, never stored, and discarded.
2. It short-circuits several `Uids()` optimizations that v24.1.2 relied on, including the
   `First` early stop and the compressed-intersection fast path.
3. Count-only queries pay for it and never read it.

Introduced in #9430. v24.1.2 has no equivalent path (`readFromDisk` there takes no
`readUids` argument).

Surfaced while triaging #9805, where an operator running `percentage=0,80,20` reported
higher `process_resident_memory_bytes` after upgrading v24.1.2 to v25.3.8.

## The path

`worker/task.go:858` (also `:2538` and `:2584`) calls `LocalCache.GetUids`, which threads
`readUids=true` down to `readFromDisk`:

```
worker/task.go:858        qs.cache.GetUids(key)
posting/lists.go:393      LocalCache.GetUids   -> getInternal(key, true, true)
posting/mvcc.go:850       getNew               -> MemLayerInstance.ReadData(..., readUids)
posting/mvcc.go:822       ReadData             -> readFromDisk(..., readUids)
posting/mvcc.go:795       readFromDisk         -> l.calculateUids()
posting/list.go:1726      calculateUids        -> l.iterate(committedUidsTime, 0, ...)
```

That `iterate` walks the whole list (mutable layer plus the packed immutable layer) with
`afterUid=0`, no count limit, and no intersect bounds, and appends every `Posting_REF` uid
into `res`.

## 1. Not gated on the posting-list cache

`initMemoryLayer` only builds a cache when `cacheSize > 0`
(`posting/mvcc.go:514`), and both `Cache.get` and `Cache.set` are nil-receiver-safe
(`posting/mvcc.go:348`, `posting/mvcc.go:365`). So when the posting-list cache share is
zero:

- `readFromCache` always misses and returns `nil`.
- `readFromDisk` still runs `calculateUids()`.
- `saveInCache` no-ops, so the array is dropped when the query's `LocalCache` goes away.

The materialization is a cache-warming optimization, but with the cache off there is
[...]

Issue #9795, opened 2026-07-29
[dgraph v25.3] mutations break reading earlier transaction


EnricoMi:
## Describe the bug

Mutations (insertions, updates, deletions) break reading earlier transactions.

## To Reproduce

1. Run [TestTransaction](https://github.com/G-Research/spark-dgraph-connector/blob/main/src/test/scala/uk/co/gresearch/spark/dgraph/connector/TestTransaction.scala#L93) of the [Spark-Dgraph-Connector](https://github.com/G-Research/spark-dgraph-connector) with Dgraph v25.3.0 or above (set env var `DGRAPH_TEST_CLUSTER_VERSION=25.3.0`).
2. Test fails

The test fetches the current transaction number and then mutates the graph:

deletion
```diff
-<12> <starring> <4> .
```

update
```diff
-<4> <name> "Princess Leia" .
+<4> <name> "Princess Leia Organa" .
```

insertion
```diff
+<13> <dgraph.type> "Person" .
+<13> <name> "Obi-Wan 'Ben' Kenobi" .
+<12> <starring> <13> .
```

The test then fetches all triples from the Dgraph cluster while setting the earlier transaction number. It expects to see only the triples that existed in that earlier transaction. However, it sees the deleted and inserted `starring` triples (though none of the other changes).

```diff
--- before-mutations.txt	2026-07-29 11:39:59.147153146 +0100
+++ after-mutations-with-transaction.txt	2026-07-29 11:39:59.244152222 +0100
@@ -8,8 +8,8 @@
 <12> <revenue> 7.75E8 .
 <12> <running_time> 121 .
 <12> <starring> <10> .
+<12> <starring> <13> .
 <12> <starring> <3> .
-<12> <starring> <4> .
 <12> <title> "Star Wars: Episode IV - A New Hope" .
 <12> <title@be> "Зорныя войны. Эпізод IV: Новая надзея" .
 <12> <title@br> "Star Wars Lodenn 4: Ur Spi Nevez" .
```

## Expected behavior

The graph of an earlier transaction should be consistent before and after mutations that happen after the transaction.

## Environment

- OS: Ubuntu
- Language: Java
- Version: v25.3.x

## Additional context

The [Spark-Dgraph-Connector](https://github.com/G-Research/spark-dgraph-connector) provides Spark read access to graph data stored in a Dgraph cluster. Transactions are vital for a Spark job to see a consistent graph while mutations might happen concurrently.

[...]

PR #9809, opened 2026-08-06
fix(query): avoid eager UID materialization on posting reads

gooohgb:
**Description**

Related  @matthewmcneely  : https://github.com/dgraph-io/dgraph/issues/9807

This PR avoids eagerly materializing the full UID slice for posting lists on read paths where that work is not useful or actively bypasses existing optimizations.

Changes:
- Gate `calculateUids()` behind the posting-list cache being enabled, so `--cache percentage=0,...` does not build and immediately discard a full `[]uint64`.
- Avoid using `calculatedUids` for bounded `Uids()` reads with `First` or `Intersect`, preserving early-stop and compressed-intersection paths.
- Reduce `Uids()` allocation size for bounded and small-intersect reads.
- Avoid `GetUids()` in worker paths that do not consume a full UID list, including count, scalar comparison, `has`, `uid_in`, facets, pagination, and intersect paths.
- Stop applying a negative `first` in `Uids()` at all, on both the memoized and the walk path. The
  worker post-filters an index read after `handleUidPostings` returns, so trimming there discarded
  rows the filter never saw; the query layer applies the count itself with `x.PageRange`.
- Clamp a negative `offset` in `uidReadFirst`, matching what `x.PageRange` does with one.

**Behavior change**

`first: -N` now returns different results for a query whose index is lossy, or whose per-token lists
are intersected, because the worker no longer trims a posting list ahead of the post-filter. The
previous answer was wrong rather than merely different:

```
name: string @index(term) .
0x1,0x2 name "great"    0x3,0x4,0x5 name "great wall"

q(func: eq(name, "great"), first: -2)
```

The term index is lossy, so the bucket for `"great"` is `{1,2,3,4,5}` and `handleCompareFunction`
is what narrows it to the two exact matches. Trimmed to the last two first, the bucket is `{4,5}`,
the filter drops both, and the query answers nothing. It now answers `0x1` and `0x2`.

A negative-`first` read also returns the whole list from the worker rather than the last N, so it
transfers more on a cold read. The equivalent problem for a positive `first` is not fixed here and
is tracked separately.

**Checklist**

- [x]  Code compiles correctly and linting (via trunk) passes locally

---
<a href="https://app.blacksmith.sh/dgraph-io/codesmith/dgraph/pr/9809"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788584060&installation_model_id=8575&pr_number=9809&repository=dgraph-io%2Fdgraph&return_to=https%3A%2F%2Fgithub.com%2Fdgraph-io%2Fdgraph%2Fpull%2F9809&signature=8ab957f5101da97830d9617f0b26f8aec6919cc1c38beaea89ef0fa18db416ed"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **Performance**
  - Improved posting-list reads by warming cached UIDs on demand and sharing the result across concurrent readers.
  - Optimized UID-only query processing and pagination handling.
  - Avoided unnecessary UID materialization for bounded, intersecting, or otherwise full-list queries.

- **Bug Fixes**
  - Corrected handling of negative pagination limits and offsets.
  - Preserved effective filtering with `AfterUid` during warmed and unwarmed reads.
  - Improved recovery after unsuccessful cache warming.

### DIFF
--- a/posting/list.go
+++ b/posting/list.go
@@ -1933,13 +1933,14 @@ func (l *List) Uids(opt ListOptions) (*pb.List, error) {
 		return out, nil, true
 	}
 
-	// Do The intersection here as it's optimized.
-	out, err, applyIntersectWith := getUidList()
-	if err != nil || !applyIntersectWith || opt.First == 0 {
+	// The bool reports whether the caller still has to intersect what came back and cut it down to
+	// First. A branch that has already intersected, or that needs neither, returns false.
+	out, err, postProcess := getUidList()
+	if err != nil || !postProcess {
 		return out, err
 	}
 
-	if opt.Intersect != nil && applyIntersectWith {
+	if opt.Intersect != nil {
 		algo.IntersectWith(out, opt.Intersect, out)
 	}
 
--- a/posting/list.go
+++ b/posting/list.go
@@ -1720,49 +1720,29 @@ func (l *List) approxLen() int {
 	return l.mutationMap.len() + codec.ApproxLen(l.plist.Pack)
 }
 
-func (l *List) calculateUids() error {
-	for {
-		l.RLock()
-		if l.mutationMap == nil || l.mutationMap.isUidsCalculated {
-			l.RUnlock()
-			return nil
-		}
-		calculatedAt := l.mutationMap.committedUidsTime
-		res := make([]uint64, 0, l.approxLen())
+func (l *List) calculateUids() (bool, error) {
+	l.Lock()
+	defer l.Unlock()
 
-		err := l.iterate(calculatedAt, 0, func(p *pb.Posting) error {
-			if p.PostingType == pb.Posting_REF {
-				res = append(res, p.Uid)
-			}
-			return nil
-		})
-
-		l.RUnlock()
-
-		if err != nil {
-			return err
-		}
-
-		l.Lock()
-		if l.mutationMap == nil || l.mutationMap.isUidsCalculated {
-			l.Unlock()
-			return nil
-		}
-		if l.mutationMap.currentEntries != nil {
-			l.Unlock()
-			return nil
-		}
-		if l.mutationMap.committedUidsTime != calculatedAt {
-			l.Unlock()
-			continue
-		}
-
-		l.mutationMap.calculatedUids = res
-		l.mutationMap.isUidsCalculated = true
-		l.Unlock()
-
-		return nil
+	if l.mutationMap == nil || l.mutationMap.isUidsCalculated ||
+		l.mutationMap.currentEntries != nil {
+		return false, nil
 	}
+
+	res := make([]uint64, 0, l.approxLen())
+	err := l.iterate(l.mutationMap.committedUidsTime, 0, func(p *pb.Posting) error {
+		if p.PostingType == pb.Posting_REF {
+			res = append(res, p.Uid)
+		}
+		return nil
+	})
+	if err != nil {
+		return false, err
+	}
+
+	l.mutationMap.calculatedUids = res
+	l.mutationMap.isUidsCalculated = true
+	return true, nil
 }
 
 // canUseCalculatedUids reports whether calculatedUids can serve a read at readTs. The slice is
--- a/posting/list.go
+++ b/posting/list.go
@@ -1784,12 +1784,13 @@ func (l *List) canUseCalculatedUids(readTs uint64) bool {
 // We have to apply the filtering before applying (offset, count).
 // WARNING: Calling this function just to get UIDs is expensive
 func (l *List) Uids(opt ListOptions) (*pb.List, error) {
+	requestedFirst := opt.First
 	if opt.First == 0 {
 		opt.First = math.MaxInt32
 	}
 
 	getUidList := func() (*pb.List, error, bool) {
-		if l.canUseCalculatedUids(opt.ReadTs) {
+		if opt.Intersect == nil && requestedFirst <= 0 && l.canUseCalculatedUids(opt.ReadTs) {
 			l.RLock()
 
 			afterIdx := 0
@@ -1816,8 +1817,6 @@ func (l *List) Uids(opt ListOptions) (*pb.List, error) {
 		// Pre-assign length to make it faster.
 		l.RLock()
 		defer l.RUnlock()
-		// Use approximate length for initial capacity.
-		res := make([]uint64, 0, l.ApproxLen())
 		out := &pb.List{}
 
 		if l.mutationMap.len() == 0 && opt.Intersect != nil && len(l.plist.Splits) == 0 {
@@ -1828,9 +1827,19 @@ func (l *List) Uids(opt ListOptions) (*pb.List, error) {
 			return out, nil, false
 		}
 
+		approxLen := l.ApproxLen()
+		resCap := approxLen
+		if opt.Intersect != nil && len(opt.Intersect.Uids) < resCap {
+			resCap = len(opt.Intersect.Uids)
+		}
+		if requestedFirst > 0 && requestedFirst < resCap {
+			resCap = requestedFirst
+		}
+		res := make([]uint64, 0, resCap)
+
 		// If we need to intersect and the number of elements are small, in that case it's better to
 		// just check each item is present or not.
-		if opt.Intersect != nil && len(opt.Intersect.Uids) < l.ApproxLen() {
+		if opt.Intersect != nil && len(opt.Intersect.Uids) < approxLen {
 			// Cache the iterator as it makes the search space smaller each time.
 			var pitr pIterator
 			for _, uid := range opt.Intersect.Uids {
--- a/posting/list.go
+++ b/posting/list.go
@@ -1721,9 +1721,18 @@ func (l *List) approxLen() int {
 }
 
 func (l *List) calculateUids() (bool, error) {
+	l.RLock()
+	skip := l.mutationMap == nil || l.mutationMap.isUidsCalculated ||
+		l.mutationMap.currentEntries != nil
+	l.RUnlock()
+	if skip {
+		return false, nil
+	}
+
 	l.Lock()
 	defer l.Unlock()
 
+	// Another reader may have warmed the list while this call waited for the write lock.
 	if l.mutationMap == nil || l.mutationMap.isUidsCalculated ||
 		l.mutationMap.currentEntries != nil {
 		return false, nil
--- a/posting/list.go
+++ b/posting/list.go
@@ -1883,7 +1883,7 @@ func (l *List) Uids(opt ListOptions) (*pb.List, error) {
 				}
 				res = append(res, p.Uid)
 
-				if opt.First != 0 && len(res) > opt.First {
+				if opt.First > 0 && len(res) > opt.First {
 					return ErrStopIteration
 				}
 			}
--- a/posting/list.go
+++ b/posting/list.go
@@ -1765,12 +1765,13 @@ func (l *List) canUseCalculatedUids(readTs uint64) bool {
 // We have to apply the filtering before applying (offset, count).
 // WARNING: Calling this function just to get UIDs is expensive
 func (l *List) Uids(opt ListOptions) (*pb.List, error) {
+	requestedFirst := opt.First
 	if opt.First == 0 {
 		opt.First = math.MaxInt32
 	}
 
 	getUidList := func() (*pb.List, error, bool) {
-		if l.canUseCalculatedUids(opt.ReadTs) {
+		if opt.Intersect == nil && requestedFirst <= 0 && l.canUseCalculatedUids(opt.ReadTs) {
 			l.RLock()
 
 			afterIdx := 0
@@ -1797,8 +1798,6 @@ func (l *List) Uids(opt ListOptions) (*pb.List, error) {
 		// Pre-assign length to make it faster.
 		l.RLock()
 		defer l.RUnlock()
-		// Use approximate length for initial capacity.
-		res := make([]uint64, 0, l.ApproxLen())
 		out := &pb.List{}
 
 		if l.mutationMap.len() == 0 && opt.Intersect != nil && len(l.plist.Splits) == 0 {
@@ -1809,9 +1808,19 @@ func (l *List) Uids(opt ListOptions) (*pb.List, error) {
 			return out, nil, false
 		}
 
+		approxLen := l.ApproxLen()
+		resCap := approxLen
+		if opt.Intersect != nil && len(opt.Intersect.Uids) < resCap {
+			resCap = len(opt.Intersect.Uids)
+		}
+		if requestedFirst > 0 && requestedFirst < resCap {
+			resCap = requestedFirst
+		}
+		res := make([]uint64, 0, resCap)
+
 		// If we need to intersect and the number of elements are small, in that case it's better to
 		// just check each item is present or not.
-		if opt.Intersect != nil && len(opt.Intersect.Uids) < l.ApproxLen() {
+		if opt.Intersect != nil && len(opt.Intersect.Uids) < approxLen {
 			// Cache the iterator as it makes the search space smaller each time.
 			var pitr pIterator
 			for _, uid := range opt.Intersect.Uids {
--- a/posting/list.go
+++ b/posting/list.go
@@ -1712,37 +1712,57 @@ func (l *List) rollup(readTs uint64, split bool) (*rollupOutput, error) {
 func (l *List) ApproxLen() int {
 	l.RLock()
 	defer l.RUnlock()
+	return l.approxLen()
+}
+
+func (l *List) approxLen() int {
+	l.AssertRLock()
 	return l.mutationMap.len() + codec.ApproxLen(l.plist.Pack)
 }
 
 func (l *List) calculateUids() error {
-	l.RLock()
-	if l.mutationMap == nil || l.mutationMap.isUidsCalculated {
-		l.RUnlock()
-		return nil
-	}
-	res := make([]uint64, 0, l.ApproxLen())
-
-	err := l.iterate(l.mutationMap.committedUidsTime, 0, func(p *pb.Posting) error {
-		if p.PostingType == pb.Posting_REF {
-			res = append(res, p.Uid)
+	for {
+		l.RLock()
+		if l.mutationMap == nil || l.mutationMap.isUidsCalculated {
+			l.RUnlock()
+			return nil
 		}
+		calculatedAt := l.mutationMap.committedUidsTime
+		res := make([]uint64, 0, l.approxLen())
+
+		err := l.iterate(calculatedAt, 0, func(p *pb.Posting) error {
+			if p.PostingType == pb.Posting_REF {
+				res = append(res, p.Uid)
+			}
+			return nil
+		})
+
+		l.RUnlock()
+
+		if err != nil {
+			return err
+		}
+
+		l.Lock()
+		if l.mutationMap == nil || l.mutationMap.isUidsCalculated {
+			l.Unlock()
+			return nil
+		}
+		if l.mutationMap.currentEntries != nil {
+			l.Unlock()
+			return nil
+		}
+		if l.mutationMap.committedUidsTime != calculatedAt {
+			l.Unlock()
+			continue
+		}
+
+		l.mutationMap.calculatedUids = res
+		l.mutationMap.isUidsCalculated = true
+		l.Unlock()
+
 		return nil
-	})
-
-	l.RUnlock()
-
-	if err != nil {
-		return err
 	}
-
-	l.Lock()
-	defer l.Unlock()
-
-	l.mutationMap.calculatedUids = res
-	l.mutationMap.isUidsCalculated = true
-
-	return nil
 }
 
 // canUseCalculatedUids reports whether calculatedUids can serve a read at readTs. The slice is
@@ -1766,12 +1786,13 @@ func (l *List) canUseCalculatedUids(readTs uint64) bool {
 // WARNING: Calling this function just to get UIDs is expensive
 func (l *List) Uids(opt ListOptions) (*pb.List, error) {
 	requestedFirst := opt.First
+	bounded := requestedFirst > 0 && requestedFirst < math.MaxInt32
 	if opt.First == 0 {
 		opt.First = math.MaxInt32
 	}
 
 	getUidList := func() (*pb.List, error, bool) {
-		if opt.Intersect == nil && requestedFirst <= 0 && l.canUseCalculatedUids(opt.ReadTs) {
+		if opt.Intersect == nil && !bounded && l.canUseCalculatedUids(opt.ReadTs) {
 			l.RLock()
 
 			afterIdx := 0
@@ -1793,7 +1814,7 @@ func (l *List) Uids(opt ListOptions) (*pb.List, error) {
 			out := &pb.List{Uids: copyArr}
 			l.RUnlock()
 
-			return out, nil, opt.Intersect != nil
+			return out, nil, false
 		}
 		// Pre-assign length to make it faster.
 		l.RLock()
@@ -1808,20 +1829,13 @@ func (l *List) Uids(opt ListOptions) (*pb.List, error) {
 			return out, nil, false
 		}
 
-		approxLen := l.ApproxLen()
-		resCap := approxLen
-		if opt.Intersect != nil && len(opt.Intersect.Uids) < resCap {
-			resCap = len(opt.Intersect.Uids)
-		}
-		if requestedFirst > 0 && requestedFirst < resCap {
-			resCap = requestedFirst
-		}
-		res := make([]uint64, 0, resCap)
+		approxLen := l.approxLen()
 
 		// If we need to intersect and the number of elements are small, in that case it's better to
 		// just check each item is present or not.
 		if opt.Intersect != nil && len(opt.Intersect.Uids) < approxLen {
 			// Cache the iterator as it makes the search space smaller each time.
+			res := make([]uint64, 0, len(opt.Intersect.Uids))
 			var pitr pIterator
 			for _, uid := range opt.Intersect.Uids {
 				ok, _, err := l.findPostingWithItr(opt.ReadTs, uid, pitr)
@@ -1844,6 +1858,12 @@ func (l *List) Uids(opt ListOptions) (*pb.List, error) {
 			uidMin = opt.Intersect.Uids[0]
 			uidMax = opt.Intersect.Uids[len(opt.Intersect.Uids)-1]
 		}
+
+		resCap := approxLen
+		if bounded && requestedFirst+1 < resCap {
+			resCap = requestedFirst + 1
+		}
+		res := make([]uint64, 0, resCap)
 
 		err := l.iterate(opt.ReadTs, opt.AfterUid, func(p *pb.Posting) error {
 			if p.PostingType == pb.Posting_REF {
--- a/posting/list.go
+++ b/posting/list.go
@@ -71,7 +71,7 @@ type List struct {
 	mutationMap  *MutableLayer
 	minTs        uint64       // commit timestamp of immutable layer, reject reads before this ts.
 	maxTs        uint64       // max commit timestamp seen for this list.
-	uidWarmState atomic.Int32 // 1 while a reader is materializing calculatedUids for this list.
+	uidWarmState atomic.Int32 // One of the uidWarm* states below. See tryStartUidWarm.
 
 	cache []byte
 }
@@ -1055,6 +1055,11 @@ func (l *List) setMutationAfterCommit(startTs, commitTs uint64, pl *pb.PostingLi
 	l.mutationMap.currentUids = nil
 	l.mutationMap.isUidsCalculated = false
 	l.mutationMap.calculatedUids = nil
+	// The list just changed, so a warm that failed against the old state may well succeed against
+	// this one. Reads reach a cached list through this path rather than through a fresh copy, since
+	// remove-on-update defaults to false, so without this a single failure would stick for the life
+	// of the cache entry.
+	l.uidWarmState.CompareAndSwap(uidWarmAbandoned, uidWarmIdle)
 
 	if pl.CommitTs != 0 {
 		l.maxTs = x.Max(l.maxTs, pl.CommitTs)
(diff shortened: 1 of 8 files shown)


gooohgb:
Hi, @matthewmcneely , could you please help review this PR? It’s affecting our production memory monitoring metrics and causing alerts. We’d really appreciate it if this could be reviewed and fixed as soon as possible. Thanks!

posting/list.go
```diff
@@ -1765,12 +1765,13 @@ func (l *List) canUseCalculatedUids(readTs uint64) bool {
...
 // WARNING: Calling this function just to get UIDs is expensive
 func (l *List) Uids(opt ListOptions) (*pb.List, error) {
+	requestedFirst := opt.First
 	if opt.First == 0 {
 		opt.First = math.MaxInt32
 	}
 
 	getUidList := func() (*pb.List, error, bool) {
-		if l.canUseCalculatedUids(opt.ReadTs) {
+		if opt.Intersect == nil && requestedFirst <= 0 && l.canUseCalculatedUids(opt.ReadTs) {
```
matthewmcneely: `requestedFirst > 0` won't distinguish bounded from unbounded here. The worker's sentinel for "no limit" is `math.MaxInt32`, not `0`:

- `calculatePaginationParams` in query/query.go starts with `count := math.MaxInt32` and returns it whenever there's a filter, an order, an excluded func, or `Params.Count == 0` (the common case).
- worker/task.go builds `opts` with `First: int(q.First + q.Offset)`, so a plain read arrives here with `opt.First == math.MaxInt32`.

So this branch is skipped for nearly every query that reaches `handleUidPostings`, and the `calculatedUids` work from #9795/#9801 stops being used at all. Note the normalization six lines up (`if opt.First == 0 { opt.First = math.MaxInt32 }`) asserts exactly the equivalence this gate contradicts.

Suggest:

```go
bounded := requestedFirst > 0 && requestedFirst < math.MaxInt32
if opt.Intersect == nil && !bounded && l.canUseCalculatedUids(opt.ReadTs) {
```

Separately: with `opt.Intersect == nil` now required to enter this branch, the `return out, nil, opt.Intersect != nil` at the bottom of it is always `false`. Worth returning `false` literally so the next reader doesn't have to prove the dead branch.

worker/task.go
```diff
@@ -778,6 +778,22 @@ func retrieveUidsAndFacets(args funcArgs, pl *posting.List, facetsTree *facetsTr
 	return uidList, fcsList, nil
 }
 
+func shouldPrecalculateUids(q *pb.Query, srcFn *functionContext, facetsTree *facetsTree,
+	opts posting.ListOptions) bool {
+	if q.DoCount || q.FacetParam != nil || facetsTree != nil {
+		return false
+	}
+	if opts.Intersect != nil || opts.First > 0 {
```
matthewmcneely: Same sentinel problem as the gate in `Uids()`: `opts.First` is `int(q.First + q.Offset)` and `q.First` defaults to `math.MaxInt32` for an unbounded query, so `opts.First > 0` is true almost always and this returns `false` for the common case.

Corroborating evidence that `MaxInt32` really is the default rather than `0`: `handleHasFunction` further down uses `len(result.Uids) >= int(q.First)` as its stop condition, which would fire immediately if `q.First` were ever `0` on an unbounded read.

```go
if opts.Intersect != nil || (opts.First > 0 && opts.First < math.MaxInt32) {
	return false
}
```

posting/mvcc.go
```diff
@@ -792,7 +796,7 @@ func (ml *MemoryLayer) readFromDisk(key []byte, pstore *badger.DB, readTs uint64
 	if err != nil {
 		return l, err
 	}
-	if readUids {
+	if readUids && ml.hasCache() {
```
matthewmcneely, line 809: The `hasCache()` gate itself is right. The interaction with the worker change is what concerns me: `calculateUids()` only ever runs on this disk path, never on the `readFromCache` hit path, and `MutableLayer.clone()` carries `isUidsCalculated`/`calculatedUids` through verbatim. So whichever query touches a key first decides whether the cached entry is warm:

1. `q(func: eq(name,"x"), first: 10)` now takes `qs.cache.Get(key)` → cache miss → `readFromDisk(readUids=false)` → `saveInCache` stores a list with `isUidsCalculated == false`.
2. Any later unbounded read of that same key → `GetUids` → cache **hit** → copy without calculated UIDs → full `iterate()`.

The entry stays cold until `maxTs < readTs` forces a re-read from disk. Before this PR `handleUidPostings` always called `GetUids`, so it was always warm. Either calculate on the cache-hit path in `ReadData` when `readUids` is set, or accept it with a comment saying so, but it shouldn't be load-order-dependent by accident.

posting/list.go
```diff
@@ -1809,9 +1808,19 @@ func (l *List) Uids(opt ListOptions) (*pb.List, error) {
 			return out, nil, false
 		}
 
+		approxLen := l.ApproxLen()
+		resCap := approxLen
+		if opt.Intersect != nil && len(opt.Intersect.Uids) < resCap {
+			resCap = len(opt.Intersect.Uids)
+		}
+		if requestedFirst > 0 && requestedFirst < resCap {
```
matthewmcneely: This cap works against the goal on two of the three paths below.

In the small-intersect branch, `res` accumulates every matching UID and the `First` truncation doesn't happen until the tail of `Uids()`, so capping at `requestedFirst` forces repeated regrowth whenever the intersection is larger than `First`.

In the iterate branch the stop check is `len(res) > opt.First` *after* the append, so `res` always reaches `First+1` and reallocs exactly once. The new test encodes that: `require.LessOrEqual(t, cap(first.Uids), 2)` for `First: 1` passes because the slice already doubled from 1 to 2.

Apply the cap only on the iterate path, and use `requestedFirst + 1`.

posting/list.go
```diff
@@ -1809,9 +1808,19 @@ func (l *List) Uids(opt ListOptions) (*pb.List, error) {
 			return out, nil, false
 		}
 
+		approxLen := l.ApproxLen()
```
matthewmcneely: Pre-existing, but you're rewriting the line so it's cheap to close out: `ApproxLen()` takes `l.RLock()`, and we already hold `RLock` from the `defer l.RUnlock()` a few lines up. Recursive `RLock` on `sync.RWMutex` deadlocks if a writer arrives between the two acquisitions. The diff already collapses two calls into one; an unlocked `approxLen()` helper (asserting `AssertRLock`) would finish the job.

worker/task.go
```diff
@@ -855,7 +871,13 @@ func (qs *queryState) handleUidPostings(
 			}
 
 			// Get or create the posting list for an entity, attribute combination.
-			pl, err := qs.cache.GetUids(key)
+			var pl *posting.List
+			var err error
+			if shouldPrecalculateUids(q, srcFn, facetsTree, opts) {
```
matthewmcneely: This is loop-invariant (`q`, `srcFn`, `facetsTree`, and `opts` are all fixed for the call), but it's evaluated once per UID inside the inner loop across `numGo` goroutines. Hoist it above `calculate` and capture the bool.

posting/list_test.go
```diff
@@ -1874,3 +1874,33 @@ func TestCalculatedUidsRespectReadTs(t *testing.T) {
...
+	require.True(t, l.canUseCalculatedUids(10))
+
+	l.Lock()
+	l.mutationMap.calculatedUids = []uint64{100, 101}
+	l.Unlock()
+
+	first, err := l.Uids(ListOptions{ReadTs: 10, First: 1})
+	require.NoError(t, err)
+	require.Equal(t, []uint64{2}, first.Uids)
+	require.LessOrEqual(t, cap(first.Uids), 2)
```
matthewmcneely: This assertion is passing because the slice regrew from cap 1 to cap 2, which is the allocation behavior flagged on the `resCap` line rather than the intended outcome.

More importantly, this test only covers the skip cases. The gap that lets the `math.MaxInt32` bug through is the missing positive case: an unbounded read (`First: 0`, no `Intersect`) should still consume `calculatedUids`, and there should be a sibling case for `First: math.MaxInt32` asserting the same thing, since that's what the worker actually sends. The existing sentinel trick (`l.mutationMap.calculatedUids = []uint64{100, 101}`) inverts nicely for both.

matthewmcneely (changes requested):

gooohgb:
Hi @matthewmcneely , thank you for the review. I’ve updated the implementation based on your feedback. Could you please take another look when you have a chance?

posting/mvcc.go
```diff
@@ -761,19 +765,24 @@ func (c *CachePL) Set(l *List, readTs uint64) {
...
 
-func (ml *MemoryLayer) readFromCache(key []byte, readTs uint64) *List {
+func (ml *MemoryLayer) readFromCache(key []byte, readTs uint64, readUids bool) (*List, error) {
 	cacheItem, ok := ml.cache.get(key)
 
 	// Issue #9597 fix: Cache is only valid if minTs <= readTs AND maxTs >= readTs.
 	// If maxTs < readTs, the cache is missing mutations committed after maxTs.
 	if ok && cacheItem.list != nil && cacheItem.list.minTs <= readTs && cacheItem.list.maxTs >= readTs {
+		if readUids && ml.hasCache() {
+			if err := cacheItem.list.calculateUids(); err != nil {
```
matthewmcneely: This races. `cacheItem.list` is the object stored in ristretto and shared by every concurrent reader, and `calculateUids()` iterates while holding only `RLock`. `List.iterate` → `pickPostings` → `MutableLayer.iterate` → `populateDeleteAll` **writes** `mm.deleteAllMarker` at posting/list.go:263. Two readers can hold `RLock` on the same shared list at once, so that write is unsynchronized.

Before this PR the cached object was never iterated: `readFromCache` only took `RLock` for `copyList`, and every reader iterated its own copy. This change is what puts a writer on the shared object.

It is not a corner case. `populateDeleteAll` only takes the writing path when `mm.deleteAllMarker == math.MaxUint64`, which is the state of any list whose mutable layer is empty, so every rolled-up posting list qualifies.

Reproduced on `f647955` with a rolled-up key, 16 goroutines calling `getNew(key, pstore, readTs, true)` at a `readTs` that satisfies the cache-hit guard:

```
WARNING: DATA RACE
Read at 0x00c000325438 by goroutine 70:
  posting.(*MutableLayer).populateDeleteAll()  posting/list.go:244
  posting.(*MutableLayer).iterate()           posting/list.go:292
  posting.(*List).pickPostings()              posting/list.go:1096
  posting.(*List).iterate()                   posting/list.go:1120
  posting.(*List).calculateUids()             posting/list.go:1733
  posting.(*MemoryLayer).readFromCache()      posting/mvcc.go:775

Previous write at 0x00c000325438 by goroutine 72:
  posting.(*MutableLayer).populateDeleteAll()  posting/list.go:263
  ... same stack ...
```

The value written happens to be the same from every goroutine, so this is unlikely to corrupt results today, but it is still a race the detector will fail on and it will bite the moment anything else memoizes on the shared layer.

Warming the cached entry is still the right goal. It needs exclusive ownership: hold `l.Lock()` across the whole compute for this call path, or warm the entry at `saveInCache` time while the list is still privately owned, rather than mutating a published object under `RLock`.

Minor, same line: the `ml.hasCache()` guard is redundant here. Reaching this point already required `ml.cache.get` to succeed, which can't happen on a nil cache.

posting/mvcc.go
```diff
@@ -761,19 +765,24 @@ func (c *CachePL) Set(l *List, readTs uint64) {
...
 }
 
-func (ml *MemoryLayer) readFromCache(key []byte, readTs uint64) *List {
+func (ml *MemoryLayer) readFromCache(key []byte, readTs uint64, readUids bool) (*List, error) {
 	cacheItem, ok := ml.cache.get(key)
 
 	// Issue #9597 fix: Cache is only valid if minTs <= readTs AND maxTs >= readTs.
 	// If maxTs < readTs, the cache is missing mutations committed after maxTs.
 	if ok && cacheItem.list != nil && cacheItem.list.minTs <= readTs && cacheItem.list.maxTs >= readTs {
+		if readUids && ml.hasCache() {
```
matthewmcneely: Separate from the race: warming the cached list in place grows `calculatedUids` on an entry whose ristretto cost was fixed at `saveInCache` time. `ApproximateSize()` counts `cap(m.calculatedUids) * 8` (posting/size.go:105), so after warming the recorded cost understates the real footprint by 8 bytes per UID, for every warmed entry.

That quietly breaks `--cache percentage`, since `MaxCost` no longer bounds what the posting-list cache actually holds. Re-`set` the entry after warming so the cost is recomputed, or move the warming to before the `set`.

posting/list.go
```diff
@@ -1712,37 +1712,57 @@ func (l *List) rollup(readTs uint64, split bool) (*rollupOutput, error) {
...
+		l.Lock()
+		if l.mutationMap == nil || l.mutationMap.isUidsCalculated {
+			l.Unlock()
+			return nil
+		}
+		if l.mutationMap.currentEntries != nil {
+			l.Unlock()
+			return nil
+		}
+		if l.mutationMap.committedUidsTime != calculatedAt {
```
matthewmcneely: Unbounded retry: every `continue` redoes a full `iterate` over the list. On a hot, frequently-committed key that can spin, and now that this runs on the shared cached object it does so under lock contention.

Since the next reader recomputes anyway and `canUseCalculatedUids` already refuses a stale slice, a plain `return nil` here is safer than looping. If you'd rather keep retrying, cap it.

While you're in here: the `currentEntries != nil` bail-out just above only runs after the full iterate has already been paid for. It's cheap to check alongside `isUidsCalculated` in the `RLock` section at the top.

posting/mvcc_test.go
```diff
@@ -273,6 +273,64 @@ func TestCacheStaleWhenMaxTsLessThanReadTs(t *testing.T) {
...
+	l, err := getNew(key, pstore, math.MaxUint64, false)
+	require.NoError(t, err)
+	require.False(t, l.mutationMap.isUidsCalculated)
+	MemLayerInstance.wait()
+
+	cacheItem, ok := MemLayerInstance.cache.get(key)
+	require.True(t, ok)
+	require.False(t, cacheItem.list.mutationMap.isUidsCalculated)
+
+	l, err = getNew(key, pstore, math.MaxUint64, true)
```
matthewmcneely: This test doesn't reach the code it's covering. `readFromCache` requires `cacheItem.list.maxTs >= readTs`, and `maxTs` here is 4 (the last commit ts from `addEdgeToUID`), so `readTs = math.MaxUint64` can never hit. Both assertions are satisfied by `readFromDisk` + `saveInCache` replacing the entry, exactly as they were before the `readFromCache` change.

Measured on this commit:

```
readFromCache(MaxUint64) returned nil? true
readFromCache(maxTs=4)   returned nil? false
```

Use the cached entry's `maxTs` as the readTs for the second `getNew`. That also gives you the setup for a concurrent variant, which is what would have caught the race on the line above.

worker/precalculate_uids_test.go
```diff
@@ -0,0 +1,90 @@
+/*
+ * SPDX-FileCopyrightText: 2017-2026 Istari Digital, Inc.
```
matthewmcneely: Missing the `©`. Repo convention is `SPDX-FileCopyrightText: © 2017-2026 Istari Digital, Inc.` (568 files); this is the only one without it.

matthewmcneely (changes requested):

gooohgb:
Hi @matthewmcneely, thank you again for the detailed review.

I have updated the implementation based on your feedback. While validating the changes, I also found and fixed a regression affecting negative pagination (`first: -N`).

The CI checks are currently stuck. Could you please take another look at the updated implementation when you have a chance? Thank you!

gooohgb:
HI, @matthewmcneely , any updates here?

posting/mvcc.go
```diff
@@ -761,19 +765,29 @@ func (c *CachePL) Set(l *List, readTs uint64) {
...
 
-func (ml *MemoryLayer) readFromCache(key []byte, readTs uint64) *List {
+func (ml *MemoryLayer) readFromCache(key []byte, readTs uint64, readUids bool) (*List, error) {
 	cacheItem, ok := ml.cache.get(key)
 
 	// Issue #9597 fix: Cache is only valid if minTs <= readTs AND maxTs >= readTs.
 	// If maxTs < readTs, the cache is missing mutations committed after maxTs.
 	if ok && cacheItem.list != nil && cacheItem.list.minTs <= readTs && cacheItem.list.maxTs >= readTs {
+		if readUids {
+			calculated, err := cacheItem.list.calculateUids()
```
matthewmcneely, line 775: Everything from both rounds looks addressed, and the concurrent test has real teeth: lifted onto `f647955` it reproduces the `populateDeleteAll` race on the first run, and it passes here. Thanks for the thorough turnaround.

One follow-on thought on this line, and I don't think it needs to block the merge. Taking `l.Lock()` does close the race, and warming the cached entry is still the right goal. The cost is that every reader of this key now waits on work that exists purely as an optimization.

Three things combine into a stall that recurs on a key that is both read-hot and write-hot:

- `setMutationAfterCommit` clears the warm state at the end of its body (`posting/list.go:1054-1055`: `isUidsCalculated = false`, `calculatedUids = nil`).
- `updateItemInCache` applies each commit in place on the published, shared list instead of evicting it (the `p.Pack == nil` branch calling `val.list.setMutationAfterCommit`).
- That in-place path is the default: `CacheDefaults` in `worker/server_state.go:44` ships `remove-on-update=false`.

So every commit disarms the flag and the next wave of readers re-warms under the write lock, for the life of the entry.

Measured on this commit, 1M-UID list, 64 concurrent readers:

```
single uncontended warm:                       7.98ms
64 readers, already warm:  p50=0s      max=78µs    wall=171µs
64 readers, post-commit:   p50=7.53ms  max=7.56ms  wall=7.57ms
plain reader (readUids=false) during a warm:   5.48ms
```

Two things stand out. Every reader in the cohort pays roughly the full warm cost, not just the one doing the work. And a reader that never asked for UIDs still stalls 5.48ms, because Go's `RWMutex` blocks new `RLock` acquisitions once a writer is pending, so the warm blocks readers that only want `copyList`.

End to end (`GetUids` plus `Uids()`, what `handleUidPostings` does per key), 9 rounds each against the merge base:

|              | per-reader p50 (median of rounds) | worst round p50 | wall (median) |
| ------------ | --------------------------------- | --------------- | ------------- |
| merge base   | 19ms                              | 77ms            | 85ms          |
| this commit  | 30ms                              | 156ms           | 36ms          |

Total work drops, which is why median wall time improves. What it gives up is independence: within each round on this commit, p50, p95, max and wall collapse to the same number (`10/15/15/15`, `28/29/29/30`), which is the lock-convoy signature. Every in-flight query touching the predicate is yoked to one warm, so a GC pause or a large split list gets broadcast to the whole cohort instead of being absorbed by one unlucky reader.

Nothing here depends on the warm for correctness, so there may be no reason to make a reader wait on it at all. One option, if you like the shape: keep it best-effort, where one caller commits to the work and everyone else serves the read unwarmed and recomputes as it does today.

```go
// posting/list.go
func (l *List) calculateUids() (bool, error) {
	l.RLock()
	skip := l.mutationMap == nil || l.mutationMap.isUidsCalculated ||
		l.mutationMap.currentEntries != nil
	l.RUnlock()
	if skip {
		return false, nil
	}

	// Warming is an optimization, never a correctness requirement, so a reader must not
	// queue behind another goroutine's warm. Exactly one caller takes the write lock; the
	// rest serve the read unwarmed.
	if !atomic.CompareAndSwapInt32(&l.warming, 0, 1) {
		return false, nil
	}
	defer atomic.StoreInt32(&l.warming, 0)

	l.Lock()
	defer l.Unlock()
	// ... unchanged from here
}
```

with a `warming int32` on `List` (and the matching word in `ApproximateSize()`). `x.SafeMutex` has no `TryLock` today, `x/lock.go` exposes only `Lock`/`RLock` and the assertions, so a CAS guard is less surgery than adding one. It also has the better failure mode: `TryLock` would keep losing to `RLock` holders on a busy key and might never warm at all, where the CAS still guarantees the warm happens.

That keeps the win this change is after, one materialization instead of N, and drops the queue. The single winner does still hold the write lock across its `iterate`, so one warm-length window where `RLock` waiters stall remains. If you want that gone too, compute the slice on the copy `readFromCache` already makes (`deleteAllMarker` lives on the clone, so iterating it is safe) and publish under a short write lock, dropping the result if `committedUidsTime` moved in the meantime.

For scope: the memory win behind #9807 comes from the `hasCache()` gate on the disk path and is unaffected by any of this.

The numbers above are a synthetic single-key microbenchmark on macOS at `GOMAXPROCS=12`, GC-noisy at roughly 500MB allocated per round, one unsplit list, no query pipeline, so treat them as directional. The recurrence in the three bullets is straight from the code and holds regardless.

Either way, happy for this to land as a follow-up if you'd rather keep the PR to the change it set out to make. Raising it now mostly so the trade-off is on the record before it ships.

gooohgb: Hi @matthewmcneely, thanks for the detailed follow-up suggestion.

I implemented the non-blocking cache-warming approach in a separate incremental PR:

https://github.com/gooohgb/dgraph/pull/1/changes

The new implementation elects a single warmer with CAS, calculates UIDs on the
private copy created by readFromCache, and publishes the result under a short
write lock only if committedUidsTime has not changed.

I kept the current PR unchanged so you can decide whether to merge #9809 first
and take this as a follow-up, or include this change before merging.

Could you please take another look when you have time? Thanks!

gooohgb:
Hi @matthewmcneely, just following up on #9809. Have you had a chance to review the latest changes and the incremental non-blocking cache-warming follow-up?

Please let me know whether you would prefer to merge the current PR first and handle the follow-up separately, or include it before merging. I’m happy to make any additional changes if needed.

Thanks again for your time and review!

matthewmcneely:
Thanks for the patience, and for the turnaround on both earlier rounds — everything from them is addressed, and the concurrent test has real teeth.

On your merge-order question: I'd fold the non-blocking warm in **before** merging. When I went looking for how bad the convoy actually gets, it turned out to reach further than the readers of one key. `calculateUids` holds the published list's write lock across the walk, and the commit path takes that same lock on the serial Raft apply loop, one step ahead of the `ProcessDelta` that releases waiting reads:

```
commitOrAbort (worker/draft.go:966)
  -> txn.UpdateCachedKeys        (draft.go:1019)
    -> updateItemInCache -> List.setMutationAfterCommit -> l.Lock()
  -> posting.Oracle().ProcessDelta  (draft.go:1022)
```

So a slow warm — a large split list is the bad case, since `iterate` reads every part from Badger under that lock — stalls commits for the whole group, not only the queries touching that predicate. That moved it from "worth doing next" to "worth doing first" for me.

Rather than leave you to do it a third time, I've built on your gooohgb/dgraph#1 and opened it against your branch: **gooohgb/dgraph#2**. Merging it into `fix-calculated-uids-materialization` folds it into this PR. It's three independent commits — take, change, or drop any of them:

1. **Non-blocking warm.** Your design from #1 (CAS-elected warmer, walk on the private copy, publish under a short write lock with a `committedUidsTime` recheck), with three changes: `needsUidWarm()` is checked on the published list under the read lock `readFromCache` already holds, because `clone()` never copies `currentEntries` (posting/list.go:133-148) so your `lCopy` check was always true; `uidWarmState` is realigned, since as written it made `gofmt -l` flag `posting/list.go` and `trunk check` would have failed; and the walk is factored into `warmCachedUids` with the ownership contract written on `calculateUids`, so it doesn't drift back onto a shared list later. Folded into the same commit: a warm failure now logs and serves the list unmaterialized instead of turning a cache hit into a query error, and the ristretto re-set skips an entry that a rollup dropped mid-warm rather than resurrecting it.

2. **Negative-first tail copy.** Your pagination fix is what makes this reachable — the walk now materializes the whole list, and returning the last N as a view pinned all of it. 8MB retained for ten uids on a million-uid list.

3. **`uid_in` call site.** worker/task.go:971 still builds `First` from `int(q.First + q.Offset)`, the arithmetic `uidReadFirst` replaces. Latent today, but the invariant should live in one place.

The new lock test deadlocks against the current behavior — it fails in 5s if `warmCachedUids` is pointed back at the shared list — so it holds the property rather than just describing it. `go test ./posting/ -race` is green in full, as is `./worker/`, and `gofmt`/`go vet` are clean.

Have a look when you get a chance. If you'd rather keep #9809 to the change it set out to make, I'm fine taking any of the three separately instead — say which and I'll move them out.

gooohgb:
Hi @matthewmcneely, the changes look great. I’ve merged them into the PR branch. Thank you for your work on this!

matthewmcneely:
Thanks for merging those. I went back over the current state properly and found six things, so there's another PR against your branch: **gooohgb/dgraph#3**. Merging it into `fix-calculated-uids-materialization` folds it into this PR. Three of the six are cleanup after my own last round.

The one worth your attention is that **my first finding was pointing the wrong way, and I only caught it by trying hard to disprove it.**

I'd reported that the memoized read path should trim a negative `first` the way the walk path does, so the two agree. Aligning them that way would have broken a query that works today:

```
name: string @index(term) .
0x1,0x2 name "great"    0x3,0x4,0x5 name "great wall"

q(func: eq(name, "great"), first: -2)
```

The term index is lossy, so the bucket for `"great"` is `{1,2,3,4,5}`, and `handleCompareFunction` is what narrows it to the two real matches — and it runs in `helpProcessTask` *after* `handleUidPostings` has returned. Trim the bucket to `{4,5}` first and the filter drops both, so the query answers nothing. `x.PageRange` behind it can't put back what the worker already discarded. `allof` is the same shape through intersection instead of post-filtering, since it isn't in `calculatePaginationParams`' exclusion list and `needsIntersect` matches it.

So the warm answer was the correct one all along. The trim comes out instead of being extended, which also fixes the cold path — wrong since `441d3033` — and makes warm and cold agree everywhere across a 28-case matrix. The positive-`first` early stop stays: it's paired with a real saving, and its soundness rests on the exclusion list, which had no test until now.

The other five:

- **`refactor`** — `opt.First` is normalized to `MaxInt32` at the top of `Uids` and never reassigned, so the `|| opt.First == 0` bail-out and the `&& applyIntersectWith` below it are both dead.
- **`perf`** — the warm give-up you asked about, plus a correction: I said it would self-heal when the cache entry was replaced. It wouldn't. `remove-on-update` defaults to `false`, so an ordinary commit applies in place and keeps the entry, and a single failure would have stuck for the life of it. `setMutationAfterCommit` now lifts the give-up.
- **`fix(worker)`** — `uidReadFirst` is mine from last round and adds `offset` unclamped. `offset` is parsed with no lower bound, so `first: 10, offset: -1` pushes down 9 and returns a uid short; `-100` drops the bound entirely.
- **`test`** — my slice-identity assertion used `require.Equal` on two `*uint64`, which `reflect.DeepEqual` resolves by following the pointers, so it passed for two distinct arrays. It was asserting nothing.
- **`docs`** — my comment claimed iterating the private copy is safe because `setMutationAfterCommit` replaces the shared maps. That's true of the one production call site, not of the function: `refresh=false` writes both in place, and that's a fatal concurrent map access, not a soft race.

Three things I left open, listed in the PR body: a dropped publish still re-arms the warm, so a key that's both read-hot and write-hot can loop on a walk that gets thrown away (mine, from last round, and I under-described that trade — it's bounded to one walk in flight per key, not one per read); the pushdown is still unsound for a *positive* `first` on `allof`, `eq` over a lossy-only index, and the geo family, which is pre-existing and wants its own change; and I didn't add a cluster test for the `eq` case above, so that chain is established by reading the code, not by a test.

Have a look when you get a chance. Sorry for the extra round — better than shipping the first version of finding 1.

gooohgb:
I took a look at fixing positive `first` pushdown for lossy `eq`, but it touches a broader set of query-planning and post-filtering paths than expected, so I agree that it should be handled separately.

I’ve reviewed and merged #3. The overall approach and test coverage look great. Thank you for the careful follow-up and for putting all of this together.

matthewmcneely:
Final pass on `2078fd69` — everything from all four rounds is closed, and the merge of #3 was clean (its tree is byte-identical to the branch tip, so it resolved nothing of its own). CI is green across the full matrix, `main`'s six new commits don't touch any file this PR touches, and I re-verified build/`gofmt`/`vet`/tests on the merged tree under Go 1.27.0, which is where `main` has since moved.

Two things done:

**Description updated**, additively — your text and the footer are untouched, I just added the two changes that arrived late and a short "Behavior change" section. The reason: as written it described only the memory work, and this PR now also changes query *results*. `first: -N` over a lossy index answers differently than before (correctly, where it was wrong), and someone writing release notes or bisecting a result change later would have had nothing to go on. Revert it freely if you'd rather word it yourself.

**The two follow-ups are filed:**

- #9829 — a dropped UID-warm publish re-arms immediately, so a read-hot and write-hot key can loop on a discarded walk. This one is mine, from the non-blocking warm, and I under-described the trade at the time: it buys "no commit stall" at the cost of this. It's bounded to one walk in flight per key rather than one per read, and reads stay correct throughout. The issue writes up the two obvious fixes and why neither works as-is.
- #9830 — the positive `first` pushdown is unsound for `allof`, `eq` over a lossy-only index, and the geo family, for the same post-filter-ordering reason. I quoted your note about it wanting its own change.

From my side this is ready to merge.

matthewmcneely (approved):

posting/list.go
```diff
@@ -1731,37 +1738,98 @@ func (l *List) rollup(readTs uint64, split bool) (*rollupOutput, error) {
...
+	l.uidWarmState.CompareAndSwap(uidWarmRunning, uidWarmIdle)
+}
+
+// abandonUidWarm gives up on materializing l's uids, so that no later reader retries it. A walk
+// fails when the list itself cannot be read -- a multi-part list with an unreadable split, say --
+// and retrying costs another full walk plus a Badger read per split, for a slice that reads do not
+// need. setMutationAfterCommit lifts this on the next commit to the list, so a failure suppresses
+// one attempt per state rather than permanently.
+func (l *List) abandonUidWarm() {
+	l.uidWarmState.Store(uidWarmAbandoned)
```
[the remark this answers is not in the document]
matthewmcneely: Valid, and it's my code. Confirmed the mechanism: during the walk the state is `uidWarmRunning`, so a concurrent commit's `CompareAndSwap(uidWarmAbandoned, uidWarmIdle)` in `setMutationAfterCommit` matches nothing, and `abandonUidWarm` then stores `uidWarmAbandoned` against a `committedUidsTime` we never actually tested. If no further commit arrives, the list stays parked for the life of the cache entry and every read walks in `Uids()` as it did before the memoization existed.

Bounded severity — it needs the walk to error at all, which means a Badger read failure on a split part, and the consequence is a lost optimization rather than a wrong answer. Worth noting the error is usually structural, so a re-walk at `T2` would fail the same way; the case your fix protects is the transient one, which is exactly the case my own comment on that branch claims to care about ("a transient error need not recur"). So the inconsistency is real.

Your proposed shape is right, and I initially mis-reasoned about it: guarding the park on the timestamp doesn't open a retry loop, because a walk that fails at `T2` parks against `T2`. Worst case is one extra walk per commit that happens to land mid-walk, which is the same bound `doRollup` gets from its own per-key dedupe.

I'm folding it into #9829 rather than patching it here. That issue covers the sibling defect on the other exit from this function — a *dropped* publish (`publishCalculatedUids` rejecting a stale result) re-arms unconditionally, which is the case that does loop — and both want the same thing: a park keyed on the state that was actually proven bad. Fixing one exit and not the other would leave the state machine half-guarded. Cross-referenced there with credit.

posting/list.go
```diff
@@ -1879,24 +1956,25 @@ func (l *List) Uids(opt ListOptions) (*pb.List, error) {
 		return out, nil, true
 	}
 
-	// Do The intersection here as it's optimized.
-	out, err, applyIntersectWith := getUidList()
-	if err != nil || !applyIntersectWith || opt.First == 0 {
+	// The bool reports whether the caller still has to intersect what came back and cut it down to
+	// First. A branch that has already intersected, or that needs neither, returns false.
+	out, err, postProcess := getUidList()
+	if err != nil || !postProcess {
```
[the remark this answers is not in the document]
matthewmcneely: The observation is mechanically correct — both the `IntersectCompressedWith` branch and the small-intersect branch return `false` and neither applies `First`, so `Uids` can return more than `First` uids. Two things stop it being a bug to fix here, though.

**It is unreachable, and pre-existing regardless.** Both `return out, nil, false` lines are unchanged from `64804bd2`, the merge base — this PR only renamed the variable and deleted a dead `opt.First == 0` term next to them. And a bounded positive `First` never coexists with a non-nil `Intersect`. `opts.Intersect` is only set when `srcFn.fnType != notAFunction && q.UidList` is non-empty, which leaves:

- a root function — `SrcUIDs` is nil at root, so no `UidList`;
- a filter SubGraph — `filterCopy` (query/query.go) never sets `Params.Count`, so `calculatePaginationParams` returns `math.MaxInt32`;
- a child carrying `first:` — `SrcFunc` is nil, so `notAFunction`, so no `Intersect`;
- a child carrying a function — `treeCopy` only calls `createSrcFunction` for an aggregator or `checkpwd`, and `needsValuePostings` routes both to `handleValuePostings`, which never calls `Uids`;
- root `func: uid(...)`, the one shape that looked live, since it populates `SrcUIDs` *and* maps to `standardFn` — but `ProcessGraph` resolves it locally (`DestUIDs = SrcUIDs`) with no worker call at all, as does a `uid()` filter.

**And the proposed fix would introduce a correctness bug.** Letting the truncation run on those branches means a bounded intersect result gets cut before the worker's own post-filter pass, which runs later in `helpProcessTask`: `handleCompareFunction` re-checks real values for a lossy tokenizer, `filterGeoFunction` re-checks real geometry. Rows dropped by the truncation are rows that filter never sees, and `x.PageRange` cannot restore them. That is exactly the defect just filed as #9830, and applying this diff would extend it to two more paths.

So the right resolution is the opposite of truncating: those branches don't apply `First` on purpose, and the invariant that keeps them safe is that no caller sends a bounded `First` with an `Intersect`. I've recorded this on #9830 as a trap for whoever fixes the positive-`first` pushdown, since the naive fix there is the same diff. Thanks — the pointer is genuinely useful even though the conclusion inverts.

PR #9812, opened 2026-08-20
fix(acl): preserve user permissions across namespaces

gooohgb:
**Description**

`AclCache.Update` clears cached permissions belonging to the namespace being
refreshed. However, when writing the new `userPredPerms`, it replaced the
user's entire predicate-permission map.

User IDs are namespace-local, so the same user ID can exist in multiple
namespaces. Refreshing ACLs for one namespace could therefore discard that
user's permissions from every other namespace.

Because `RefreshACLs` iterates over a Go map of namespaces, the last namespace
loaded could differ between Alpha instances and restarts. This could leave
different Alpha instances with different allowed-predicate sets and affect
predicate discovery paths such as `expand(_all_)`.

This change preserves the existing per-user map and merges the newly refreshed
namespaced predicates into it after removing only the permissions belonging to
the namespace being refreshed.

Regression tests cover:

- Loading the same user ID in both namespace orders.
- Refreshing permissions in one namespace without affecting another.
- Clearing permissions in one namespace while preserving another namespace.

This addresses the same-user collision not covered by the different-user
multi-namespace scenario added in #8418.

**Checklist**

- [x] The PR title follows the Conventional Commits syntax.
- [x] Regression tests for the bug have been added.
- [ ] Code compiles correctly and linting via trunk passes locally.

---
<a href="https://app.blacksmith.sh/dgraph-io/codesmith/dgraph/pr/9812"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789813502&installation_model_id=8575&pr_number=9812&repository=dgraph-io%2Fdgraph&return_to=https%3A%2F%2Fgithub.com%2Fdgraph-io%2Fdgraph%2Fpull%2F9812&signature=21ad1c1984c00a7a3b1e0ea1ad39609409c7910de9ca65e81c63ceeb80ed7c8c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/worker/acl_cache.go
+++ b/worker/acl_cache.go
@@ -51,9 +51,18 @@ var AclCachePtr = &AclCache{
 }
 
 func (cache *AclCache) GetUserPredPerms(userId string) map[string]int32 {
-	cache.Lock()
-	defer cache.Unlock()
-	return cache.userPredPerms[userId]
+	cache.RLock()
+	defer cache.RUnlock()
+
+	perms, found := cache.userPredPerms[userId]
+	if !found {
+		return nil
+	}
+	snapshot := make(map[string]int32, len(perms))
+	for predicate, permission := range perms {
+		snapshot[predicate] = permission
+	}
+	return snapshot
 }
 
 func (cache *AclCache) Update(ns uint64, groups []acl.Group) {
@@ -135,11 +144,14 @@ func (cache *AclCache) Update(ns uint64, groups []acl.Group) {
 		}
 	}
 
-	for _, v := range AclCachePtr.userPredPerms {
-		for k := range v {
+	for userID, perms := range AclCachePtr.userPredPerms {
+		for k := range perms {
 			if x.ParseNamespace(k) == ns {
-				delete(v, k)
+				delete(perms, k)
 			}
+		}
+		if len(perms) == 0 {
+			delete(AclCachePtr.userPredPerms, userID)
 		}
 	}
 
--- a/worker/acl_cache.go
+++ b/worker/acl_cache.go
@@ -148,8 +148,17 @@ func (cache *AclCache) Update(ns uint64, groups []acl.Group) {
 		AclCachePtr.predPerms[k] = v
 	}
 
-	for k, v := range userPredPerms {
-		AclCachePtr.userPredPerms[k] = v
+	// User IDs are namespace-local, so the same ID can exist in multiple namespaces. Merge the
+	// namespaced predicates instead of replacing permissions collected from another namespace.
+	for userID, newPerms := range userPredPerms {
+		perms, found := AclCachePtr.userPredPerms[userID]
+		if !found {
+			perms = make(map[string]int32)
+			AclCachePtr.userPredPerms[userID] = perms
+		}
+		for predicate, permission := range newPerms {
+			perms[predicate] = permission
+		}
 	}
 }
 
--- a/edgraph/access.go
+++ b/edgraph/access.go
@@ -647,9 +647,10 @@ func authorizePreds(ctx context.Context, userData *userData, preds []string,
 		return &authPredResult{allowed: nil, blocked: blockedPreds}
 	}
 	// User can have multiple permission for same predicate, add predicate
-	allowedPreds := make([]string, 0, len(worker.AclCachePtr.GetUserPredPerms(userId)))
+	userPredPerms := worker.AclCachePtr.GetUserPredPerms(userId)
+	allowedPreds := make([]string, 0, len(userPredPerms))
 	// only if the acl.Op is covered in the set of permissions for the user
-	for predicate, perm := range worker.AclCachePtr.GetUserPredPerms(userId) {
+	for predicate, perm := range userPredPerms {
 		if (perm & aclOp.Code) > 0 {
 			allowedPreds = append(allowedPreds, predicate)
 		}
--- a/worker/acl_cache_test.go
+++ b/worker/acl_cache_test.go
@@ -14,10 +14,20 @@ import (
 	"github.com/dgraph-io/dgraph/v25/x"
 )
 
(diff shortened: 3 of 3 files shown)


worker/acl_cache.go
```diff
@@ -148,8 +148,17 @@ func (cache *AclCache) Update(ns uint64, groups []acl.Group) {
...
+	// User IDs are namespace-local, so the same ID can exist in multiple namespaces. Merge the
+	// namespaced predicates instead of replacing permissions collected from another namespace.
+	for userID, newPerms := range userPredPerms {
+		perms, found := AclCachePtr.userPredPerms[userID]
+		if !found {
+			perms = make(map[string]int32)
+			AclCachePtr.userPredPerms[userID] = perms
+		}
+		for predicate, permission := range newPerms {
+			perms[predicate] = permission
```
matthewmcneely, line 172: The merge itself is correct, and this is pre-existing rather than something you introduced. But the new write loop widens it, so it's worth knowing about.

`GetUserPredPerms` (line 53) returns the live inner map and then drops the lock. `edgraph.authorizePreds` then ranges it twice with no lock held (`edgraph/access.go` L650-L655), while the `SubscribeForAclUpdates` goroutine calls `Update` and mutates those same inner maps. I ran a probe under `-race` against both commits:

- base: `mapdelete` (the clear loop above) vs `mapIterStart`
- this branch: `mapassign_faststr` (this line) vs `mapIterStart`

Delete-vs-iterate was already there. Insert-vs-iterate is the nastier form, because a grow/rehash mid-iteration is what trips Go's unrecoverable `concurrent map read and map write` throw, which takes the Alpha down. This change also makes the maps genuinely larger (the bug was collapsing each user down to a single namespace), so the iteration window widens along with it.

Cheap to close while you're in here:

```go
func (cache *AclCache) GetUserPredPerms(userId string) map[string]int32 {
	cache.RLock()
	defer cache.RUnlock()
	perms := make(map[string]int32, len(cache.userPredPerms[userId]))
	for pred, perm := range cache.userPredPerms[userId] {
		perms[pred] = perm
	}
	return perms
}
```

While you're there, `authorizePreds` calls `GetUserPredPerms` twice per query, once for the `len()` in the `make` and again for the range. One call would do.

worker/acl_cache.go
```diff
@@ -148,8 +148,17 @@ func (cache *AclCache) Update(ns uint64, groups []acl.Group) {
 		AclCachePtr.predPerms[k] = v
 	}
 
-	for k, v := range userPredPerms {
-		AclCachePtr.userPredPerms[k] = v
+	// User IDs are namespace-local, so the same ID can exist in multiple namespaces. Merge the
+	// namespaced predicates instead of replacing permissions collected from another namespace.
+	for userID, newPerms := range userPredPerms {
+		perms, found := AclCachePtr.userPredPerms[userID]
```
matthewmcneely, line 166: Non-blocking design note, since what you have is correct.

The collision is possible at all because `userPredPerms` is keyed by bare user ID while every value key is namespace-qualified. Keying the outer map by `x.NamespaceAttr(ns, userID)` would remove it at the source, and take two costs with it that this change slightly worsens:

- The clear loop (L138-L144) walks every user's full predicate map on each refresh, and `RefreshACLs` calls `Update` once per namespace, so a full refresh is O(namespaces x total entries).
- `authorizePreds` now iterates a user's predicates across all namespaces to build `allowedPreds`, and `expand(_all_)` builds a hash map over that slice on every query. For a user ID present in many namespaces (common in multi-tenant setups) that grows linearly with namespace count.

`authorizePreds` already has `ns := userData.namespace` in hand, so `GetUserPredPerms(ns, userId)` should be a fairly contained change. Happy to take it as a follow-up if you'd rather keep this PR tight.

worker/acl_cache.go
```diff
@@ -148,8 +148,17 @@ func (cache *AclCache) Update(ns uint64, groups []acl.Group) {
...
+		if !found {
+			perms = make(map[string]int32)
+			AclCachePtr.userPredPerms[userID] = perms
+		}
+		for predicate, permission := range newPerms {
+			perms[predicate] = permission
+		}
 	}
```
matthewmcneely, line 174: Low priority, and pre-existing: once a user's last permission in their only namespace is cleared, the entry sticks around forever with an empty inner map. The merge branch doesn't prune it either. One line at the end of the loop:

```go
if len(perms) == 0 {
	delete(AclCachePtr.userPredPerms, userID)
}
```

Callers behave the same either way, since `allowedPreds` ends up `[]string{}` whether the lookup returns nil or an empty map.

matthewmcneely (changes requested):

gooohgb:
thank you for identifying the concurrency issue.

GetUserPredPerms now copies the user’s permission map under RLock and returns an independent snapshot instead of exposing the cache’s live inner map. authorizePreds obtains that snapshot once and uses it throughout the authorization check.

I also added the suggested cleanup for empty user entries and regression coverage confirming that cache updates and caller-side mutations cannot change a previously returned snapshot.

The focused race test, the full worker package, and the full edgraph package all pass. I left the namespace-qualified outer-key redesign for a separate follow-up, as suggested.

PR #9818, opened 2026-08-27
fix(dgraphtest): give Zero the admin whitelist the Alpha already gets

matthewmcneely:
**Description**

Zero guards `/moveTablet` and `/removeNode` with `adminAuthHandler(strict=true)`, so they are
protected whether or not `--security` is configured: with neither a token nor a whitelist, only a
loopback caller is admitted. `dgraphtest` passes `--security` to the Alpha and never to Zero, so a
test process reaching Zero through a published container port cannot call those endpoints at all.

Pre-v21 is deliberately excluded: those binaries predate both the `--security` superflag and the
admin guard, so passing it there would fail an upgrade test with an unknown-flag error while fixing
nothing. The Alpha's command makes the same split for the same reason.

**Testing**

Asserted on the command rather than end to end, on purpose. An integration test that calls
`MoveTablet` passes on Docker Desktop no matter what the harness passes, because loopback is
admitted unconditionally — I wrote that test first, it passed without this fix, and a test that
cannot fail is worse than no test.

`adminAuthHandler`'s own behaviour is already covered by the unit tests in `dgraph/cmd/zero`,
including the loopback and remote cases. What was untested is the harness configuring Zero, so
that is what these three tests pin. Two of them fail without this change; the third (pre-v21)
passes either way by design, guarding against over-applying the flag.

**Checklist**

- [x] The PR title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary)

### DIFF
--- a/dgraphtest/dgraph.go
+++ b/dgraphtest/dgraph.go
@@ -142,6 +142,21 @@ func (z *zero) cmd(c *LocalCluster) []string {
 	if c.lowerThanV21 {
 		zcmd = append(zcmd, fmt.Sprintf(`--idx=%v`, z.id+1), "--telemetry=false")
 	} else {
+		// Zero's destructive admin endpoints (/moveTablet, /removeNode) are guarded
+		// unconditionally: with neither a token nor a whitelist configured, only a
+		// loopback caller is admitted. A test reaching Zero through a published
+		// container port is not loopback, so without this the harness cannot exercise
+		// those endpoints at all — see adminAuthHandler in dgraph/cmd/zero/admin.go.
+		//
+		// Same value the Alpha already gets, for the same reason: the harness is not
+		// the place to enforce network policy, and a test cluster that refuses its own
+		// control plane fails in a way that looks nothing like its cause.
+		security := `--security=whitelist=0.0.0.0/0`
+		if c.conf.securityToken != "" {
+			security += fmt.Sprintf(`;token=%s`, c.conf.securityToken)
+		}
+		zcmd = append(zcmd, security)
+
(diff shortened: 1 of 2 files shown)


matthewmcneely:
Closing this — the premise was wrong, and I want the record to say so rather than leave a
plausible-looking fix lying around.

Zero already gets its admin whitelist, via `DGRAPH_ZERO_SECURITY=whitelist=0.0.0.0/0` in
`zero.env()` (#9788), and deliberately by environment rather than by flag because an older zero
binary in an upgrade test would fail to start on an unrecognized one. The comment sitting eight
lines from where I made this change says exactly that. Adding `--security` on top duplicated the
setting and Zero began exiting 1 within half a second, which is what broke `TestCheckUpgrade` and
`TestDuplicateUserWithLiveLoader` here.

The 401 I was chasing was self-inflicted. #9813 had overwritten three `dgraphtest` files wholesale
from another tree instead of patching them, silently reverting `zero.env()`, `alpha.env()`, the
`env` method on the `dnode` interface, and the `Env: dc.env(c)` wiring. `DGRAPH_ZERO_SECURITY`
appears 3 times in the log of the run that passed and 0 times in the run that failed.

Fixed in #9813 by restoring those files to `main` and re-applying only the two lines it means to
change. Verified by auditing every deleted line on that branch, not by checking it builds.

For anyone who lands here from a search: `/moveTablet` and `/removeNode` are guarded with
`strict=true`, so they are protected whether or not `--security` is set — with neither a token nor a
whitelist, only a loopback caller is admitted. If a test cannot reach them, check that Zero actually
received `DGRAPH_ZERO_SECURITY`.

PR #9820, opened 2026-09-01
fix(unique): include language in value identity

### LINKED ISSUE #9815: fix(unique): @unique and @lang disagree on whether language is part of value identity
## What

With `@unique` on a `@lang` predicate, the same value under two different language tags is rejected
when both edges are in one mutation and accepted when they are in separate mutations.

Verified on v25.3.8.

## Repro

Schema:

```
gxid: string @unique @index(exact) @lang .
```

Both in one mutation:

```
{ set {
  _:a <gxid> "same"@en .
  _:b <gxid> "same"@fr .
} }
```

gives `could not insert duplicate value [same] for predicate [gxid]`.

The same two values in separate mutations are both accepted, landing on two nodes:

```
{ set { _:a <gxid> "sep"@en . } }
{ set { _:b <gxid> "sep"@fr . } }

{ q(func: eq(gxid@en,"sep")) { uid gxid@en } }  =>  [{"uid":"0xb","gxid@en":"sep"}]
{ q(func: eq(gxid@fr,"sep")) { uid gxid@fr } }  =>  [{"uid":"0xc","gxid@fr":"sep"}]
```

## Cause

The two uniqueness checks disagree about whether language is part of a value's identity.

`addQueryIfUnique` scopes the DB-level query per language:

```go
predicateName = fmt.Sprintf("%v@%v", predicateName, pred.Lang)
```

`verifyUniqueWithinMutation` compares predicate, value, and subject, and never looks at `Lang`:

```go
if pred2.Predicate == pred1.Predicate && dql.TypeValFrom(pred2.ObjectValue).Value == pred1Value &&
    pred2.Subject != pred1.Subject {
```

So the in-request check treats `"x"@en` and `"x"@fr` as one value while the DB check treats them as
two.

## Which side should change

Probably the in-request check. `@lang` values are stored per language independently, so `gxid@en`
and `gxid@fr` are distinct values and uniqueness should be scoped per language, which is what
`addQueryIfUnique` already does. That makes the fix adding `Lang` to the comparison in
`verifyUniqueWithinMutation`.

Either resolution is defensible, but the two paths should agree.

## Minor, related

The generated `__dgraph_uniquecheck_<n>__` variable names leak into user-facing DQL errors when an
upsert fails validation for an unrelated reason:

```
Some variables are defined but not used
Defined:[__dgraph_uniquecheck_0__ __dgraph_uniquecheck_4294967296__ v]
```

gooohgb:
**Description**

Fixes #9815.

For predicates using both `@unique` and `@lang`, equal values under different
language tags were handled inconsistently.

The database-level uniqueness query generated by `addQueryIfUnique` already
scopes values by language, such as `gxid@en` and `gxid@fr`. However,
`verifyUniqueWithinMutation` compared only the predicate and value, causing
equal values under different languages in the same mutation to be incorrectly
rejected as duplicates.

This change includes the language tag when checking value identity within a
mutation. It also applies the same rule to `isSwap`, preventing a mutation of
one language from being mistaken for releasing a unique value stored under
another language.

Predicates without language tags retain their existing behavior because both
language values are empty.

Regression coverage includes:

- Equal values under the same language are rejected.
- Equal values under different languages are accepted.
- Equal untagged values are rejected.
- Tagged and untagged values are treated independently.
- Swap detection only considers mutations under the same language.
- The reported DQL mutation is covered in `TestUniqueForLangDirective`.

**Checklist**

- [ ] Code compiles correctly and linting (via trunk) passes locally

---
<a href="https://app.blacksmith.sh/dgraph-io/codesmith/dgraph/pr/9820"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790821462&installation_model_id=8575&pr_number=9820&repository=dgraph-io%2Fdgraph&return_to=https%3A%2F%2Fgithub.com%2Fdgraph-io%2Fdgraph%2Fpull%2F9820&signature=359d5d4a3199ec9cf5d29b947d19c482d0eec464bdebf8ba726b90f6a5779216"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/edgraph/server_test.go
+++ b/edgraph/server_test.go
@@ -253,6 +253,11 @@ func TestVerifyUniqueWithinMutationLanguageIdentity(t *testing.T) {
 			err := verifyUniqueWithinMutation(qc)
 			if tc.wantError {
 				require.ErrorContains(t, err, "could not insert duplicate value")
+				predicateName := "gxid"
+				if tc.firstLang != "" {
+					predicateName += "@" + tc.firstLang
+				}
+				require.ErrorContains(t, err, "predicate ["+predicateName+"]")
 			} else {
 				require.NoError(t, err)
 			}
--- a/edgraph/server_test.go
+++ b/edgraph/server_test.go
@@ -208,6 +208,81 @@ func TestGetHash(t *testing.T) {
 	require.Equal(t, hex.EncodeToString(h.Sum(nil)), getHash(10, 20))
 }
 
+func TestVerifyUniqueWithinMutationLanguageIdentity(t *testing.T) {
+	tests := []struct {
+		name       string
+		firstLang  string
+		secondLang string
+		wantError  bool
+	}{
+		{name: "same language", firstLang: "en", secondLang: "en", wantError: true},
+		{name: "different languages", firstLang: "en", secondLang: "fr"},
+		{name: "both untagged", wantError: true},
+		{name: "tagged and untagged", firstLang: "en"},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			qc := &queryContext{
+				gmuList: []*dql.Mutation{{
+					Set: []*api.NQuad{
+						{
+							Subject:   "_:a",
+							Predicate: "gxid",
+							ObjectValue: &api.Value{
+								Val: &api.Value_StrVal{StrVal: "same"},
+							},
+							Lang: tc.firstLang,
+						},
+						{
+							Subject:   "_:b",
+							Predicate: "gxid",
+							ObjectValue: &api.Value{
+								Val: &api.Value_StrVal{StrVal: "same"},
+							},
+							Lang: tc.secondLang,
+						},
+					},
+				}},
+				uniqueVars: map[uint64]uniquePredMeta{
+					encodeIndex(0, 0): {},
+					encodeIndex(0, 1): {},
+				},
+			}
+
+			err := verifyUniqueWithinMutation(qc)
+			if tc.wantError {
+				require.ErrorContains(t, err, "could not insert duplicate value")
+			} else {
+				require.NoError(t, err)
+			}
+		})
+	}
+}
+
+func TestIsSwapLanguageIdentity(t *testing.T) {
+	qc := &queryContext{
+		gmuList: []*dql.Mutation{{
+			Set: []*api.NQuad{{
+				Subject:   "0x1",
+				Predicate: "gxid",
+				Lang:      "fr",
+			}},
+		}},
+		uniqueVars: map[uint64]uniquePredMeta{
+			encodeIndex(0, 0): {},
+		},
+	}
+
+	got, err := isSwap(qc, 1, "gxid", "en")
+	require.NoError(t, err)
+	require.False(t, got, "updating another language does not release the existing value")
+
+	got, err = isSwap(qc, 1, "gxid", "fr")
+	require.NoError(t, err)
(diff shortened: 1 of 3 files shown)


edgraph/server_test.go
```diff
@@ -208,6 +208,81 @@ func TestGetHash(t *testing.T) {
 	require.Equal(t, hex.EncodeToString(h.Sum(nil)), getHash(10, 20))
 }
 
+func TestVerifyUniqueWithinMutationLanguageIdentity(t *testing.T) {
```
shiva-istari, line 211: @gooohgb Thanks for the fix.
One ask: three of the important cases are only covered by unit tests right now. Unit tests check the new comparison logic in isolation, but they don't prove the whole flow (mutation → uniqueness query → verify) behaves correctly against a real cluster. Could you add integration tests for these three?

**1. Tagged + untagged value in separate mutations (both orders)**
First commit `_:a <name> "x"` (no language), then in a new mutation try `_:b <name@en> "x"` —and also the other way around. Your unit test says these two are different values and both should be accepted. But whether the *database-side* check agrees depends on how `eq(name, "x")` behaves when the value is stored under a language tag  and that's never actually tested. If the DB side sees them as the same value, we end up with the same "works in separate mutations, fails in one mutation" inconsistency this PR is fixing  just for the tagged/untagged pair instead of en/fr.

**2. Swap across two languages (must now FAIL)**
Node X already has `"v"@en`. One mutation gives `"v"@en` to a new node and at the same time changes X's `@fr` value. Before this PR that slipped a real duplicate past `@unique`; with your fix it must be rejected. That's the most valuable behavior change in the PR, and it currently has no end-to-end test.

**3. Swap within the same language (must still SUCCEED)**
Node X changes its `"v"@en` to `"w"@en` while a new node takes `"v"@en`  the classic legal swap, just on a `@lang` predicate. The existing swap tests only use a predicate without `@lang`.

Two small side notes:
- Now that language is part of the value's identity, the duplicate error could say which one: `predicate [gxid@en]` instead of `predicate [gxid]`. Cheap to add and makes the error self-explanatory.
- The PR says "Fixes #9815", which will auto-close the issue  but the issue's minor second point (the `__dgraph_uniquecheck_N__` variable names leaking into unrelated error messages) isn't addressed here. Maybe split that into its own issue so it doesn't get closed along with this.

gooohgb:
Hi @shiva-istari, thank you for the review. I’ve addressed the requested cases by adding end-to-end integration coverage for:

Tagged and untagged values in separate mutations, in both orders.

A swap across different languages, which must fail.

A legal swap within the same language, which must succeed.

Duplicate-value errors now include the language-qualified predicate, such as name@en.

I verified the changes against a temporary Zero/Alpha cluster. go test ./edgraph -count=1 and all TestUnique* integration tests passed.

shiva-istari (approved):

Issue #9670, opened 2026-03-30
Dgraph v25.1.0 alpha Crashing


foman:
## Describe the bug

Hi team, we're running dgraph v25.1.0 in kubenetes,  3 alpha nodes. we noticed that all the 3 alpha nodes crashed almost at the same time. the following logs are from the crash alpha pod.  this happend twice in 10 days. please help take a look, thanks in advance.

Error log:
alpha-0:
```
I0330 08:23:03.856631       1 groups.go:153] Informed Zero about tablets I have: OK
I0330 08:23:03.856747       1 groups.go:156] Upserted Schema and Types: OK
I0330 08:23:03.856759       1 groups.go:159] Server is ready: OK
I0330 08:23:03.856770       1 access.go:419] InitializeAcl closed
I0330 08:23:04.575569       1 log.go:22] 1 [term: 43] received a MsgVote message with higher term from 2 [term: 44]
I0330 08:23:04.575625       1 log.go:22] 1 became follower at term 44
I0330 08:23:04.575646       1 log.go:22] 1 [logterm: 43, index: 471104830, vote: 0] cast MsgVote for 2 [logterm: 43, index: 471104830] at term 44
I0330 08:23:04.575908       1 log.go:22] 1 [logterm: 43, index: 471104830, vote: 2] rejected MsgVote from 3 [logterm: 43, index: 471104830] at term 44
I0330 08:23:05.577528       1 log.go:22] raft.node: 1 elected leader 2 at term 44
I0330 08:23:06.318047       1 access.go:354] RefreshAcls closed
I0330 08:23:07.619510       1 admin.go:846] namespace: 0. No GraphQL schema in Dgraph; serving empty GraphQL API
I0330 08:23:44.652189       1 draft.go:773] Creating snapshot at Index: 471105243, ReadTs: 807437427
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x1d8ae12]

goroutine 26443 [running]:
github.com/dgraph-io/dgraph/v25/dql.TypeValFrom(0xc00bc6f2c0?)
        /home/runner/_work/dgraph/dgraph/dql/mutation.go:53 +0x12
github.com/dgraph-io/dgraph/v25/edgraph.verifyUniqueWithinMutation(0xc00ba806c0)
        /home/runner/_work/dgraph/dgraph/edgraph/server.go:2162 +0x10a
[...]

PR #9822, opened 2026-09-03
perf(unique): make verifyUniqueWithinMutation linear

### LINKED ISSUE #9814: perf(unique): verifyUniqueWithinMutation is O(N^2), dominates @unique cost on batched mutations
## What

`@unique` write cost grows quadratically with the number of unique-predicate edges in a single
mutation. At 16k edges in one mutation a write takes 9.7s against 142ms for the same mutation on a
plain `@index(exact)` predicate.

The cost sits entirely in `verifyUniqueWithinMutation`, not in the injected uniqueness queries.

## Measurements

v25.3.8, one mutation, `p: string @index(exact) .` against `p: string @unique @index(exact) .`:

| edges/mutation | plain | `@unique` | ratio | growth per doubling |
| --- | --- | --- | --- | --- |
| 500 | 7.1 ms | 19.4 ms | 2.7x | |
| 1,000 | 12.6 ms | 60.4 ms | 4.8x | |
| 2,000 | 24.0 ms | 203.6 ms | 8.5x | |
| 4,000 | 39.2 ms | 554.5 ms | 14.1x | 2.91x |
| 8,000 | 74.1 ms | 2,242.7 ms | 30.3x | 4.04x |
| 16,000 | 141.6 ms | 9,762.4 ms | 68.9x | 4.35x |

Plain grows about 2x per doubling. `@unique` converges on 4x, which is O(N^2).

Single-edge mutations are unaffected: 1.66 ms/mutation with `@unique` against 1.77 ms plain over 300
sequential inserts. The cost scales with edges-per-mutation rather than with total write volume or
how hot the predicate is.

## Where the time goes

The server latency breakdown separates the two candidate costs, because `addQueryIfUnique` runs
during parsing and the duplicate scan runs during processing:

| N | phase | plain | `@unique` |
| --- | --- | --- | --- |
| 8,000 | parsing | 1.1 ms | 1.4 ms |
| 8,000 | processing | 93.4 ms | 2,249.4 ms |

Parsing stays flat, so injecting one `eq()` query per edge is cheap. The per-pair constant confirms
the nested loop as the source: 64M pairs / 2,249 ms at N=8k and 256M pairs / 9,762 ms at N=16k both
work out to 35-38 ns per pair.

## Cause

`verifyUniqueWithinMutation` in `edgraph/server.go` iterates `qc.uniqueVars` (a
`map[uint64]uniquePredMeta`) inside itself. Two things make it worse than the algorithm requires:

- the inner loop starts at `for j := range qc.uniqueVars` rather than `i+1`, so every pair is
  compared twice
- `dql.TypeValFrom(pred2.ObjectValue)` is called on every pair before any match is established

Nested map iteration also explains why the per-pair constant is 35 ns rather than a few.

## Suggested fix

Hash `(predicate, value)` into a set and make a single linear pass, keeping the first-seen subject so
the existing error message is preserved. Starting `j` at `i+1` is a free 2x if a smaller change is
wanted first.

## Impact

Live loader's default `--batch` is 1000, which lands at the ~5x point and never reaches the bad part
of the curve. Application code that builds one large mutation from an import is where this bites.

Numbers are single-threaded HTTP against a `dgraph/standalone:v25.3.8` container, so treat the
absolute values as indicative. The scaling exponent is the durable part, and it is CPU-bound work
that will contend across concurrent writers rather than disappear.

shiva-istari:
Fixes #9814.

The in-request duplicate check for @unique predicates compared every unique-predicate edge against every other, calling dql.TypeValFrom once per pair: O(N^2) time and allocations in the number of edges per mutation. At 8k edges one check took 1.67s and 64M allocations, dominating batched writes on @unique predicates.

Replace the nested scan with a single pass over a seen-map keyed on (predicate, value), remembering the first subject that set each value. Semantics are unchanged: duplicate values from the same subject remain allowed, nil ObjectValues are skipped, entries pruned by updateMutations are still ignored, and the error message is identical. Value identity still uses the interface{} produced by TypeValFrom, so type identity participates in the comparison exactly as it did with ==.

Measured (M4 Pro, benchstat over 6 runs, all p=0.002): 5.20ms -> 32.6us at 500 edges, 1.67s -> 585us at 8000 edges (-99.96%); allocs/op drops from N^2 (64M at 8k) to ~N (8k).

---
<a href="https://app.blacksmith.sh/dgraph-io/codesmith/dgraph/pr/9822"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791012521&installation_model_id=8575&pr_number=9822&repository=dgraph-io%2Fdgraph&return_to=https%3A%2F%2Fgithub.com%2Fdgraph-io%2Fdgraph%2Fpull%2F9822&signature=d46f7704069adb1a08393a7679932a0d28dff361b3ac467c0647926e773eda89"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of unique values within mutations.
  * Duplicate values across different subjects are now detected consistently, including byte, floating-point, and string values.
  * Equivalent string values from different input formats are now recognized as duplicates.
  * Repeated values for the same subject remain supported.
  * Prevented errors when comparing non-scalar values.


edgraph/server.go
```diff
@@ -2440,41 +2440,48 @@ func isDropOperation(op *api.Operation) bool {
...
+// uniqueValueKey identifies a value for the in-request duplicate check: two edges
+// collide iff they write the same value to the same predicate. The value is held as
+// the interface{} produced by dql.TypeValFrom, so type identity participates in the
+// comparison exactly as it did in the previous ==-based check (int64(1) and "1" are
+// distinct). NOTE for #9820: when language becomes part of value identity, add the
+// NQuad's Lang here.
+type uniqueValueKey struct {
+	predicate string
+	value     interface{}
+}
```
matthewmcneely, line 2461: `uniqueValueKey.value` holds whatever `dql.TypeValFrom` returned, and five of its ten branches return an unhashable dynamic type: `[]byte` for `BytesVal`/`GeoVal`/`DatetimeVal`/`BigfloatVal`, and `[]float32` for `Vfloat32Val`. Hashing one of those is `panic: hash of unhashable type`, and there's no `recover()` anywhere in the mutation path (`edgraph/`, `worker/`, `dgraph/cmd/alpha/`), so it takes the alpha process down.

The old code had the same hazard — `==` on uncomparable interfaces panics too — but the `pred2.Predicate == pred1.Predicate` short-circuit meant it took two edges on the same `@unique` predicate before anything got compared. The new code hashes on the first edge, unconditionally.

It's reachable from a plain JSON mutation, with no vector predicate and no ACL involved. `chunker/json_parser.go:224` runs `types.ParseVFloat` on every string value before it consults the schema, so any string shaped like `[...]` becomes a `Vfloat32Val`:

```go
// schema: email: string @unique @index(hash) .
nqs, _, _ := chunker.ParseJSON([]byte(`[{"uid":"_:a","email":"[1.0, 2.0]"}]`), chunker.SetNquads)
// ObjectValue is *api.Value_Vfloat32Val, so TypeValFrom(...).Value is []float32
verifyUniqueWithinMutation(qc)
// panic: hash of unhashable type: []float32
//   edgraph.verifyUniqueWithinMutation  edgraph/server.go:2477
```

Simplest fix I found is to key on a hashable encoding that keeps the dynamic-type half of the old `==`:

```go
type uniqueValueKey struct {
	predicate string
	value     string
}

v := dql.TypeValFrom(pred.ObjectValue).Value
key := uniqueValueKey{predicate: pred.Predicate, value: fmt.Sprintf("%T\x00%v", v, v)}
```

I prototyped that locally: the vfloat case returns `could not insert duplicate value [[1 2]] for predicate [email]` instead of panicking, and `TestVerifyUniqueWithinMutationBoundsChecks` still passes. One `Sprintf` per edge keeps it O(N), and `addQueryIfUnique` already pays a `fmt.Sprintf("%v", ...)` per unique edge anyway.

Worth being deliberate about that `%T`. Keeping it preserves today's behavior where `int64(1)` and `"1"` stay distinct while `DefaultVal "x"` still equals `StrVal "x"` (both land on Go `string`), which is what `==` did. Dropping it would instead align this check with the injected `eq()` query, which stringifies with plain `%v` and already conflates the two. That divergence predates this PR so I'd keep `%T` and leave it alone, but the key encoding is where the choice gets made.

edgraph/server.go
```diff
@@ -2440,41 +2440,48 @@ func isDropOperation(op *api.Operation) bool {
...
+
+// verifyUniqueWithinMutation rejects a request in which two edges set the same value on
+// the same @unique predicate for different subjects. A single linear pass over a
+// seen-map replaces the earlier every-pair scan, which was O(N^2) in the number of
+// unique-predicate edges and dominated large batched mutations (issue #9814).
 func verifyUniqueWithinMutation(qc *queryContext) error {
```
matthewmcneely, line 2489: No new tests here, and this function has history — #9670 was a panic in the same spot. Worth adding alongside `TestVerifyUniqueWithinMutationBoundsChecks`:

- a non-scalar value on a `@unique` predicate (regression for the panic above)
- duplicate value from the same subject is allowed
- duplicate value from a different subject is rejected

matthewmcneely (changes requested):

edgraph/server.go
```diff
@@ -2448,42 +2448,72 @@ func isDropOperation(op *api.Operation) bool {
...
+// content equality here replaces a crash rather than changing any working behavior.
+func uniqueValueKeyFrom(predicate, lang string, tv types.Val) uniqueValueKey {
+	switch v := tv.Value.(type) {
+	case []byte:
+		return uniqueValueKey{predicate: predicate, lang: lang, tid: tv.Tid, value: string(v)}
+	case []float32:
+		return uniqueValueKey{predicate: predicate, lang: lang, tid: tv.Tid,
+			value: string(types.FloatArrayAsBytes(v))}
+	default:
+		return uniqueValueKey{predicate: predicate, lang: lang, tid: tv.Tid, value: tv.Value}
```
matthewmcneely, line 2480: This is the trap I was pointing at with the `%T` note last round, and the `default` arm is where it hits: `tid` is now part of the key for *every* value, not just the slice branches.

`DefaultID` and `StringID` both produce a Go `string`, so `==` treated them as equal. Keying on `Tid` splits them:

```
_:a <email> "x" .                 → DefaultVal → Tid=DefaultID, Value=string("x")
_:b <email> "x"^^<xs:string> .    → StrVal     → Tid=StringID,  Value=string("x")
```

Same worktree, same test, both commits:

| commit | result |
| --- | --- |
| `014fe43` (base) | `could not insert duplicate value [x] for predicate [email]` |
| `5bd79b1` (this) | `<nil>` — both edges written |

And the everyday trigger isn't typed RDF literals, it's mixing RDF and JSON mutations in one `api.Request`: `chunker/rdf_parser.go:215` emits `DefaultVal` for untyped literals while the JSON path emits `StrVal`. For two new nodes both injected `eq()` queries come back empty, so this check is the only guard, and it now passes. Two nodes land on a `@unique` predicate with the same value.

So the commit message's "preserving the type identity the previous ==-based check had" holds for `int64(1)` vs `"1"`, but not here.

Fix is to keep type separation only where it's actually needed. For everything that isn't a slice the `interface{}` already carries its own Go dynamic type, which is exactly what `==` compared:

```go
type uniqueValueKey struct {
	predicate string
	lang      string
	value     interface{}
}

// byteContent keys slice-typed values by exact bytes. Its own Go type keeps them from
// colliding with a plain string, and Tid separates the four []byte-backed types.
type byteContent struct {
	tid   types.TypeID
	bytes string
}

func uniqueValueKeyFrom(predicate, lang string, tv types.Val) uniqueValueKey {
	value := tv.Value
	switch v := tv.Value.(type) {
	case []byte:
		value = byteContent{tid: tv.Tid, bytes: string(v)}
	case []float32:
		value = byteContent{tid: tv.Tid, bytes: string(types.FloatArrayAsBytes(v))}
	}
	return uniqueValueKey{predicate: predicate, lang: lang, value: value}
}
```

I prototyped that on this commit: the `DefaultVal`/`StrVal` duplicate is caught again and every subtest you added still passes, `a string never collides with equal bytes` and the whole vfloat group included.

edgraph/server_test.go
```diff
@@ -288,6 +288,173 @@ func TestIsSwapLanguageIdentity(t *testing.T) {
...
+	})
+
+	t.Run("same value on different predicates is allowed", func(t *testing.T) {
+		require.NoError(t, verifyUniqueWithinMutation(qcFor(
+			nquad("_:a", "email", str("dup@example.com")),
+			nquad("_:b", "backup_email", str("dup@example.com")),
+		)))
+	})
+
+	t.Run("value type participates in identity", func(t *testing.T) {
```
matthewmcneely, line 345: This subtest passes under both the old `==` and the new `tid` key, so it doesn't actually pin the behavior it's named for. Every string in these tests is built with `Value_StrVal`, which is why the `DefaultVal` regression slipped through green CI.

Worth adding a case that puts both string representations in one request, since that's the one the key encoding can break:

```go
t.Run("equal strings from different representations collide", func(t *testing.T) {
	require.Error(t, verifyUniqueWithinMutation(qcFor(
		nquad("_:a", "email", &api.Value{Val: &api.Value_DefaultVal{DefaultVal: "x"}}),
		nquad("_:b", "email", str("x")),
	)))
})
```

matthewmcneely (changes requested):

matthewmcneely (approved):
Both findings are resolved at `aa7f597`. LGTM.

`byteContent` puts the key back on Go dynamic-type identity for everything comparable, so map equality is the old `==` equality again, and only the five slice branches get content-normalized. Re-ran the repros in a clean worktree:

- single vfloat edge: no panic
- `DefaultVal` vs `StrVal` in one RDF mutation: duplicate rejected
- RDF mutation + JSON mutation in one request: duplicate rejected

I also swept all five slice-returning branches of `dql.TypeValFrom` (bytes, geo, datetime, bigfloat, vfloat), single edge and duplicate pair each — no panic, duplicates caught by content. Everything reaching the `default` arm is `int64`, `string`, `bool`, or `float64`, so the hashability hole is closed.

The parser-driven tests are the right shape. Those anti-vacuity guards mean the suite fails loudly if `rdf_parser.go` or the chunker ever stops producing the value kinds these findings rested on, rather than passing for the wrong reason.

`gofmt`, `go vet`, and the full `edgraph` package are clean locally, and CI is green. Thanks for the thorough turnaround on both rounds.

Two bits of bot noise worth ignoring rather than chasing: CodeRabbit's ast-grep `uint32(len(...))` warnings land on the bounds-check line this PR only moved, and the `api.NQuad` lock-copy vet hit at `edgraph/server.go:1114` predates this change.

Issue #9824, opened 2026-09-04
Orphaned posting-list split parts remain live after deletion, rollup, and compaction


KeisukeYamashita:
## Describe the bug

When deletion makes a posting-list split part empty, rollup drops its reference from the parent but does not supersede the existing child record in Badger.

The logical graph reads correctly, but the unreferenced child remains a live, non-empty `BitCompletePosting` record. Advancing the discard timestamp and explicitly compacting the database does not reclaim its contents.

I reproduced this on upstream main at `166d22d54c75749a38ddb41da592e84be77e3434`, with no implementation changes. The reproduction uses synthetic data and the normal posting-list mutation and rollup paths to create the initial splits.

## To Reproduce

1. Check out the tested revision:

   ```sh
   git clone https://github.com/dgraph-io/dgraph.git
   cd dgraph
   git checkout --detach 166d22d54c75749a38ddb41da592e84be77e3434
   ```

2. Save the Go code below as `posting/split_cleanup_test.go`.

3. Run:

   ```sh
   go test ./posting -run '^TestRollupNativeSplitCleanup$' -count=3 -v -timeout=5m
   ```

The test creates a fresh managed Badger database and:

- Adds 1,024 UID edges with 2 KiB synthetic facets at commit timestamp 10.
- Calls rollup using the default split threshold, producing parent splits `[1, 257, 513, 769]`.
- Deletes either the upper half of the edges or the whole list at timestamp 20.
- Calls rollup at read timestamp 30 and persists every returned KV through `TxnWriter`.
- Reopens the database, compacts with discard timestamp 15, and verifies historical and current reads.
- Replays the same rollup output to provide overlapping SSTs, reopens again, advances discard timestamp to 100, and calls `Flatten(1)`.
- Scans all retained versions of child keys no longer referenced by the parent.

On unmodified main, both cases fail on each of three runs:

```text
native split creation: edges=1024, parent splits=[1 257 513 769]
native removed split=513 retained payload=532562
Should be zero, but was 532562

native split creation: edges=1024, parent splits=[1 257 513 769]
native removed split=1 retained payload=532561
Should be zero, but was 532561
```

These are retained value bytes, not compressed filesystem sizes. The test stops at the first failing child in each case.

The fixture uses `NumVersionsToKeep=MaxInt32`, disables background compactors, and sets the L0 table threshold to 1 to exercise explicit compaction. There are no external readers or Raft state.

<details>
[...]

PR #9825, opened 2026-09-04
fix(posting): mark removed split parts for reclamation

KeisukeYamashita:
Refs dgraph-io/dgraph#9824

**Description**

Write `BitEmptyPosting` records for split parts removed during rollup, allowing Badger compaction to reclaim their old payloads.

Use the replacement parent's timestamp, including the `+1` for finite read timestamps, to preserve historical reads.

Add regression tests covering:
- Partial and whole-list deletion for data, reverse, and index keys.
- Splits created through normal mutations and rollup.
- Historical reads, timestamp boundaries, reinsertion, delayed rollup writes, database reopen, and backup conversion.

Validation: the posting test suite, targeted race tests (three runs), targeted debug tests, and Dgraph build pass. Trunk has not yet been run.

This change does not remove empty-key metadata or retroactively clean up already-orphaned parts.

**Checklist**

- [x] The PR title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) syntax.
- [ ] Code compiles correctly and linting (via trunk) passes locally.
- [x] Regression tests added for the bug fix.

---
<a href="https://app.blacksmith.sh/dgraph-io/codesmith/dgraph/pr/9825"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791102655&installation_model_id=8575&pr_number=9825&repository=dgraph-io%2Fdgraph&return_to=https%3A%2F%2Fgithub.com%2Fdgraph-io%2Fdgraph%2Fpull%2F9825&signature=55d7e1058398f040f90a08758d3998e71e1b4bfce4e8f82aefd99593bb18b2c5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved posting list rollups so reads referencing older split parts continue to work after compaction.
  * Preserved access to historical data during rollups, deletions, concurrent reads, and backup operations.
  * Reclaimed unreferenced split data after compaction while maintaining valid reads.

* **Tests**
  * Added comprehensive coverage for split creation, cleanup, deletion modes, key types, timestamp boundaries, and historical reads.

### DIFF
--- a/posting/list.go
+++ b/posting/list.go
@@ -1471,6 +1471,20 @@ func (l *List) Rollup(alloc *z.Allocator, readTs uint64) ([]*bpb.KV, error) {
 		kvs = append(kvs, kv)
 	}
 
+	// Retire removed parts at the parent's timestamp (+1 for finite readTs)
+	// so compaction preserves reads using the old parent.
+	for _, startUid := range l.plist.Splits {
+		if _, retained := out.parts[startUid]; retained {
+			continue
+		}
+		part, err := out.marshalPostingListPart(alloc, l.key, startUid, &pb.PostingList{})
+		if err != nil {
+			return nil, errors.Wrapf(err, "cannot marshal removed posting list part")
+		}
+		part.Version = kv.Version
+		kvs = append(kvs, part)
+	}
+
 	// Sort the KVs by their key so that the main part of the list is at the
 	// start of the list and all other parts appear in the order of their start UID.
 	sort.Slice(kvs, func(i, j int) bool {
--- a/posting/split_cleanup_test.go
+++ b/posting/split_cleanup_test.go
@@ -1,0 +1,393 @@
+/*
+ * SPDX-FileCopyrightText: © 2017-2026 Istari Digital, Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package posting
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"math"
+	"reflect"
+	"testing"
+
+	"github.com/dgraph-io/badger/v4"
+	bpb "github.com/dgraph-io/badger/v4/pb"
+	"github.com/dgraph-io/dgo/v250/protos/api"
+	"github.com/dgraph-io/dgraph/v25/codec"
+	"github.com/dgraph-io/dgraph/v25/protos/pb"
+	"github.com/dgraph-io/dgraph/v25/x"
+	"github.com/dgraph-io/ristretto/v2/z"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/proto"
+)
+
+// Tests share pstore; do not run in parallel.
+type splitCleanupFixture struct {
+	db   *badger.DB
+	opt  badger.Options
+	key  []byte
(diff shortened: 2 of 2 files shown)


posting/list.go
```diff
@@ -1471,6 +1471,20 @@ func (l *List) Rollup(alloc *z.Allocator, readTs uint64) ([]*bpb.KV, error) {
 		kvs = append(kvs, kv)
 	}
 
+	// Retire removed parts at the parent's timestamp (+1 for finite readTs)
+	// so compaction preserves reads using the old parent.
+	for _, startUid := range l.plist.Splits {
```
matthewmcneely, line 1476: The crash-safety of this depends on a write-ordering invariant that's currently implicit. If one of these tombstones ever became durable before the new parent KV and the process died in between, the old parent would still be the latest visible version and would reference a part that now reads as empty. Worse, the next rollup would read that part as empty and persist the loss.

Today the ordering holds: the sort below puts the main key first (`DefaultPrefix` 0x00 sorts before `ByteSplit` 0x04), `TxnWriter` submits commits in call order, and badger's WAL replays a strict prefix, so a tombstone can't survive a crash without its parent. But nothing in the code says that order is load-bearing (the sort comment only mentions readability). Could you add a short comment here or on the sort noting that the main-list KV must be written before these tombstones?

posting/list.go
```diff
@@ -1471,6 +1471,20 @@ func (l *List) Rollup(alloc *z.Allocator, readTs uint64) ([]*bpb.KV, error) {
...
+		if _, retained := out.parts[startUid]; retained {
+			continue
+		}
+		part, err := out.marshalPostingListPart(alloc, l.key, startUid, &pb.PostingList{})
+		if err != nil {
+			return nil, errors.Wrapf(err, "cannot marshal removed posting list part")
+		}
+		part.Version = kv.Version
```
matthewmcneely, line 1484: This works only because the retained-parts loop above shadows `kv` in its own scope, so the outer `kv` here is still the main-list KV with the `+1` applied. That's fragile: a refactor of that loop could silently change which version these tombstones get. Suggest capturing `parentVersion := kv.Version` right after it's computed and using it here.

While you're at it, a line on why removed parts get the parent's `+1` version while retained parts stay at `out.newMinTs` would help future readers: a reader at exactly `newMinTs` still resolves the old parent, which must still see the removed part's data.

matthewmcneely (changes requested):

